PR #525, opened 2014-10-17
extended gen/mailaliases

Laliska:
- 1. works with U:D:rootMailAliasesMail (if defined)
- 2. generates file 'mailaliases-root'


licehammer:
We won't be using this functionality at the end.

PR #615, opened 2014-12-09
SERVICES: report used facility name and DNS alias

zlamalp:
- When data are rejected by the slave script because of wrong
  facility name or dns alias, show what value was used in error message.


zlamalp:
Merged by Slávek locally in 400d836e. Closing this pull-request.

PR #619, opened 2014-12-10
Optimization of getData() methods

zlamalp:
- Added methods getRequiredAttributes() which can handle list of members or users
  and retrieved all attributes at once for all members/users. Returned value is HashMap
  with key "member/user" and value "list of attributes".
- Use above methods in getData() like methods to skip iteration over all members/users
  on resource.
- When resolving authorization on attributes, always load user and
  facility (if necessary by method). Don't do it for each retrieved attribute.
  Like this data are loaded everytime, but only once (not per each attribute).
- Fixed typos on various places, removed some whitespace.
- Fixed indentation of ServicesManagerEntryIntegrationTest.

This pull request should resolve issues from #574.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
@@ -450,7 +450,7 @@ public class AttributesManagerEntry implements AttributesManager {
 
 		//Choose to which attributes has the principal access
 		for(Attribute attr: attributes) {
-			if(!AuthzResolver.isAuthorizedForAttribute(sess, ActionType.WRITE, new AttributeDefinition(attr), facility , null)) throw new PrivilegeException("Principal has no access to set attribute = " + new AttributeDefinition(attr));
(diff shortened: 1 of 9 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
```diff
@@ -1327,6 +1324,8 @@ public void deleteAttribute(PerunSession sess, AttributeDefinition attributeDefi
 		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
 		List<Attribute> attributes = getAttributesManagerBl().getRequiredAttributes(sess, resource, member, workWithUserAttributes);
 		Iterator<Attribute> attrIter = attributes.iterator();
+		User user = getPerunBl().getUsersManagerBl().getUserByMember(sess, member);
+		Facility facility = getPerunBl().getResourcesManagerBl().getFacility(sess, resource);
```
licehammer: This is not optimal. You get these data even when you don't need them (e.g. member-resource attributes)

zlamalp:
Now solved in #692. Closing this pull request.

PR #622, opened 2014-12-14
Notifications: dynamic user locale recognition

jirmauritz:
- the locale and message language can be dynamically choosen
- with the key $user.preferredLanguage in locale of NotifReceiver will be messages sent with respect to the user attribute preferredLanguage

- bug fix: while removing NotifReceiver, it is removed from cache map  allTemplatesByRegexId too

- unnecessary log message error changed to info


jirmauritz:
closed due to few changes according to Slavek's comments

PR #633, opened 2015-01-15
sshkeys_root and k5login_root - the way of creating destination file and setting file permissions has been changed

Laliska:
if destination file doesn't exist, permissions are set on the file in working directory and then propagated by diff_mv on destination file

if destination file exists, any changes are propagated by diff_mv from file in working directory to destination file

all the changes are described in changelog

### DIFF
--- a/perun-services/slave/process-k5login_root.sh
+++ b/perun-services/slave/process-k5login_root.sh
@@ -9,17 +9,23 @@ function process {
 	### Status codes
 	I_CHANGED=(0 "${DST_FILE} updated")
 	I_NOT_CHANGED=(0 "${DST_FILE} has not changed")
+	E_CHOWN=(50 'Cannot chown on ${FROM_PERUN}')
+	E_CHMOD=(51 'Cannot chmod on ${FROM_PERUN}')
 
 	FROM_PERUN="${WORK_DIR}/k5login_root"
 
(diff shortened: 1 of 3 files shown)


perun-services/slave/process-k5login_root.sh
```diff
@@ -14,12 +14,20 @@ function process {
 
 	create_lock
 
+	if [ ! -f ${DST_FILE} ]; then
+		# Destination file doesn't exist
+		chown root.root $FROM_PERUN
+		chmod 0644 $FROM_PERUN
+
+		if [`stat -c "%a" ${FROM_PERUN}` -eq 0644 -a `stat -c "%U" ${FROM_PERUN}` == "root" -a `stat -c "%G" ${FROM_PERUN}` == "root" ]; then
```
licehammer: In general it's sufficient to check return codes of commands (in this case chown and chmod) to tell if the action was performed successfully. You don't need to check the result.
To check the return code you can use catch_error function.

licehammer: To compare two strings in  [ ] use only one '='  (see 'man [').

licehammer:
Merged locally: c7d6b117d2ae46a48cb83096156e511f1d932576

PR #652, opened 2015-01-23
gen+send scripts hadoop_base and hadoop_hdfs have been created

Laliska:
services return list of logins of the facility members


licehammer:
Merged locally with slight modification. Both gen scripts generates files without new line at the end.

PR #662, opened 2015-01-29
Fix of problem with setting more than 1 GID at once

stavamichal:
- change in unixGroup_namespaces module in method changedAttributeHook
   because first we need to set all needed gids and after that can be
   call check and check dependencies. The reason is that there can be
   relation between gids which do not allow to set them separately.
 - add new constructor to WrongReferenceAttributeValueException (with
   cause)
 - add complex test of this functionality


licehammer:
Merged locally with modification of commit message heading.

PR #667, opened 2015-02-04
attributes for fs_replicas created

Simcsa:
- R:D:replicaSourcePath, R:D:replicaDestination and R:D:replicaDestinationPath were created
- added control that value of replicaSourcePath must be the same as one of the F:D:homeMountPoints
- replicaDestination and replicaDestinationPath are checked by regexes

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_replicaDestination.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_replicaDestination.java
@@ -18,11 +18,11 @@ public class urn_perun_resource_attribute_def_def_replicaDestination extends Res
 	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Attribute attribute) throws WrongAttributeValueException {
 
 		if(attribute.getValue() == null) {
-			throw new WrongAttributeValueException(attribute);
+			throw new WrongAttributeValueException(attribute, resource, "This attribute can't be empty");
 		}
 
 		if (!perunSession.getPerunBl().getModulesUtilsBl().isFQDNValid(perunSession, (String) attribute.getValue())) {
-			throw new WrongAttributeValueException(attribute, "Bad replicaDestination attribute format " + attribute.getValue() + ". It should be" +
+			throw new WrongAttributeValueException(attribute, resource, "Bad replicaDestination attribute format " + attribute.getValue() + ". It should be" +
 					"fully qualified domain name.");
 		}
 	}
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_replicaDestination.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * @author Simona Kruppova, Oliver Mrazik
+ */
+public class urn_perun_resource_attribute_def_def_replicaDestination extends ResourceAttributesModuleAbstract implements ResourceAttributesModuleImplApi {
+
+	@Override
+	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Attribute attribute) throws WrongAttributeValueException {
+
+		if(attribute.getValue() == null) {
+			throw new WrongAttributeValueException(attribute);
```
licehammer: Please add some message. It don't have to be clear for user that this attribute can't be empty.

Also add attribute holder to exception - it enables user to know which attribute for which object failed when he execute operation which works with multiple attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_replicaSourcePath.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * @author Simona Kruppova, Oliver Mrazik
+ */
+public class urn_perun_resource_attribute_def_def_replicaSourcePath extends ResourceAttributesModuleAbstract implements ResourceAttributesModuleImplApi {
+
+	@Override
+	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+
+		Facility facility = perunSession.getPerunBl().getResourcesManagerBl().getFacility(perunSession, resource);
+		List<Attribute> attributes = perunSession.getPerunBl().getAttributesManagerBl().getAttributes(perunSession, facility);
```
licehammer: You can get concrete attribute using getAttribute(PerunSession, Facility, String attrName)

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_replicaDestination.java
```diff
@@ -0,0 +1,39 @@
...
+	@Override
+	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Attribute attribute) throws WrongAttributeValueException {
+
+		if(attribute.getValue() == null) {
+			throw new WrongAttributeValueException(attribute, resource, "This attribute can't be empty");
+		}
+
+		if (!perunSession.getPerunBl().getModulesUtilsBl().isFQDNValid(perunSession, (String) attribute.getValue())) {
+			throw new WrongAttributeValueException(attribute, resource, "Bad replicaDestination attribute format " + attribute.getValue() + ". It should be" +
+					"fully qualified domain name.");
```
licehammer: Missing space in string. Result will be "...should befully..." instead of "...should be fully..."

licehammer:
replicaSourcePath module contains bug. Right behavior: if F:D:homeMountPoints is not set, then replicaSourcePath can't be set. Currently it can be set to any value.

zlamalp:
Closing this pull since there is another with fixed implementation #691.

PR #677, opened 2015-02-10
CORE: Fixed setting of attributes in synchronizer

zlamalp:
- Always use nested transaction when setting attributes to member/user
  in synchronization to prevent storing value on any exception.

  Reason - invalidation of member is not enough, because validation checks
  only required attributes. We could invalidate member because of
  exception related to non-required attribute which then stayed in a wrong
  state, because value was kept in DB, since whole sync succeeded.

- Added method setAttributeInNestedTransaction() for user/member and
  mergeAttributeInValueNestedTransaction() for user to AttributesManagerBl.
  They are set to run in nested transaction by spring config in perun-beans.xml.


zlamalp:
Tested locally with manual run of synchronization (merged user as happened in production). With new implementation, already taken logname in same namespace is not set, with old implementation is.

PR #682, opened 2015-02-11
Optimalization of get*Data* methods.

Natrezim:
- loops through members in ServiceManagerBlImpl get*Data* methodes removed
- added 4 new getRequiredAttributes to AttributesManagerImpl which take
  list of members. They return map of member as a key and list of their
  attributes as value.
- added new getRequiredAttributes which takes list of members, it calls
  the 4 new getRequiredAttributes methods and combines their output into
  one hashMap.
- Added changes made by Pavel Zlámal:
  -- Fixed SQL when selecting user required attributes
  -- Method for creating string with list of IDs changed
     from handling only Users to all PerunBeans.
  -- Methods getRequiredAttributes() for list of users now return
     HashMap<User, List<Attribute>> instead of HashMap<Integer,
     List<Attribute>>,
     to be the same as member-like methods.
  -- Added missing authorization on retrieved attributes for new
     getRequiredAttributes() methods.
  -- Updated javadoc to respect above changes.
- Added facility attribute to UserAttributeExtractor because processing of
  virtual attributes(NS_USER_FACILITY_ATTR_VIRT) require it

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
@@ -1,5 +1,6 @@
 package cz.metacentrum.perun.core.api;
 
+import java.util.HashMap;
 import java.util.List;
 
 import cz.metacentrum.perun.core.api.exceptions.*;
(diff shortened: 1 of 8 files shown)


zlamalp:
There is a new pull #692 with fixed implementation from todays hackaton. Closing this one.

PR #698, opened 2015-02-17
facility overview: load hosts directly after start

ondrejvelisek:
facility overview: load hosts directly after start


ondrejvelisek:
Add comit with dynamicly resolve param in links like:
/ic-fed/index-fed-en.html?target=https://einfra.cesnet.cz/apps/user-profile/

PR #704, opened 2015-02-21
created gen and send script for fs_replicas service

Laliska:


licehammer:
Merged locally as 0df95c7ef1b76ffbaa8b87ab025edb949cccc577 with following modifications:
- remove two following '/' chars in output (output is unix path so there is no nedd for sequence "//")
- clearer work with Perl references:  $x->{$y}{$z} usually works the same as $x->{$y}->{$z} thanks to Perl's automatic dereferencing, but the second usage is easier to comprehend.

PR #706, opened 2015-02-25
CORE, GUI: Membership expiration is now without time

zlamalp:
- Create new date formatter without time in beans utils.
 - Update all places in code, where new formatter must be used.
 - Change old date with time to new date without time.
 - AuditparserTest change because using timestamp in example audit
   messages.
 - MembershipExpiration module, use beansUtils formatter without date
   instead of new explicit formatter with date.
 - MembershipExpiration module test, test only on date, not on time.
- When handling membership expiration in GUI, use only date part
  of the timestamp.


zlamalp:
test this please

PR #708, opened 2015-02-25
CORE: Added method getMembersByExpiration()

zlamalp:
- Added method getMembersByExpiration() to searcher. It returns
  all members with expiration date set, relative to input.
  Default is "=" and current date. You can specify different
  comparison sign and set shift in days relative to current date.

  Method ignores members state or VO, everything is based on
  attribute value comparison only (in DB).

- Added test for new method.

- Synchronizer now doesn't iterate through all VOs and their members,
  and uses method getMembersByExpiration() to select only such
  members, which are required to take action on.

- Synchonizer now doesn't touch members in INVALID, SUSPENDED
  and DISABLED state, because their status is not relevant to
  the expiration.
  E.g. from now DISABLED members are not switched to
  EXPIRED state or back to VALID state, since disabling is
  manual action by administrator and shouldn't be overridden
  by some automatic tool.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
@@ -166,10 +166,6 @@ public class SearcherImpl implements SearcherImplApi {
 
 			return jdbcTemplate.query(query.toString(), MembersManagerImpl.MEMBER_MAPPER, def.getId());
 
-		} catch (EmptyResultDataAccessException e) {
-			return new ArrayList<Member>();
-		} catch (RuntimeException e) {
-			throw new InternalErrorException(e);
 		} catch (Exception e) {
 			throw new InternalErrorException(e);
 		}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
@@ -143,13 +143,13 @@ public class SearcherImpl implements SearcherImplApi {
 	}
 
 	@Override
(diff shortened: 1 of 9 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
```diff
@@ -133,4 +141,39 @@ public SearcherImpl(DataSource perunPool) {
...
+		try {
+
+			AttributeDefinition def = ((PerunBl) sess.getPerun()).getAttributesManagerBl().getAttributeDefinition(sess, "urn:perun:member:attribute-def:def:membershipExpiration");
+
+			String query = "select distinct " + MembersManagerImpl.memberMappingSelectQuery + " from members left join member_attr_values val on " +
+					"val.member_id=members.id and val.attr_id=? where TO_DATE(val.attr_value, 'yyyy-MM-dd')"+operator+compareDate;
+
+			return jdbcTemplate.query(query.toString(), MembersManagerImpl.MEMBER_MAPPER, def.getId());
+
+		} catch (EmptyResultDataAccessException e) {
```
licehammer: This exception can't raise here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
```diff
@@ -133,4 +141,39 @@ public SearcherImpl(DataSource perunPool) {
...
+			return new ArrayList<Member>();
+		} catch (RuntimeException e) {
```
licehammer: ```catch(Exception e) ``` which is bellow catches also ```RuntimeException```, so this catch is redundant.

zlamalp:
test this please

PR #739, opened 2015-03-20
Remove members group from list of groups in method getMemberGroups

stavamichal:
- getMemberGroups should not return members group (javadoc and usecase)


stavamichal:
Before merging this pull request, please control that there is no missusing of this method (from now does not return members group, javadoc is correct now).

zlamalp:
Removing members group is implemented in entry, BL layer is used by some more methods and seems to expect to get also members group. We should do this properly. Closing it and making issue in redmine.

PR #740, opened 2015-03-20
Use safer move (diff move) for k5login file

stavamichal:
- first create temporary k5login file in the same home (on the same
   file system), then move it to the right location (diff move for
   respecting linux acl)
 - new trap management (add method add_on_exit and on_exit for this
   puprose)


stavamichal:
Move to pull request: https://github.com/CESNET/perun-services/pull/8

PR #745, opened 2015-03-24
Special behavior in fs_replicas

stavamichal:
- use new attribute isReplicaSourcePathStandaloneDir
 - if this attribute is true, then do not use login in source path
 - if this attribute is false, then use login in source path


licehammer:
Merged locally (updated commit message)

PR #746, opened 2015-03-24
Set status of member before loging

stavamichal:
- we need to have current status of member in auditer log
   message


licehammer:
Merged locally with modified (more verbose) commit message.

PR #750, opened 2015-03-26
Fix of SQL IN clause max size

stavamichal:
- max size of sql IN clause (select in) is 1000, need to be separate
   in more IN clause with 'or' between them (ex.: 'in (..) or in (..) ..')


licehammer:
Merged locally as 80055cdd09857890dfb699742eea9d2e368e2e9b

PR #752, opened 2015-03-26
Fix minor bugs in unixGID modules

stavamichal:
- check allowed range for gid always when check attribute value (not only if it is new GID)
 - remove self object (group or resource) from list of groups and resources with the same GID set (we don't want it there)
 - fix bad seting value for groupGIDAttribute in resource_unixGID_namespace module, we need to take value from the correct place, not from newly created empty attributeDefinition


licehammer:
Please rebase to current master.
Rewrite commit message to be more comprehensive.

stavamichal:
Test this please

stavamichal:
Test this please

stavamichal:
Test fixed.

PR #767, opened 2015-04-05
Safe IN sql clause usage in notifications

jirmauritz:
- method removeAllPoolMessages now uses BeansUtil to carefully build sql
  query where IN clause with big number of argument is used
- new method in BeansUtil, takes List\<Integer> instead of List\<PerunBean>
- logging sender resolution


jirmauritz:
Closing due to bad method name (prepareInSQLCaluse).
Opening new one with fixed name.

PR #768, opened 2015-04-06
Fix drupal elixir slave script

tauceti2:
$groupStruc->{$gid}->{$g_parent_id} must be checked, top-level groups don't have such ID.


zlamalp:
Required change of Drupal service moved to own pull-request #771. Closing this one.

PR #785, opened 2015-04-21
DISPATCHER,ENGINE: Removed unused checkInJob

zlamalp:
- We do not register engine/dispatcher using checkInJob anymore.
- Fixed wrong application context due to wrong comments.


zlamalp:
Closing since it will be fixed by #790.

PR #807, opened 2015-05-04
ENGINE: Rename perun-engine-new to perun-engine

zlamalp:
- Removed old unused perun-engine module.
- Renamed perun-engine-new to perun-engine to take its place.


zlamalp:
test this please.

PR #814, opened 2015-05-06
SERVICES: Added slave part of ldap_vsb_vi service

zlamalp:
- Modified gen part of ldap_vsb_vi. Groups are now
  in subgroup "ou=perun" just like users and it's entry
  is also created. Users and groups are now split
  between two ldif files.

- Added slave script for ldap_vsb_vi. It updates LDAP
  twice, once for users, once for groups.

- Updated description in example pre-script for ldap service.
- Updated deb package changelog and version.


perun-services/gen/ldap_vsb_vi
```diff
@@ -162,7 +179,7 @@ for my $g (@gs) {
 	for my $r (@res) {
 
 		# PRINT RESOURCE
-	    print FILE "dn: cn=" . $r . ",ou=" . $g . ",ou=groups," . $facilityAttributes{$A_F_BASE_DN} . "\n";
+	    print FILE "dn: cn=" . $r . ",ou=" . $g . ",ou=perun,ou=groups," . $facilityAttributes{$A_F_BASE_DN} . "\n";
 	    print FILE "cn: " . $r . "\n";
```
licehammer: indentation

PR #832, opened 2015-05-18
new external source for csv files

Laliska:
- new dependency in pom.xml file for opencsv library
- new class ExtSourceCSV.java in perun/core with implemented methods: findSubjects(), getSubjectByLogin(), getGroupSubjects()
- for now, the class supports searching for local files only

### DIFF
--- a/perun-core/pom.xml
+++ b/perun-core/pom.xml
@@ -352,6 +352,12 @@
 			<scope>provided</scope>
 		</dependency>
 
+                <dependency>
+			<groupId>com.opencsv</groupId>
+			<artifactId>opencsv</artifactId>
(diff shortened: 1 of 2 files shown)


perun-core/pom.xml
```diff
@@ -344,14 +344,20 @@
 			<artifactId>serviceloader-maven-plugin</artifactId>
 			<version>1.0.2</version>
 		</dependency>
-
-		<dependency>
+             
+                <dependency>
```
stavamichal: There are spaces instead of tabs. Please replace spaces by tabs.

Laliska: Should be ok now

PR #838, opened 2015-05-22
SERVICES: Skip empty groups (resources) for ldap_vsb_vi

zlamalp:
- When there are no members on resource, skip it's creation
  in resulting ldif for service ldap_vsb_vi.


zlamalp:
Moved to the own repository: https://github.com/CESNET/perun-services/pull/1

Closing this one.

PR #850, opened 2015-06-03
CLI tools for member-group attributes + Attribute and AttributeDefinition contains also displayName

Laliska:
-new method setMemberGroupAttribute
-new method listOfMemberGroupAttributes
-new method listOfRequiredMemberGroupAttributes
-new method removeMemberGroupAttribute
-new method removeAllMemberGroupAttributes

-method createAttribute now works also with displayName


zlamalp:
Can you please fix also createAttribute CLI tool ? It should allow to set optional display name param. Thank you.

Laliska:
Hi, I've changed it, it should be ok now.

PR #853, opened 2015-06-04
apps: delete metacentrum logo from header of meta applications

ondrejvelisek:


ondrejvelisek:
I've added MetaLogo back to the Metacenturm mailing list app.

PR #856, opened 2015-06-07
Notifications: methods now use only bl layer

jirmauritz:
- all core methods use bl layer
- Role.NOTIFICATION no more has PERUNADMIN rights
- optimalization of perun session initialization


jirmauritz:
Closing due to changes and creating new PR.

PR #880, opened 2015-06-26
New mechanic to set any contacts for facilities

stavamichal:
- new object contactGroup with facility, name and owner's, user's and
   group's contacts
 - change schema for oracle, postgre, hsqldb and also table order
 - add new methods for working with new table to facility impl and api
   managers
 - add methods also to bl, blimpl, api and entry managers
 - add also tests of all methods (basic funcionality)
 - add all methods from api to RPC

### DIFF
--- a/perun-beans/src/main/java/cz/metacentrum/perun/core/api/ContactGroup.java
+++ b/perun-beans/src/main/java/cz/metacentrum/perun/core/api/ContactGroup.java
@@ -1,0 +1,142 @@
+package cz.metacentrum.perun.core.api;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ *
+ * Represents group of contacts for Facility (users, owners and groups)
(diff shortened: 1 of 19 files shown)


stavamichal:
Before merging this commit to devel, production or virtual app, it is needed to create table and everything around by db changelog for version 26.

zlamalp:
I don't see any mechanic to delete contacts from facility when either owner, group or user is deleted from Perun. Since you have constraint (far-key) it would prevent from deletion of owner, group or user. I'm not sure, if this contact information is a priority and should be behind "force" option when deleting group/user or you should better use "on delete cascade" for far-key definition. I suggest second option as I use it on application_form tables where keeping form/items and data for deleted VO makes no sense.

stavamichal:
Fixed. After discussion with Slavek, all removing methods (User, Group, Owner) has it's own part of removing all facilities contacts if forceDelete choosed.

PR #881, opened 2015-06-26
make some scheduling parameters configurable by properties

mvocu:
Make some scheduling parameters (cron expressions, thread pool size) configurable by properties instead of hardcoded. Proper defaults are set in spring context, no change to configuration files is required.


licehammer:
Merged locally. There was still one "facility.type" left, so i fix that during merge.

PR #892, opened 2015-07-07
login namespace module works also for elixir-persistent attribute

Laliska:
- checkAttributeValue() and fillAttribute() now work with elixir-persistent attr.
- changedAttributeHook() has been implemented for setting UserExtSource after elixirID is generated by fillAttribute() method
- new testing class has been created for login namespace module


zlamalp:
Hi, I've merged #899 which allows us to define attribute module per login-namespace (or namespace for any other attribute). Can you please move your functionality into own module class. I believe you can extend base module and override only necessary methods. I'll do the same for ceitec.

Laliska:
Hi, ok I just moved it into its own new class.

zlamalp:
Hi, can you please squash all commits into a single one ? I would merge it afterwards. Thank you.

Laliska:
Commits have been squashed into 1 commit.

PR #896, opened 2015-07-08
modul a test urn_perun_member_group_attribute_def_virt_isGroupAdmin 

kk616:


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_group_attribute_def_virt_isGroupAdmin.java
```diff
@@ -0,0 +1,64 @@
...
+    public Attribute getAttributeValue(PerunSessionImpl sess, Member member, Group group, AttributeDefinition attributeDefinition) throws InternalErrorException {
+
+        List<User> userAdminList;
+        User newUser;
+        Attribute attribute = new Attribute(attributeDefinition);
+
+        try{
+            userAdminList = sess.getPerunBl().getGroupsManagerBl().getAdmins(sess, group, true);
+            newUser = sess.getPerunBl().getUsersManagerBl().getUserByMember(sess, member);
+        }catch (NullPointerException ex){
```
licehammer: There is no need to catch NullPointerException anywhere in perun-core.  (Also, you should never pass null as an argument to any method in core unless it's specifically allowed in documentation.)

zlamalp:
test this please.

PR #902, opened 2015-07-16
Restriction of attribute friendlyName by regular expression

Holdo:
Restriction of attribute friendlyName by regular expression.
Some tests needed to be adjusted, because friendlyNames of attributes contained '_', which is not supported, it was changed to '-'.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
@@ -87,6 +87,8 @@ public interface AttributesManager {
 
 	public static final String LOGIN_NAMESPACE = "login-namespace";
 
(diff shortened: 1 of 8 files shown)


Holdo:
Which checks have failed? I cannot see into the Jenkins.

zlamalp:
I gave you access right now, it will take some time to reflect this change. As for the test of perun-core, you can run them locally (it uses in memory db) to see whats wong. For now, other tests (outside AttributeManager) are creating new attributes with wrong names.

Holdo:
I fixed all the other tests, Jenkins checks passed. Please review.

zlamalp:
Hi, code seems ok, but can you please remove merge commits? If there is no conflict, you can make pull-request based on old master (git will resolve it). If there is a conflict, for our workflow we use rebase. Rebase changes history, but only in your branch which nobody should have checked-out. It keeps our commit history more clear to others. Thank you.

Holdo:
Merge commits removed. Is it ok now?

PR #908, opened 2015-07-21
Fixed deleting of old TaskResults.

D3jph:
Changed SQL statement in TaskResultDaoJdbc because it was throwing SQLException.
Added tests for this statement.


zlamalp:
Merged locally with change in SQL to be compliant to all 3 DBs, tests run in memory now and fixed equals() in TaskResult object. Thanks for the test, which pointed it out.

PR #912, opened 2015-07-24
PERUN: Implemented proper usage of Spring profiles

zlamalp:
Please note, that this is a big change and should be thoroughly tested on devel.
All tests in Jenkins on all DBs passed (it failed to update pull status on github).

- Bumped Spring version 3.1.1 -> 3.1.4.
- Renamed maven module perun-beans -> perun-base 
  (java package name kept for now).
- Renamed all spring app-context files to match same pattern
  module-name.xml or module-name-jdbc.xml etc.
- Dropped almost all module-name-test.xml files, since tests
  are configured using spring profiles.
- Removed all unnecessary .sql/.txt/.sh files from some modules since
  we keep db and test config in perun-core for most of cases.
- Instead of redefining same beans (usually dataSource) in each module,
  use beans defined in perun-core.

- By default Perun now runs in-memory using hsql DB and predefined
  properties in appcontext xml files. It contains single user perun
  with perun admin rights and mostly empty db.

- When building we can tell spring, which "profile" will be the default.
  When not specified, "default" is the default. This means, that all tests
  and also perun-rpc deployed to tomcat will run using in-memory db.

  We can pass: -Dspring.profiles.default=[devel/production]

  during build to make sure, that properties are read from /etc/perun hence
  tests and perun-rpc runs on top o DB defined in /etc/perun/jdbc.properties.

  For now only differences between devel and production are pool sizes for
  perun-dispatcher component and auto-synchronization of group members in
  perun-core.

  By default devel build no longer logs in a "trace", it can be activated
  adding "trace" profile to the list of default/active profiles.

  -Dspring.profiles.default=devel,trace

  Please note, that spring profiles are in a fact run-time parameter. If you pass
  -Dspring.profiles.active=[list of wanted profiles] you can run any build
  of perun in a required mode (in memory / real db / trace). For tomcat you
  must set this value to JAVA_OPTS and perform restart, for jars simply put
  this parameter after your run command.

CABINET

- Fixed test when contacting MU publications fails without credentials.
- Print test messages like perun-core.

CONTROLLER

- Fixed appcontext in a tests, but they are just empty stubs and should be implemented
  in a near future.

CORE

- Unified devel/production resources using spring profile.
- Updated test-data.sql to create also initial perun admin user.
- Fixed data type conversion for oracle NVARCHAR2 -> VARCHAR2 when
  comparing column/row value with data passed to sql.
- Fixed searching for user using exact fields.

DISPATCHER

- Reworked test to conform perun-core standard (Each test extends
  class with shared configuration).
- This module was using spring profiles, but:
  - removed @ActiveProfile(test) since it was not different from default.
  - kept "integration" since it starts some scheduler and was probably used
    for runtime testing ?
  - also some test are runned only on specific profile, which I didn't test.
- Fixed one test, which was performing the action, but didn't test the results.
- Set tests to rollback after transaction, since they were keeping data in DB
  (it was not relevant for inmemory, but we still want to test in real env.).
- Please note that tests are not passing when dispatcher is "enabled" in configuration
  since it starts some periodic threads. I'm not sure if it should be just removed from
  tests or not so tests of dispatcher should be revisited in a near future.

ENGINE

- Removed all tests for old version of engine.
- All tests run by default without need to specify any profile.
- Removed dependency on perun-rpc-lib.
- Cleaned pom.xml, indentation fixes in sources.
- Version of Spring dependency is now shared from our main pom.xml (now 3.1.4).
- Deleted unnecessary bash, txt , sql files which were simulating services run.
- Unified appcontext. Where not configurable, previous "production" values are used.
  Removed definition of unused dataSource, using the one from perun-core and local one.
  Specific data source definition for tests.
- DependenciesResolver uses HashSet instead of a List in dependencies, since
  multiple dependency on same Task doesn't make sense.
- Fixed reloading of previously running Tasks on engine startup / restart
  (was reloading Tasks with wrong engine id).
- Added javadoc to some classes in Engine plus tests.
- ExecutorEngineWorkerImpl - done fixme, so path to gen/send scripts is now
  configurable by "engine.genscript.path" or "engine.sendscript.path" property.
  Default is "gen" and "send".  It was required for unit test to properly run "/bin/true" 
  exec service command since java trigger this relatively to working directory, which
  is always different (test in IDE, maven, runtime).

Since Engine is using two data sources and Spring can handle transaction only above one,
tests must clean up "dataSourceLocal" after each test. Local data source contains Tasks processed by the Engine (schedulingPool) while data source shared with perun-core is used to store TaskResults.

In tests, dataSourceLocal is always "in memory" while normally it's either this (default) or in a file (devel,production).

Since writing TaskResults requires Task to exists in a shared db (as if it would be created by the Dispatcher in a first place), test do have two TaskDao beans (TaskDao and TaskDaoCore) for this purpose.

NOTIFICATIONS

- Fixed classpath of all classes in tests.
- Remove jabber test, since it was not testing inner component, but just
  connection to jabber.org.
- Removed dependency on h2 db, we are using hsql now defined in root pom.xml

RPC

- Default spring profile value is set during maven build to web.xml file
  so webapp should run in a same mode as was tested during the build.
  You can override this value by setting it in JAVA_OPTS for tomcat or by
  setting default/active profile on a command line. 
- Fixed import of appcontext in a web.xml.

TASKS-LIB

- Tests run in a transaction and cleans up.
- SimpleDateFormat is not thread safe, hence we can't used shared
  static property, but must provide new instance for each usage.

Currently we use different date/time format for Tasks and TasksResults which doesn't make sense.
We should switch to using SimpleDateFormat provided by BeansUtils class from perun-base.

VOOT

- Print proper test messages.
- Fixed tests, which should end in exception.
- Fixed creation of VOs and other objects to not collide with existing.


zlamalp:
test this please.

PR #914, opened 2015-07-24
Fix disabled service propagation

mvocu:
Check if the service is not disabled globally, for the facility or for the particular destination, add checks both to dispatcher and engine before the propagation is sent out/executed.


zlamalp:
test this please.

PR #915, opened 2015-07-24
try to recover from InternalErrorException in AuditerConsumer, log error

mvocu:
Do not quit auditer thread when exception occurs, try to reinitialize auditer consumer instead and log error message (should probably employ some strategy to not hammer on it indefinitely). Also log more verbose message when interrupted.


zlamalp:
test this please.

zlamalp:
Merged locally with added logging of InterruptedException too.

PR #926, opened 2015-08-12
Add facility Manager rights to work with getData methods

stavamichal:
- add facility manager rights to work with all getData methods in
   service manager (only for facility where he is already manager)


zlamalp:
test this please.

PR #928, opened 2015-08-13
Check for allowed destination types + new destination type "service-specific"

Holdo:
Feature #2561 - Add check for allowed destination types + add new destination type "service-specific"

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
@@ -22,6 +22,7 @@ import cz.metacentrum.perun.core.api.exceptions.ServiceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ServicesPackageExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ServicesPackageNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongPatternException;
 
 /**
  * @author Michal Prochazka <michalp@ics.muni.cz>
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -938,4 +938,18 @@ public static String cipherInput(String plainText, boolean decrypt) throws Inter
 
 	}
 
+	public static void checkDestinationType(Destination destination) throws WrongPatternException {
+		String destinationType = destination.getType();
+		if (
```
stavamichal: There seems to be too much tabulators in my opinion. Please use 1 instead of 2 or even 3.
And also there is probably no need to use new line on the row 943. You can inspire yourself in VoEntryManager on the line:97 method getAllVos.
Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -938,4 +938,18 @@ public static String cipherInput(String plainText, boolean decrypt) throws Inter
 
 	}
 
+	public static void checkDestinationType(Destination destination) throws WrongPatternException {
```
stavamichal: Add there also possibility, that destination or type is null. Better than let throw NullPointerException.
You can throw InternalErrorException in this case.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
```diff
@@ -570,7 +571,7 @@
 	 * @throws FacilityNotExistsException
 	 * @throws DestinationAlreadyAssignedException
```
stavamichal: Add new exception also to the javadoc of all changed methods.

Holdo:
Please check now.

Holdo:
Please check now.

PR #933, opened 2015-08-19
Added methods for sending SMS through Perun

jirmauritz:
- methods prepared in Utils class
- updated HttpClient library from Apache to version 4.5 since version 4.1 has communication problems with server sms.cesnet.cz


zlamalp:
Can you please update also perun-cabinet module so we don't have two versions of httpclient ? See its ```pom.xml```, ```CabinetServiceImpl``` and ```HttpServiceImpl```. Thank you.

jirmauritz:
Closed to update HttpClient in perun-cabinet module.

PR #939, opened 2015-08-24
Security Teams feature

ondrejvelisek:
+ two bug fixes


ondrejvelisek:
Change InternalErrorException to SecurityTeamAlreadyExists when security team with the same name already exists
Return null when user is not blacklisted instead of false.

ondrejvelisek:
fix compareTo method in SecurityTeams and rebase to current master

ondrejvelisek:
Update postgre and Oracle DB schema

stavamichal:
Fixed in pull request #953

PR #940, opened 2015-08-24
Delete "attributes" attribute from ExtSource class

Holdo:
Feature #2567 - delete "attributes" attribute from ExtSource class

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/ExtSource.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/ExtSource.java
@@ -12,11 +12,9 @@ public class ExtSource extends Auditable {
 
 	private String name;
 	private String type;
-	private Map<String, String> attributes;
 
 	public ExtSource() {
 		super();
-		attributes = new HashMap<>();
 	}
 
 	public ExtSource(int id, String name, String type) {
@@ -55,18 +53,6 @@ public class ExtSource extends Auditable {
 
 	public void setType(String type) {
 		this.type = type;
-	}
-
-	public Map<String, String> getAttributes() {
-		return attributes;
-	}
-
-	public void setAttributes(Map<String, String> attributes) {
-		this.attributes = attributes;
-	}
-
-	public void setAttribute(String attrName, String attrValue) {
-		this.attributes.put(attrName, attrValue);
 	}
 
 	@Override
--- a/perun-base/src/main/java/cz/metacentrum/perun/rpc/deserializer/JsonDeserializer.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/rpc/deserializer/JsonDeserializer.java
@@ -64,9 +64,6 @@ public class JsonDeserializer extends Deserializer {
 	@JsonIgnoreProperties({"shortName", "beanName"})
 	private interface GroupMixIn {}
 
-	@JsonIgnoreProperties({"attributes", "beanName"})
-	private interface ExtSourceMixIn {}
-
 	private interface MemberMixIn {
 		@JsonIgnore
 		void setStatus(String status);
@@ -93,7 +90,6 @@ public class JsonDeserializer extends Deserializer {
 		mapper.getDeserializationConfig().addMixInAnnotations(PerunException.class, PerunExceptionMixIn.class);
 		mapper.getDeserializationConfig().addMixInAnnotations(Destination.class, DestinationMixIn.class);
 		mapper.getDeserializationConfig().addMixInAnnotations(Group.class, GroupMixIn.class);
-		mapper.getDeserializationConfig().addMixInAnnotations(ExtSource.class, ExtSourceMixIn.class);
 
 		mapper.getDeserializationConfig().addMixInAnnotations(Application.class, PerunBeanMixIn.class);
 		mapper.getDeserializationConfig().addMixInAnnotations(ApplicationForm.class, PerunBeanMixIn.class);
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ExtSourcesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ExtSourcesManagerBl.java
@@ -38,13 +38,14 @@ public interface ExtSourcesManagerBl {
 	 *
 	 * @param perunSession
 	 * @param extSource
+	 * @param attributes
 	 *
 	 * @return ExtSource object with newly associated ID.
 	 *
 	 * @throws InternalErrorException
 	 * @throws ExtSourceExistsException
 	 */
-	ExtSource createExtSource(PerunSession perunSession, ExtSource extSource) throws InternalErrorException, ExtSourceExistsException;
+	ExtSource createExtSource(PerunSession perunSession, ExtSource extSource, Map<String, String> attributes) throws InternalErrorException, ExtSourceExistsException;
 
 	/**
 	 * Deletes an external source.
@@ -202,7 +203,7 @@ public interface ExtSourcesManagerBl {
 	 * @throws CandidateNotExistsException
 	 * @throws ExtSourceUnsupportedOperationException
 	 */
-	Candidate getCandidate(PerunSession perunSession, ExtSource source, String login) throws InternalErrorException, ExtSourceNotExistsException, CandidateNotExistsException,ExtSourceUnsupportedOperationException;
+	Candidate getCandidate(PerunSession perunSession, ExtSource source, String login) throws InternalErrorException, ExtSourceNotExistsException, CandidateNotExistsException, ExtSourceUnsupportedOperationException;
 
 	/**
 	 * Get the candidate from subjectData where at least login must exists.
@@ -220,7 +221,7 @@ public interface ExtSourcesManagerBl {
 	 * @throws CandidateNotExistsException
 	 * @throws ExtSourceUnsupportedOperationException
 	 */
-	Candidate getCandidate(PerunSession perunSession, Map<String,String> subjectData ,ExtSource source, String login) throws InternalErrorException, ExtSourceNotExistsException, CandidateNotExistsException,ExtSourceUnsupportedOperationException;
+	Candidate getCandidate(PerunSession perunSession, Map<String,String> subjectData ,ExtSource source, String login) throws InternalErrorException, ExtSourceNotExistsException, CandidateNotExistsException, ExtSourceUnsupportedOperationException;
 
 	void checkExtSourceExists(PerunSession sess, ExtSource extSource) throws InternalErrorException, ExtSourceNotExistsException;
 
@@ -243,4 +244,12 @@ public interface ExtSourcesManagerBl {
 	 * @param sess
 	 */
 	void loadExtSourcesDefinitions(PerunSession sess);
+
+	/**
+	 * Gets attributes for external source.
+	 *
+	 * @param extSource	External Source
+	 * @return			Map of attributes for external source
+	 */
+	Map<String, String> getAttributes(ExtSource extSource);
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
@@ -121,7 +121,7 @@ public interface MembersManagerBl {
 	 */
 	public Member createMember(PerunSession sess, Vo vo, String extSourceName, String extSourceType, String login, Candidate candidate) throws InternalErrorException, WrongReferenceAttributeValueException, WrongAttributeValueException, AlreadyMemberException, ExtendMembershipException;
 
-/**
+	/**
 	 * Creates a new member and sets all member's attributes from the candidate.
 	 * It can be called in synchronous or asynchronous mode
 	 * Also stores the associated user if doesn't exist. This method is used by the registrar.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
@@ -56,9 +56,9 @@ public class ExtSourcesManagerBlImpl implements ExtSourcesManagerBl {
 		this.extSourcesManagerImpl.initialize(sess);
 	}
 
-	public ExtSource createExtSource(PerunSession sess, ExtSource extSource) throws InternalErrorException, ExtSourceExistsException {
+	public ExtSource createExtSource(PerunSession sess, ExtSource extSource, Map<String, String> attributes) throws InternalErrorException, ExtSourceExistsException {
 		getPerunBl().getAuditer().log(sess, "{} created.", extSource);
-		return getExtSourcesManagerImpl().createExtSource(sess, extSource);
+		return getExtSourcesManagerImpl().createExtSource(sess, extSource, attributes);
 	}
 
 	public void deleteExtSource(PerunSession sess, ExtSource extSource) throws InternalErrorException, ExtSourceAlreadyRemovedException {
@@ -107,7 +107,7 @@ public class ExtSourcesManagerBlImpl implements ExtSourcesManagerBl {
 			extSource.setName(extSourceName);
 			extSource.setType(extSourceType);
 			try {
-				return this.createExtSource(sess, extSource);
+				return this.createExtSource(sess, extSource, null);
 			} catch (ExtSourceExistsException e1) {
 				throw new ConsistencyErrorException("Creating existing extSource", e1);
 			}
@@ -187,7 +187,7 @@ public class ExtSourcesManagerBlImpl implements ExtSourcesManagerBl {
 	}
 
 	@Override
-	public Candidate getCandidate(PerunSession sess, ExtSource source, String login) throws InternalErrorException, ExtSourceNotExistsException, CandidateNotExistsException, ExtSourceUnsupportedOperationException  {
+	public Candidate getCandidate(PerunSession sess, ExtSource source, String login) throws InternalErrorException, ExtSourceNotExistsException, CandidateNotExistsException, ExtSourceUnsupportedOperationException {
 		// New Canddate
 		Candidate candidate = new Candidate();
 
@@ -281,7 +281,7 @@ public class ExtSourcesManagerBlImpl implements ExtSourcesManagerBl {
 						try {
 							// Create new one if not exists
 							additionalExtSource = new ExtSource(additionalExtSourceName, additionalExtSourceType);
-							additionalExtSource = getPerunBl().getExtSourcesManagerBl().createExtSource(sess, additionalExtSource);
+							additionalExtSource = getPerunBl().getExtSourcesManagerBl().createExtSource(sess, additionalExtSource, null);
 						} catch (ExtSourceExistsException e1) {
 							throw new ConsistencyErrorException("Creating existin extSource: " + additionalExtSourceName);
 						}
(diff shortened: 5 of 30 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/ExtSource.java
```diff
@@ -57,18 +55,6 @@ public void setType(String type) {
...
-		this.attributes = attributes;
-	}
-
-	public void setAttribute(String attrName, String attrValue) {
-		this.attributes.put(attrName, attrValue);
-	}
-
 	@Override
 	public String serializeToString() {
 		StringBuilder str = new StringBuilder();
```
Holdo: This is what I had to delete. Everything else is just about fixing the mess.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -1567,7 +1567,15 @@ public void run() {
...
+				try {
+					throw e;
+				} catch (PrivilegeException pe) {
+					//FIXME Don't know why, but this catch block for PrivilegeException must be here
+					failedDueToException = true;
+					exceptionMessage = "Cannot synchronize group ";
+					log.error(exceptionMessage + group, pe);
+					exceptionMessage+= "due to unexpected exception: " + pe.getClass().getName() + " => " + pe.getMessage();
+				}
 			} finally {
```
Holdo: I have no idea why the compiler forced me to add the "throw e" into a try block with PrivilegeException catching. It is catching the general Exception just few lines before. Exception extends PerunException, which extends PrivilegeException. No idea why the catching of the general Exception doesn't work for the PrivilegeException.

perun-base/src/main/java/cz/metacentrum/perun/rpc/deserializer/JsonDeserializer.java
```diff
@@ -64,9 +64,6 @@
 	@JsonIgnoreProperties({"shortName", "beanName"})
 	private interface GroupMixIn {}
 
-	@JsonIgnoreProperties({"attributes", "beanName"})
-	private interface ExtSourceMixIn {}
```
stavamichal: Do not remove all ExtSourceMixIn, just attributes. We still want to ignore beanName!

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ExtSourcesManagerBl.java
```diff
@@ -19,6 +19,7 @@
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Do not use Privilege Exception in other layers then Entry! This one is just for testing privileges in the entry layer for users using RPC of perun.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ExtSourcesManagerBl.java
```diff
@@ -243,4 +250,16 @@
...
+	 * Gets attributes for external source. Must be Perun Admin.
+	 *
+	 * @param extSource	External Source
+	 * @param sess		Current Session
+	 * @return			Map of attributes for external source
+	 * @throws ExtSourceNotExistsException
+	 * @throws InternalErrorException
+	 * @throws PrivilegeException
+	 */
+	Map<String, String> getAttributes(ExtSource extSource, PerunSession sess) throws InternalErrorException, PrivilegeException, ExtSourceNotExistsException;
```
stavamichal: Please add perun session always as first argument of method.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/GroupsManagerBl.java
```diff
@@ -33,6 +33,7 @@
 import cz.metacentrum.perun.core.api.exceptions.NotGroupMemberException;
 import cz.metacentrum.perun.core.api.exceptions.NotMemberOfParentGroupException;
 import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -24,6 +24,7 @@
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotValidYetException;
 import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/VosManagerBl.java
```diff
@@ -13,6 +13,7 @@
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -27,6 +27,7 @@
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.ParserException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: This is really wierd and we don't want to do this there. Remove Privilege Exception from the body of this method and also from throws, we don't want to use it there.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -53,6 +53,7 @@
 import cz.metacentrum.perun.core.api.exceptions.NotGroupMemberException;
 import cz.metacentrum.perun.core.api.exceptions.NotMemberOfParentGroupException;
 import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -390,7 +391,7 @@ public void setSearcherBl(SearcherBl searcherBl) {
 	/**
 	 * Call managers' initialization methods
 	 */
-	public void initialize() throws InternalErrorException {
+	public void initialize() throws InternalErrorException, PrivilegeException {
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -34,6 +34,7 @@
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.LoginNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceCSV.java
```diff
@@ -282,4 +287,13 @@ private boolean compareRowToQuery(String[] row, String query) throws InternalErr
         }
         return lineAsMap;
     }
+
+	protected Map<String,String> getAttributes(ExtSource extSource) {
```
stavamichal: This is just the copy of the same method in every ExtSource, better use the one in BlImpl of ExtSource manager. If needed, let Spring to fill object perun to these extSources and call it by perun.getExtSourcesManagerBl....

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -39,6 +40,7 @@
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.rt.InternalErrorRuntimeException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/ExtSourcesManagerImplApi.java
```diff
@@ -12,6 +13,7 @@
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
stavamichal: Also there do not use Privilege Exception.

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -286,7 +286,7 @@ private Candidate setUpCandidate() {
 
 	}
 
-	private Member createSomeMember(final Vo createdVo) throws ExtendMembershipException, AlreadyMemberException, WrongAttributeValueException, WrongReferenceAttributeValueException, InternalErrorException {
+	private Member createSomeMember(final Vo createdVo) throws ExtendMembershipException, AlreadyMemberException, WrongAttributeValueException, WrongReferenceAttributeValueException, InternalErrorException, PrivilegeException {
```
stavamichal: Also there do not use Privilege Exception. You are calling Bl method so no need throw Privilege Exception at all.

Holdo:
Fixed, please check

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -138,10 +139,9 @@ public ExtSource createExtSource(PerunSession sess, ExtSource extSource) throws
 			es.setId(extSource.getId());
 			es.setName(extSource.getName());
 			es.setType(extSource.getType());
-			es.setAttributes(extSource.getAttributes());
 
 			// Now store the attributes
-			Map<String, String> attributes = extSource.getAttributes();
+			Map<String, String> attributes = getAttributes(extSource);
```
stavamichal: This  is not correct. These attributes do not exists before creating extSource. We need to save them separately. For example in method createExtSource use Map<String,String> attributes like argument. And then save them to DB there.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -207,16 +207,16 @@ private void updateExtSource(PerunSession sess, ExtSource extSource) throws Inte
 		}
 
 		// Check the attributes
-		if (!extSourceDb.getAttributes().equals(extSource.getAttributes())) {
+		if (!getAttributes(extSourceDb).equals(getAttributes(extSource))) {
```
stavamichal: There must be DB vs local data in argument. This is DB vs DB data.

Holdo:
Added Map<String, String> attributes parameter for methods from last two comments. Some tests needed to be adjusted.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -281,7 +281,7 @@ public Candidate getCandidate(PerunSession sess, ExtSource source, String login)
 						try {
 							// Create new one if not exists
 							additionalExtSource = new ExtSource(additionalExtSourceName, additionalExtSourceType);
-							additionalExtSource = getPerunBl().getExtSourcesManagerBl().createExtSource(sess, additionalExtSource);
+							additionalExtSource = getPerunBl().getExtSourcesManagerBl().createExtSource(sess, additionalExtSource, attributes);
```
stavamichal: These attributes are not extSource attributes. They are attributes of new candidate. There must be null.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -386,7 +386,7 @@ public Candidate getCandidate(PerunSession perunSession, Map<String,String> subj
 						try {
 							// Create new one if not exists
 							additionalExtSource = new ExtSource(additionalExtSourceName, additionalExtSourceType);
-							additionalExtSource = getPerunBl().getExtSourcesManagerBl().createExtSource(perunSession, additionalExtSource);
+							additionalExtSource = getPerunBl().getExtSourcesManagerBl().createExtSource(perunSession, additionalExtSource, attributes);
```
stavamichal: These attributes are not extSource attributes. They are attributes of new candidate. There must be null.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -207,16 +206,16 @@ private void updateExtSource(PerunSession sess, ExtSource extSource) throws Inte
...
-		if (!extSourceDb.getAttributes().equals(extSource.getAttributes())) {
+		if (!getAttributes(extSourceDb).equals(attributes)) {
 			log.debug("There is a change in the extSource attributes");
-			try {
+				try {
 				// Firstly delete all attributes, then store new ones
 				jdbc.update("delete from ext_sources_attributes where ext_sources_id=?", extSource.getId());
 
-				for (String attrName: extSource.getAttributes().keySet()) {
+				for (String attrName: getAttributes(extSource).keySet()) {
```
stavamichal: This is not correct. You already deleted all attributes from DB. Now you need to add new attributes from map in parameter of method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -207,16 +206,16 @@ private void updateExtSource(PerunSession sess, ExtSource extSource) throws Inte
...
 					jdbc.update("insert into ext_sources_attributes (ext_sources_id, attr_name, attr_value, created_by, created_at, modified_by, modified_at, created_by_uid, modified_by_uid) " +
 							"values (?,?,?,?," + Compatibility.getSysdate() + ",?," + Compatibility.getSysdate() + ",?,?)", extSource.getId(), attrName,
-							extSource.getAttributes().get(attrName), sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getUserId(),
+							getAttributes(extSource).get(attrName), sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getUserId(),
```
stavamichal: Use attribute from map instead of not existing attribute from database.

Holdo:
Fixed, please review

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -207,16 +206,16 @@ private void updateExtSource(PerunSession sess, ExtSource extSource) throws Inte
 		}
 
 		// Check the attributes
-		if (!extSourceDb.getAttributes().equals(extSource.getAttributes())) {
+		if (!getAttributes(extSourceDb).equals(attributes)) {
 			log.debug("There is a change in the extSource attributes");
-			try {
+				try {
```
stavamichal: Indentation here is not correct.

perun-core/src/main/resources/perun-core.xml
```diff
@@ -270,6 +270,25 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 		<constructor-arg ref="perun" />
 	</bean>
 
+    <bean id="extSourceCSV" class="cz.metacentrum.perun.core.impl.ExtSourceCSV">
```
zlamalp: Hi, Spring somehow doesn't like it on run-time causing null pointer exception. Please change all new extSource beans like this:
```xml
<bean class="cz.metacentrum.perun.core.impl.ExtSourceLdap" factory-method="setPerunBlImpl">
	<constructor-arg ref="perun" />
</bean>
```
It's a little hack, which makes spring call static method. Notice removed *id* which is not necessary if bean is not used anywhere else for inclusion. Also code part in each ExtSource class needs to be changed to this:

```java
private static PerunBlImpl perunBl;

// filled by spring (perun-core.xml)
public static PerunBlImpl setPerunBlImpl(PerunBlImpl perun) {
	perunBl = perun;
	return perun;
}
```
Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -207,16 +206,16 @@ private void updateExtSource(PerunSession sess, ExtSource extSource) throws Inte
 		}
 
 		// Check the attributes
-		if (!extSourceDb.getAttributes().equals(extSource.getAttributes())) {
+		if (!getAttributes(extSourceDb).equals(attributes)) {
 			log.debug("There is a change in the extSource attributes");
```
zlamalp: I know, this is an unchanged log message, but can you add ExtSource object to tell, for which attributes has changed. Thanks.
```java
log.debug("There is a change in attributes for {}", extSource);
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcesManagerImpl.java
```diff
@@ -514,4 +504,13 @@ public boolean extSourceExists(PerunSession perunSession, ExtSource extSource) t
 	public void checkExtSourceExists(PerunSession perunSession, ExtSource es) throws InternalErrorException, ExtSourceNotExistsException {
 		if(!extSourceExists(perunSession, es)) throw new ExtSourceNotExistsException("ExtSource: " + es);
 	}
+
+	public Map<String,String> getAttributes(ExtSource extSource) {
+		Map<String, String> attributes = new HashMap<>();
+		SqlRowSet srs = jdbc.queryForRowSet("select attr_name, attr_value from ext_sources_attributes where ext_sources_id = " + extSource.getId());
```
zlamalp: Using ``queryForRowSet()`` is unsafe for HSQL DB (used in tests), since it by default doesn't publish renamed column names. I know, columns are not renamed in this sql query, but we should use standard ``jdbc.query()``. You can select multiple-lines as one object using own ``ResultSetExtractor`` and implementing its ``extractData()`` method. See e.g. ``AttributeRightsExtractor`` defined inside ``AttributesManagerImpl``. Thanks.

Holdo:
Fixed again :smile:

PR #942, opened 2015-08-25
Transactions - new Auditer Messages behaviour

jirmauritz:
- PerunConnection created to catch important events such as setSavepoint, releaseSavepoint, rolback
- Auditer Messages are stored in list of lists where inner lists represent one nested transaction
 - when a transaction is successful, all the Auditer Messages from last list are copied to the previous list
 - when a transaction is not successful, the last list is deleted

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
@@ -392,6 +392,7 @@ public class PerunBlImpl implements PerunBl {
 	 */
 	public void initialize() throws InternalErrorException {
 		this.extSourcesManagerBl.initialize(this.getPerunSession());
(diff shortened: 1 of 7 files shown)


zlamalp:
test this please.

zlamalp:
Hi, tests are not passing since they use springTransactionManager instead of perunTransactionManager bean. I'm not sure if we will ditch the first one (without auditer) or not, since log() method is used in a code a lot and requires auditer when transaction starts not after (when is active). Switching bean definition in abstract classes of tests like this: ```@TransactionConfiguration(defaultRollback = true, transactionManager = "perunTransactionManager")
``` should be enough. Or you can somehow fix the log() method and recognize that auditer messages resource is not initialized (null).

jirmauritz:
Hi, I fixed the log method. I had this fix already on the testing branch, I don't know why I didn't merge it with this branch. Closing pull request and opening new, fixed, one.

PR #950, opened 2015-09-02
CLI tools for managing Facility Contact Groups

zoraseb:


perun-rpc/src/main/perl/Perun/beans/ContactGroup.pm
```diff
@@ -0,0 +1,113 @@
...
+}
+
+sub setUsers
+{
+	my $self = shift;
+	$self->{_users} = shift;
+
+	return;
+}
+
```
licehammer: Please add getCommonArrayRepresentationHeading and getCommonArrayRepresentation methods. 
You can use them later for printing ContactGroups to output. Method printTable uses them internally.

licehammer:
Why does name of some tools start with "get" and other with "list"? (e.g. getContactsOfFacility and listOfAllFacilitiesGroupContacts)

zoraseb:
Fixed, please check it.

PR #955, opened 2015-09-03
clear recurrence for updated tasks

mvocu:
Clear recurrence for tasks with updated data.


licehammer:
Could you please explain reason for this change in commit message?

mvocu:
Explained, hopefully ;-)

PR #961, opened 2015-09-09
Core: fix AuditerMessages behaviour to handle top-level transactions

jirmauritz:
- fixed problem with more top-level transactions when synchronizing groups in perun


zlamalp:
Closing, merged locally by Slávek in e2ef71ad5136. Thanks.

PR #971, opened 2015-09-14
Add compareTo method to perunBean and all objects with needs to sort

stavamichal:
- add compareTo method to all objects we need to compare (perunBean,
   ContactGroup etc.)

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/ExtSource.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/ExtSource.java
@@ -8,7 +8,7 @@ import java.util.Map;
  *
  * @author Michal Prochazka michalp@ics.muni.cz
  */
-public class ExtSource extends Auditable {
+public class ExtSource extends Auditable implements Comparable<PerunBean>{
 
 	private String name;
 	private String type;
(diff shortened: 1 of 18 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/Group.java
```diff
@@ -120,14 +120,18 @@ public void setShortName(String shortName) {
...
-			return this.name.toLowerCase().compareTo(group.getName().toLowerCase());
+	@Override
+	public int compareTo(PerunBean perunBean) {
+		if(perunBean == null) throw new NullPointerException("PerunBean to compare with is null.");
+		if(perunBean instanceof Group) {
+			Group group = (Group) perunBean;
+			if (this.getName() == null && group.getName() != null) return -1;
+			if (group.getName() == null && this.getName() != null) return 1;
+			if (this.getName() == null && group.getName() == null) return 0;
+			return this.getName().toLowerCase().compareToIgnoreCase(group.getName());
```
zlamalp: Any reason to convert ``this.getName()`` to lowercase when you are comparing using ignore case ? Also this method should honor ``equals()`` when possible. => Two groups with same name but different case are different for Perun based on equals().

perun-base/src/main/java/cz/metacentrum/perun/core/api/ExtSource.java
```diff
@@ -103,4 +103,28 @@ public boolean equals(Object obj) {
 		return true;
 	}
 
+	@Override
+	public int compareTo(PerunBean perunBean) {
+		if(perunBean == null) throw new NullPointerException("PerunBean to compare with is null.");
+		if(perunBean instanceof ExtSource) {
+			ExtSource extSource = (ExtSource) perunBean;
+			int compare;
+			//Compare on last Name
```
zlamalp: Missleading inline comment (copy pasted from User probably). Same for line 118.

zlamalp:
I'm missing compare for ``ResourceTag`` bean, it has ``tagName`` param we can use to sort by. Same goes for ``ServicePackage``, it has ``name``.

Otherwise last one missing is ``UserExtSource``. But I'm not sure, if we can sort simply by login or we should get its ExtSource to compare and then compare on logins from same ExtSource.

PR #986, opened 2015-09-29
Init manager and read only perun

stavamichal:


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -5091,8 +5091,16 @@ protected void initialize() throws InternalErrorException {
 		attr.setDisplayName("Last Synchronization timestamp");
 		attributes.add(attr);
 
+		if(perun.isThisPerunReadOnly()) log.debug("Loading attributes manager init in readOnly version.");
+
 		for(AttributeDefinition attribute : attributes) {
-			if(!checkAttributeExistsForInitialize(attribute)) createAttributeExistsForInitialize(attribute);
+			if(!checkAttributeExistsForInitialize(attribute)) {
+				if(perun.isThisPerunReadOnly()) {
+					log.error("One of required attributes not exists {}.", attribute);
```
zlamalp: Maybe I'm missing it, but shouldn't we stop Perun startup here ? Because here we can't create attributes required by core functionality and they don't exists in DB.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -407,6 +408,11 @@ public void setSearcherBl(SearcherBl searcherBl) {
 		this.searcherBl = searcherBl;
 	}
 
+	@Override
+	public boolean isThisPerunReadOnly() {
```
zlamalp: Is this method really required herte ? It's already in BeansUtils as static method and can be called from anywhere. Also, I would drop "This" in the name of method. But it's not that important.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
```diff
@@ -158,12 +160,19 @@ public AuthzRoles getRoles(User user) throws InternalErrorException {
 
 	public void initialize() throws InternalErrorException {
 
+		if(perun.isThisPerunReadOnly()) log.debug("Loading authzresolver manager init in readOnly version.");
```
zlamalp: Same here, shouldn't we stop startup, when roles are missing and can't be created ?

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/DatabaseManagerImplApi.java
```diff
@@ -1,7 +1,10 @@
 package cz.metacentrum.perun.core.implApi;
 
-import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.DBVersion;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import java.sql.SQLException;
```
zlamalp: I believe this is unused import.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/DatabaseManagerImplIntegrationTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.List;
+
+/**
+ * @author Simona Kruppova 410315
+ * @author Oliver Mrazik 410035
+ */
+public class DatabaseManagerImplIntegrationTest extends AbstractPerunIntegrationTest{
```
zlamalp: These tests would probably cause problems on real DB. From the SQL I'm not sure, that they correctly cleanup when they finish in error/ok or before they start. Also SQL seems to be valid for postgres/hsqldb only (e.g. oracle doesn't know column type text). I'll test it locally. Quick solution is to run these tests only in a "default" profile using class/test annotation:

```java
@ActiveProfiles("default")
public class DatabaseManagerImplIntegrationTest extends AbstractPerunIntegrationTest {
```

perun-base/src/main/java/cz/metacentrum/perun/core/api/BeansUtils.java
```diff
@@ -617,6 +617,37 @@ public static String getPropertyFromCustomConfiguration(String propertyFile, Str
...
+			log.debug("Read only configuration is null, set to false.");
+			return false;
+		}
+
+		if(readOnly.contains("true")) {
+			log.debug("Read only configuration found='{}', set to true.", readOnly);
+			return true;
+		}
+
+		log.debug("Read only configuration found='{}', set to false.", readOnly);
```
zlamalp: I'm not sure, how many times is ``PerunBasicDataSource.getConnection()`` called, but it will probably spawn log with debug message since it check if perun is read only everytime.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/DatabaseManagerImplIntegrationTest.java
stavamichal: These tests just try if parsing of DB version is OK. Do not work with Database itself.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
stavamichal: In my opinion is better to have this method also in object Perun for easier usage.

stavamichal:
Fixed.

stavamichal:
This one is prepared to be merged. (maybe test it on all DBs before)

PR #987, opened 2015-09-30
Use prefixes in user virt CertDNs

stavamichal:
- add prefix to getValue method of userCertDNs virt attribute module
 - reason: possibility to existence more than 1 duplicit subject DNs
   with different values (unique is only subject DN + user cert DN)
 - update also Initialization of LDAP from core and updating of this
   attribute in LDAPc


stavamichal:
Do not merge without merging and use also services in: https://github.com/CESNET/perun-services/pull/22

PR #992, opened 2015-10-06
Fix of possibility to throw indexOutOfBoundException in group sync

stavamichal:
- better throw InternalErrorException with information than throwing
   indexOutOfBoundException when array has not enough big size


zlamalp:
We agreed, that this is more like hotfix and we should fix implementation when handling string arrays. LoA part including "|" should be optional here (based on inline comment). Also there is no need to set ``additionalExtLoa`` to -1 since when not used in ``UserExtSource`` it will be 0. Keeping the pull for later update.

stavamichal:
Fixed.

PR #993, opened 2015-10-07
Fix force propagation

mvocu:
This patchset moves processing of forced propagations into special thread both in dispatcher and engine and leaves out all scheduling/execution intervals present for normal propagations. Forcing the propagation now also restarts currently processing tasks, both SEND and GEN.

### DIFF
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/processing/impl/EventProcessorImpl.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/processing/impl/EventProcessorImpl.java
@@ -192,6 +192,7 @@ public class EventProcessorImpl implements EventProcessor {
 						task.setDestinations(null);
 						// signal that task needs to regenerate data
 						task.setSourceUpdated(true);
+						task.setPropagationForced(false);
 						task.setRecurrence(0);
 					} else {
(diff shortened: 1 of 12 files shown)


mvocu:
Failed tasks are not reported immediately, they undergo the usual procedure of retry/resubmission (as described by max_retry and delay task parameters) - see PropagationMaintainerImpl::rescheduleErrorTasks(). This is done every two minutes (scheduling cycle for PropagationMaintainer).

zlamalp:
Please update test ``EventParserImplTest.parseEventTest()``. It is checking message format on hardcoded string, but its structure (parser) has changed (added forced flag). Thanks.

zlamalp:
Hi, about "Failed tasks are not reported immediately" you are right. I didn't realize that. Thanks for clarification.

mvocu:
Regarding the "missing" `taskStatus.isTaskFinished()` - this is covered by testing `task.getStatus().equals(cz.metacentrum.perun.taskslib.model.Task.TaskStatus.DONE)`. Status `DONE` is set (in preceding lines) only for tasks that have finished.

mvocu:
The event parser test is updated.

zlamalp:
Thanks for clarification and fixing the test. Otherwise it seems ok, merging.

PR #1014, opened 2015-10-19
Facility detail new option "Not selected" 

kk616:
new option "Not selected" - gets attributes with values independent of service

### DIFF
--- a/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/facilitiestabs/FacilitySettingsTabItem.java
+++ b/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/facilitiestabs/FacilitySettingsTabItem.java
@@ -5,6 +5,8 @@ import com.google.gwt.event.dom.client.ChangeEvent;
 import com.google.gwt.event.dom.client.ChangeHandler;
 import com.google.gwt.event.dom.client.ClickEvent;
 import com.google.gwt.event.dom.client.ClickHandler;
+import com.google.gwt.event.logical.shared.ValueChangeEvent;
+import com.google.gwt.event.logical.shared.ValueChangeHandler;
 import com.google.gwt.resources.client.ImageResource;
 import com.google.gwt.user.cellview.client.CellTable;
 import com.google.gwt.user.client.ui.*;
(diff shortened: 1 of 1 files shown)


perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/facilitiestabs/FacilitySettingsTabItem.java
```diff
@@ -121,10 +125,30 @@ public Widget draw() {
...
-
 		// get empty table
 		final CellTable<Attribute> table = reqAttrs.getEmptyTable();
 
+		final GetAttributesV2 attrs = new GetAttributesV2(new JsonCallbackEvents(){
+			@Override
+			public void onFinished(JavaScriptObject jso){
+				List<Attribute> list = JsonUtils.<Attribute>jsoAsList(jso);
+				for(int i = 0; i < list.size(); i++){
+					reqAttrs.addToTable(list.get(i));
```
zlamalp: Please use ``setList();`` instead of ``addToTable()`` and then call ``sortTable();``. Thanks.

zlamalp:
"Not selected" option is not present if you uncheck "Offer available only" checkbox. It should be present in both modes. While "All" is present only in a first mode.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/facilitiestabs/FacilitySettingsTabItem.java
```diff
@@ -136,19 +160,43 @@ public Widget draw() {
 
 		// services listbox
 		final ListBoxWithObjects<Service> servList = new ListBoxWithObjects<Service>();
+		if (switchServicesChb.getValue() == true) {
+			lastServiceId = 1;
```
zlamalp: Please do not set "lastServiceId" to 1 since it supposed to be ID of a real Service object (not its index in a listbox). It worked previously with 0, since by DB sequence definition we start counting services with 1. Like this you couldn't automatically select service with id=1 and "all" option would be selected instead.

Create own global variable where you can store last selected index and if relevant, use it instead of last service id (e.g. when service id=0, read index which must be either 0 (Not selected) or 1 (All - only if present in listbox) ). Thank you.

PR #1028, opened 2015-11-05
Add option to be able overwrite user attributes while synchronize group

stavamichal:
- new attribute in extSources overwriteUserAttributes, if exists it is
   row of attribute names like: "full_name_of:attribute:withNamespace,
   secondAttribute" which will be separated by "," and then used to
   recognize which attributes for user will be always overwrite for
   this extSource (default settings is updating attributes)
 - if attribute not exists in extSource, all attributes will be updated
 - also if new member for existing user will be created, overwrite only
   named attributes, update all other


stavamichal:
Ask me before merging this pull request please.

Thank you.

PR #1029, opened 2015-11-06
Only javaDoc correction

ondrejvelisek:
Core: Edit wrong javaDoc for user_blacklisted attribute


ondrejvelisek:
I've melt it with another pullrequest #1030

PR #1031, opened 2015-11-09
Added code from bachelors thesis about engine monitoring

dsarman:


licehammer:
Will be done in a different way.

PR #1053, opened 2015-11-30
Fix membership expiration

stavamichal:
- problem with expiration at 29.2. (only once in 4 years)
 - we are using period without year and in callendar there is
   always a year (ex. 1970) and this should not be the leap year,
   so date can be converted bad from 29.2. to 1.3.
 - fix it in vo membershipExpirationRules module
 - fix it in manageMembershipExpiration method
 - rework method standardFormatDate in module (to be more readable)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -1786,9 +1786,10 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 					}
 
 					// Set the date to which the membershi should be extended, can be changed if there was grace period, see next part of the code
-					calendar.set(year, month-1, day); // month is 0-based
 					if (extensionInNextYear) {
-						calendar.add(Calendar.YEAR, 1);
+						calendar.set(year+1, month-1, day);
+					} else {
+						calendar.set(year, month-1, day);
 					}
 
 					// ***** GRACE PERIOD *****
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1786,9 +1786,10 @@ protected boolean canBeMemberInternal(PerunSession sess, Vo vo, User user, Strin
 					}
 
 					// Set the date to which the membershi should be extended, can be changed if there was grace period, see next part of the code
-					calendar.set(year, month-1, day); // month is 0-based
 					if (extensionInNextYear) {
-						calendar.add(Calendar.YEAR, 1);
+						calendar.set(year+1, month-1, day);
```
licehammer: Isn't it exactly the same code as it is in else branch?

stavamichal: Yes, this is typo. I will fix it.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_vo_attribute_def_def_membershipExpirationRules.java
```diff
@@ -164,17 +170,20 @@ private boolean isAllowedParameter(String parameter) {
...
-			date = date.substring(0,position2) + date.substring(position2+1);
+	private String standardFormatDate(String date) throws InternalErrorException {
+		if(date == null || date.isEmpty()) return date;
+		//remove all white spaces
+		date = date.replaceAll("\\s", "");
+
+		Matcher standardFormatDateMatcher = standardFormatDatePattern.matcher(date);
+
+		if(!standardFormatDateMatcher.matches()) {
+			throw new InternalErrorException("Date is not in supported format: " + date);
```
licehammer: Is it surely internal error? Doesn't it mean that the user tried to store bad value to attribute?

stavamichal: It is little more complex. In this case the method not get right format of date, but it is not the same like the format for purpose of module. And I want to show that problem is in the method itself.

PR #1056, opened 2015-12-02
Change mail in user profile

ondrejvelisek:


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -1029,6 +1029,7 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 	 *
 	 * @param i String encrypted request parameter
 	 * @param m String encrypted request parameter
+	 * @param u String id of user you want to validate preferred email request
```
zlamalp: Param ``u`` is type of ``int``, see method implementation.

ondrejvelisek: Yeah, I've just copied it. :-) I've fixed it.

PR #1057, opened 2015-12-03
New more effective group synchronization method

stavamichal:


stavamichal:
Contains changes from #1028

PR #1059, opened 2015-12-08
New EGI-SSO extSource

stavamichal:
- extends LDAP extSource
 - can read all people from EGI SSO LDAP
 - it is very specific extSource with no other use
 - using external openssl tool to get issuer and cert subject from DER
   format of certificate (binary format)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceEGISSO.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceEGISSO.java
@@ -1,0 +1,215 @@
+package cz.metacentrum.perun.core.impl;
+
+import cz.metacentrum.perun.core.api.GroupsManager;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceEGISSO.java
```diff
@@ -0,0 +1,215 @@
...
+			base = "ou=People,dc=egi,dc=eu";
+		}
+
+		if(query == null || query.isEmpty()) throw new InternalErrorException("Query can't be null when searching through EGI SSO.");
+
+		try {
+			SearchControls controls = new SearchControls();
+			controls.setTimeLimit(5000);
+			if (maxResults > 0) {
+					controls.setCountLimit(maxResults);
```
licehammer: Wrong indent. Should be one tab less.

PR #1072, opened 2016-01-06
catch integer parsing exception in event parsing and report it

mvocu:
Patches:
  - regular expression matching events (the last '|' separator should be escaped)
  - uncaught runtime exceptions when parsing wrongly formatted events are now properly detected and logged, faulty events are ignored
  - when dispatcher finds '|' character in data to be sent, it encodes it using base64; engine decodes it if necessary


mvocu:
Human readability should not be harmed - the decoded message is logged a couple of lines later. Escaping | could be better for readability but would slow down the parsing in most cases.

The InvalidEventMessageException exception thrown by parser is caught and logged by the caller, unlike the runtime exception thrown before this fix.

Null task in EventProcessorImpl means there was a parsing exception that was logged in the code above; nevertheless the code needs a little cleanup (task is checked for null later on again, unnecessarily). I will fix that.

PR #1076, opened 2016-01-14
Fix logging of attribute moduel not found

stavamichal:
- log it only if not exists attribute module with friendlyName and also
   not exists attribute module with baseFriendlyName


licehammer:
Merged locally.

PR #1078, opened 2016-01-16
CORE: Use batch update for audit inserts into the DB

Holdo:
CORE: Use batch update for audit inserts into the DB

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
@@ -15,6 +15,7 @@ import cz.metacentrum.perun.core.api.PerunSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.helpers.MessageFormatter;
+import org.springframework.jdbc.core.BatchPreparedStatementSetter;
 import org.springframework.jdbc.core.JdbcPerunTemplate;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -537,11 +536,52 @@ public void setLastProcessedId(String consumerName, int lastProcessedId) throws
...
+							@Override
+							public void setValues(PreparedStatement ps, int i) throws SQLException {
+								final AuditerMessage auditerMessage = messages.get(i);
+								final String message = auditerMessage.getMessage();
+								final PerunSession session = auditerMessage.getOriginaterPerunSession();
+								log.info("AUDIT: {}", message);
+								try {
+									ps.setInt(1, Utils.getNewId(jdbc, "auditer_log_id_seq"));
+								} catch (InternalErrorException e) {
+									log.error("Cannot get unique id for new auditer log message ['{}'], exception: {}", message, e);
```
licehammer: Consider pass the exception up the calling chain. Without ID, the SQL query will fail later anyway.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -537,11 +536,52 @@ public void setLastProcessedId(String consumerName, int lastProcessedId) throws
...
+	/**
 	 * Store the message to the DB.
 	 *
+	 * @deprecated for performance reasons. Use storeMessagesToDb method instead
+	 * @param sess
 	 * @param message
-	 * @throws InternalErrorException
 	 */
+	@Deprecated
 	public void storeMessageToDb(final PerunSession sess, final String message) {
```
licehammer: You can delete this method if it's not used anywhere.

Holdo:
1. Exception is now thrown instead of being logged
2. Depreciated method is no longer depreciated, because it is still in use in the Auditer class
3. Added Spring AOP transaction advisor for new method

PR #1082, opened 2016-01-25
Fix uncaught exception

mvocu:
This pull request removes multiple forced script execution when another instance of GEN/SEND script is running.


licehammer:
merged locally

PR #1083, opened 2016-01-26
Fix creation cookie

stavamichal:
Create cookie if specific path or home exists in env
   
 - if there is enviromental variable PERUN_COOKIE set, use this path to save cookie
 - if not and there is at least enviromental variable HOME with path to HOME directory, use this path to save cookie
 - in other case do not save cookie at all


licehammer:
Please clarify commit message.

stavamichal:
Clarification done.

PR #1084, opened 2016-01-27
New sponsored user type

stavamichal:
- user can be normal, service or sponsored
 - need changes in database, add type column with 'service' or
   'sponsored' info
 - all methods like 'getServiceUser' need to work with specificUser
   instead so they are renamed to something like 'getSpecificUser'
 - rename existing exceptions from service to specific
 - add new enum object specificUserType with values normal, service and
   specific to choose which type we want to use for user
 - add new role SPONSOR to define relationship between sponsor and his
   sponsored users
 - modify row mappers in usersManager and authzResolverManager using new
   database schema for this funcionality
 - add new attribute sponsoredUser to User object and also add all
   needed methods for working with it
 - modify parsing of user in auditParser due to sponsoredUser attribute
 - add this behavior also to LDAP (attribute sponsoredUser) to recognize
   who is sponsored, who is normal and who is service user in LDAP
 - add this behavior also to LDAP initializator to be able initialize
   all information about user specification to LDAP
 - add new tests for testing work with sponsored users


perun-core/src/main/resources/hsqldbChangelog.txt
```diff
@@ -3,6 +3,30 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.35
+alter table user add sponsored_acc char(1) default '0' not null;
```
zlamalp:
```sql
alter table users add sponsored_acc char(1) default '0' not null;
```
(user -> users)

perun-core/src/main/resources/oracleChangelog.txt
```diff
@@ -3,6 +3,25 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.35
+alter table user add sponsored_acc char(1) default '0' not null;
```
zlamalp:
```sql
alter table users add sponsored_acc char(1) default '0' not null;
```
(user -> users)

perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -3,6 +3,30 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.35
+alter table user add sponsored_acc char(1) default '0' not null;
```
zlamalp:
```sql
alter table users add sponsored_acc char(1) default '0' not null;
```
(user -> users)

zlamalp:
Fixed in pull #1087 with bundled GUI changes. Closing this one.

PR #1087, opened 2016-02-01
CORE, GUI: Support for sponsored user type

zlamalp:
- user can be normal, service or sponsored
 - need changes in database, add type column with 'service' or
   'sponsored' info
 - all methods like 'getServiceUser' need to work with specificUser
   instead so they are renamed to something like 'getSpecificUser'
 - rename existing exceptions from service to specific
 - add new enum object specificUserType with values normal, service and
   specific to choose which type we want to use for user
 - add new role SPONSOR to define relationship between sponsor and his
   sponsored users
 - modify row mappers in usersManager and authzResolverManager using new
   database schema for this funcionality
 - add new attribute sponsoredUser to User object and also add all
   needed methods for working with it
 - modify parsing of user in auditParser due to sponsoredUser attribute
 - add this behavior also to LDAP (attribute sponsoredUser) to recognize
   who is sponsored, who is normal and who is service user in LDAP
 - add this behavior also to LDAP initializator to be able initialize
   all information about user specification to LDAP
 - add new tests for testing work with sponsored users

CORE: Fixed bugs in sponsored user logic …
- We now allow sponsored users to own service user.
- Fixed table name in sql changelog.
- Fixed enum name in Role "sponsor".

GUI: Modified naming from Service -> Specific user
- Modified api methods and params since they were renamed in core.
- Added filters when listing users and related specific users.
- Allow to connect sponsors and sponsored users on perun admin user detail.
- List relation to sponsors on member and self user detail.
   For now only perun admin can add/remove sponsorship relation.
- Allow to create sponsored account using "create service member" hack, where
  perun admin can choose type of new user (service / sponsored).
- For now users can't see or modify details of users they sponsor.
- Added missing param to RPC javadoc of createSpecificMember


zlamalp:
test all

zlamalp:
test all.

PR #1105, opened 2016-03-01
REGISTRAR: Url encode Group and VO names in notifications

zlamalp:
- Some VOs/Groups names can contain characters not suitable for URL,
  like spaces, which malform clickable link in mail clients.
  Now they are URL encoded and should be safe.


zlamalp:
test this please

PR #1109, opened 2016-03-04
Fixed: Refactoring of the setAttribute method in impl layer

jirmauritz:
The setAttribute method in the impl layer refactored into two methods: first take one object (PerunBean or String for entityless attribute) and second take two PerunBeans. The implementation is broken into more methods where some of them are transactional. This fixes troubles with postgres database when inserting value after catched exception.


jirmauritz:
Rebased on master with conflicts resolved on 15.08.2016 at 11:50.

stavamichal:
More in PR #1266

PR #1110, opened 2016-03-07
Get members of synchronized group before start of sync process

stavamichal:
- getMembers of group call before get source data
 - add behavior when user make manual changes of group in perun
  - when updating and member not exists, skip him
  - when removing and member is already removed, skip him
  - when adding and member is already added, skip him


stavamichal:
Closing: merged by manual cherrypick.

PR #1111, opened 2016-03-07
DISPATCHER: Fix reading of input messages

zlamalp:
- Move sleep out of the inner loop handling messages to the loop polling for available input.
- Restart internal HornetQ JMS server in case of JMSExceptions

Changes made by @mvocu in a single new commit.


zlamalp:
test this please

PR #1113, opened 2016-03-09
CORE: Service user can't be owner of any other specific user

zlamalp:
- We don't want service users to be owners of other service users
  or guests. Allowed is user -> service, user -> guest and guest -> service.


zlamalp:
test this please

PR #1122, opened 2016-03-14
CORE: Fixed delete specific user

zlamalp:
- We do not delete them, but fixed SQL anyway. Was referring to
  old table "service_user_users".
- Fixed DB version in comments of our DB init files.


zlamalp:
Test this please

PR #1124, opened 2016-03-17
Remove attribute value if extSource return null for it

stavamichal:
- if extSource return null for some attribute, remove this attribute if
   it is member attribute or user with overwrite process
 - if it is user attribute without overwrite process, method
   mergeUserAttributes will not remove current value in Perun


licehammer:
This doesn't work properly with member attributes. If null is returned from extSource as a member attribute value, null is stored to the attribute regardless of  overwriteUserAttributesList.

With user attributes it is OK, because "setAttribute" method is called if attribute is in overwriteUserAttributesList  and "merge" method is called otherwise.

stavamichal:
Ok, I removed the part for member and let just part for user and overwrite attribute list.

licehammer:
Merged manually.

PR #1138, opened 2016-04-06
RPC: Added protection against cookie stealing

zlamalp:
RPC: Added protection against cookie stealing

- When authz/remote user changes, recreate the session to prevent
stealing of identity by cookie stealing.

- When using LOCAL ext source (non-authz access) Apache won't send
  any REMOTE_USER, hence we can't check it against value from Perun,
  which was generated by us.

- When cookies in Tomcat are set to "secured" and "httpOnly",
  which is the default, nobody should be able to steal cookie from a browser.
  Our checks are for cases, when cookie is stolen the hard way (security hole
  in a browser) and we can find out when user is using any kind of authz,
  except the non-authz access.


zlamalp:
I'm unsure about local ext source case. By comments if ExtSource is LOCAL then generate REMOTE_USER name on the fly. But it's only in Perun, hence from Apache it will aways be blank ? Right know it would be consider as changed and regenerated. If I keep it, I don't know if somebody stealed a cookie. Any suggestions ?

PR #1147, opened 2016-04-21
CORE: Speed up getFlatData() method

zlamalp:
- Load all user / user-facility attributes at once for all users allowed
  on facility. It was already used by more complex versions of getData() methods.


zlamalp:
Test this please

zlamalp:
We have only 2 services, which are using it. Tested locally and it produces same content. Will merge.

PR #1148, opened 2016-04-21
CORE: Improved getAssignedResources() in Bl and Impl

zlamalp:
- We can now optionally specify VO or Service when retrieving
  assigned resources from facility. Implemented on DB layer,
  so no more multiple calls and for cycling when used by
  others like getAllowedUsers() method.


zlamalp:
Closing, main changes merged in #1149.

PR #1151, opened 2016-05-02
CORE: Added two optimized select methods

zlamalp:
- getAssignedGroupsToFacility(sess, facility)
- getAllowedMembers(sess, facility)

- Removed many for cycles on related entities in order
  to get assigned groups or allowed members for facility.
  Use new methods instead.

- Added tests on both methods + getAssignedResourcesWithVoOrServiceFilter()
  proving, that single select result matches original implementation
  (getting related entities and joining results in a for loop).

- Source formatting, removed trailing whitespace.


zlamalp:
WIP - do not merge, tests needs to be added to prove it returns same results as previous implementation.

zlamalp:
I added test for comparing old and new implementation. @stavamichal or @licehammer, please review my pull-request. Thank you.

stavamichal:
Checked. It seems to be good for me.

PR #1152, opened 2016-05-02
logging messages added to MailManager & RegistrarManager

Mirco333:


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -659,6 +659,7 @@ public ApplicationFormItem addFormItem(PerunSession user, ApplicationForm form,
 		// set new properties back to object & return
 		item.setOrdnum(ordnum);
 		item.setId(itemId);
+		perun.getAuditer().log(user, "Application form item ID=" + form.getId() + " voID=" + form.getVo().getId() + " groupID=" + form.getGroup().getId() + " has been added");
```
zlamalp: Hi, please always check if group is not null, since it's not mandatory part of an application form (unlike VO, which is always there). You can use one-liner like:
```java
...." voID=" + form.getVo().getId() + (form.getGroup() != null) ? (" groupID=" + form.getGroup().getId()) : "" + " has been added");
```

Otherwise messages look OK. Thanks.

zlamalp:
Hi, please add auditer log message also in ``MailManager.setSendingEnabled()`` method.

Please remove auditer log message from createApplication() since it's already logged in it's internal call to ``createApplicationInternal()``. Sorry I didn't notice in a first place.

Thank you.

zlamalp:
I did those small changes by myself. Merged locally. Thanks for your work.

PR #1154, opened 2016-05-03
New attribute module for redmine project ID

balcirakpeter:
- Created attribute module to controle value for defined regular expression

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_redmineProjectID.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_redmineProjectID.java
@@ -1,0 +1,51 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_redmineProjectID.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class urn_perun_resource_attribute_def_def_redmineProjectID extends ResourceAttributesModuleAbstract implements ResourceAttributesModuleImplApi {
+
+	private static final Pattern pattern = Pattern.compile("^[a-z][-_a-z0-9]+$");
+
+	@Override
+	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		String id = (String) attribute.getValue();
+		if (id == null) {
+			throw new WrongAttributeValueException(attribute, resource, "Attribute can't be empty.");
```
licehammer: Please add suggestion what attribute value is correct to error message. Something like you have few line later is suitable.

PR #1155, opened 2016-05-05
Do not use unlimited bulk of ids to one SQL query

stavamichal:
- Oracle does not support unlimited number of IDs for 'in clause' even
   with or between them, so we need to separate one query to more of
   them. For now we use default value 10000 records per one query.
 - new method getRequiredAttributesForBulk which calls more SQL queries
   for one bulk of IDs (member or user ids)
 - for example: if there are 33000 records, than we will call 4 queries
   (10000+10000+10000+3000) instead of 1 (33000)


zlamalp:
Merging this fix since we need to deploy production. Better solution will be merged after proper testing.

PR #1158, opened 2016-05-08
Module for attribute freeipaGroupName

balcirakpeter:
- Checks if attribute match pattern ^[a-zA-Z0-9_.][a-zA-Z0-9_.-]{0,252}[a-zA-Z0-9_.$-]?$
- Checks if group name is unique (case insensitive)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_freeipaGroupName.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_freeipaGroupName.java
@@ -1,0 +1,94 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.Facility;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.Resource;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.ResourceGroupAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.ResourceGroupAttributesModuleImplApi;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_freeipaGroupName.java
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		Matcher match = pattern.matcher(groupName);
+
+		if (!match.matches()) {
+			throw new WrongAttributeValueException(attribute, group, "Bad format of attribute freeipaGroupName. It has to match pattern ^[a-zA-Z0-9_.][a-zA-Z0-9_.-]{0,252}[a-zA-Z0-9_.$-]?$");
+		}
+
+		//get all groups from vo and check if attribute group name is already used in any of the groups
+		try{
```
stavamichal: This is not correct. We need to check uniqueness of groupName on facility.
Facility can have multiple resources. Every resource can be from different VO.
Two Groups in one VO can be assigned to different resources on different facilities.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_freeipaGroupName.java
```diff
@@ -0,0 +1,89 @@
...
+			for(Group gr : resourceGroups){
+				Attribute freeipaGroupNameAttribute = new Attribute();
+
+				try{
+					freeipaGroupNameAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, rs, gr, A_GR_freeipaGroupName);
+				} catch(AttributeNotExistsException ex) {
+					throw new ConsistencyErrorException("Attribute "+ A_GR_freeipaGroupName +" does not exists for group " + gr + " and resource " + rs ,ex);
+				}
+
+				if (freeipaGroupNameAttribute != null){
```
stavamichal: This should never happen. Attribute will be not null there if exists. Only value can be null.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_freeipaGroupName.java
```diff
@@ -0,0 +1,89 @@
...
+					String name = (String) freeipaGroupNameAttribute.getValue();
+
+					if (name.toLowerCase().equals(groupName.toLowerCase())){
```
stavamichal: Be careful, there is chance to throw NullPointerException on "name". On the line 68 test not null value instead and it will be ok.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_freeipaGroupName.java
```diff
@@ -0,0 +1,89 @@
...
+
+		//Get facility for the resource
+		Facility facility = sess.getPerunBl().getResourcesManagerBl().getFacility(sess, resource);
+
+		// Get all resources from the facility
+		List<Resource> facilityResources =  sess.getPerunBl().getFacilitiesManagerBl().getAssignedResources(sess, facility);
+
+		//For each resource get all groups
+		for(Resource rs : facilityResources){
+			List<Group> resourceGroups =  sess.getPerunBl().getResourcesManagerBl().getAssignedGroups(sess, rs);
```
stavamichal: I am afraid there can be (in some situation) also our value so we can find self there. If rs.id = resource.id call resourceGroups.remove(group).

PR #1165, opened 2016-05-17
RPC: Make local Tomcat to keep session in a cookie

zlamalp:
- Locally started Tomcat didn't kept session (opposite to
  production deployment). Now it stores exactly same session
  and we can better debug any session relative code.


zlamalp:
test this please.

PR #1167, opened 2016-05-20
Tool for export member's names and e-mails from group

zoraseb:


perun-rpc/src/main/perl/exportGroupMembers
```diff
@@ -0,0 +1,74 @@
...
+    }
+    my $group = $groupsAgent->getGroupByName(vo => $voId, name => $groupName);
+    $groupId = $group->getId;
+}
+
+my @richMembers = $groupsAgent->getGroupRichMembers(group => $groupId, pageSize => 0, pageNum => 0);
+unless(@richMembers) { printMessage "No member found", $batch; exit 0; }
+
+#output
+my $table = Text::ASCIITable->new();
```
zlamalp: UTF-8 mode must be disabled for ASCIITable, since data is already decoded:
```perl
my $table = Text::ASCIITable->new({reportErrors => 0, utf8 => 0});
```

perun-rpc/src/main/perl/exportGroupMembers
```diff
@@ -0,0 +1,74 @@
...
+$table->setCols('Name', 'e-mail' );
+
+foreach my $richMember (sort $sortingFunction @richMembers) {
```
zlamalp: Please do not "get email for each user". Instead of `groupsManager.getGroupRichMembers()` used above, you can use `membersManager.getCompleteRichMembers()`. See docs at:

https://perun.cesnet.cz/web/rpc-javadoc-MembersManager.shtml#MembersManagergetCompleteRichMembers1
https://perun.cesnet.cz/web/rpc-javadoc-MembersManager.shtml#MembersManagergetCompleteRichMembers3

for required method params. Maybe we could get generic export tool out of this, since described methods can get members of VO or group with specified member status and specified list of user/member attributes. 

There might be need to create such function in ``MembersManager.pm`` and update ``RichMember`` object in order to read it's user/member attributes. Then table could write column for each retrieved attribute.

For default use you should provide list of all member statuses, so function gets really all vo/group members and also print this status as column.

perun-rpc/src/main/perl/exportGroupMembers
```diff
@@ -0,0 +1,74 @@
...
+    unless(defined $voId) {
+        unless(defined $voShortName) { die "ERROR: VO specification required.\n"; }
+        my $vo = $agent->getVosAgent->getVoByShortName(shortName => $voShortName);
+        $voId = $vo->getId;
+    }
+    my $group = $groupsAgent->getGroupByName(vo => $voId, name => $groupName);
+    $groupId = $group->getId;
+}
+
+my @richMembers = $groupsAgent->getGroupRichMembers(group => $groupId, pageSize => 0, pageNum => 0);
```
zlamalp: There shouldn't be ``pageSize => 0, pageNum => 0`` anywhere in a client code, since these parameters were removed on server side long time ago. If you find them anywhere, you can safely remove them.

PR #1168, opened 2016-05-24
Info about messages count to statistics about Perun

stavamichal:
- add info about messages count from auditer_log to statistics about
   Perun
 - the only way to easy check if there is some problem with tablespace
   of auditer_log


zlamalp:
We should discuss access rights to methods for retrieving statistics, it's not really private that we have x vos in system etc. Also when we want probe from nagios to check new stat param, we would have to give it perun admin rights, which is not desired.

PR #1173, opened 2016-05-30
New extSource ISMUXML

stavamichal:
- extends ExtSourceXML
 - create own query for ssl connection
 - need new attributes "user", "password", "workplaceId" and "groupName"
 - sending xml file with request

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceISXML.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceISXML.java
@@ -1,0 +1,143 @@
+package cz.metacentrum.perun.core.impl;
+
+import cz.metacentrum.perun.core.api.GroupsManager;
+import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceISXML.java
```diff
@@ -0,0 +1,103 @@
...
+
+	/**
+	 * Prepare request for getting group information.
+	 *
+	 * @return request with specific settings for group (by XML setting in perun-extSource.xml file)
+	 */
+	private String getQueryForGroup() {
+		return	"<?xml version=\"1.0\" encoding=\"utf-8\"?>\n" +
+				"<request>\n" +
+				"<skupina_clenove reqid=\"5\">\n" +
```
tauceti2: Please generate a random reqid code here and log it into the Perun logs. If there will be an error on IS side, they will tag that error with the reqid.

tauceti2: Please generate a random reqid code here and log it into the Perun logs. If there will be an error on IS side, they will tag that error with the reqid.

PR #1188, opened 2016-06-08
fix possible exception when reading TaskResult message from engine

mvocu:
fix code where dispatcher tries to access bean in empty list parsed from task destination result message


mvocu:
Please merge this and deploy to production as soon as possible.

zlamalp:
Merged and deployed. Thanks for the fix.

PR #1189, opened 2016-06-08
Oauth2 Support

ondrejvelisek:
Add support for OAuth 2.0. It parses user authenticated identity from OIDC header which is send by Apache Openidc module.


ondrejvelisek:
I would like to deploy it on perun.elixir-czech.cz

PR #1199, opened 2016-06-17
GUI,REGISTRAR: Added new RegistrarModule logic canBeApproved() and beforeApprove()

zlamalp:
- We can now use custom beforeApprove() method, which is called
  before application is approved. More specifically, before password is
  validated since it's non reversible change.
- Implemented filter for Educational && Research community which prevents
  approving of application if proper IdP attributes are not passed to form.
- Added check canBeApproved() - for any application can be performed on demand.
- Call this check from GUI before actually approving application. If it throws
   exception - VO/Group admin can still override it and approve registration.
   If it's ok, process continues as before.
   If check must be enforced, same logic needs to be implemented in a module
   method beforeApproval() which makes approving fail the hard way.
- Above logic satisfy DU and Meta different requirements.
- Allow LoA = 2 from hostel IdP for Meta and DU.


zlamalp:
While DU use-case will be revisited, there is no need to hold this, since until explicitly set, nothing is used for their VOs. MetaCenter can have it working right now. Merging.

PR #1206, opened 2016-06-28
Fix lightweight synchronization when source are different

stavamichal:
- use member source to looking for existing user instead of login
   source
 - in method convertSubjectsToCandidates use right order of source
   (member source and login source were switched by mistake)
 - let method convertSubjectsToCandidates get all subject attributes

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -1463,35 +1463,19 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 				continue;
 			}
 
-			//try to find user from perun by login and extSource
+			//try to find user from perun by login and member extSource (need to use memberSource because loginSource is not saved by synchronization)
 			User user = null;
 			Candidate candidate = null;
 			try {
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -1463,35 +1463,19 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
 				continue;
 			}
 
-			//try to find user from perun by login and extSource
+			//try to find user from perun by login and member extSource (need to use memberSource because loginSource is not saved by synchronization)
 			User user = null;
 			Candidate candidate = null;
 			try {
-				UserExtSource userExtSource = getPerunBl().getUsersManagerBl().getUserExtSourceByExtLogin(sess, loginSource, login);
+				UserExtSource userExtSource = getPerunBl().getUsersManagerBl().getUserExtSourceByExtLogin(sess, memberSource, login);
```
licehammer: Please, add comment to the code explaining why there must be memberSource and not loginSource. It's a bit tricky, so I rather have it stressed out in the code.

stavamichal: It is already done on line 1466. It is clear or should I change the position and text a little?

licehammer: OK, solved.

PR #1207, opened 2016-06-28
Fix task assignment

mvocu:
Remove the smart matcher from event processing, tasks are assigned to engine(s) before sending out.

### DIFF
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/processing/impl/EventProcessorImpl.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/processing/impl/EventProcessorImpl.java
@@ -54,8 +54,6 @@ public class EventProcessorImpl implements EventProcessor {
 	@Autowired
 	private EventLogger eventLogger;
 	@Autowired
-	private SmartMatcher smartMatcher;
-	@Autowired
 	private TaskExecutor taskExecutor;
 	private EvProcessor evProcessor;
 	@Autowired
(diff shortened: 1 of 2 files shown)


perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/processing/impl/EventProcessorImpl.java
```diff
@@ -82,41 +80,12 @@ public void run() {
...
-								log.debug("NO MATCH (took "
-										+ (timeStamp2 - timeStamp) + "ms) for "
-										+ dispatcherQueue.getClientID()
-										+ " AND " + event.toString());
-							}
-						}
-						if (orphan) {
-							eventLogger.logEvent(event, -1);
-						}
+						createTask(null, event);
```
zlamalp: Hi, this will consequently call ``SchedulingPoolImpl.addToPool()`` and for newly created Tasks set engineId=-1. ``SchedulingQueue`` is then assigned to it later in ``scheduleTask()`` or ``sendToEngine()`` methods, but I don't see anywhere setting the engineId property back to the Task (both memory-pool and DB) once queue is resolved. Can you please check this too ? I believe correct engineId is necessary for some independent jobs which takes engineId based on queue. Thank you.

mvocu:
Not sure what Task property you have in mind...there is no engineId property in the Task class? Engine assignment is obtained by calling SchedulingPool.getQueueForTask()...

However, there is an engine_id column in table tasks, but is is set only when creating new task record. The TaskDao contains no support for updating the column once the task has been created. I do not know if there are other components using this column, but dispatcher relies only on its (internal) dynamic association between tasks and engines...

zlamalp:
We talked off the record, that there is engine_id in DB (not in Task object) which is set on object creation (assigning it to the first queue). So when we remove initial assigning, we need to add method, which will update enigne_id property in DB, when the queue is changed for the Task later.

zlamalp:
Hi @mvocu, it seems ok now. Setting engine ID was merged in own pull-request. I'd just like to know, what are future plans for SmartMatcher ? Maybe question is for @licehammer. After this, it seems to be used only to reload engines/rules from DB to create queues on startup. But rules are not used anyway (we have round-robin now and we still use single engine). So how about dropping a lot of unnecessary code about registered engines/rules and also clear DB a bit ?

PR #1211, opened 2016-06-30
New workflow for ability to synchronize members by list of their logins

stavamichal:
- new method getGroupSubjects which can get subjects from extSource by
   list of Logins in one or more bulks (not just one by one)
 - redesing of the synchronization process, new method
   categorizeMembersForSynchronization and changes in other methods to
   clear the differences between lightweight and normal synchronization
 - prepare behavior for extSourceSQL and extSourceSQLComplex to be able
   to work with bulk of logins
 - add methods to beansUtils for creating SQLInCaluse for values (not
   just of ids)

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/BeansUtils.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/BeansUtils.java
@@ -455,7 +455,7 @@ public class BeansUtils {
 		for(PerunBean pb: beans) {
 			beansIds.add(pb.getId());
 		}
-		return BeansUtils.prepareInSQLClause(beansIds, identifier);
+		return BeansUtils.prepareInSQLClauseForIds(beansIds, identifier);
 	}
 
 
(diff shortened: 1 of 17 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/BeansUtils.java
```diff
@@ -506,6 +506,52 @@ public static String prepareInSQLClause(List<Integer> beansIds, String identifie
 	}
 
 	/**
+	 * Create a string with set of IN clause. Every in clause has maximum 1000 values.
+	 * Identifier means for what IN clause is calling (Like 'table.value')
+	 *
+	 * Reason for using is compatibility with oracle and other dbs.
+	 *
+	 * Example: " ( in ('10','15',...) or in (...) or ... ) "
```
licehammer: Add "identifier" to example. It doesn't make much sense without it.

licehammer:
Please add test for prepareInSQLClauseForValues method.

stavamichal:
Actual state of this commit is in pull request #1247

PR #1215, opened 2016-07-12
Add persistent flag to UserExtSource

ondrejvelisek:
from perun.properties file.
Please check UserExtSource model bean. I'm not sure if I should add property to other methods such as toString, serializeToString, equals, hashCode etc. Or should I rewrite tests?

Otherwise it works fine. I've tested it on new WUI user profile.

### DIFF
--- a/perun-base/src/main/resources/perun-base.xml
+++ b/perun-base/src/main/resources/perun-base.xml
@@ -89,6 +89,7 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 					<prop key="perun.sms.program"></prop>
 					<prop key="perun.readOnlyPerun">false</prop>
 					<prop key="perun.DBInitializatorEnabled">false</prop>
+					<prop key="perun.userExtSources.persistent">PERUN,[\w\d]*</prop>
 				</props>
 			</property>
 		</bean>
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
@@ -3,6 +3,7 @@ package cz.metacentrum.perun.core.impl;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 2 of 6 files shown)


perun-base/src/main/resources/perun-base.xml
```diff
@@ -89,6 +89,7 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 					<prop key="perun.sms.program"></prop>
 					<prop key="perun.readOnlyPerun">false</prop>
 					<prop key="perun.DBInitializatorEnabled">false</prop>
+					<prop key="perun.userExtSources.persistent">PERUN,[\w\d]*</prop>
```
ondrejvelisek: I guess it is used only in tests.

ondrejvelisek:
Format of property is:  `ExtSourceName1;ExtSourceName2,regex2;ExtSourceName3`
FOr example:  `perun.userExtSources.persistent=PERUN;https://engine.elixir-czech.org/authentication/idp/metadata,.*@elixir-europe.org$`

perun-base/src/main/resources/perun-base.xml
zlamalp: It's used always as a backup, when no other config file is provided to Perun (``/etc/perun/perun.properties``) or such property is missing in it. So it's used by default for in-memory tests.

zlamalp:
For now just skip ``persistent`` param in ``toString()/serializeToString()``. Not everything needs to be in this method. If it will be required, we will add it later.

Also please see ``perun-base`` module, package ``rpc.deserializer`` class ``JsonDeserializer`` and add mixin for ignoring new property (when received by RPC from client apps).

```java
@JsonIgnoreProperties({"persistent"})
private interface UserExtSourceMixIn {}

mapper.getDeserializationConfig().addMixInAnnotations(UserExtSource.class, UserExtSourceMixIn.class);
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -67,6 +72,32 @@
...
+					userExtSourcePersistentPatterns.put(extSource, Pattern.compile(".*"));
+				}
+			}
+		} catch (InternalErrorException e) {
+			log.info("Error when reading property perun.userExtSources.persistent. Serving default behavior.", e);
+			// If error occurred no persistent user ext sources are considered. e.g. property is not set.
+			userExtSourcePersistentPatterns = new HashMap<>();
+		} catch (PatternSyntaxException e) {
+			// cannot use checked exception because of static context
+			throw new InternalErrorRuntimeException("Error when parsing perun.userExtSources.persistent property. Probably regex is in wrong format.", e);
```
licehammer: The PatternSyntaxException can raise, only if "split" argument (which should be regex) is wrongly formatted. In other words, it's not dependent on input (i.e. property from configuration).  
I can guarantee you, that ";" is correct regex :-), therefor this exception never raises.  
Please either delete this "catch" section or at least change the error message in InternalErrorRuntimeException.

ondrejvelisek: Done. I've deleted whole catch block.

PR #1216, opened 2016-07-13
CORE: Allow configurable list of logins for which we don't lookup users

zlamalp:
- To lessen number of unnecessary exceptions in a log we can now configure
  a list of logins for which perun doesn't lookup users in DB.
  This also allows us to assign basic auth identity to a real user.
- Default list of logins is taken from config /etc/perun/perun.properties
  perun.dont.lookup.users and should match other internal logins in same
  config file.


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -95,17 +98,35 @@
 
 	final static Logger log = LoggerFactory.getLogger(PerunBlImpl.class);
 
+	final static List<String> dontLookupUsersForLogins = new ArrayList<String>();
```
licehammer: Rather to use HashSet instead of ArrayList. HashSet can perform contains method in O(1), but for ArrayList it's O(n)

PR #1218, opened 2016-07-13
Disable adding/removing members from groups synchronized with external source

zwejra:
Adding member to synchronized group or removing member from synchronized group is no more allowed.

Changed methods addMember() and removeMember() in GroupsManagerEntry.class
to block adding/removing member to/from synchronized group.

Methods addMember() and removeMember() now throws exceptions WrongAttributeAssignmentException,
AttributeNotExistsException and ExternalyManagedException.

Test addAndRemoveMemberInSynchronizedGroup() was written.

Exception ExternalyManagedException has been made.

This exception is used in GroupsManagerEntry.java in methods addMember and removeMember,
when group is synchronized with external source.

Changed formatting.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
@@ -23,6 +23,7 @@ import cz.metacentrum.perun.core.api.VosManager;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyMemberException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ExternallyManagedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedFromResourceException;
 import cz.metacentrum.perun.core.api.exceptions.GroupExistsException;
@@ -49,6 +50,7 @@ import cz.metacentrum.perun.core.bl.GroupsManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.impl.Utils;
 import cz.metacentrum.perun.core.implApi.GroupsManagerImplApi;
+import java.util.Objects;
 
 /**
  * GroupsManager entry logic
@@ -225,35 +227,47 @@ public class GroupsManagerEntry implements GroupsManager {
 		return group;
 	}
 
-	public void addMember(PerunSession sess, Group group, Member member) throws InternalErrorException, MemberNotExistsException, PrivilegeException, AlreadyMemberException, GroupNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException, NotMemberOfParentGroupException {
+	public void addMember(PerunSession sess, Group group, Member member) throws InternalErrorException, MemberNotExistsException, PrivilegeException, AlreadyMemberException, GroupNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException, NotMemberOfParentGroupException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternallyManagedException {
 		Utils.checkPerunSession(sess);
 		getGroupsManagerBl().checkGroupExists(sess, group);
 		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
-
+		
 		// Authorization
 		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, group)
-				&& !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
+		    && !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
 			throw new PrivilegeException(sess, "addMember");
-				}
+		}
 
 		// Check if the member and group are from the same VO
 		if (member.getVoId() != (group.getVoId())) {
 			throw new MembershipMismatchException("Member and group are form the different VO");
 		}
-
+		
+		// Check if the group is externally synchronized
+		Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+		if (Objects.equals("true", (String) attrSynchronizeEnabled.getValue())) {
+			throw new ExternallyManagedException("Adding of member is not allowed. Group is externally managed.");
+		}
+		
 		getGroupsManagerBl().addMember(sess, group, member);
 	}
 
-	public void removeMember(PerunSession sess, Group group, Member member) throws InternalErrorException, MemberNotExistsException, NotGroupMemberException, PrivilegeException, GroupNotExistsException {
+	public void removeMember(PerunSession sess, Group group, Member member) throws InternalErrorException, MemberNotExistsException, NotGroupMemberException, PrivilegeException, GroupNotExistsException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternallyManagedException {
 		Utils.checkPerunSession(sess);
 		getGroupsManagerBl().checkGroupExists(sess, group);
 		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
 
 		// Authorization
 		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, group)
-				&& !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
+		    && !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
 			throw new PrivilegeException(sess, "removeMember");
-				}
+		}
+		
+		// Check if the group is externally synchronized
+		Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+		if (Objects.equals("true", (String) attrSynchronizeEnabled.getValue())) {
+			throw new ExternallyManagedException("Removing of member is not allowed. Group is externally managed.");
+		}
 
 		getGroupsManagerBl().removeMember(sess, group, member);
 	}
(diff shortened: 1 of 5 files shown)


zwejra:
Sorry, I have changed formatting badly.

zwejra:
All should be okay now.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -1424,6 +1425,47 @@ public void convertGroupsToRichGroupsWithAttributesWithListOfNamesTest() throws
...
+
+		vo = setUpVo();
+		setUpGroup(vo);
+
+		Attribute synchroAttr = new Attribute(perun.getAttributesManager().getAttributeDefinition(sess, "urn:perun:group:attribute-def:def:synchronizationEnabled"));
+		synchroAttr.setValue("true");
+		perun.getAttributesManager().setAttribute(sess, group, synchroAttr);
+
+		Member member = setUpMember(vo);
+		groupsManager.addMember(sess, group, member);
```
zlamalp: Hi, test will throw ExternalyManagedException on this line (as expected) and will end. Any further code is not executed. In order to check other use-cases of add/remove methods you must create separate test for each case.

zwejra: Oh, thanks, I didn't realize that. It should be okay now.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -1450,6 +1451,61 @@ public void convertGroupsToRichGroupsWithAttributesWithListOfNamesTest() throws
...
+	}
+
```
zlamalp: Hi, there should be a ``@Test`` annotation above the method, otherwise it's not started during testing.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -1450,6 +1451,61 @@ public void convertGroupsToRichGroupsWithAttributesWithListOfNamesTest() throws
...
+		synchroAttr.setValue("false");
+		perun.getAttributesManager().setAttribute(sess, group, synchroAttr);
+
+		Member member = setUpMember(vo);
+		groupsManager.addMember(sess, group, member);
+		List<Member> members = groupsManager.getGroupMembers(sess, group);
+		assertTrue("List of members should contain member", members.contains(member));
+		
+		groupsManager.removeMember(sess, group, member);
+		members.clear();
```
zlamalp: You don't have to ``clear()`` list variable here since it's re-assigned on the next line -> all list content is replaced by the new assignment and old content is discarded by java garbage collector (if not referenced from anywhere else).

zwejra: Yeah, thank you. I have changed it.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
zwejra: Fixed.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -1450,6 +1451,61 @@ public void convertGroupsToRichGroupsWithAttributesWithListOfNamesTest() throws
...
 	}
 
+	@Test(expected = ExternalyManagedException.class)
+	public void addMemberToSynchronizedGroup() throws Exception {
+		System.out.println(CLASS_NAME + "addMemberToSynchronizedGroup");
+
+		vo = setUpVo();
+		setUpGroup(vo);
+
+		Attribute synchroAttr = new Attribute(perun.getAttributesManager().getAttributeDefinition(sess, "urn:perun:group:attribute-def:def:synchronizationEnabled"));
```
zlamalp: When passing attribute name, it would be better to use constant ``GroupsManager.GROUPSYNCHROENABLED_ATTRNAME``  (as you do in entry logic code) to prevent breaking changes in a future.

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ExternalyManagedException.java
```diff
@@ -0,0 +1,23 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+/**
+ * This exception raises when group is externally managed.
+ *
+ * @author Jan Zvěřina <zverina.jan@email.cz>
+ */
+public class ExternalyManagedException extends PerunException {
```
zlamalp: Externaly -> Externally

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -225,16 +226,25 @@ public Group getGroupByName(PerunSession sess, Vo vo, String name) throws GroupN
...
 		Utils.checkPerunSession(sess);
 		getGroupsManagerBl().checkGroupExists(sess, group);
 		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
+		Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+
+		if (attrSynchronizeEnabled.getValue() != null) {
+			String synchronizeEnabled = (String) attrSynchronizeEnabled.getValue();
+			// Check if group is synchronized
+			if (synchronizeEnabled.equals("true")) {
+				throw new ExternalyManagedException("Adding of member is not allowed. Group is externaly managed.");
```
zlamalp: Externaly -> Externally in a exception message.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -244,16 +254,25 @@ public void addMember(PerunSession sess, Group group, Member member) throws Inte
...
 		Utils.checkPerunSession(sess);
 		getGroupsManagerBl().checkGroupExists(sess, group);
 		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
+		Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+
+		if (attrSynchronizeEnabled.getValue() != null) {
+			String synchronizeEnabled = (String) attrSynchronizeEnabled.getValue();
+			// Check if group is synchronized
+			if (synchronizeEnabled.equals("true")) {
+				throw new ExternalyManagedException("Removing of member is not allowed. Group is externaly managed.");
```
zlamalp: Externaly -> Externally in a exception message.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -244,16 +254,25 @@ public void addMember(PerunSession sess, Group group, Member member) throws Inte
 		getGroupsManagerBl().addMember(sess, group, member);
 	}
 
-	public void removeMember(PerunSession sess, Group group, Member member) throws InternalErrorException, MemberNotExistsException, NotGroupMemberException, PrivilegeException, GroupNotExistsException {
+	public void removeMember(PerunSession sess, Group group, Member member) throws InternalErrorException, MemberNotExistsException, NotGroupMemberException, PrivilegeException, GroupNotExistsException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternalyManagedException {
 		Utils.checkPerunSession(sess);
 		getGroupsManagerBl().checkGroupExists(sess, group);
 		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
+		Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
```
zlamalp: It's probably better to put this block of code after authorization block (``AuthzResolver.isAuthorized(...``). Right now, SQL select is performed everytime, even if user is not authorized to call this method. Switching it can save up this unnecessary select. Same goes for ``addMember()`` method.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -244,16 +254,25 @@ public void addMember(PerunSession sess, Group group, Member member) throws Inte
...
+
+		if (attrSynchronizeEnabled.getValue() != null) {
+			String synchronizeEnabled = (String) attrSynchronizeEnabled.getValue();
+			// Check if group is synchronized
+			if (synchronizeEnabled.equals("true")) {
```
zlamalp: This can be very nicely simplified to:
```java
if (Objects.equals("true", (String) attrSynchronizeEnabled.getValue()) {
 throw new ExternalyManagedException(...)
}
```
same for ``addMember()`` method.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
zwejra: Changed. Next time I will think about it more.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
zwejra: Changed.

zlamalp:
Hi, it looks great. I'm merging it. Thanks for your work.

I leave for @licehammer consideration, that we could also try/catch attribute exceptions ``WrongAttributeAssignmentException, AttributeNotExistsException``.

Right now, not all attributes required by core functionality of group sync are created on empty Perun instance. So there are two possibilities. To cover it up and let people manage group members or throw exception always, which prevents them to manage any group, until it's fixed by Perun admin.

PR #1229, opened 2016-07-20
Add method resolveVirtualAttributeValueChange for attribute isBanned

stavamichal:
- if ban is set or remove, we need to immediately react on this change

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
@@ -3,17 +3,28 @@ package cz.metacentrum.perun.core.impl.modules.attributes;
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.BanOnFacility;
+import cz.metacentrum.perun.core.api.BanOnResource;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.Resource;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
```diff
@@ -62,6 +149,29 @@ public AttributeDefinition getAttributeDefinition() {
...
+	 * @param perunSession
+	 * @param user affected user
+	 * @param facility affected facility
+	 * @return list of pairs of REsource and Members affected by Ban
+	 * @throws InternalErrorException
+	 */
+	private List<Pair<Resource, Member>> getAffectedMemberResourceObjects(PerunSessionImpl perunSession, User user, Facility facility) throws InternalErrorException {
+		List<Pair<Resource, Member>> listOfAffectedObjects = new ArrayList<>();
+		List<Resource> assignedResources = perunSession.getPerunBl().getFacilitiesManagerBl().getAssignedResources(perunSession, facility);
+		List<Member> userMembers = perunSession.getPerunBl().getMembersManagerBl().getMembersByUser(perunSession, user);
```
zlamalp: You don't have to retrieve all user members, since Member contains ``userId``, which you can check against assigned members for each resource.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
```diff
@@ -52,6 +72,73 @@ public Attribute getAttributeValue(PerunSessionImpl sess, Resource resource, Mem
 
     }
 
+	@Override
+	public List<String> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, String message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<String> resolvingMessages = new ArrayList<String>();
+		if(message == null) return resolvingMessages;
+
+		//getPerunBl().getAuditer().log(sess, "{} removed for {}", attribute, facility);
```
zlamalp: Any reason to keep these commented lines ?

PR #1235, opened 2016-07-25
Fix security bug in adding hosts

stavamichal:
- first check, then add (not in bad order)


licehammer:
Merged locally with more verbose commit message.

PR #1240, opened 2016-07-28
fixed bug while parsing names in voot extSource

Laliska:


licehammer:
Could you describe in commit message what was the bug about?

Laliska:
@licehammer I changed commit message and added one more commit, that includes eppn recognition in JSON obtained from voot external source. Branch is up to date with master now as well.

PR #1246, opened 2016-08-03
Checkup the name of service when the service is created

balcirakpeter:
- Added check whether the name of service fulfills defined pattern when the service is created
- Defined pattern "^[a-zA-Z0-9_]+$"


zlamalp:
test this please. Hi, this is a notice for Jenkins to run automatic tests. From know on, it should work automatically.

zlamalp:
Hi, some tests are failing for module ``perun-tasks-lib`` in ``ExecServiceDaoTest.setUp()`` since there is a probably now not allowed service name used in a test. Please fix it. Thank you.

zlamalp:
Merged locally with fix to the test.

PR #1253, opened 2016-08-08
2958 Restriction to create resource with existing name in same Facili…

Natrezim:
…ty in same VO.

- added new exception ResourceExistsException

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ResourceExistsException.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ResourceExistsException.java
@@ -1,5 +1,7 @@
 package cz.metacentrum.perun.core.api.exceptions;
 
+import cz.metacentrum.perun.core.api.Resource;
+
 /**
  * Checked version of ResourceExistsException.
  *
@@ -9,6 +11,8 @@ public class ResourceExistsException extends PerunException {
 
 	private static final long serialVersionUID = -255958501797585251L;
 
+	private Resource resource;
+	
 	public ResourceExistsException(String message) {
 		super(message);
 	}
@@ -19,5 +23,14 @@ public class ResourceExistsException extends PerunException {
 
 	public ResourceExistsException(Throwable cause) {
 		super(cause);
+	}
+
+	public ResourceExistsException(Resource resource) {
+		super("Resource with name \"" + resource.getName() + "\" already exists in Facility and Vo.");
+		this.resource = resource;
+	}
+
+	public Resource getResource() {
+		return resource;
 	}
 }
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ResourceExistsException.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ResourceExistsException.java
@@ -1,0 +1,23 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+/**
+ * Checked version of ResourceExistsException.
+ *
+ * @author Oliver Mrázik
+ */
+public class ResourceExistsException extends PerunException {
+
+	private static final long serialVersionUID = -255958501797585251L;
+
+	public ResourceExistsException(String message) {
+		super(message);
+	}
+
+	public ResourceExistsException(String message, Throwable cause) {
+		super(message, cause);
+	}
+
+	public ResourceExistsException(Throwable cause) {
+		super(cause);
+	}
+}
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -78,9 +79,16 @@ public Resource getResourceByName(PerunSession sess, Vo vo, Facility facility, S
...
-	public Resource createResource(PerunSession sess, Resource resource, Vo vo, Facility facility) throws InternalErrorException, FacilityNotExistsException {
-		resource = getResourcesManagerImpl().createResource(sess, vo, resource, facility);
-		getPerunBl().getAuditer().log(sess, "{} created.", resource);
+	public Resource createResource(PerunSession sess, Resource resource, Vo vo, Facility facility) throws InternalErrorException, FacilityNotExistsException, ResourceExistsException {
+		try{
+			getResourcesManagerImpl().getResourceByName(sess, vo, facility, resource.getName());
+			throw new ResourceExistsException("Resource with name \"" + resource.getName() + "\" already exists in " +
+					"facility " + facility + " in VO " + vo + ".");
+		} catch (ResourceNotExistsException e) {
+			resource = getResourcesManagerImpl().createResource(sess, vo, resource, facility);
```
Natrezim: @licehammer It is ok to do it this way? (I mean doing the logic in catch block after the exception is thrown) I don't like it but it seemed to me as the easiest way, because the method getResourceByName already existed, but it throws exception when resource with that name not exists.

licehammer: You can use "resourceExists" method which works in the same way as checkResourceExists, but it returns boolean instead of raising an exception.

Natrezim: But the "resourceExists" method take only Resource as a parameter. I need to check if the method with the same name exists in facility and in VO. And the method checks only the id of a resource. Am I missing something?

zlamalp: I believe your processing is OK, since there is no other method for checking all params and we probably cant rely on DB constraint because we might have multiple same Resources already in DB. Just please fix trailing whitespace.

zlamalp:
Please add same check to the bl layer of ``updateResource()`` since we can now freely change name to something that already exists.

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ResourceExistsException.java
```diff
@@ -0,0 +1,23 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+/**
+ * Checked version of ResourceExistsException.
+ *
+ * @author Oliver Mrázik
+ */
+public class ResourceExistsException extends PerunException {
```
zlamalp: It would be nice, if this exception could contain Resource param (set from constructor and with getter method). It would allow me to print better exception text in GUI. When necessary fill this with Resource object in conflict (already existing) so use can see, which resource is already there.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -885,7 +905,7 @@ public void updateResource() throws Exception {
 		facility = setUpFacility();
 		resource = setUpResource();
 
-		Resource resourceToUpdate = resourcesManager.createResource(sess, resource, vo, facility);
```
zlamalp: I not sure, if this test does anything useful (wasn't before either). It should create 2 resources, which are different, then change one of them and see, it property was really updated. Also now it should check (maybe in another test method) that you can't update resource name to the one, that already exists. Update method itself returns updated object and it should match getById like method results. Can you please rewrite this test. Thank you.

Natrezim: I do not understand why I need 2 resources in this test? I created one and modified it, then checked returned resource with getResourceById result, but I do not see a reason for second resource.

Natrezim:
Also I had to add additional check into update method, because if I was updating something other than name it throw ResourceAlreadyExistsException because the resource with the same name already exists (because I updating it.) So the code looks a little bit worse.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
zlamalp: Well, you are right, that it's not needed for simple update use-case. But you added restriction, that two resources can't have same name and you don't test it here or in another test. So you actually need two resources to check new behavior of update method.

Please remove assigning ``Resource resourceToUpdate = resource;`` since they are the same object after that. You can use ``resource`` in all code to prevent confusion (setName() on resourceToUpdate but setDescription() on resource). Thank you.

PR #1255, opened 2016-08-09
APPS/facility-overview: XSS protection

ondrejvelisek:
html encodes Facility name, Owner names, Facility descriptions and host names. So no html tags (especially <script>...</script>) are executed.


ondrejvelisek:
I've generalized solution. Now all text cells in tables encodes its values. 
I've also checked all apps and didn't find any other xss vulnerability outside tables.

PR #1121, opened 2016-03-14
Added redmineRole attribute

hrasnam:

PR #1259, opened 2016-08-11
New attribute module for redmine role

balcirakpeter:
- Created attribute module to controle value for defined regular expression
- Fixed: erased double space and changed error message in case of null attribute


balcirakpeter:
Please close pull request #1121

zlamalp:
Is there anything preventing merging this attribute module for redmine @licehammer @tauceti2 ? I believe @zwejra will need it for a redmine service. Code seems to be ok by me.

PR #1266, opened 2016-08-15
Postgres compatibility optimization of sql rollbacking

stavamichal:
This pull request contains:
 - Fixed: Refactoring of the setAttribute method in impl layer #1109
 - Nested transactions on all update (not delete) and insert db calls on core.impl layer


zlamalp:
Merged locally by @licehammer yesterday, closing this one.

PR #1275, opened 2016-08-17
Tests for methods addHost()/addHosts() and addDestination() were made.

zwejra:
Tests for addHost(s) are in FacilitiesManagerEntryIntegrationTest.java.
Tests for addDestination() are in ServicesManagerEntryIntegrationTest.java.
Some tests for addHost(s) covers tests for addDestination() as well, so I did not do them again.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntryIntegrationTest.java
@@ -33,11 +33,14 @@ import cz.metacentrum.perun.core.api.Host;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.Owner;
 import cz.metacentrum.perun.core.api.OwnerType;
+import cz.metacentrum.perun.core.api.PerunBean;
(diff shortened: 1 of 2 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntryIntegrationTest.java
```diff
@@ -1655,6 +1660,291 @@ public void removeExpiredBansIfNotExist() throws Exception {
 		assertTrue(bansOnFacility.size() == 1);
 	}
 
+	@Test
+	public void addHostAndDestinationSameNameSameAdmin() throws Exception {
+		System.out.println(CLASS_NAME + "addHostAndDestinationSameNameSameAdmin");
+		
+		// Initialize host, destination and service
+	 	String hostName = "TestHost";
+		Host hostOne = new Host(999, hostName);
```
zlamalp: Hi, can you please replace all 999 and 998 ids to 0? During create, we forcefully set object ID to the one retrieved from the sequence, but in case of update call, it could cause problems when running tests against live DB with actual data. Also 0 is a better indication, that object is not initialized (we can't have id=0 in perun DB). After this small fix it can be merged. Thank you.

zlamalp:
Merged locally with required change. Thanks.

PR #1279, opened 2016-08-18
#3008 Create method groupsManager.getGroupUnions()

Natrezim:
- removed duplicit method (getRelatedGroupsIds) which returns all result groups ids which have relation with given group.
- now there are two new methods on Impl which returns result groups or operand groups of given group
- created new method in Bl and Entry (getGroupUnions) which returns operand or result groups based on reverseDirection parameter
- test for new method

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
@@ -884,7 +884,7 @@ public interface GroupsManager {
 	 * @return list of groups.
 	 *
 	 * @throws GroupNotExistsException
-	 * @throws InternalErrorException when relation does not exist
+	 * @throws InternalErrorException
 	 * @throws PrivilegeException
 	 */
 	List<Group> getGroupUnions(PerunSession sess, Group group, boolean reverseDirection) throws InternalErrorException, GroupNotExistsException, PrivilegeException;
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
@@ -874,4 +874,18 @@ public interface GroupsManager {
 	 * @throws PrivilegeException
 	 */
 	void removeGroupUnion(PerunSession sess, Group resultGroup, Group operandGroup) throws InternalErrorException, GroupNotExistsException, PrivilegeException, GroupOperationsException;
+
+	/**
+	 * Get list of group unions for specified group.
+	 * @param sess perun session
+	 * @param group group
+	 * @param reverseDirection if false get all operand groups of requested result group
+	 *                         if true get all result groups of requested operand group   
+	 * @return list of groups.
+	 *
+	 * @throws GroupNotExistsException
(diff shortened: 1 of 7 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
```diff
@@ -874,4 +874,18 @@
...
+	/**
+	 * Get list of group unions for specified group.
+	 * @param sess perun session
+	 * @param group group
+	 * @param reverseDirection if false get all operand groups of requested result group
+	 *                         if true get all result groups of requested operand group   
+	 * @return list of groups.
+	 *
+	 * @throws GroupNotExistsException
+	 * @throws InternalErrorException when relation does not exist
```
zlamalp: Explanation of InternalErrorException here is false. Method returns empty list if no relation for group is found and not the exception.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/GroupsManagerImplApi.java
```diff
@@ -570,13 +559,22 @@
...
+	 * Return all result groups of requested operand group.
 	 *
 	 * @param sess perun session
 	 * @param groupId group id
 	 * @return list of group ids
 	 * 
 	 * @throws cz.metacentrum.perun.core.api.exceptions.InternalErrorException
 	 */
-	List<Integer> getGroupRelations(PerunSession sess, int groupId) throws InternalErrorException;
+	List<Integer> getResultGroups(PerunSession sess, int groupId) throws InternalErrorException;
```
zlamalp: Honestly I'm confused about the original method names. We had ``getGroupRelations()`` and ``getRelatedGroupsIds()`` which both returned list of result groups ids by operand gid.
I believe first should return Groups and second Ints. But they both returned ints. This was probably wrong in original implementation and is reason we didn't have getter method on relations.

So, right now it seems to be ok to use ``getResultGroups()`` and ``getOperandGroups()``but I was hoping for a single joining select for groups, instead of using ``getGroupsByIds()`` in BL layer. Can you please bring back method ``getResultGroupsIds()`` (changed name!) which can be used for simple checks in member handling methods ? But please without ``operand_gid=" + groupId`` like constructs in calling db from java (sql injection). And than convert new methods to return List<Group> from Impl. Thank you.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -651,6 +651,26 @@ public void transitiveGroupMembershipCheck() throws Exception {
 
 	}
 
+	@Test
+	public void getGroupUnions() throws Exception {
```
zlamalp: We have also tests: ``createGroupUnion()`` and ``removeGroupUnion()``. They for now check results only by members. Can you please add call ``groupsManagerBl.getGroupUnions()`` and check if they are ok by group relations too. Thank you. Also it would be nice, if they check ``membershipType`` and ``sourceGroupId`` to be the correct one. Not sure if we test it anywhere (I do in a one test for ``membershipType``, but group are related by hierarchy there, not by weak relation (inclusion)).

PR #1295, opened 2016-08-26
Core: setAttribute methods now handle rich objects + exceptions

jirmauritz:
- setAttribute methods now handle rich objects 
- exceptions wrapped by InternalErrorException


zlamalp:
Tested on devel by group sync. Merging.

PR #1299, opened 2016-09-01
Created gui for group unions

m-krajcovic:
It is accessible through Group Manager under Relations tab. You can view/add/remove group unions


zlamalp:
Merged locally with some fixes and changes. Thank you.

PR #1322, opened 2016-09-16
Hack for GID range in EINFRA

stavamichal:
- there are 10 groups with BAD GID not in range of EINFRA Namespace so
   we need to skip checking for these artefact gids
 - change is in method checkIfGIDIsWithinRange just for group_gid and
   for namespace einfra


licehammer:
We don't want to merge this to master. It is used only as a temporary hack.

PR #1324, opened 2016-09-16
Added support for group mail footers

m-krajcovic:


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -1836,9 +1837,14 @@ private String substituteCommonStrings(Application app, List<ApplicationFormItem
 			String footer = "";
 			// get proper value from attribute
 			try {
-				Attribute attrFooter = attrManager.getAttribute(registrarSession, app.getVo(), URN_VO_MAIL_FOOTER);
-				if (attrFooter != null && attrFooter.getValue() != null) {
-					footer = BeansUtils.attributeValueToString(attrFooter);
+				Attribute attrVoFooter = attrManager.getAttribute(registrarSession, app.getVo(), URN_VO_MAIL_FOOTER);
+				if (attrVoFooter != null && attrVoFooter.getValue() != null) {
```
zlamalp: Hi, this imply, that vo mail footer must exists, which is not true. Group footer can be defined independently on vo footer. If it's group application, we should try to fill group attr, if not available we should try vo attr. If it's vo application, we should try to fill vo attr. Thanks.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/EditMailFooterTabItem.java
```diff
@@ -224,11 +242,10 @@ public void open()
 	}
 
 	public boolean isAuthorized() {
-
 		if (session.isVoAdmin(voId)) {
 			return true;
 		} else {
-			return false;
+			return group != null && session.isGroupAdmin(groupId);
```
zlamalp: Hi, since existing object ids start with 1 and you define voId/groupId=0 by default, then you can safely use something like: ``session.isVoAdmin(voId) || session.isGroupAdmin(groupId)`` in the first IF.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
m-krajcovic: i fixed this to check for group mail footer first, it should be ok now

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -1836,9 +1837,14 @@ private String substituteCommonStrings(Application app, List<ApplicationFormItem
 			String footer = "";
 			// get proper value from attribute
 			try {
-				Attribute attrFooter = attrManager.getAttribute(registrarSession, app.getVo(), URN_VO_MAIL_FOOTER);
-				if (attrFooter != null && attrFooter.getValue() != null) {
-					footer = BeansUtils.attributeValueToString(attrFooter);
+				Attribute attrGroupFooter = attrManager.getAttribute(registrarSession, app.getGroup(), URN_GROUP_MAIL_FOOTER);
```
zlamalp: Hi, before getting group attribute you should check if application has group in itself ``app.getGroup() != null``. Right now it would fail for vo notifications, since it would call getAttribute() with null object.
IF group -> fill group footer -> if empty, fill vo footer
IF vo -> fill vo footer

m-krajcovic: this should be finally fixed right

zlamalp:
Merged locally with fix to vo/group attr friendly name reference.

PR #1325, opened 2016-09-20
update engine ID in database when assigned in dispatcher

mvocu:
* adds a method to TaskManager (and TaskDao) to update engine id in database
* updates the engine id database column when calling setQueueForTask() in dispatcher

### DIFF
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/SchedulingPoolImpl.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/SchedulingPoolImpl.java
@@ -293,6 +293,7 @@ public class SchedulingPoolImpl implements SchedulingPool {
 		} else {
 			tasksById.get(task.getId()).put(task, queueForTask);
 		}
+		taskManager.updateTaskEngine(task, queueForTask.getClientID());
 	}
 
 	@Override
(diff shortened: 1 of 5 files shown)


perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/SchedulingPoolImpl.java
```diff
@@ -293,6 +293,7 @@ public void setQueueForTask(Task task, DispatcherQueue queueForTask) throws Inte
 		} else {
 			tasksById.get(task.getId()).put(task, queueForTask);
 		}
+		taskManager.updateTaskEngine(task, queueForTask.getClientID());
```
zlamalp: Hi, I don't like the fact, that this method doesn't handle runtime exceptions from DB. I believe we generally want to recover from most actions. If action fails here, it bubbles up, but nothing (except ``InternallErrorException``) is catched and logged in ``scheduleTask()``, ``sendToEngine()``, etc., which will probably stop upper logic unexpectedly. For example I believe, that force propagation would just stop without any proper log (``EventProcessorImpl``): 
```java
taskExecutor.execute(new Runnable() {
	@Override
	public void run() {
		log.debug("  Force scheduling the task.");
		taskScheduler.scheduleTask(task_final);
	}
});
```
I think this is a general problem in ``perun-tasks-lib`` module, which should be fixed. But for now, just add try/catch for Exception or RuntimeException to ``updateTaskEngine()`` in ``TaskDaoJdbc`` and re-throw it as InternalErrorException, since it's properly catched in upper logic of dispatcher. Thanks.

mvocu:
Exception in that single method are now transformed into InternalErrorException.

PR #1336, opened 2016-09-30
 Increase length for login from eduPersonTargetedID/NameID

zoraseb:
- changed column length for ext_login in user_ext_sources
- changed columns length in all tables containing columns created_by, createdBy and modified_by


zlamalp:
Hi @licehammer, we will need to bump all created_by/modified_by columns too, as noted in redmine issue 3068. Change is probably not yet pushed by @zoraseb. Then DB version will change and pull-request/commit messages will be cleaned up.

licehammer:
Agreed. Let's wait until this pull request is complete.

PR #1337, opened 2016-09-30
Method for user:virt:kerberosLogins

balcirakpeter:
resolveVirtualAtributeValueChange in urn_perun_user_attribute_def_virt_kerberosLogins added
Method generates message when UserExtSourceKerberos is added or removed

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_kerberosLogins.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_kerberosLogins.java
@@ -1,20 +1,19 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
-import cz.metacentrum.perun.core.api.Attribute;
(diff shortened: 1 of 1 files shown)


balcirakpeter:
Related to pull request in CESNET/perun-services/Gen scripts switched to user:virt:kerberosLogins

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_kerberosLogins.java
```diff
@@ -77,4 +105,27 @@ public AttributeDefinition getAttributeDefinition() {
 		attr.setDescription("Logins in kerberos (including realm and kerberos UserExtSources)");
 		return attr;
 	}
+
+	/**
+	 * Get User from message if exists or if there is only one. In other case return null instead.
```
zlamalp: Please fix javadoc to: "Get User from message if exists AND there is only one".

balcirakpeter: Fixed

PR #1348, opened 2016-10-12
OIDC: Structured claims support in config.

ondrejvelisek:
Config structure of file perun-oidc-scopes.properties has changed. It uses JSON now.

Config file has to be replaced when change is deployed. Example file is n devel  perun-oidc-scopes.properties.new

### DIFF
--- a/perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
+++ b/perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
@@ -11,13 +11,19 @@ import cz.metacentrum.perun.core.api.exceptions.RpcException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.rpc.deserializer.Deserializer;
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.map.ObjectMapper;
+import org.codehaus.jackson.node.NullNode;
+import org.codehaus.jackson.node.ObjectNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Arrays;
+import java.io.IOException;
 import java.util.Collection;
-import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 
 /**
@@ -29,14 +35,18 @@ public class OIDC {
 
 	private final String USERINFO_METHOD = "userinfo";
 	private final String USERINFO_FOR_USER_METHOD = "getSpecificUserinfo";
-	private final String PROPERTIES_FILE = "perun-oidc-scopes.properties";
+	private final String CONFIG_FILE = "perun-oidc-scopes.properties";
+
+	// if config value matches this regex it is consider as perun attribute and is replaced by real value.
+	private Pattern attrRegex = Pattern.compile("^urn:perun:.*");
+
 
 	public Object process(PerunSession sess, String method, Deserializer params) throws InternalErrorException, WrongAttributeAssignmentException, UserNotExistsException, PrivilegeException {
 
 		if (USERINFO_METHOD.equals(method)) {
 
 			User user = sess.getPerunPrincipal().getUser();
-			Map<String, String> properties = BeansUtils.getAllPropertiesFromCustomConfiguration(PROPERTIES_FILE);
+			Map<String, String> properties = BeansUtils.getAllPropertiesFromCustomConfiguration(CONFIG_FILE);
 
 			if (sess.getPerunClient().getScopes().contains(PerunClient.SCOPE_ALL)) {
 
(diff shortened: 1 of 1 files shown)


ondrejvelisek:
Change regex for attrbute to `urn:perun:.*`
Remove empty lines

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
```diff
@@ -70,44 +80,88 @@ public Object process(PerunSession sess, String method, Deserializer params) thr
...
+		}
+
+		return userinfo;
+
+	}
 
-					try {
-						userinfo.put(claimName, sess.getPerun().getAttributesManager().getAttribute(sess, user, attrName).getValue());
 
+	private JsonNode dfs(PerunSession sess, User user, JsonNode node) throws InternalErrorException, WrongAttributeAssignmentException, PrivilegeException, UserNotExistsException {
```
licehammer: Please, rename this method.

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
```diff
@@ -70,44 +80,88 @@ public Object process(PerunSession sess, String method, Deserializer params) thr
...
+			if (node.isTextual()) {
+				Matcher m = attrRegex.matcher(node.getTextValue());
+				if (m.matches()) {
+
+					try {
+						Object value = sess.getPerun().getAttributesManager().getAttribute(sess, user, node.getTextValue()).getValue();
+						if (value != null) {
+							return new TextNode(value.toString());
+						} else {
+							log.info("Attribute "+node.getTextValue()+" defined in config "+CONFIG_FILE+" is not present for user "+user);
```
licehammer: How can you be sure, that the problem is in CONFIG_FILE?

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
```diff
@@ -29,14 +35,18 @@
 
 	private final String USERINFO_METHOD = "userinfo";
 	private final String USERINFO_FOR_USER_METHOD = "getSpecificUserinfo";
-	private final String PROPERTIES_FILE = "perun-oidc-scopes.properties";
+	private final String CONFIG_FILE = "perun-oidc-scopes.properties";
+
+	// if config value matches this regex it is consider as perun attribute and is replaced by real value.
+	private Pattern attrRegex = Pattern.compile("urn:perun:.*");
```
licehammer: "^urn:perun:.*" regex would be slightly better.

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
```diff
@@ -70,44 +80,88 @@ public Object process(PerunSession sess, String method, Deserializer params) thr
...
+		if (!node.isObject()) {
+
+			if (node.isTextual()) {
+				Matcher m = attrRegex.matcher(node.getTextValue());
+				if (m.matches()) {
+
+					try {
+						Object value = sess.getPerun().getAttributesManager().getAttribute(sess, user, node.getTextValue()).getValue();
+						if (value != null) {
+							return new TextNode(value.toString());
```
licehammer: Are you sure you want to get attribute value like this even if the value type is List or Map?

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
ondrejvelisek: I think text does not say problem is there. But attribute is configured there and is not set for the user.

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
ondrejvelisek: Well, you are right. I should convert it to JSON.

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
ondrejvelisek: OK

ondrejvelisek:
I would say 'i' for Iterator is quite standard, isn't it?

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
licehammer: I tried to point out that log msg claims it get something form CONFIG_FILE, but in reality data are obtained from argument of the method (therefore they don't have to originate in CONFIG_FILE).

licehammer:
> I would say 'i' for Iterator is quite standard, isn't it?

Not in java :-)

To be honest, I don't mind if "cryptic name" is used in very small piece of code (e.g. one-liner). If you have it in more complex code or if you start using such names for more variable it's gets messy. 

For example if you see line like this:
```
String res = someMethod(i.next().getValue())
```
You have no idea what it is about unless you backtrack in the code to find out meaning of variables.

perun-oidc/src/main/java/cz/metacentrum/perun/oidc/OIDC.java
ondrejvelisek: You are right.

ondrejvelisek:
Ok. I think all of your change requests are in place now.

PR #1349, opened 2016-10-12
#3071 - Assigning services on facilities

Natrezim:
- added new DB table facility_services
- supporting methods across all 3 layers
- updated Resources "assignService" method to check if assigning service is already on facility
- updated tests for reflecting these changes
- database updated to version 3.1.39


perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -160,7 +169,7 @@ create index idx_fk_res_ban_member_res on resources_bans (member_id, resource_id
 create index idx_fk_fac_ban_user on facilities_bans (user_id);
 create index idx_fk_fac_ban_fac on facilities_bans (facility_id);
 create index idx_fk_fac_ban_user_fac on facilities_bans (user_id, facility_id);
-grant all on resources_bans to perun;
+grant all on resources_bans to perun;
```
zlamalp: Please remove trailing whitespace.

perun-core/src/main/resources/schema.sql
```diff
@@ -164,6 +164,18 @@ create table facility_contacts (
 	group_id integer
 );
 
+create table facility_services (
```
zlamalp: There is also ```test-data.sql``` file, where you need to update DB version in order for tests to run correctly.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/FacilitiesManagerImpl.java
```diff
@@ -1022,6 +1052,18 @@ public void removeSecurityTeam(PerunSession sess, Facility facility, SecurityTea
 	}
 
 	@Override
+	public void removeService(PerunSession sess, Facility facility, Service service) throws InternalErrorException, ServiceNotExistsException {
+		try {
+			if (0 == jdbc.update("DELETE FROM facility_services WHERE service_id = ? AND facility_id = ?", 
+					service.getId(), facility.getId())) {
+				throw new ServiceNotExistsException(service);
```
zlamalp: I believe, that when sql delete returns "0 rows were deleted", then the case is most probably ```ServiceNotAssignedException``` and not ```ServiceNotExistsException```. Also can be caused by the fact, that service or facility with passed ID doesn't exists, but it should be checked using own select - there are probably already existing methods for checking service/facility existence by id. Please fix javadoc for this too, your original exception is missing there. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -911,18 +917,82 @@ public void checkFacilityContactExists(PerunSession sess, Facility facility, Str
 	}
 
 	@Override
+	public List<Service> getAssignedServices(PerunSession sess, Facility facility) throws InternalErrorException {
```
zlamalp: There is no reason for this complex logic in Bl. You should always generally prevent iterating over a list of ids and getting each item by its id. Just select list of Services directly in impl layer using sql join with services table, like we do for Resource. We already have service mapper and other necessary stuff for it.

Natrezim: Yes, I agree, I just blindly copied it from ResourceManager getAssignedResources. I changed it there too.

Natrezim:
@zlamalp I have a problem with 2 test in Perun-engine. runGenTest and sendGenTest. Both are failing, but I can't find a connection to my changes... Am I missing something?

zlamalp:
@Natrezim If you can repeat test faliure before and after your change, than leave fixing it to @mvocu.

PR #1354, opened 2016-10-14
CLI tool which removes selected hosts

zoraseb:
CLI tool which removes selected hosts (optionally destinations) from facilities.

### DIFF
--- a/perun-rpc/src/main/perl/removeHostsFromAllFacilities
+++ b/perun-rpc/src/main/perl/removeHostsFromAllFacilities
@@ -1,0 +1,90 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Getopt::Long qw(:config no_ignore_case);
+use Perun::Agent;
+use Perun::Common qw(printMessage);
+
+sub help {
+	return qq{
+	Removes selected hosts (optionally destinations) from facilities. One or more hostnames is required field.
+	------------------------------------
+	Available options:
+	--hosts        | -H list Of Host names
+	--dest         | -d remove appropriated destinations
+	--batch        | -b batch
+	--help         | -h prints this help
+
+	};
+}
+
+my (@hosts, $dest, $type, $batch);
+GetOptions ("help|h" => sub { print help(); exit 0;} ,
+	"batch|b" => \$batch,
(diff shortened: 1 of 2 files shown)


perun-rpc/src/main/perl/removeHostsFromAllFacilities
```diff
@@ -0,0 +1,74 @@
...
+
+use strict;
+use warnings;
+use Getopt::Long qw(:config no_ignore_case);
+use Perun::Agent;
+use Perun::Common qw(printMessage);
+
+sub help {
+	return qq{
+	Removes selected hosts (optional and destinations) from facilities. One or more hostnames is required field.
```
zlamalp: (optional and destinations) -> (optionally destinations)

perun-rpc/src/main/perl/removeHostsFromAllFacilities
```diff
@@ -0,0 +1,74 @@
...
+	------------------------------------
+	Available options:
+	--hosts        | -H list Of Host names
+        --dest         | -d remove appropriated destinations
```
zlamalp: Please fix indentation (space -> tab).

perun-rpc/src/main/perl/removeHostsFromAllFacilities
```diff
@@ -0,0 +1,74 @@
...
+    # zruseni destinations
+    if (defined $dest) {
+        @facilities = $facilitiesAgent->getFacilitiesByDestination(destination => $host);
+        unless (@facilities) { printMessage "No facility found for destination $host", $batch; exit 0;}
+        while (@facilities) {
+            my $fac=shift(@facilities);
+            my $facilityId=$fac->getId;
+            my @services = $servicesAgent->getAssignedServices(facility => $facilityId);
+            my $service;
+            my $typ="host";
```
zlamalp: Are we sure, that ``$typ="host";`` will be always only "host" ( @licehammer )? We have also ``"user@host", "user@host:port", "service-specific"`` and other types. Does a tool caller have a knowledge of exact destination type ? I guess it should be either passed on input or based on destination retrieved from facility destinations by name and then deleted using original object.

Also please fix Czech variable name to English (including some comments in script). Thank you.

PR #1355, opened 2016-10-15
#3071 - Assigning services on facilities

Natrezim:
- added new DB table facility_services
- supporting methods across all 3 layers
- updated Resources "assignService" method to check if assigning service is already on facility
- updated all tests for reflecting these changes
- database updated to version 3.1.40

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
@@ -8,6 +8,12 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import cz.metacentrum.perun.core.api.ServicesPackage;
+import cz.metacentrum.perun.core.api.exceptions.ServiceAlreadyAssignedException;
+import cz.metacentrum.perun.core.api.exceptions.ServiceNotAssignedException;
+import cz.metacentrum.perun.core.api.exceptions.ServiceNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ServicesPackageNotExistsException;
+import cz.metacentrum.perun.core.bl.AttributesManagerBl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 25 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -911,18 +917,67 @@ public void checkFacilityContactExists(PerunSession sess, Facility facility, Str
...
 
 	@Override
+	public void assignService(PerunSession sess, Facility facility, Service service) throws InternalErrorException, ServiceAlreadyAssignedException {
+		facilitiesManagerImpl.assignService(sess, facility, service);
+		getPerunBl().getAuditer().log(sess, "{} assigned to {}", service, facility);
+
+		AttributesManagerBl attributesManagerBl = getPerunBl().getAttributesManagerBl();
+		// call check of facility's resource's member's user's attributes
+		try {
+			attributesManagerBl.checkAttributesValue(sess, facility, attributesManagerBl.getRequiredAttributes(sess, facility));
```
zlamalp: This seems to be correct from consistency point of view, but I'm afraid, that we could end up with service, which will never be assigned due to wrong value in attribute dependency with objects already connected with facility (user-member-resource). Some services shares attributes, also it's highly dependent on attribute modules check implementation. What do you think @licehammer ? @stavamichal suggested we use service connection on facility only to limit, what services can be configured on its resources. Usually until there is a resource and some groups assigned, there are no data to be pushed to service destination.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/FacilitiesManagerImpl.java
```diff
@@ -1000,6 +1002,20 @@ public void removeFacilityContact(PerunSession sess, Facility facility, String n
 		}
 	}
 
+//	return jdbc.query("select " + ResourcesManagerImpl.resourceMappingSelectQuery + " from resources where facility_id=?",
+//	ResourcesManagerImpl.RESOURCE_MAPPER, facility.getId());
```
zlamalp: I'm not sure, why is there this commented code ?

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -1150,6 +1152,23 @@ public void removeFacilityContact(PerunSession sess, ContactGroup contactGroupTo
 	}
 
 	@Override
+	public List<Service> getAssignedServices(PerunSession sess, Facility facility) throws InternalErrorException, FacilityNotExistsException, PrivilegeException {
+		Utils.checkPerunSession(sess);
+		getFacilitiesManagerBl().checkFacilityExists(sess, facility);
+		
+		// Authorization
+		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, facility)
```
zlamalp: I don't think, that VOADMIN and VOOBSERVER should be able to use this. For Resource it's ok, since they can somehow manage Resource, but facility settings belong to facility manager.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -1150,6 +1152,23 @@ public void removeFacilityContact(PerunSession sess, ContactGroup contactGroupTo
...
+				&& !AuthzResolver.isAuthorized(sess, Role.VOOBSERVER, facility)
+				&& !AuthzResolver.isAuthorized(sess, Role.ENGINE)
```
zlamalp: Regarding this, we should think, if/when/how can dispatcher/engine use this method instead of getting services from resource. Could it somehow speed up ExecService/Facility resolving from event ?This will be probably up to @mvocu.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
zlamalp: We agreed with @licehammer that check on attributes will stay in place.

zlamalp:
Please rebase to current master, we added two db versions in the mean time, so bump is necessary to 3.1.42. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
Natrezim: OK, I left only Facility manager here.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
Natrezim: OK, I left only Facility manager here.

zlamalp:
Regarding the fact, that service on resource must be on facility first, we must ensure, that on update all services from resources are assigned to their facilities. Please add this logic to DB changelog (similar like for group arithmetic). Thank you.

perun-core/src/main/resources/oracleChangelog.txt
```diff
@@ -72,7 +82,7 @@ alter table USERS modify created_by nvarchar2(1300);
 alter table USER_ATTR_VALUES modify created_by nvarchar2(1300);
 alter table USER_EXT_SOURCES modify created_by nvarchar2(1300);
 alter table USER_FACILITY_ATTR_VALUES modify created_by nvarchar2(1300);
-alter table USER_EXT_SOURCE_ATTR_VALUES modify created_by nvarchar2(1300);
+alter_table USER_EXT_SOURCE_ATTR_VALUES modify created_by nvarchar2(1300);
```
zlamalp: Please remove ``_`` in ``alter_table`` (twice). It was fixed after you rebased, so maybe rebase on master one more time or fix it manually. Thank you.

Natrezim:
@zlamalp is it enough to just assign all services on particular resource to facility based on resource.facility_id?

zlamalp:
@Natrezim Yes, you should assign all services from current facility resources to facility. You should get unique set of services at once or create conditional insert, since same service can be on more facility resources.

zlamalp:
I didn't check SQL yet, but moving data should be part of the same DB version, like schema change requiring it (after schema change, before version update). Please put it to 3.1.42 changelog. Thank you.

stavamichal:
Please, rebase this pull request to master.

stavamichal:
@Natrezim please, can you create also database change-log for setting default for existing facilities? For every service on resource need to exist such service in the new table between facility and service. Thank you.

zlamalp:
This needs rebase and than rework and implementation of GUI. Maybe we should close it (save  interesting bits) and do it again when there will be more time for proper design of UI workflows.

PR #1358, opened 2016-10-17
allow null values in tasks.engine_id column

mvocu:
* modifies DB schema to allow null values in tasks.engine_id
* modifies row mapper to map null engine_id to -1
* modifies update methods to map engine_id -1 to null

### DIFF
--- a/perun-core/src/main/resources/hsqldbChangelog.txt
+++ b/perun-core/src/main/resources/hsqldbChangelog.txt
@@ -4,6 +4,9 @@
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
 -- this update is not supported on hsql since its used only as in-memory db
+3.1.40
+alter table tasks alter column engine_id type integer;
+update configurations set value='3.1.40' where property='DATABASE VERSION';
 
 3.1.39
 alter table user_ext_sources alter column login_ext type varchar(1300);
--- a/perun-core/src/main/resources/schema.sql
+++ b/perun-core/src/main/resources/schema.sql
@@ -857,7 +857,7 @@ create table tasks (
 	status varchar(16) not null,
 	start_time timestamp,
 	end_time timestamp,
-	engine_id integer not null,
+	engine_id integer,
 	created_at timestamp default now not null,
 	err_message varchar(4000),
 	created_by_uid integer,
--- a/perun-db/oracle.sql
+++ b/perun-db/oracle.sql
@@ -1,4 +1,4 @@
--- database version 3.1.39 (don't forget to update insert statement at the end of file)
+-- database version 3.1.40 (don't forget to update insert statement at the end of file)
 
 create user perunv3 identified by password;
 grant create session to perunv3;
@@ -866,7 +866,7 @@ create table tasks (
 	status nvarchar2(16) not null,
 	start_time date,
 	end_time date,
-	engine_id integer not null,
+	engine_id integer,
 	created_at date default sysdate not null,
 	err_message nvarchar2(4000),
 	created_by_uid integer,
@@ -1800,7 +1800,7 @@ constraint UESATTRVAL_ATTR_FK foreign key (attr_id) references attr_names(id)
 );
 
 -- set initial Perun DB version
-insert into configurations values ('DATABASE VERSION','3.1.39');
+insert into configurations values ('DATABASE VERSION','3.1.40');
 
 -- insert membership types
 insert into membership_types (id, membership_type, description) values (1, 'DIRECT', 'Member is directly added into group');
--- a/perun-db/postgres.sql
+++ b/perun-db/postgres.sql
@@ -1,4 +1,4 @@
--- database version 3.1.39 (don't forget to update insert statement at the end of file)
+-- database version 3.1.40 (don't forget to update insert statement at the end of file)
 
 -- VOS - virtual organizations
 create table "vos" (
@@ -927,7 +927,7 @@ create table "tasks" (
 	status varchar(16) not null,        --state of task
 	start_time timestamp,                    --real start time of task
 	end_time timestamp,                      --real end time of task
-	engine_id integer not null, --identifier of engine which executing the task (engines.id)
+	engine_id integer, --identifier of engine which executing the task (engines.id)
 	created_at timestamp default now() not null,
 	err_message varchar(4000),          --return message in case of error
 	created_by_uid integer,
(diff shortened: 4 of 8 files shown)


perun-tasks-lib/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskDaoJdbc.java
```diff
@@ -115,7 +119,7 @@ public int scheduleNewTask(Task task, int engineID) throws InternalErrorExceptio
 			newTaskId = Utils.getNewId(this.getJdbcTemplate(), "tasks_id_seq");
 			this.getJdbcTemplate().update(
 					"insert into tasks(id, exec_service_id, facility_id, schedule, recurrence, delay, status, engine_id) values (?,?,?, " + Compatibility.toDate("?","'DD-MM-YYYY HH24:MI:SS'") + ",?,?,?,?)",
-					newTaskId, task.getExecServiceId(), task.getFacilityId(), getDateFormatter().format(task.getSchedule()), task.getRecurrence(), task.getDelay(), task.getStatus().toString(), engineID);
+					newTaskId, task.getExecServiceId(), task.getFacilityId(), getDateFormatter().format(task.getSchedule()), task.getRecurrence(), task.getDelay(), task.getStatus().toString(), engineID < 0 ? "null" : engineID);
```
zlamalp: You can't store string ``"null"`` in ``integer`` type column. It doesn't correspond with your change in DB, where ``engine_id`` can be null.

perun-core/src/main/resources/schema.sql
```diff
@@ -857,7 +857,7 @@ create table tasks (
 	status varchar(16) not null,
 	start_time timestamp,
 	end_time timestamp,
-	engine_id integer not null,
+	engine_id integer,
```
zlamalp: Please update schema in all required files in ``perun-db`` module and also in all changelogs in ``perun-core``. Also please bump DB version.
There is ``engine_id`` column in ``tasks_results`` table too! I think we should keep it consistent.

perun-tasks-lib/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskDaoJdbc.java
```diff
@@ -176,7 +180,7 @@ public Task getTask(int execServiceId, int facilityId, int engineID) {
 						", " + ExecServiceDaoJdbc.execServiceMappingSelectQuery + ", " + ServicesManagerImpl.serviceMappingSelectQuery + " from tasks left join exec_services on tasks.exec_service_id = exec_services.id " +
 						"left join facilities on facilities.id = tasks.facility_id left join services on services.id = exec_services.service_id " +
 						"where exec_services.id = ? and tasks.facility_id = ? and tasks.engine_id = ?",
-						new Integer[] { execServiceId, facilityId, engineID }, TASK_ROWMAPPER);
+						new Object[] { execServiceId, facilityId, engineID < 0 ? "null" : engineID}, TASK_ROWMAPPER);
```
zlamalp: Regarding the fact, that you can't store string in integer column, you should modify select instead to distinct between ``tasks.engine_id is null`` and ``tasks.engine_id = "+engineId+"``. I now joining strings is not generally safe, but here value is a generic ``int``, which is harmless regarding sql injection.

perun-core/src/main/resources/schema.sql
mvocu: engine_id in tasks_results can not be null...there must be an engine that processed given task to obtain task result.

mvocu: Committed schema and changelog modification (hopefully at the right place now).

perun-tasks-lib/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskDaoJdbc.java
mvocu: Fixed.

perun-tasks-lib/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskDaoJdbc.java
mvocu: Fixed (but hope that array parameters may be longer than required)...

perun-core/src/main/resources/schema.sql
zlamalp: Please update DB version also in ``test-data.sql`` in perun-core resources. Test DB can't update its own version, since version mismatch prevents perun to startup.

zlamalp:
Please update DB version also in ``test-data.sql`` in perun-core resources. It's failing tests right now. Thank you.

perun-core/src/main/resources/hsqldbChangelog.txt
```diff
@@ -4,6 +4,8 @@
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
 -- this update is not supported on hsql since its used only as in-memory db
+3.1.40
+alter table tasks alter column engine_id type integer;
```
zlamalp: DB version bump must be part of each version changelog. Please add
```sql
insert into configurations (property,value) values ('DATABASE VERSION','3.1.40');
```
to ``hsqldbChangelog.txt``, ``oracleChangelog.txt`` and ``postgresChangelog``.

perun-db/oracle.sql
```diff
@@ -866,7 +866,7 @@ create table tasks (
 	status nvarchar2(16) not null,
 	start_time date,
 	end_time date,
-	engine_id integer not null,
+	engine_id integer,
```
zlamalp: Please update DB version in a comment on first line of a file and at the end in initial version update. Thank you.

perun-db/postgres.sql
```diff
@@ -927,7 +927,7 @@ create table "tasks" (
 	status varchar(16) not null,        --state of task
 	start_time timestamp,                    --real start time of task
 	end_time timestamp,                      --real end time of task
-	engine_id integer not null, --identifier of engine which executing the task (engines.id)
+	engine_id integer, --identifier of engine which executing the task (engines.id)
```
zlamalp: Please update DB version in a comment on first line of a file and at the end in initial version update. Thank you.

PR #1375, opened 2016-11-10
Perl CLI tools - using printTable

Natrezim:
- all listOf tools rewritten to use printTable
- reformat of all perl files.

### DIFF
--- a/perun-rpc/src/main/perl/Perun/beans/User.pm
+++ b/perun-rpc/src/main/perl/Perun/beans/User.pm
@@ -21,7 +21,7 @@ sub TO_JSON
 
 	my $id;
 	if (defined($self->{_id})) {
-		$id = $self->{_id}*1;
+		$id = $self->{_id} * 1;
 	} else {
 		$id = 0;
 	}
@@ -67,7 +67,7 @@ sub TO_JSON
 	} else {
 		$isServiceUser = undef;
 	}
-	
+
 	my $isSponsoredUser;
 	if (defined($self->{_isSponsoredUser})) {
 		$isSponsoredUser = $self->{_isSponsoredUser};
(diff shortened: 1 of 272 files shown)


perun-rpc/src/main/perl/Perun/beans/User.pm
```diff
@@ -177,11 +222,11 @@ sub getSortingName {
 
 sub getCommonArrayRepresentation {
 	my $user = shift;
-	return ($user->getId, $user->getDisplayName);
+	return ($user->getId, $user->getDisplayName, $user->getIsServiceUser, $user->getIsSponsoredUser);
```
zlamalp: In order to get nice print of booleans you can use implementation used in Attribute object, method getValueAsScalar(). 
```perl
($value) ? 'true' : 'false';
```
You can also check, that value is actually JSO boolean 
```perl
if (ref $value eq "JSON::PP::Boolean")
```
If you would need to set it back (since you can't pass boolean from terminal console) then you can use: 
```perl
my $true = JSON::PP::true;
my $false = JSON::PP::false;
```
and pass it to setter method after checking input for 1/0 or 'true'/'false'. Undef also means 'false'.

perun-rpc/src/main/perl/listOfRequiredVoAttributes
```diff
@@ -22,7 +22,8 @@ sub help {
 	};
 }
 
-my ($voId, $voShortName, $sortingFunction, $batch);
+our $batch;
+my ($voId, $voShortName, $sortingFunction);
```
zlamalp: API method requires also ``service`` property as you said, so please update this tool and call ``getRequiredAttributes(vo =>$voId);`` also with ``service => $serviceId`` param. Thank you.

perun-rpc/src/main/perl/Perun/beans/User.pm
```diff
@@ -181,8 +181,8 @@ sub setIsServiceUser
 	my $self = shift;
 	my $value = shift;
 	$value eq 'true' || $value eq 1 || ref $value eq "JSON::PP::Boolean"
```
zlamalp: Sorry I didn't realize previously, but this will not work. Since it sets 'true' when passed value is boolean (but we don't know if original value is true or false, since we check value type (reference) and not value).
You should check the reference first, if boolean, then set passed value to object immediately. Then 'elsif' branch check if value is 'true' or 1, set boolean true to object, otherwise set false. Otherwise always set false.

Same goes for setIsSponsoredUser() method.

zlamalp: Please rebase this branch on current master. Zora updated listOfPrincipalRoles tool, which was not working properly.

PR #1381, opened 2016-11-23
DISPATCHER,ENGINE: Cleanup and speedup normal propagation

zlamalp:
DISPATCHER

- Removed long unused CheckInJob.
- Removed EngineDao (was used only by CheckInJob).
- Added javadoc and spring annotations to Jobs in use.
- Added javadoc to DispatcherStarter, fixed indents.
- Fixed spring context (data for jobs, cron triggers), now
  perun-dispatcher.xml contains only quartz jobs definitions and
  perun-dispatcher-scheduler.xml contains perunScheduler definition,
  which actually starts the jobs.
- Change cron delay for propagation jobs to half of original value.
- Maintenance job for checking runtime state against DB changed
  from "at 2 everyday" to "every 2 hours".

ENGINE

- Change cron delay for propagation jobs to half of original value.
- Increase max concurrent running gen scripts from 10 to 15.


zlamalp:
Tested locally that jobs are actually running, no exception seen in spring context. Merging.

PR #1382, opened 2016-11-24
Avoid not setting required attributes for member in group

stavamichal:
- there is a situation when assigning group to resource and in the same
   time add member to this group, this member can exist there without
   setted required attributes (by resource/facility)
 - we need to control state of group after first cycle of attributes
   setting and if there is someone new, do the process for him too


stavamichal:
We will use locks on Group and Members instead.

Closing.

PR #1391, opened 2016-12-05
planServicePropagation added

zuz-schwarzova:


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GeneralServiceManagerMethod.java
```diff
@@ -392,6 +392,30 @@ public Integer call(ApiCaller ac, Deserializer parms) throws PerunException {
 	},
 
 	/*#
+	 * Plans service propagation on defined facility.
```
zlamalp: Please add javadoc twice and fix it for forcePropagation() method too. First javadoc should hold "facility/service" combination, like it already does. Second javadoc should cover "only service" case. Doubling javadoc is necessary for our documentation generator to recognize it as two methods with different set of params. Thank you.
```
/*#
* first Javadoc
*/
/*#
* second javadoc
*/
```

zlamalp: Please add javadoc twice and fix it for forcePropagation() method too. First javadoc should hold "facility/service" combination, like it already does. Second javadoc should cover "only service" case. Doubling javadoc is necessary for our documentation generator to recognize it as two methods with different set of params. Thank you.
```
/*#
* first Javadoc
*/
/*#
* second javadoc
*/
```

PR #1397, opened 2016-12-12
Updated Spring framework to version to 3.2

zlamalp:
- Spring 3.1 support ended long time ago.
- Update to 3.2, which seems not to require change in a code.
  Yet its support ends on Dec 2016, so next year we should move
  to Spring 4.


zlamalp:
test this please.

PR #1403, opened 2016-12-14
New utilities for locking on objects and actions in perun

stavamichal:
- new Class PerunLockUtils (static one)
 - it contains methods and containers for purpose of working with locks
   on specific actions, objects or combinations of objects
 - for now just 2 locking methods for groupMembership and one for
   unlocking all defined locks are implemented here

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
@@ -1,0 +1,154 @@
+package cz.metacentrum.perun.core.impl;
+
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
+
+import java.util.Map;
+import java.util.List;
+import java.util.Collections;
+import java.util.ArrayList;
+import java.util.Set;
+import java.util.Random;
+import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * Methods and structures for working with locks on objects and actions.
+ *
+ * Created by Michal Stava stavamichal@gmail.com
+ */
+public class PerunLocksUtils {
+
+	private final static Logger log = LoggerFactory.getLogger(PerunLocksUtils.class);
+
+	//This empty object is used just for purpose of saving and identifying all locks for separate transaction
+	public static Object uniqueKey = new Object();
+
+	//Maps for saving and working with specific locks
+	private static ConcurrentHashMap<Group, ReadWriteLock> groupsLocks = new ConcurrentHashMap<>();
+	private static ConcurrentHashMap<Group, ConcurrentHashMap<Member, Lock>> groupsMembersLocks = new ConcurrentHashMap<>();
+
+	/**
+	 * Create transaction locks for combination of group and member (from list of members)
+	 * and also bind them to the transaction (as resource by Object uniqueKey)
+	 *
+	 * @param group the group
+	 * @param members list of members
+	 * @throws InternalErrorException
+	 * @throws InterruptedException
+	 */
+	public static void lockGroupMembership(Group group, List<Member> members) throws InternalErrorException, InterruptedException {
+		if(group == null) throw new InternalErrorException("Group can't be null when creating lock for group and list of members.");
+		if(members == null) throw new InternalErrorException("Members can't be null or empty when creating lock for group and list of members.");
+
+		//Sort list of members strictly by ids (there is compareTo method in perunBean using ids for comparing)
+		Collections.sort(members);
+
+		List<Lock> returnedLocks = new ArrayList<>();
+
+		try {
+			//TODO - On java8 use computeIfAbsent instead
+			//try to lock all needed locks there
+			ReadWriteLock groupReadWriteLock = groupsLocks.get(group);
+			if(groupReadWriteLock == null) {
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
```diff
@@ -0,0 +1,124 @@
...
+
+		//Sort list of members strictly by ids (there is compareTo method in perunBean using ids for comparing)
+		Collections.sort(members);
+
+		List<Lock> returnedLocks = new ArrayList<>();
+		//ReentrantReadWriteLock readwriteLock = new ReentrantReadWriteLock();
+
+		try {
+			//try to lock all needed locks there
+			groupsLocks.putIfAbsent(group, new ReentrantReadWriteLock(true));
```
licehammer: The constructor of ReentrantReadWriteLock is called each time which brings unnecessary overhead because the object have to be constructed and later cleared by garbage collector.

zlamalp: @licehammer so, we should use something like:
```java
synchronized (groupsLocks) {
  if (groupsLocks.get(group) == null) groupsLocks.put(group, new ReentrantReadWriteLock(true))
}
```
and so on whenever we add/update value in a map (groupsLocks / groupsMembersLocks) ? Because we need atomicity in checking and putting value to make sure no two threads are putting same value together. Reading should be OK, all ended operation are reflected by ``ConcurrentHashMap`` javadoc.

licehammer: Your solution will work, but it creates unnecessary critical (i.e. synchronized) section which will be entered every time. I would suggest something like this:
```java
Lock readGroupLock = groupsLocks.get(group).readLock();
if(groupsLocks == null) {
    Lock lock = new ReentrantReadWriteLock(true);
    groupsLocks = groupsLocks.putIfAbsent(group, lock);
    if(groupsLocks == null) groupsLocks = lock;
}

Comment (stavamichal) [inline perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java]: Thanks, it is fixed now.

File: perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java [hunk 2]
```diff
@@ -0,0 +1,127 @@
...
+				groupsLocks.putIfAbsent(group, new ReentrantReadWriteLock(true));
+				readGroupLock = groupsLocks.get(group).readLock();
+			}
+			readGroupLock.lock();
+			returnedLocks.add(readGroupLock);
+			for (Member member : members) {
+				Map<Member, Lock> newLockForMember = new HashMap<>();
+				newLockForMember.put(member, new ReentrantLock(true));
+				groupsMembersLocks.putIfAbsent(group, newLockForMember);
+				Lock memberLock = groupsMembersLocks.get(group).get(member);
```
licehammer: It faster if you check for existing Lock in groupsMembersLocks first and only if it doesn't exist then create a new one.

stavamichal: Fixed for all occurrences.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
```diff
@@ -0,0 +1,141 @@
...
+
+		//Sort list of members strictly by ids (there is compareTo method in perunBean using ids for comparing)
+		Collections.sort(members);
+
+		List<Lock> returnedLocks = new ArrayList<>();
+		//ReentrantReadWriteLock readwriteLock = new ReentrantReadWriteLock();
+
+		try {
+			//try to lock all needed locks there
+			Lock readGroupLock = groupsLocks.get(group).readLock();
```
licehammer: May cause NullPointerException if groupsLocks.get(group) returns null.
The same problem is at line 111.

stavamichal: Also fixed. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
```diff
@@ -0,0 +1,141 @@
...
+			for (Member member : members) {
+				//Get members lock map by group if exists or create a new one
+				ConcurrentHashMap<Member, Lock> newLockForMember = groupsMembersLocks.get(group);
+				if(newLockForMember == null) {
+					groupsMembersLocks.putIfAbsent(group, new ConcurrentHashMap<Member, Lock>());
+					newLockForMember = groupsMembersLocks.get(group);
+				}
+
+				//Get concrete member lock from members lock map or create a new one if not exists
+				Lock memberLock = groupsMembersLocks.get(group).get(member);
```
licehammer: You could use:
```java
Lock memberLock = newLockForMember.get(member);
```
Otherwise I see no point in code at line 73.

stavamichal: Yes, you are right, I overlooked this. It is fixed now and also variable is renamed to be more suitable.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
```diff
@@ -0,0 +1,150 @@
...
+		Collections.sort(members);
+
+		List<Lock> returnedLocks = new ArrayList<>();
+
+		try {
+			//try to lock all needed locks there
+			ReadWriteLock groupReadWriteLock = groupsLocks.get(group);
+			if(groupReadWriteLock == null) {
+				groupsLocks.putIfAbsent(group, new ReentrantReadWriteLock(true));
+				groupReadWriteLock = groupsLocks.get(group);
```
martin-kuba: Instead of putIfAbsent(key,value) immediately followed by  get(key),  simple computeIfAbsent(key, key -> value) may be used

stavamichal: Good point! I added it to TODO comment.

PR #1407, opened 2016-12-20
update task information before scheduling

mvocu:
* refetch task data (facility, exec service) every time before scheduling (should fix issue #3164)
* check if task is fit for propagation every time before scheduling and remove them if 
  they are not (facility or exec service does not exist anymore, service is not assigned to facility, is blocked etc.) (should fix issues #3190, #3140)

### DIFF
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/TaskSchedulerImpl.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/TaskSchedulerImpl.java
@@ -11,6 +11,7 @@ import org.apache.commons.codec.binary.Base64;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
 
 import cz.metacentrum.perun.core.api.Destination;
 import cz.metacentrum.perun.core.api.Facility;
@@ -23,6 +24,7 @@ import cz.metacentrum.perun.core.api.exceptions.FacilityNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.ServiceNotExistsException;
+import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.dispatcher.jms.DispatcherQueue;
 import cz.metacentrum.perun.dispatcher.jms.DispatcherQueuePool;
 import cz.metacentrum.perun.dispatcher.scheduling.DenialsResolver;
@@ -45,7 +47,8 @@ public class TaskSchedulerImpl implements TaskScheduler {
 	@Autowired
 	private DependenciesResolver dependenciesResolver;
 	@Autowired
-	private Perun perun;
+	private ApplicationContext appCtx;
+	private PerunBl perun;
 	private PerunSession perunSession;
 	@Autowired
 	private Properties dispatcherPropertiesBean;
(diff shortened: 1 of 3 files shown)


perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/TaskSchedulerImpl.java
```diff
@@ -246,14 +289,23 @@ public void scheduleTask(Task task) {
...
+							// state.
+							log.info("   Last Task [dependency:"
+									+ dependency.getId() + ", facility:"
+									+ facility.getId()
+									+ "] was NULL, we are gonna propagate.");
+							if(scheduleItAndWait(dependency, facility, execService,
+									dispatcherQueue, time)) {
+								// task sucessfully scheduled, nothing to do
+							} else {
+								// TODO: task aborted - maybe set this one to error?
```
zlamalp: I believe something should be done in else branch (e.g. aborting SEND Task too). If above code causes GEN Task to be aborted, then it's removed from pool completely. So when we goes through pool once again, SEND is ``proceed=true`` (not aborted until here). Here we are ok not to start (create new GEN task), but this whole part sets ``proceed=false`` at the end => we won't reach line 433 where we actually abort the SEND Task. So SEND Task would stay in a pool as PLANNED or NONE? forever.

Above case is when we delete Facility or Service is not assigned. If we delete Service, it should work since GEN-SEND dependencies will be empty and we won't set ``proceed=false``.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/TaskSchedulerImpl.java
```diff
@@ -111,33 +116,71 @@ public void scheduleTask(Task task) {
...
-		log.debug("Is the execService ID:" + execService.getId() + " enabled globally?");
-		if (execService.isEnabled()) {
-			log.debug("   Yes, it is globally enabled.");
-		} else {
-			log.debug("   No, execService ID: "+ execService.getId() + " is not enabled globally. Task will not run.");
-			return;
+		Boolean abortTask = false;
+		try {
+			refetchTaskInformation(task);
+			List<Service> assignedServices = perun.getServicesManager().getAssignedServices(perunSession, task.getFacility());
```
mvocu: But  there is no method getServicesManagerBl() in cz.metacentrum.perun.core.api.Perun? I can typecast `perun` bean to `PerunBlImpl` explicitely (ouch!), but for what reason when `PerunBlImpl.getAssignedServices()` contains just 
`return servicesManagerImpl.getAssignedServices(sess, facility);`?
Moreover, the `getServicesManager()` returns object of class `ServicesManagerEntry`, which (apart from checking authorization first) implements `getAssignedServices()` by going through `getServicesManagerBl().getAssignedServices()`. Or am I missing something?

zlamalp: We usually autowire bean ``PerunBl`` and not ``Perun`` which give us access to perun managers respecting current transaction, including ``ServicesManagerBl``. The goal here is not to start new transaction (which is always done when Entry method is called). Also you don't have to check access rights, if it's all called by inner component. Hence using ``PerunBl`` is a better choice.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/TaskSchedulerImpl.java
mvocu: Fixed by last commit.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/impl/TaskSchedulerImpl.java
mvocu: There is no bean `PerunBl`, there is only `perun`, typecast using Spring mechanism for getting beans (not `@Autowire`). I used that, though personally I feel it is a kind of hack.

zlamalp:
Merged locally with slight change - we can directly autowire ``PerunBl`` bean in property ``perun`` without usage of ``ApplicationContext`` in init or constructor.

PR #1410, opened 2017-01-02
CLI: Fixed User object boolean properties

zlamalp:
- Fixed naming of serviceUser and sponsoredUser properties of User object
  to match actual names send from RPC (removed "is").


zlamalp:
test this please.

PR #1422, opened 2017-01-10
CORE: Fix of extSourceName for BBMRI

jirmauritz:


jirmauritz:
Not relevant, the ext source name is correct. Closing.

PR #1427, opened 2017-01-23
Fix null pointer exception in gid extSource

stavamichal:
- if replacement is null, that means attribute not exists or has null
   value, so throw InternalErrorException

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcePerun.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcePerun.java
@@ -139,6 +139,9 @@ public class ExtSourcePerun extends ExtSource implements ExtSourceApi {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourcePerun.java
```diff
@@ -139,6 +139,9 @@ public static PerunBlImpl setPerunBlImpl(PerunBlImpl perun) {
 			}
 		}
 
+		//Null login is not allowed there
+		if(login == null) throw new InternalErrorException("Login can't be null.");
```
licehammer: It would be handy to log related richUser as well. It might help with debug if this error ever occurs.

stavamichal: Fixed.

PR #1441, opened 2017-02-08
CORE: Send expiration notifications month before (not 30 days)

zlamalp:
- Notifications about membership expiration will be sent month
  before expiration, not 30 days. Notification regex must be updated
  on new version deployment.
- Before creating audit message check, if user can extend membership.
  If not and reason is "OUTSIDEEXTENSIONPERIOD", do not proceed.
  There is no point in sending notification if user can't extend yet.
- Notify only VALID or SUSPENDED members. For backward notification
  (after 7 days) include also EXPIRED state.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Synchronizer.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Synchronizer.java
@@ -5,6 +5,7 @@ import java.util.Calendar;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import cz.metacentrum.perun.core.api.*;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Synchronizer.java
```diff
@@ -90,32 +93,67 @@ public void checkMembersState() {
...
+					if (!Objects.equals(ex.getReason(), ExtendMembershipException.Reason.OUTSIDEEXTENSIONPERIOD)) {
+						// we don't care about other reasons (LoA), user can update it later
+						getPerun().getAuditer().log(sess, "{} will expire in a month in {}.", m, vosMap.get(m.getVoId()));
+					}
+				}
 			}
 
 			// log message for all members which will expire in 14 days
-			compareDate.add(Calendar.DAY_OF_MONTH, -16);
+			compareDate.add(Calendar.DAY_OF_MONTH, 14);
```
stavamichal: I would prefer to use always a new variable with specific current date + or - some day. Recycling this variable is not bad, but it is harder to read and debug.

PR #1445, opened 2017-02-24
Modify group locks - add timeout, use group/member locks in Perun

stavamichal:
- modify locks for Group, GroupMember - add timeout (4 hours)
 - add locks to the code in Perun where is any manipulation with
   groups (creating, deleting, adding and removing) and manipulation
   with members (adding to Group, remove from Group)
 - add better thread local variable for purpose of seeing only locks for
   current thread
 - unbind all already used locks from ending thread

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
@@ -2,6 +2,8 @@ package cz.metacentrum.perun.core.impl;
 
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.exceptions.GroupNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 
 import org.slf4j.Logger;
@@ -16,6 +18,8 @@ import java.util.Set;
 import java.util.Random;
 import java.util.HashMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -31,7 +35,7 @@ public class PerunLocksUtils {
 	private final static Logger log = LoggerFactory.getLogger(PerunLocksUtils.class);
 
 	//This empty object is used just for purpose of saving and identifying all locks for separate transaction
-	public static Object uniqueKey = new Object();
+	public static ThreadLocal<Object> uniqueKey = ThreadLocal.withInitial(Object::new);
 
 	//Maps for saving and working with specific locks
 	private static ConcurrentHashMap<Group, ReadWriteLock> groupsLocks = new ConcurrentHashMap<>();
@@ -46,7 +50,7 @@ public class PerunLocksUtils {
 	 * @throws InternalErrorException
 	 * @throws InterruptedException
 	 */
-	public static void lockGroupMembership(Group group, List<Member> members) throws InternalErrorException, InterruptedException {
+	public static void lockGroupMembership(Group group, List<Member> members) throws InternalErrorException {
 		if(group == null) throw new InternalErrorException("Group can't be null when creating lock for group and list of members.");
 		if(members == null) throw new InternalErrorException("Members can't be null or empty when creating lock for group and list of members.");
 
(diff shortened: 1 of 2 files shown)


stavamichal:
There are "TODO" parts in this commit. I need to first discuss if they are correct thoughts and than I can remove them from code at all.

stavamichal:
Please @zlamalp  look at this commit when you'll have a time. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
```diff
@@ -56,46 +60,53 @@ public static void lockGroupMembership(Group group, List<Member> members) throws
...
 
 		try {
-			//TODO - On java8 use computeIfAbsent instead
-			//try to lock all needed locks there
-			ReadWriteLock groupReadWriteLock = groupsLocks.get(group);
-			if(groupReadWriteLock == null) {
-				groupsLocks.putIfAbsent(group, new ReentrantReadWriteLock(true));
-				groupReadWriteLock = groupsLocks.get(group);
+			try {
+				//TODO - On java8 use computeIfAbsent instead
```
zlamalp: We are on java8 now, maybe we could implement suggested solution.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -606,7 +662,10 @@ protected void removeDirectMember(PerunSession sess, Group group, Member member)
 		// check all relations with this group and call processRelationMembers to reflect changes of removing member from group
 		List<Integer> relations = groupsManagerImpl.getResultGroupsIds(sess, group.getId());
 		for (Integer groupId : relations) {
-			processRelationMembers(sess, groupsManagerImpl.getGroupById(sess, groupId), Collections.singletonList(member), group.getId(), false);
+			Group relationGroup =  groupsManagerImpl.getGroupById(sess, groupId);
+			//Create lock for affected group and member
+			PerunLocksUtils.lockGroupMembership(relationGroup, Arrays.asList(member));
```
zlamalp: We might want to check javadoc of ``Collections.singletonList(member)`` used in a code for group relations and compare it with ``Arrays.asList(member)`` if it has any benefits. Probably depends on later usage. Second method is backed by fixed sized array, hence we can't add more items to the list.

stavamichal: Both methods create immutable list of objects, but Collections.singletonList(Object) seems to be less complicated (it is just an immutable list of 1 Object) so it is probably more suited for this operation.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
stavamichal: Ok.

stavamichal:
Ok. I add new commit with fixes of mentioned problems. @zlamalp can you please look on it again? Please also try to think about my "TODO" in GroupsManagerBlImpl. I need to remove them before merging this pull request but I still don't know if the behavior around them is correct.

stavamichal:
Rebase of this pull request was little complicated. For this reason I ask for new review please.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2763,6 +2866,10 @@ public void removeRelationMembers(PerunSession sess, Group resultGroup, List<Mem
 	@Override
 	public Group createGroupUnion(PerunSession sess, Group resultGroup, Group operandGroup, boolean parentFlag) throws WrongReferenceAttributeValueException, WrongAttributeValueException, GroupNotExistsException, InternalErrorException, GroupRelationAlreadyExists, GroupRelationNotAllowed {
 
+		//Lock affected groups
+		PerunLocksUtils.lockGroupMembership(resultGroup);
+		PerunLocksUtils.lockGroupMembership(operandGroup);
```
Vojtech-Sassmann: If I understand changes in this PR correctly, then groups in operations like "updateGroup" are locked in reverse order of their names. This code could be cause of deadlock. 

Imagine that method "updateGroup" wants to lock groups ABC in their names reverse order. It locks groups C, B and than this method is called with groups A and B where A is result group and B is operand group. It manages to lock A and then waits because "updateGroup" method locked group B. And method "updateGroup" is waiting because this method locked group A. I believe these two locks should be done as well in reverse order of given groups names.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2808,6 +2915,10 @@ public void removeGroupUnion(PerunSession sess, Group resultGroup, Group operand
 			throw new GroupRelationDoesNotExist("Union does not exist between result group " + resultGroup + " and operand group" + operandGroup + ".");
 		}
 
+		//Lock affected groups
+		PerunLocksUtils.lockGroupMembership(resultGroup);
+		PerunLocksUtils.lockGroupMembership(operandGroup);
+
```
Vojtech-Sassmann: Possible deadlock cause.

stavamichal:
This whole PR was changed:
 - application of groups locks in the code was removed, because we need to test how this change the whole Perun performance
 - commits were squashed to one and message was changed
 - PR was rebased to actual master

Please @Vojtech-Sassmann, @licehammer check it and add review or set it as ready to merge. Thank you.

PR #1460, opened 2017-03-21
CABINET: Reworking locking in cabinet

zlamalp:
- Fixed deleteThanks() processing order.

- Previous fix didn't work. Synchronized on method level still
  causes deadlock, even when items are sorted.
  Synchronized inside method (on manager class/instance) didn't help either.
  Working solution is synchronized on manager class outside
  the deadlocking method -> create/deleteThanks().

- Same solution used for create/deleteAuthorship(), updatePublication()
  and updateCategory() when we recalculate priority coefficient.

- It should work also for concurrent calls of createThanks() and createAuthorship()
  for same user since we lock on manager class and always use same order of locks,
  so first come first served.

### DIFF
--- a/perun-cabinet/src/main/java/cz/metacentrum/perun/cabinet/bl/impl/AuthorshipManagerBlImpl.java
+++ b/perun-cabinet/src/main/java/cz/metacentrum/perun/cabinet/bl/impl/AuthorshipManagerBlImpl.java
@@ -110,12 +110,16 @@ public class AuthorshipManagerBlImpl implements AuthorshipManagerBl {
 			throw new CabinetException(ErrorCodes.USER_NOT_EXISTS, e);
 		}
 		log.debug("{} created.", authorship);
+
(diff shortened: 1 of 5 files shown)


zlamalp:
Hi @martin-kuba, can you please review this? @licehammer can you check github settings, Martin is not offered in "reviewers", but others outside devel team are.

licehammer:
I added @martin-kuba to the Perun team. It should work now.

zlamalp:
Thanks @licehammer. Without this fix I can reproduce deadlock almost everytime, since GUI allow multiple selection when adding Thanks. After this fix, it works (tested on 200 authors and 20 thanks sent parallel).

I basically need to make sure, that whenever there is a change in authors or rank (publication/category), priorityCoefficient or thanks attribute for one user (all authors) is calculated and set correctly. I don't mind, that after fix only single thread can make the calculation at the time, since action is pretty quick and users don't usually do this on parallel (with exception to thanks, which they can submit multiple times).

PR #1462, opened 2017-03-24
Fix parsing of attribute with mapping rules for extSource

stavamichal:
- use parsing by ",\n" instaed of just ","


stavamichal:
There is an info in technical change-log on redmine wiki what need to be done before deploying this change.

PR #1469, opened 2017-04-10
Registrar: module for BBMRI registration added

jirmauritz:
The module
1. reads input field with collection IDs and checks, whether the collections exist in Perun as groups
2. adds users to the appropriate groups
Feature is already deployed on https://perun.bbmri-eric.eu/fed/gui/

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
@@ -1,0 +1,161 @@
+package cz.metacentrum.perun.registrar.modules;
+
+import cz.metacentrum.perun.core.api.*;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.PerunException;
+import cz.metacentrum.perun.registrar.RegistrarManager;
+import cz.metacentrum.perun.registrar.RegistrarModule;
+import cz.metacentrum.perun.registrar.exceptions.CantBeApprovedException;
+import cz.metacentrum.perun.registrar.model.Application;
+import cz.metacentrum.perun.registrar.model.ApplicationFormItemData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
+
+/**
+ * Module for BBMRI instance.
+ * The module
+ * 1. reads input with collection IDs and checks, whether the collections exist in Perun as groups
+ * 2. adds users to the appropriate groups
+ *
+ * @author Jiri Mauritz <jirmaurtiz@gmail.com>
+ */
+@SuppressWarnings("Duplicates")
+public class BBMRI implements RegistrarModule {
+
+	final static Logger log = LoggerFactory.getLogger(BBMRI.class);
+	public static final String BIOBANK_IDS_FIELD = "Comma or new-line separated list of collection IDs you are representing:";
+
+	private RegistrarManager registrar;
+
+	@Override
+	public void setRegistrar(RegistrarManager registrar) {
+		this.registrar = registrar;
+	}
+
+	@Override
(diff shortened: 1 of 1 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
```diff
@@ -0,0 +1,161 @@
...
+
+		// get the field of application with the collections
+		Set<String> collectionIDsInApplication = getCollectionIDsFromApplication(session, app);
+
+		// get non-existing collections
+		collectionIDsInApplication.removeAll(collectionIDsInPerun);
+
+		// difference must be empty
+		if (!collectionIDsInApplication.isEmpty()) {
+			throw new CantBeApprovedException("Collections " + collectionIDsInApplication + " does not exists.", "", "", "", true);
```
jirmauritz: CantBeApprovedException should be generalized because parameters are too specific for the Metacentrum VO purpose.

zlamalp: I'll look into this, so it can be merged and new production can be deployed to bbmri.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}
+
+		if (collectionsString == null) {
+			throw new InternalErrorException("There is no field with biobank IDs on the registration form.");
+		}
+
+		// get set of collection IDs from application
+		Set<String> collectionIDsInApplication = new HashSet<>();
+		for (String collection : collectionsString.split("[,\n ]+")) {
```
zlamalp: I might not be reading regex right, but is splitting by whitespace correct ? It's not stated as allowed splitter in a ``BIOBANK_IDS_FIELD``.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
```diff
@@ -0,0 +1,161 @@
...
+	 */
+	public Application beforeApprove(PerunSession session, Application app) throws PerunException {
+		// get perun and beans from session
+		Perun perun = session.getPerun();
+		Vo vo = app.getVo();
+
+		// get all collection IDs from Perun
+		Set<String> collectionIDsInPerun = new HashSet<>();
+		Group collectionsGroup = perun.getGroupsManager().getGroupByName(session, vo, "collections");
+		for (Group group : perun.getGroupsManager().getSubGroups(session, collectionsGroup)) {
```
zlamalp: Just for consideration, you can also use: 
```java
List<RichGroup> groups = perun.getGroupsManager().getAllRichSubGroupsWithAttributesByNames(session, collectionsGroup, Arrays.asList("urn:perun:group:attribute-def:def:collectionID"));
```
which will return whole sub-group hierarchy sorted. It probably depends whether you expect group structure to change in a future or "representatives" groups should be always on a third level. If group structure will be fixed, your solution might be even quicker.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
zlamalp: I checked and if you don't specify ``reason`` param in ``CantBeApprovedException`` as ``"NOT_ACADEMIC"`` it will server default behavior in GUI - printing your exception message and allowing to "approve anyway" if it's specified as soft exception. Your code seems to be correct.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRI.java
jirmauritz: The regex is correct. Petr proposed the ``BIOBANK_IDS_FIELD`` text and told me to split the collections on all three: new line, space and comma.

PR #1481, opened 2017-04-26
DISPATCHER: Temporary revert back property resolving

zlamalp:
Merge or re-use if original spring context solution is not fixed soon and
there will be need to deploy production.

- With current config, devel and production spring profiles ignored properties
   which were stored in a file /etc/perun/perun-dispatcher.properties even if they
   were correctly referenced in location tag. Only properties from xml bean definition
   were visible to application.
- This fix restore previous behavior in properties handling and allows HornetQ server to start.


zlamalp:
Solved by #1483. Closing this one.

PR #1483, opened 2017-04-26
fix reading of properties from files in dispatcher

martin-kuba:
fixed reading of properties from /etc/perun/perun-dispatcher.properties in perun-dispatcher module, which was broken by previous change


zlamalp:
Tested locally and works. Thanks.

PR #1485, opened 2017-04-28
Try to recover from db constraint error on membership in synchronizator

stavamichal:
- When sychronizator have a new user (not in Vo already), it tries to
   create a member for this user in the VO. This process can ends with
   an error because of race condition of two or more synchronization
   processes in one VO at the same time. If this happens, we will try
   to found this member again. Only if this attempt fails again, it will
   throw consistency exception, because it is probably some bigger
   problem.


zlamalp:
I don't see anything wrong here. It simply retry previous action - find user/member by ues - and updates attributes if found. Also finally it throws ``AlreadyMemberException`` if it fails on DB constraint on insert. Merging.

PR #1497, opened 2017-05-15
REGISTRAR: Call canBeApproved() check in auto-approval case

zlamalp:
- When application form is set to auto-approval, there was missing check
  of attribute module, if application can be approved. It's called manually
  from GUI, if form is in manual approval mode.
- Now, if application can't be auto-approved exception is thrown and mail
  sent to VO manager.


zlamalp:
test this please.

PR #1499, opened 2017-05-15
Add attribute module for phone in Perun

stavamichal:
- add check to existing member attribute module for phone
 - create a new module for user attribute module for phone
 - only international numbers starting with '+' are correct
 - min length of number is 4 and max 16 (we want to create some
   reasonoble technical limitation)


stavamichal:
@zlamalp Please check this review and if needed, add parts for perun gui or perun registrar (it will be needed to add info about pattern into applications where phone number is required).

zlamalp:
@stavamichal Registrar changes are needed only on deployment (update restriction on registration form).
We must check usages of attributes in services and perun instances. e.g. u:phone is used at VŠUP and filled from external source with probably no restrictions (we can make them new attribute - it's in a process).
Just a generic question. Should we always check object class of attribute value in checkAttribute() method? Value type should be defined by attribute, but I understand, in a code any object can be set to attribute value.

stavamichal:
@zlamalp There is no need to check objectClass there, it is just "copy+paste" part of code used in other email attributes. But it is probably no big problem to have this check there.

PR #1504, opened 2017-05-17
Notifications: prevent from crash when postgre restart

jirmauritz:


zlamalp:
Closing this, since we talked about more solid solution and this won't help us much right now. We still need to check, what messages were sent and what messages are waiting to be sent.

PR #1505, opened 2017-05-18
 Refresh of all attributes in setRequiredAttributes

stavamichal:
Problem:
 If some attribute module alters values of other attributes inside
 changedAttributeHook() method and it happens to be non-virtual
 attribute, which is also required by the same service/facility,
 then setRequiredAttributes() method might fail on checkAttributeValue()
 call if you don't allow null value or check itself permits old value.

Solution:
 - we need to refresh all attributes in setRequiredAttributes method,
   not just virtual ones
 - javadoc for this method was changed too (to be more correct)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
@@ -1400,15 +1400,15 @@ public interface AttributesManagerBl {
 	void setRequiredAttributes(PerunSession sess, Service service, Facility facility, Resource resource, User user, Member member) throws InternalErrorException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeValueException, MemberResourceMismatchException;
 
 	/**
-	 * Get and set required attributes from arrayList for member, resource, user and facility.
+	 * Take list of required attributes and set those which are empty and can be filled, then check them all.
 	 *
-	 * IMPORTANT: set all attrs from arrayList, set not required attrs too if they are in arrayList
+	 * Important: this method DO NOT set non-empty attributes in list, just refresh their values and check them
 	 *
 	 * Procedure:
-	 * 1] Get all attrs from arrayList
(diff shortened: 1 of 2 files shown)


stavamichal:
Should be test better before merge into production code.

zlamalp:
Explanation of requested change:

If some attribute module alters values of other attributes inside ``changedAttributeHook()`` method and it happens to be non-virtual attribute, which is also required by the same service/facility, then ``setRequiredAttributes()`` method might fail on ``checkAttributeValue()`` call if you don't allow null value or check itself permits old value.

Like at VŠUP, where "login" is filled from users real name and on saving it fills other (cca 5 attributes) - usually mail addresses generated from login etc.
In our case we can't implement check on null value (prevent deletion or emptiness), because attribute present in list of attributes passed to the "setRequiredAttributes" is empty at that time (is not filled, if no login is present).

We must the either:
- ommit check on null in attribute module
- should perform cyclic ``fill()`` on all empty attributes until all their values stop changing.
- re-fetch actual values of all non-core attributes before we perform check on generated values.

3rd solution was chosen in this pull-request.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -1371,15 +1371,15 @@
 	void setRequiredAttributes(PerunSession sess, Service service, Facility facility, Resource resource, User user, Member member) throws InternalErrorException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeValueException, MemberResourceMismatchException;
 
 	/**
-	 * Get and set required attributes from arrayList for member, resource, user and facility.
+	 * Get list of required attributes and set those which are empty and can be filled, then check them all.
```
licehammer:
```suggestion
	 * Take list of required attributes from method's parameter and set those which are empty and can be filled, then check them all.
```

stavamichal: Changed.

stavamichal:
Commit message was also changed

PR #1506, opened 2017-05-23
Core: new method: getFacilitiesbyAttribute

jirmauritz:
- changed unused method in BL with the same name to take attributeName
  and attributeValue instead of attribute as an object
- rpc method created
- tested

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
@@ -3,6 +3,7 @@ package cz.metacentrum.perun.core.api;
 import java.util.List;
 
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.BanAlreadyExistsException;
 import cz.metacentrum.perun.core.api.exceptions.BanNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.FacilityAlreadyRemovedException;
@@ -104,6 +105,21 @@ public interface FacilitiesManager {
 	 * @throws PrivilegeException
 	 */
 	List<Facility> getFacilitiesByDestination(PerunSession perunSession, String destination) throws InternalErrorException, FacilityNotExistsException, PrivilegeException;
+
+	/**
+	 * Returns all facilities who have set the attribute with the value. Searching by attributeName. Searching only def and opt attributes.
+	 * Can find only attributes with String Value by this way! (not Integer, Map or List)
+	 *
+	 * @param sess
+	 * @param attributeName name of the attribute
+	 * @param attributeValue value of the attribute
+	 * @return list of facilities
+	 * @throws InternalErrorException
+	 * @throws PrivilegeException
+	 * @throws AttributeNotExistsException
+	 */
+	List<Facility> getFacilitiesByAttribute(PerunSession sess, String attributeName, String attributeValue) throws InternalErrorException, PrivilegeException, AttributeNotExistsException, WrongAttributeAssignmentException;
+
 
 	/**
 	 * Get count of all facilities.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
@@ -107,8 +107,8 @@ public interface FacilitiesManager {
 	List<Facility> getFacilitiesByDestination(PerunSession perunSession, String destination) throws InternalErrorException, FacilityNotExistsException, PrivilegeException;
 
 	/**
-	 * Returns all facilities who have set the attribute with the value. Searching by attributeName. Searching only def and opt attributes.
-	 * Can find only attributes with String Value by this way! (not Integer, Map or List)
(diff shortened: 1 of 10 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -163,6 +164,20 @@ public Facility getFacilityByName(PerunSession sess, String name) throws Interna
 		return facilities;
 	}
 
+	public List<Facility> getFacilitiesByAttribute(PerunSession sess, String attributeName, String attributeValue)
+			throws InternalErrorException, PrivilegeException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		Utils.checkPerunSession(sess);
+
+		// Authorization
+		if(!AuthzResolver.isAuthorized(sess, Role.SERVICEUSER)) {
```
zlamalp: I believe ``Role.SERVICEUSER`` would allow any service user to call this. Regarding rest of our API, you should check for FACILITYADMIN role and you should filter resulting list of facilities to return only those, where user is facility manager. Or to be more correct you should check, if user has READ right on attribute he search by on that facility. If not, do not return the facility.

perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
```diff
@@ -106,6 +107,21 @@
 	List<Facility> getFacilitiesByDestination(PerunSession perunSession, String destination) throws InternalErrorException, FacilityNotExistsException, PrivilegeException;
 
 	/**
+	 * Returns all facilities who have set the attribute with the value. Searching by attributeName. Searching only def and opt attributes.
+	 * Can find only attributes with String Value by this way! (not Integer, Map or List)
```
zlamalp: This is not true, it can handle string, integer, boolean and even list attributes, but the list must be in exactly same order of values as stored value. In a fact, it can't handle large attributes (map, large string, large array/list), because select compares only ``attr_value`` column. Please state, that large attributes are not supported in javadoc through all tree layers. Also please throw exception in Bl and Impl, if passed attribute is of large type.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -94,6 +94,22 @@ public Facility call(ApiCaller ac, Deserializer parms) throws PerunException {
 	},
 
 	/*#
+	 * Searches for the Facilities by theirs attribute.
```
zlamalp: Please state here, that large attributes are not supported, and that only def and opt attribute types are supported.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -94,6 +94,23 @@ public Facility call(ApiCaller ac, Deserializer parms) throws PerunException {
 	},
 
 	/*#
+	 * Returns all facilities that have set the attribute 'attributeName' with the value 'attributeValue'.
+	 * Searching only def and opt attributes. Large attributes are not supported.
+	 *
+	 * @param attributeName String
+	 * @param attributeValue String
+	 * @return Facilities Found facilities
```
zlamalp: It should return: ``List<Facility>`` not ``Facilities`` - javadoc is parsed and it would fail.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -404,10 +404,19 @@ public PerunBl getPerunBl() {
...
 
-	public List<Facility> getFacilitiesByAttribute(PerunSession sess, Attribute attribute) throws InternalErrorException, WrongAttributeAssignmentException {
-		getPerunBl().getAttributesManagerBl().checkNamespace(sess, attribute, AttributesManager.NS_FACILITY_ATTR);
-		if(!(getPerunBl().getAttributesManagerBl().isDefAttribute(sess, attribute) || getPerunBl().getAttributesManagerBl().isOptAttribute(sess, attribute))) throw new WrongAttributeAssignmentException("This method can process only def and opt attributes");
-		return getFacilitiesManagerImpl().getFacilitiesByAttribute(sess, attribute);
+
+	public List<Facility> getFacilitiesByAttribute(PerunSession sess, String attributeName, String attributeValue) throws InternalErrorException, WrongAttributeAssignmentException {
+		try {
+			AttributeDefinition attributeDef = getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attributeName);
+			Attribute attribute = new Attribute(attributeDef);
```
zlamalp: There is still missing check in a code, that passed attribute is a large type. Please use ``AttributesManagerImpl.isLargeAttribute(sess, attrDef)`` and throw some exception in such case - ``InternalErrorException`` probably.

jirmauritz: How should I get AttributesManagerImpl? I have access only to Bl managers and AttributeManagerBl has no such method as isLargeAttribute. And the method is not static.

zlamalp: If it can't be accessed directly, try to move ``isLargeAttribute()`` method to some kind of Utils class (core/base). It can be static method, since it only check type of passed attribute.

PR #1511, opened 2017-05-25
Replaced log4j with logback

martin-kuba:
Switched logging to logback, or more precisely, to spring-boot-starter-logging which uses bridges for all other logging frameworks to logback. 
Default config files logback.xml and logback-test.xml were created in perun-base module and thus they are inherited in all other modules. 
The shade plugin (creates a large executable jar file with all libraries packaged inside) was removed from perun-core and added to perun-engine, perun-ldapc and perun-ldapc-initializer.
Module duplicit dependencies were removed among pom.xml files.


perun-ldapc/src/main/resources/log4j.properties
```diff
@@ -1,39 +0,0 @@
-# log4j Configuration
```
zlamalp: How is this replaced in a logback version ? I believe, own file was used on purpose - e.g. to make ``cz.metacentrum.perun.core`` log to ``perun-ldapc.log``. For now, start stop scripts do not pass logging config on command line.

perun-engine/src/test/java/cz/metacentrum/perun/engine/AbstractEngineTest.java
```diff
@@ -63,9 +64,9 @@ public void setup() throws Exception {
 		engineId = Integer.parseInt(propertiesBean.getProperty("engine.unique.id"));
 
 		// create session
-
+		User user = perun.getUsersManagerBl().getUserById(null, 1);
```
zlamalp: I'm not fond with this change. In-memory version of tests might work, since we added SQL to create initial user (perun@BA), but when you run this against real DB, tests might fail, since you can't be sure, there is a User with ID=1. Maybe it would be better to modify methods, which caused a lot of exceptions in logs - createFaciliy() ? We might use simple check, if user is not null in session and add admin only then. I'm not sure, if it caused unnecessary exceptions in other methods.

perun-controller/pom.xml
```diff
@@ -108,25 +81,15 @@
...
-			<groupId>org.aspectj</groupId>
-			<artifactId>aspectjweaver</artifactId>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>perun-base</artifactId>
+			<version>${project.version}</version>
+			<type>test-jar</type>
+			<scope>test</scope>
 		</dependency>
-
+
```
zlamalp: Please remove whitespace.

perun-cabinet/pom.xml
```diff
@@ -86,49 +86,29 @@
...
-			<groupId>org.apache.httpcomponents</groupId>
-			<artifactId>httpmime</artifactId>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>perun-base</artifactId>
+			<version>${project.version}</version>
+			<type>test-jar</type>
+			<scope>test</scope>
 		</dependency>
-
+
```
zlamalp: Please remove whitespace.

perun-base/src/main/resources/logback.xml
```diff
@@ -0,0 +1,318 @@
...
+		scanPeriod="30 seconds" ... how often are changes scanned
+	-->
+
+	<contextName>perun-base</contextName>
+
+	<!-- variable defining the directory where log files will be put, takes system property perun.log -->
+	<variable name="LOGDIR" value="${perun.log:-/var/log/perun/}"/>
+
+	<!-- production setting for format of log lines -->
+    <variable name="ENCODER_PATTERN" value="%date [%thread] %-5level %logger{35} - %msg%n"/>
```
zlamalp: Please fix indentation (spaces to tabs).

pom.xml
```diff
@@ -64,7 +64,7 @@
 
 		<!-- redefine versions of some libraries defined by Spring platform -->
 		<postgresql.version>42.0.0</postgresql.version>
-		<tomcat.version>7.0.77</tomcat.version>
+		<tomcat.version>7.0.56</tomcat.version>
```
martin-kuba: The really used version of Tomcat on most instances is 7.0.56, not the latest. It probably has no influence, as the difference in Tomcat API among patchlevels is zero, but just to be on the safe side, the really used version is set here.

perun-web-gui/pom.xml
```diff
@@ -167,6 +167,15 @@
 			<scope>test</scope>
 		</dependency>
 
+		<!-- imports logback-test.xml for test from perun-base -->
```
martin-kuba: OK, removed.

perun-registrar-lib/pom.xml
```diff
@@ -76,74 +76,32 @@
...
-			<groupId>org.slf4j</groupId>
-			<artifactId>slf4j-log4j12</artifactId>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>perun-base</artifactId>
+			<version>${project.version}</version>
+			<type>test-jar</type>
+			<scope>test</scope>
 		</dependency>
-
+
```
martin-kuba: removed

perun-controller/pom.xml
martin-kuba: done

perun-cabinet/pom.xml
martin-kuba: done

perun-engine/src/test/java/cz/metacentrum/perun/engine/AbstractEngineTest.java
martin-kuba: reverted back, replaced by lowering log level from error to warn

perun-engine/src/main/resources/logback.xml
```diff
@@ -0,0 +1,29 @@
...
+	<variable name="LOGDIR" value="${perun.log:-/var/log/perun/}"/>
+
+	<!-- production setting for format of log lines -->
+	<variable name="ENCODER_PATTERN" value="%date [%thread] %-5level %logger{35} - %msg%n"/>
+
+	<!-- variable defining the max number of rotated files -->
+	<variable name="MAXHISTORY" value="30"/>
+
+	<appender class="ch.qos.logback.core.rolling.RollingFileAppender" name="perun-engine">
+		<file>${LOGDIR}perun-ldapc.log</file>
```
zlamalp: perun-ldapc.log -> perun-engine.log

perun-engine/src/main/resources/logback.xml
```diff
@@ -0,0 +1,29 @@
+<configuration packagingData="false" debug="false" scan="false" scanPeriod="30 seconds">
+
+	<contextName>perun-engine</contextName>
+
+	<!-- variable defining the directory where log files will be put, takes system property perun.log -->
+	<variable name="LOGDIR" value="${perun.log:-/var/log/perun/}"/>
```
zlamalp: Default for engine should be ``/var/log/perun-engine/``

perun-base/src/main/resources/logback-default.xml
```diff
@@ -0,0 +1,317 @@
+<!--
+	This is the default config for the Logback logging framework in Perun.
+	It is inherited by transitive dependencies into all Perun modules.
+	
+    You can load another file by e.g.
+     -Dlogback.configurationFile=file:///etc/perun/logback.xml
+    and change the directory for log files by e.g.
+      -Dperun.log=/var/log/perun
```
zlamalp: I believe there should be slash at the end ``/var/log/perun/`` or it might not work (it's like that in config and pom.xml).

zlamalp:
test this please

zlamalp:
Since we changed logging, this look weird. NullPointerException in engine test, but tests are OK and build is success. We should look at it later.

```
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 5.337 sec - in cz.metacentrum.perun.engine.unit.TaskStatusManagerImplTest
Running cz.metacentrum.perun.engine.unit.TaskExecutorEngineImplTest
TaskExecutorEngineImpl.beginExecutingTest
2017-05-26 21:02:11,186 [main] ERROR cz.metacentrum.perun.engine.scheduling.impl.ExecutorEngineWorkerImpl ExecutorEngineWorkerImpl.java:167 - ERROR with TASK ID: 19058 , Exception:java.lang.NullPointerException
java.lang.NullPointerException: null
	at cz.metacentrum.perun.engine.jms.JMSQueueManager.reportFinishedDestination(JMSQueueManager.java:181) ~[classes/:na]
```

perun-ldapc/src/main/resources/log4j.properties
martin-kuba: I have introduced cz.metacentrum.perun.logging.PerunLogbackConfigurator into perun-base module which gets executed if no logback.xml is found on classpath. In that case it tries to load ${perun.conf.custom/logback.xml, /etc/perun/logback.xml and classpath:logback-default.xml in that order. Thus if a module (like perun-ldapc) contains its own logback.xml, it takes precedence.

perun-engine/src/main/resources/logback.xml
martin-kuba: changed

perun-engine/src/main/resources/logback.xml
martin-kuba: fixed

perun-base/src/main/resources/logback-default.xml
martin-kuba: fixed

martin-kuba:
The problem in TaskExecutorEngineImpl.beginExecutingTest is caused by null in session variable.

PR #1512, opened 2017-05-25
Optimalization of import from external source

zwejra:
- ExtSource*:
   - Methods getGroupSubjects now return string with status. It fills subjects through given parameter now.
   - Private methods changed to support change mentioned above.
---

 - Methods categorizeMembersForLightweightSynchronization and categorizeMembersForSynchronization were changed.
 - Now it uses map where key is login from needed external source and value is RichMember attached to this login.
 - Thanks to this is checking if user is already in group faster.
---

 - Methods to getCandidate were refactored. Now they count hash code of subject obtained from external source.
 - Hashcodes of subjects are stored in member-group attribute. If it has same value in the next synchronization cycle, we don't need to update this member's attributes.
---

 - Communication with external sources SQL, SQLComplex and LDAP was changed.
 - Now they can send us only modified members of group, when it is possible.
---

 - UserDuplicateException was created. It is thrown when candidate, which should be added to group, represents more than one User.


perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -16,24 +16,7 @@
...
-import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.MemberNotValidYetException;
-import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.PasswordCreationFailedException;
-import cz.metacentrum.perun.core.api.exceptions.PasswordOperationTimeoutException;
-import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthFailedException;
-import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
-import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.*;
```
licehammer: Unnecessary change. You didn't even do any other change in this file. Please remove it from the commit.

PR #1514, opened 2017-05-26
replaced 'now' with statement_timestamp() for postgres

martin-kuba:
In the Compatibility class, replaced the 'now' string which returns time of transaction start with statement_timestamp() function that returns time of statement execution


zlamalp:
Tested locally, works great. Now we need to update postgres DB definition, since timestamp columns deafult to now():
```sql
modified_at timestamp default now() not null
```

PR #1522, opened 2017-06-12
Api attributes

martin-kuba:
Implemented feature requested in https://homeproj.cesnet.cz/issues/3226 - Api.java reads request attributes instead of HTTP headers, and configured set of attributes of UserExtSOurce is updated.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/AttributeDefinition.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/AttributeDefinition.java
@@ -1,14 +1,8 @@
 package cz.metacentrum.perun.core.api;
 
-import cz.metacentrum.perun.core.api.Auditable;
-import cz.metacentrum.perun.core.api.BeansUtils;
-
 /**
  * This class represents definition of attribute. All attributes comes from some definition.
  * Attribute definition is attribute without connection to some object.
- * TODO
- *
- *
  *
  * @author Slavek Licehammer <glory@ics.muni.cz>
  */
@@ -134,8 +128,8 @@ public class AttributeDefinition extends Auditable implements Comparable<PerunBe
 	public String getFriendlyNameParameter() {
 		int index = friendlyName.indexOf(':');
 
-		if (index != -1 && index < friendlyName.length()-1) {
-			return friendlyName.substring(index+1);
+		if (index != -1 && index < friendlyName.length() - 1) {
+			return friendlyName.substring(index + 1);
 		} else return "";
 	}
 
@@ -150,10 +144,10 @@ public class AttributeDefinition extends Auditable implements Comparable<PerunBe
 	}
 
 	public int compareTo(PerunBean perunBean) {
-		if(perunBean == null) throw new NullPointerException("PerunBean to compare with is null.");
-		if(perunBean instanceof AttributeDefinition) {
+		if (perunBean == null) throw new NullPointerException("PerunBean to compare with is null.");
+		if (perunBean instanceof AttributeDefinition) {
 			AttributeDefinition attrDef = (AttributeDefinition) perunBean;
-			if (this.getFriendlyName()== null && attrDef.getFriendlyName() != null) return -1;
+			if (this.getFriendlyName() == null && attrDef.getFriendlyName() != null) return -1;
 			if (attrDef.getFriendlyName() == null && this.getFriendlyName() != null) return 1;
 			if (this.getFriendlyName() == null && attrDef.getFriendlyName() == null) return 0;
 			return this.getFriendlyName().compareToIgnoreCase(attrDef.getFriendlyName());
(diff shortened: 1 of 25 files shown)


martin-kuba:
It is ready to be merged. 

Before deploying, Shibboleth config in /etc/shibboleth/shibboleth2.xml must be modified by adding attribute **attributePrefix="AJP_"** to <ApplicationDefaults> element. This change enables passing Shibboleth attributes to Tomcat through mod_proxy_ajp module.

martin-kuba:
deleted a branch by mistake, reopening

perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
```diff
@@ -305,4 +306,95 @@ public String getUserExtSourcesPersistent() {
...
+					attr.setDisplayName("certificate DN");
+					attr.setDescription("Distinguished Name from X509 digital certificate");
+					break;
+				case "cadn":
+					attr.setDisplayName("CA DN");
+					attr.setDescription("Distinguished Name of Certificate Authority");
+					break;
+				case "certificate":
+					attr.setDisplayName("X509 certificate");
+					attr.setDescription("PEM encoded X509 certificate");
```
zlamalp: This attribute should be probably set to LargeString data type, otherwise on Oracle instance value is limited to 4000 bytes.

perun-base/src/main/java/cz/metacentrum/perun/core/api/AttributeDefinition.java
```diff
@@ -176,39 +170,32 @@ public int hashCode() {
...
 	public String serializeToString() {
-		StringBuilder str = new StringBuilder();
-
-		return str.append(this.getClass().getSimpleName()).append(":[").append(
-				"id=<").append(getId()).append(">").append(
-				", friendlyName=<").append(getFriendlyName() == null ? "\\0" : BeansUtils.createEscaping(getFriendlyName())).append(">").append(
-				", namespace=<").append(getNamespace() == null ? "\\0" : BeansUtils.createEscaping(getNamespace())).append(">").append(
-				", type=<").append(getType() == null ? "\\0" : BeansUtils.createEscaping(getType())).append(">").append(
-				']').toString();
+		return this.getClass().getSimpleName() + ":[" +
```
zlamalp: Any reason for changing usage of ``StringBuilder`` to string concatenation ? I believe we once changed all beans ``toString()`` and ``serializeToString()`` methods to use ``StringBuilder`` , since it should be quicker or less memory consuming.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
```diff
@@ -244,35 +268,24 @@ else if (extLogin == null && req.getAttribute("SSL_CLIENT_VERIFY") != null && ((
...
-				}
 			}
 		}
 
 		// Read all headers and store them in additionalInformation
-		String headerName = "";
-		for(Enumeration<String> headerNames = req.getHeaderNames(); headerNames.hasMoreElements();){
-			headerName = (String)headerNames.nextElement();
+		String headerName;
+		for (Enumeration<String> headerNames = req.getHeaderNames(); headerNames.hasMoreElements(); ) {
```
zlamalp: If original problem was, that headers can be malicious, we probably shouldn't save them all. Since like this we can overwrite trusted values from request attributes, which were stored to the same map previously.

perun-base/src/main/java/cz/metacentrum/perun/core/api/AttributeDefinition.java
martin-kuba: Yes, actually the compiler converts the concatenation of strings into calling a StringBuilder, so during runtime both versions are identical, and this one is simpler to read in source code. The only place where using StringBuilder explicitly makes sense is when the string is created in a loop.

This change was suggested by Intellij IDEA and its description said that it is suggested only if it is at least as efficient as the original code.

perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
martin-kuba: changed to LargeString

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
martin-kuba: Good point. I have checked where the additionalInformations is used, and it was always used to retrieve attributes from the federation or certificates, so I have removed storing the headers completely.

perun-base/src/main/java/cz/metacentrum/perun/core/api/AttributeDefinition.java
zlamalp: Thanks for the explanation :-).

martin-kuba:
Requested changes were pushed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
zlamalp: I checked and if needed, ``referer`` header is taken from servlet request. Also servlet request is present in deserializer. On side of GUI and WUI, there doesn't seem to be any usage of ``additionalAttributes`` outside federated ones. So I believe this change is safe. But as you noted, before deployment custom CGI scripts must be checked on each instance (for usage headers and perun session attributes).

zlamalp:
Please also fix ``cz.metacentrum.perun.webgui.json.JsonErrorHandler:L411``, where name of exception must be changed: ``AttributeExistsException`` -> ``AttributeDefinitionExistsException`` to match changes in perun-core. Thank you.

PR #1532, opened 2017-06-21
CORE: Fix read only connection for SQLite

zlamalp:
- SQLite doesn't support standard JDBC interface to set connection
  to read only mode. Hence we check DB type based on connection URL.
  If SQLite, we set connection properties, before creating actual connection.
  Otherwise we use standard API.
- Switch to using DriverManagerConnectionFactory constructor with properties.
- Fix indentation of createConnection() method.


zlamalp:
This should be tested first. I'm unsure about the case, when DB doesn't require any user/pass, then is probably shouldn't consume properties object at all.

zlamalp:
Tested locally, implementation is consistent in its behavior with previous version. Merging.

PR #1533, opened 2017-06-21
Add and modify method getGroupsWhereUserIsAdmin

stavamichal:
- modify existing version of method with user as parameter by adding
   also all groups, where user is groupAdmin by authorizedGroup, also
   change javadoc for this method and modify existing test
 - add new method with parameters user and vo which get the same result
   like the other method withou vo, but filter all returned groups by
   vo in parameter, also add new test for this method
 - add this new one also to perun-rpc

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
@@ -723,17 +723,32 @@ public enum UsersManagerMethod implements ManagerMethod {
 	},
 
 	/*#
-	 * Returns list of Groups, where the user is an Administrator.
+	 * Returns list of Groups in Perun, where the User is a direct Administrator
+	 * or he is a member of any group which is Administrator of some of these Groups.
 	 *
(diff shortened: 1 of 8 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -723,17 +723,32 @@ public User call(ApiCaller ac, Deserializer parms) throws PerunException {
...
-	 * Returns list of Groups, where the user is an Administrator.
+	 * Returns list of Groups in Perun, where the User is a direct Administrator
+	 * or he is a member of any group which is Administrator of some of these Groups.
 	 *
 	 * @param user int User <code>id</code>
 	 * @return List<Group> Found Groups
 	 */
+	/*#
+	 * Returns list of Groups in selected Vo, where the User is a direct Administrator
+	 * or he is a member of a group (in this Vo) which is Administrator of some of these Groups.
```
licehammer: Are you sure that the administrator group have to be from the same VO?

stavamichal:
You are right. I removed this specification from javadoc. User can be member of any group in Perun there.

PR #1550, opened 2017-07-13
Refactoring of attribute dependencies logic

stavamichal:
- instead one big method use a bulk of special private methods where
   every method takes care about specific way how to obtain
   richAttributes from input parameters
 - this is easier for adding new funcionality and testing existing one
   (lost of small tests than one big)
 - add all missing logic for dependencies and perunBean UserExtSource
   (newly also userExtSources can have attributes and these attributes
   of course can have dependencies to other attributes)
 - add new public method to MembersManagerBl and BlImpl to be able to
   say if some member is allowed or not (has status DISABLED or INVALID)


stavamichal:
We should add tests for every small method in this whole commit before merging it.

stavamichal:
This functionality with added tests and modifications is now in pull request https://github.com/CESNET/perun/pull/1598 Closing this one.

PR #1553, opened 2017-07-20
New method getMemberByExtSourceNameAndExtLogin

stavamichal:
- it is copy of method getUserByExtSourceNameAndExtLogin, just with Vo
   in parameter and return Member instead of User
 - this method try to find member by Vo, extSourceName and extLogin
 - add it to the rpc and add also test for it

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -993,6 +993,20 @@ public class MembersManagerEntry implements MembersManager {
 					 return getMembersManagerBl().canBeMemberWithReason(sess, vo, user, loa);
 	}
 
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -993,6 +993,19 @@ public boolean canBeMemberWithReason(PerunSession sess, Vo vo, User user, String
 					 return getMembersManagerBl().canBeMemberWithReason(sess, vo, user, loa);
 	}
 
+	public Member getMemberByExtSourceNameAndExtLogin(PerunSession sess, Vo vo, String extSourceName, String extLogin) throws ExtSourceNotExistsException, UserExtSourceNotExistsException, MemberNotExistsException, UserNotExistsException, InternalErrorException, VoNotExistsException, PrivilegeException {
+		Utils.checkPerunSession(sess);
+		Utils.notNull(extSourceName, "extSourceName");
+		Utils.notNull(extLogin, "extLogin");
+		getPerunBl().getVosManagerBl().checkVoExists(sess, vo);
+
+		if(!AuthzResolver.isAuthorized(sess, Role.VOADMIN)) {
```
zlamalp: I guess VOOBSERVER role should be allowed too. Otherwise code seems ok.

PR #1566, opened 2017-07-31
Added a method to RPC.

zdenek-strm:
The method returns group-resource attributes. Selection of specific attributes is possible.

The method is already available in BL, this commit adds it to RPC.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
@@ -498,6 +498,29 @@ public class AttributesManagerEntry implements AttributesManager {
 		return attributes;
 	}
 
+	public List<Attribute> getAttributes(PerunSession sess, Resource resource, Group group, List<String> attrNames, boolean workWithGroupAttributes) throws PrivilegeException, InternalErrorException, ResourceNotExistsException, GroupNotExistsException, GroupResourceMismatchException, WrongAttributeAssignmentException {
+		Utils.checkPerunSession(sess);
+		getPerunBl().getResourcesManagerBl().checkResourceExists(sess, resource);
+		getPerunBl().getGroupsManagerBl().checkGroupExists(sess, group);
+		List<Attribute> attributes = getAttributesManagerBl().getAttributes(sess, resource, group, attrNames, workWithGroupAttributes);
+		Iterator<Attribute> attrIter = attributes.iterator();
+		//Choose to which attributes has the principal access
+		while(attrIter.hasNext()) {
+			Attribute attrNext = attrIter.next();
+			if(getAttributesManagerBl().isFromNamespace(sess, attrNext, NS_GROUP_ATTR)) {
+				if(!AuthzResolver.isAuthorizedForAttribute(sess, ActionType.READ, attrNext, group, null)) attrIter.remove();
+				else attrNext.setWritable(AuthzResolver.isAuthorizedForAttribute(sess, ActionType.WRITE, attrNext, group, null));
+			} else if(getAttributesManagerBl().isFromNamespace(sess, attrNext, NS_GROUP_RESOURCE_ATTR)) {
+				if(!AuthzResolver.isAuthorizedForAttribute(sess, ActionType.READ, attrNext, group, resource)) attrIter.remove();
+				else attrNext.setWritable(AuthzResolver.isAuthorizedForAttribute(sess, ActionType.WRITE, attrNext, group, resource));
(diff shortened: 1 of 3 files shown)


licehammer:
Test this please

perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
```diff
@@ -498,6 +498,32 @@ public AttributesManagerEntry() {
 		return attributes;
 	}
 
+	public List<Attribute> getAttributes(PerunSession sess, Resource resource, Group group, List<String> attrNames, boolean workWithGroupAttributes) throws PrivilegeException, InternalErrorException, ResourceNotExistsException, GroupNotExistsException, GroupResourceMismatchException, WrongAttributeAssignmentException {
+		Utils.checkPerunSession(sess);
+		getPerunBl().getResourcesManagerBl().checkResourceExists(sess, resource);
+		getPerunBl().getGroupsManagerBl().checkGroupExists(sess, group);
+		if(!getPerunBl().getGroupsManagerBl().getVo(sess, group).equals(getPerunBl().getResourcesManagerBl().getVo(sess, resource))) {
```
zlamalp: There is exactly same check in Bl using private method from Bl. But if you still want to check, that both objects belong to the same VO, you can simply compare voId stored in both of them, like private method ``checkGroupIsFromTheSameVoLikeResource()`` in Bl does. There is no need to load objects from DB.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -269,12 +269,21 @@
...
 					}
-				}  else if (parms.contains("group")) {
+				} else if (parms.contains("group")) {
 					if (parms.contains("workWithGroupAttributes")) {
-						return ac.getAttributesManager().getAttributes(ac.getSession(),
-								ac.getResourceById(parms.readInt("resource")),
-								ac.getGroupById(parms.readInt("group")),
-								parms.readBoolean("workWithGroupAttributes"));
+						if (parms.contains("attrNames")) {
+							return ac.getAttributesManager().getAttributes(ac.getSession(),
```
zlamalp: When you add new combination of parameters for RPC method, you must add it to the special kind of javadoc above. See above ``getAttributes`` enum, find existing combination for group_resource attributes and add new entry below with added param ``attrNames``. Thank you.

zdenek-strm: Ok, thank you for explanation, javadoc is now added.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -111,6 +111,18 @@
 	 * @throw ResourceNotExistsException When Resource with <code>id</code> doesn't exist.
 	 */
 	/*#
+	 * Returns specified  Group-Resource attributes for selected Group and Resource.
+	 * If <code>workWithGroupAttributes == true</code> then also all non-empty Group attributes are returned.
```
zlamalp: This is not true, since this method takes list of ``attrNames`` => it returns also empty attributes for both ``group_resource`` and ``group`` namespace if ``workWithGroupAttributes == true``. If passed list of names is empty, it should return all non-empty attributes. Please also fix spaces in above line. Thank you.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -111,6 +111,19 @@
 	 * @throw ResourceNotExistsException When Resource with <code>id</code> doesn't exist.
 	 */
 	/*#
+	 * Returns all specified Group-Resource attributes for selected Group and Resource.
+	 * If <code>attrNames</code> is empty, it returns all non-empty attributes.
+	 * If <code>workWithGroupAttributes == true</code> then also Group attributes are returned. <code>False</code> is default.
```
zdenek-strm: *This is not true, since this method takes list of attrNames => it returns also empty attributes for both group_resource and group namespace if workWithGroupAttributes == true. If passed list of names is empty, it should return all non-empty attributes. Please also fix spaces in above line. Thank you.*

The javadoc was updated. Again it seems I deleted your previous comment. 
I will not use --amend in the future, but rather squash commits at the end of review.

PR #1567, opened 2017-08-01
Add member-group attributes to getRichMembersWithAttributesByNames

zdenek-strm:
Updated two overloaded methods convertMembersToRichMembersWithAttributes, so they will work with member-group attributes.

The goal is to enable the methods listed below (that call the convertMembersToRichMembersWithAttributes) work with member group attributes:

getRichMembersWithAttributesByNames(PerunSession sess, Group group, Resource resource, List<String> attrsNames)
getRichMembersWithAttributesByNames(PerunSession sess, Group group, List<String> attrsNames)
getRichMembersWithAttributes(PerunSession sess, Group group, List<AttributeDefinition> attrsDef)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
@@ -944,7 +944,7 @@ public interface AttributesManager {
 	 * @param group group to set on
 	 * @param resource resource to set on
 	 * @param attributes attribute to set
-	 * @param workWithUserAttributes true/false If true, we can use group attributes too
+	 * @param workWithGroupAttributes true/false If true, we can use group attributes too
 	 *
 	 * @throws PrivilegeException if privileges are not given
 	 * @throws GroupNotExistsException if the group doesn't exists in the underlying data source
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
@@ -164,9 +164,9 @@ public interface AttributesManager {
 	 * @throws PrivilegeException if privileges are not given
 	 * @throws ResourceNotExistsException if the resource doesn't exists in underlying data source
 	 * @throws MemberNotExistsException if the member doesn't have access to this resource
-	 * @throws WrongAttributeAssignmentException
+	 * @throws MemberResourceMismatchException
 	 */
-	List<Attribute> getAttributes(PerunSession sess, Resource resource, Member member) throws PrivilegeException, ResourceNotExistsException, InternalErrorException, MemberNotExistsException, ResourceNotExistsException, WrongAttributeAssignmentException;
+	List<Attribute> getAttributes(PerunSession sess, Resource resource, Member member) throws PrivilegeException, InternalErrorException, MemberNotExistsException, ResourceNotExistsException, MemberResourceMismatchException;
 
 	/**
 	 * Gets all <b>non-empty</b> attributes associated with the member on the resource and if workWithUserAttributes is
@@ -183,11 +183,11 @@ public interface AttributesManager {
 	 * @throws ResourceNotExistsException if the resource doesn't exists in underlying data source
 	 * @throws InternalErrorException if an exception raise in concrete implementation, the exception is wrapped in InternalErrorException
 	 * @throws MemberNotExistsException if the member doesn't have access to this resource
-	 * @throws WrongAttributeAssignmentException
+	 * @throws MemberResourceMismatchException
 	 *
 	 * !!WARNING THIS IS VERY TIME-CONSUMING METHOD. DON'T USE IT IN BATCH!!
 	 */
-	List<Attribute> getAttributes(PerunSession sess, Resource resource, Member member, boolean workWithUserAttributes) throws PrivilegeException, ResourceNotExistsException, InternalErrorException, MemberNotExistsException, WrongAttributeAssignmentException;
+	List<Attribute> getAttributes(PerunSession sess, Resource resource, Member member, boolean workWithUserAttributes) throws PrivilegeException, ResourceNotExistsException, InternalErrorException, MemberNotExistsException, MemberResourceMismatchException;
 
 	/**
 	 * Gets selected <b>non-empty</b> attributes associated with the member and the resource.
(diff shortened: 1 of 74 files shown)


zdenek-strm:
Thanks, I did not realize that. I will look into it again.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -236,6 +236,8 @@ public AttributesManagerBlImpl(AttributesManagerImplApi attributesManagerImpl) {
 				memberResourceAttributeNames.add(attributeName);
 			} else if(attributeName.startsWith(AttributesManager.NS_USER_FACILITY_ATTR)) {
 				userFacilityAttirbuteNames.add(attributeName);
+			} else if(attributeName.startsWith(AttributesManager.NS_MEMBER_GROUP_ATTR)) {
```
zlamalp: Please remove this, since this method doesn't provide member_group attributes and it belongs to the else branch where error is logged and silently skipped.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -254,6 +256,31 @@ public AttributesManagerBlImpl(AttributesManagerImplApi attributesManagerImpl) {
...
+		if(attrNames.isEmpty()) {
+			attributes.addAll(getAttributes(sess, member, group));
+			return attributes;
+		}
+
+		List<String> memberGroupAttributeNames = new ArrayList<>();
+
+		//possible inefficiency - additional iteration over attrNames
+		for(String attributeName: attrNames) {
+			if(attributeName.startsWith(AttributesManager.NS_MEMBER_GROUP_ATTR)) {
```
zlamalp: You can simplify this by using only positive branch. Because rest of attribute namespaces are already retrieved by above code (if possible and present in params), it makes no sense to log error "unsupported" in else branch when method as a whole returns it. Lists of attribute names are expected to be reasonable short, so iterating over it is OK.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -175,6 +175,26 @@
 	List<Attribute> getAttributes(PerunSession sess, Resource resource, Member member, List<String> attrNames, boolean workWithUserAttributes) throws InternalErrorException, WrongAttributeAssignmentException;
 
 	/**
+	 * Gets selected <b>non-empty</b> attributes associated with the member, group and the resource.
```
zlamalp: This is not true, when you pass list of attrNames, even empty attributes are retrieved. Non-empty attributes case is only when list of attrNames is empty.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -938,6 +970,48 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
 		return richMembers;
 	}
 
+	//also adds group-member attributes
+	public List<RichMember> convertMembersToRichMembersWithAttributes(PerunSession sess, Group group, List<RichMember> richMembers, List<AttributeDefinition> attrsDef)  throws InternalErrorException, WrongAttributeAssignmentException {
+		List<AttributeDefinition> usersAttributesDef = new ArrayList<AttributeDefinition>();
+		List<AttributeDefinition> membersAttributesDef = new ArrayList<AttributeDefinition>();
+		List<AttributeDefinition> groupAttributesDef = new ArrayList<AttributeDefinition>();
```
zlamalp: Better use "member_group" naming consistently, since we have also "group" namespace and it's confusing, if it contains group attributes or not. So, something lik: ``memberGroupAttributesDef``

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -906,6 +907,37 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
 		return richMembers;
 	}
 
+	// works with RESOURCE and GROUP
```
zlamalp: I guess rest of such methods doesn't have proper javadoc either, but it would help to determine, what method does or should do. Please add it for new methods at least.

zlamalp:
@licehammer what do you think about usage of ``WrongAttributeAssignmentException`` ? It was added to the methods, since attributes manager method returning member_group attributes throws it. I removed it in my pull #1575 but we use it generally, when objects (member-group) are not in a relation. I don't think it's OK, since exception covers case, when namespace of requested attribute doesn't match passed params. But this case is, they are not in relation, but Impl would simply return none or empty attributes. I guess we could provide different exception (I believed there already was).

zlamalp:
Hi, @zdenek-strm, please remove ``WrongAttributeAssignmentException`` from ``@throws`` and method annotation, it shouldn't be thrown by new getAttributes() method used inside anymore. Thank you.

zlamalp:
We merged original changes in #1602, removal of ``WrongAttributeAssignmentException`` needs to be revisited. I'll comment necessary changes in this pull-request, but when implementing, please create new branch and pull-request based on current master a cherry-picked your last commit. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
```diff
@@ -665,8 +661,10 @@
 	 * @throws WrongAttributeValueException if the attribute value is illegal
 	 * @throws WrongAttributeAssignmentException if attribute is not vo attribute
 	 * @throws WrongReferenceAttributeValueException
+	 * @throws GroupResourceMismatchException
+	 * @throws MemberResourceMismatchException
 	 */
-	void setAttributes(PerunSession sess, Vo vo, List<Attribute> attributes) throws PrivilegeException, InternalErrorException, VoNotExistsException, AttributeNotExistsException, WrongAttributeValueException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException;
+	void setAttributes(PerunSession sess, Vo vo, List<Attribute> attributes) throws PrivilegeException, InternalErrorException, VoNotExistsException, AttributeNotExistsException, WrongAttributeValueException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException, GroupResourceMismatchException, MemberResourceMismatchException;
```
zlamalp: Such methods shouldn't throw ``GroupResourceMismatchException`` and ``MemberResourceMismatchException``, because Group/Resource/Member is not method param. It seems, that in most of methods, it's caused by ``checkAttributeDependencies()`` called in ``AttributesManagerBlImpl.`` Please catch exception in this method since is uses data retrieved inside the logic and they should be correct. If not, wrap it to ``ConsistencyErrorException`` or ``InternalErrorException``. This should lower the span of new exceptions in most methods. Please make it as a new pull-request, I will review the rest of changes ofter this is fixed. Thank you.

zlamalp:
Changes were partially merged. Last commit will be reworked in own pull-request. Closing this on.

--- original comment
Such methods shouldn't throw GroupResourceMismatchException and MemberResourceMismatchException, because Group/Resource/Member is not method param. It seems, that in most of methods, it's caused by checkAttributeDependencies() called in AttributesManagerBlImpl. Please catch exception in this method since is uses data retrieved inside the logic and they should be correct. If not, wrap it to ConsistencyErrorException or InternalErrorException. This should lower the span of new exceptions in most methods. Please make it as a new pull-request, I will review the rest of changes ofter this is fixed. Thank you.

PR #1568, opened 2017-08-01
Creation of unix permission atrribute module and its test

Vojtech-Sassmann:
This class checks if the attribute value is correct unix permission mask. If not, exception is thrown.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_facility_attribute_def_def_homeDirUmaskTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_facility_attribute_def_def_homeDirUmaskTest.java
@@ -1,0 +1,69 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.Facility;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Test of unix permission mask attribute.
+ *
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+public class urn_perun_facility_attribute_def_def_homeDirUmaskTest {
+
+	private static PerunSessionImpl session;
+	private static Facility facility;
+	private static urn_perun_facility_attribute_def_def_homeDirUmask classInstance;
+
+	@Before
+	public void setUp() {
+		classInstance = new urn_perun_facility_attribute_def_def_homeDirUmask();
+		facility = new Facility();
+		session = mock(PerunSessionImpl.class, RETURNS_DEEP_STUBS);
+	}
+
+	@Test
+	public void testCheckAttributeValue() throws Exception {
+		System.out.println("testCheckAttributeValue()");
+
+		Attribute attributeToCheck = new Attribute();
+
+		attributeToCheck.setValue(null);
+		classInstance.checkAttributeValue(session, facility, attributeToCheck);
+
+		attributeToCheck.setValue("0542");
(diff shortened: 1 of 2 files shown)


licehammer:
Code is OK.

Before the merge to upstream, please discard the unnecessary merge commit and either squash the other two commits into a single one or rewrite the commit messages to fit the content of the commits.

Vojtech-Sassmann:
I have just fixed the problem with commits.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_homeDirUmask.java
```diff
@@ -0,0 +1,45 @@
...
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.FacilityAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.FacilityAttributesModuleImplApi;
+
+/**
+ * Checks if the value is valid unix permission mask.
+ *
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+public class urn_perun_entityless_attribute_def_def_homeDirUmask extends FacilityAttributesModuleAbstract implements FacilityAttributesModuleImplApi {
```
zlamalp: If this supposed to be facility attribute, class name must match it, right now it references "entityless" attribute, which is wrong. Please note, that during class instantiation, they are referenced by exact attribute URN, converted to class name string. So for ``urn:perun:facility:attribute-def:def:homeDirUmask`` you must have ``urn_perun_facility_attribute_def_def_homeDirUmask`` module.

Please fix test class name too.

zlamalp:
Just generic note. I believe, that ``umask != permissions``, we probably should update attribute displayName and description in a code and in production instances, @licehammer ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_homeDirUmask.java
Vojtech-Sassmann: Class names should be ok now.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_facility_attribute_def_def_homeDirUmaskTest.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class urn_perun_facility_attribute_def_def_homeDirUmaskTest {
+
+	private static PerunSessionImpl session;
+	private static Facility facility;
+	private static urn_perun_entityless_attribute_def_def_homeDirUmask classInstance;
+
+	@Before
+	public void setUp() {
+		classInstance = new urn_perun_entityless_attribute_def_def_homeDirUmask();
```
zlamalp: Please fix class name here too.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_facility_attribute_def_def_homeDirUmaskTest.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Test of unix permission mask attribute.
+ *
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+public class urn_perun_facility_attribute_def_def_homeDirUmaskTest {
+
+	private static PerunSessionImpl session;
+	private static Facility facility;
+	private static urn_perun_entityless_attribute_def_def_homeDirUmask classInstance;
```
zlamalp: Please rename variable too.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_facility_attribute_def_def_homeDirUmaskTest.java
Vojtech-Sassmann: Ok, now it should be fine.

PR #1571, opened 2017-08-02
Remove parameters from createSponsoredMember

Vojtech-Sassmann:
Changed the way of identifying the sponsor user. Now this method uses the sponsor's ID.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
@@ -108,14 +108,13 @@ public enum MembersManagerMethod implements ManagerMethod {
 	 * Creates a new sponsored member in a given VO and namespace.
 	 *
 	 * Can be called either by a user with role SPONSOR, in that case the user becomes the sponsor,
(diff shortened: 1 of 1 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -126,7 +125,12 @@ public RichMember call(ApiCaller ac, Deserializer params) throws PerunException
 			String password = params.readString("password");
 			Vo vo =  ac.getVoById(params.readInt("vo"));
 			String namespace = params.readString("namespace");
-			User sponsor = ac.getUsersManager().getUserByExtSourceNameAndExtLogin(ac.getSession(), params.readString("extSourceName"), params.readString("extLogin"));
+			User sponsor;
+			try {
+				sponsor = ac.getUserById(params.readInt("sponsor"));
+			} catch (RpcException exp) {
```
licehammer: You can use params.contains("sponsor") to check if the parameter is available. Using that, you don't have to deal with the exception anymore.

Vojtech-Sassmann: Ok, now it should be ok.

PR #1573, opened 2017-08-02
Added namespace validation in createAttribute method

Vojtech-Sassmann:
The createAttribute method now checks if the attribute.nameSpace is valid nameSpace.
Also, new ENTITY_TYPE constant with entity types has been created.


perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
```diff
@@ -95,6 +95,9 @@
 
 	public static final String ATTRIBUTES_REGEXP = "^[-a-zA-Z0-9.]+([:][-a-zA-Z0-9.]+)?$";
 
+	public static final String[] ENTITY_TYPES = {"facility", "facility:cluster", "resource", "member_resource", "member_group",
```
zlamalp: Please remove ``facility:cluster`` attribute type and all other references to it (NS_* constants). It's not used and never was. Thank you.

PR #1581, opened 2017-08-14
Implemented attribute value nullity check

Vojtech-Sassmann:
Created convertEmptyAttributeValueToNull to convert
empty values to null. Removed methods convertEmptyStringIntoNullInAttrValue
and convertBooleanFalseIntoNullInAttrValue.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -497,6 +497,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		// clasification of attributes to attributes to remove and attributes to set
 		List<Attribute> attributesToRemove = new ArrayList<Attribute>();
 		List<Attribute> attributesToSet = new ArrayList<Attribute>();
+		convertEmptyAttrValueToNull(attributes);
 		for(Attribute attribute : attributes) {
 			if (attribute.getValue() == null) {
 				attributesToRemove.add(attribute);
@@ -529,6 +530,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		// clasification of attributes to attributes to remove and attributes to set
 		List<Attribute> attributesToRemove = new ArrayList<Attribute>();
 		List<Attribute> attributesToSet = new ArrayList<Attribute>();
+		convertEmptyAttrValueToNull(attributes);
 		for(Attribute attribute : attributes) {
 			if (attribute.getValue() == null) {
 				attributesToRemove.add(attribute);
@@ -559,6 +561,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		// clasification of attributes to attributes to remove and attributes to set
 		List<Attribute> attributesToRemove = new ArrayList<Attribute>();
 		List<Attribute> attributesToSet = new ArrayList<Attribute>();
+		convertEmptyAttrValueToNull(attributes);
 		for(Attribute attribute : attributes) {
 			if (attribute.getValue() == null) {
 				attributesToRemove.add(attribute);
@@ -588,6 +591,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		// clasification of attributes to attributes to remove and attributes to set
 		List<Attribute> attributesToRemove = new ArrayList<Attribute>();
 		List<Attribute> attributesToSet = new ArrayList<Attribute>();
+		convertEmptyAttrValueToNull(attributes);
 		for(Attribute attribute : attributes) {
 			if (attribute.getValue() == null) {
 				attributesToRemove.add(attribute);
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -6058,6 +6087,43 @@ public void checkAttributeDependencies(PerunSession sess, RichAttribute richAttr
...
+			}
+		} else if (attribute.getValue() instanceof Boolean) {
+			if(attribute.getValue().equals(Boolean.FALSE)) {
+				attribute.setValue(null);
+			}
+		} else if (attribute.getValue() instanceof ArrayList) {
+			if(((ArrayList)attribute.getValue()).isEmpty()) {
+				attribute.setValue(null);
+			}
+		} else if (attribute.getValue() instanceof HashMap) {
```
zlamalp: We use ``LinkedHashMap`` as attribute value for a map. I would probably prefer solution, which checks against attribute type stored in a object (class names including large+className). What do you think @licehammer ? 
Now this method would ignore values of different java types, maybe use generic List/Map instead on ArrayList/HashMap ?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -6058,6 +6087,43 @@ public void checkAttributeDependencies(PerunSession sess, RichAttribute richAttr
...
+	public void convertEmptyAttrValueToNull(List<Attribute> attributes) {
+		attributes.forEach(this::convertEmptyAttrValueToNull);
+	}
+
+	/**
+	 * Checks if the attribute represents empty value. If so, converts it into null.
+	 *
+	 * @param attribute attribute to be checked if is null
+	 */
+	public void convertEmptyAttrValueToNull(Attribute attribute) {
```
zlamalp: Please add usage of this method to 
```java
public Attribute mergeAttributeValue(PerunSession sess, User user, Attribute attribute)
```
in ``AttributeManagerBlImpl`` where you should use input attribute object and call it before line

```java
if(attribute.getValue() == null) return storedAttribute;
```
so we can be sure, that empty attribute is really null.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
zlamalp: Hi @Vojtech-Sassmann, we decided to use your solution, it will be used as consistency check. So please fix ``HashMap`` to ``LinkedHashMap``. Also create else branch, where you will throw ``ConsistencyErrorException`` telling us, that somebody used invalid value type for attribute. Thank you.

Vojtech-Sassmann: Hi @zlamalp, I have just done these changes and after that some tests failed because they use an Integer attribute value type. Should I add another branch for Integer?

zlamalp: Hi, @Vojtech-Sassmann, yes, Integer is a valid data type for perun attributes. Sorry I didn't notice earlier.

Vojtech-Sassmann: @zlamalp and what values are considered as an empty value? Only null or even some others like 0 for instance?

zlamalp: Only ``null`` is considered as null/empty.

Vojtech-Sassmann: Now it should be all OK.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -6058,6 +6088,50 @@ public void checkAttributeDependencies(PerunSession sess, RichAttribute richAttr
...
+		} else if (attribute.getValue() instanceof ArrayList) {
+			if(((ArrayList)attribute.getValue()).isEmpty()) {
+				attribute.setValue(null);
+			}
+		} else if (attribute.getValue() instanceof LinkedHashMap) {
+			if(((LinkedHashMap)attribute.getValue()).isEmpty()) {
+				attribute.setValue(null);
+			}
+		} else {
+			throw new ConsistencyErrorException("Invalid attribute value type: " + attribute.getType());
```
zlamalp: Please include also actual value java class to the exception message, since our data types doesn't necessarily match it (LargeString, LargeArray for example).

Vojtech-Sassmann: Done.

PR #1584, opened 2017-08-14
User deletion and getUserExtSourceByExtLogin fix

Vojtech-Sassmann:
*Call to delete user's userExtSources moved to BI, also changed the way
of getting a list of userExtSources from DB. Now it uses the user's ID.
*In getUserExtSourceByExtLogin changed the SQL selecet so it can
directly return result.


Vojtech-Sassmann:
@zlamalp Done.

PR #1589, opened 2017-08-19
Repaired getAttributes(member,group)

VMecko:
Repaired getAttributes(sess, member, group, attrNames, workWithUserAttributes) method in AttributesManagerEntry. Implemented this method in BL layer.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
@@ -216,6 +216,25 @@ public interface AttributesManagerBl {
 	List<Attribute> getAttributes(PerunSession sess, Member member, Group group, boolean workWithUserAttributes) throws InternalErrorException, WrongAttributeAssignmentException;
 
 	/**
+	 * Get all attributes associated with the member in the group which have name in list attrNames.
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -216,6 +216,25 @@
 	List<Attribute> getAttributes(PerunSession sess, Member member, Group group, boolean workWithUserAttributes) throws InternalErrorException, WrongAttributeAssignmentException;
 
 	/**
+	 * Get all attributes associated with the member in the group which have name in list attrNames.
+	 * If workWithUserAttributes is true, get also all <b>non-empty</b> user and member attributes
```
zlamalp: Right now javadoc is not consistent with implementation. In case of ``workWithUserAttributes==false`` it returns member-group + member attributes, but javadoc state:  _"If workWithUserAttributes is true, get also all <b>non-empty</b> user and member attributes"_, which means, by default it should return only member-group attributes.

Based on API layer javadoc implementation is correct, so you should update javadoc in Bl layer to match it.

VMecko: Good point, thanks. Repaired in next commit.

PR #1591, opened 2017-08-21
Module for login-namespace:cesnet-eduroam

zwejra:
This module fills value of login in namespace cesnet-eduroam with value from entityLess attribute cesnetEduroamIdsCounter.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_cesnet_eduroam.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_cesnet_eduroam.java
@@ -1,0 +1,78 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.*;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+
+/**
+ * Class for filling logins in the cesnet-eduroam namespace.
+ *
+ * @author Jan Zvěřina <zverina@cesnet.cz>
+ */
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_cesnet_eduroam.java
```diff
@@ -0,0 +1,103 @@
...
+	 * @return Filled attribute
+	 * @throws InternalErrorException
+	 * @throws WrongAttributeAssignmentException
+	 */
+	@Override
+	public Attribute fillAttribute(PerunSessionImpl perunSession, User user, AttributeDefinition attribute) throws InternalErrorException, WrongAttributeAssignmentException {
+
+		// Get attribute with login of user in namespace cesnet-eduroam
+		Attribute filledAttribute = new Attribute(attribute);
+		try {
```
zwejra: Is this part to line 54 necessary? I saw it in login namespace for eduroam-vsup. I am not sure if this method is called when login already exists.

zlamalp: I think not, since ``eduroam-vsup`` is filled from existing ``vsup`` login-namespace. There you fill attribute from itself. I can't say, it's wrong, but ``fillAttribute()`` is called only, if value is empty - hence not filled anyway.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_cesnet_eduroam.java
```diff
@@ -0,0 +1,103 @@
...
+		if (idsCounter.getValue() == null) {
+			counterValue = new Integer(1000);
+		} else {
+			counterValue = (Integer) idsCounter.getValue();
+		}
+
+		// Set login of user in namespace cesnet-eduroam with value of counter
+		filledAttribute.setValue(counterValue);
+		try {
+			perunSession.getPerunBl().getAttributesManagerBl().setAttribute(perunSession, user, filledAttribute);
```
zlamalp: You are not supposed to store ``filledAttribute`` inside ``fillAttribute()`` method, just return it. Storing entityless attribute few lines below is OK.

PR #1592, opened 2017-08-22
RemoveSponsor method was added to the entry and also to the RPC.

zdenek-strm:


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -155,6 +155,25 @@ public RichMember call(ApiCaller ac, Deserializer params) throws PerunException
...
+	 * Removes sponsor of existing member.
+	 *
+	 * Can be called only by VO admin.
+	 *
+	 * @param sponsored int id of sponsored member, optional
+	 * @param sponsor int id of sponsoring user that is to be removed, optional
+	 */
+	removeSponsor {
+		@Override
+		public Void call(ApiCaller ac, Deserializer params) throws PerunException {
```
zlamalp: RPC/Core API is not consistent with opposite method sponsorMember().
```
sponsorMember(userId,userId)
removeSponsor(memberId,userId)
```
@licehammer do we really want it that way ? When params are named the same (sponsored/sponsor) ?

licehammer: After discussion with Pavel, we agreed on using parameters member and sponsor. Please change it accordingly.

zdenek-strm: Done. Do we still want to change sponsorMember(userId,userId) to accept member as a first parameter nstead of a user as discussed previously?

licehammer: Yes. Preferably in separate pull request.

licehammer:
Squash the commits into a single one and then we can merge it.

PR #1595, opened 2017-08-23
Add limit of concurent active synchronization

stavamichal:
- limit by default with value 10, can be changed by property
   perun.group.maxConcurentGroupsToSynchronize in perun.properties file
 - instead of execute all not-already running groups, add them to the
   queue and they will be executed as soon as there will be place for
   them (less running threads than current maximum)
 - forcing group adds group at the first place of the queue (it means
   it can be outrun by another forcing of group)
 - add some methods to work with this concurent queue of waiting groups


stavamichal:
Behavior of this pull request will be changed.

PR #1596, opened 2017-08-23
add another destination type (DUMMY), that makes the whole propagatio…

mvocu:
…n sucessfull without actually running SEND task

* adds propagation type DUMMY to the Destination
* dispatcher now sets tasks with no destinations to the error state
* engine does not try to propagate tasks that have DUMMY destination, they are considered done
* engine sets tasks with no destinations to the error state

See RT#214867.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
@@ -79,7 +79,8 @@ public class ServicesManagerImpl implements ServicesManagerImplApi {
 		"services.id as services_id, services.name as services_name, " +
 		"services.created_at as services_created_at, services.created_by as services_created_by, " +
 		"services.modified_by as services_modified_by, services.modified_at as services_modified_at, " +
-		"services.created_by_uid as services_created_by_uid, services.modified_by_uid as services_modified_by_uid";
+		"services.created_by_uid as services_created_by_uid, services.modified_by_uid as services_modified_by_uid, " +
+		"facility_service_destinations.propagation_type as f_s_des_propagation_type ";
 
 	public static final RowMapper<Service> SERVICE_MAPPER = new RowMapper<Service>() {
 		public Service mapRow(ResultSet rs, int i) throws SQLException {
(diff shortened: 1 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
```diff
@@ -131,8 +131,11 @@ public Destination mapRow(ResultSet rs, int i) throws SQLException {
 			destination.setModifiedAt(rs.getString("destinations_modified_at"));
 			destination.setModifiedBy(rs.getString("destinations_modified_by"));
 			try { // do not mind if the column is not in the results
-				if(rs.getString("f_s_des_propagation_type").equals(Destination.PROPAGATIONTYPE_SERIAL)) {
-					destination.setPropagationType(Destination.PROPAGATIONTYPE_SERIAL);
+				String ptype = rs.getString("f_s_des_propagation_type");
+				if(ptype.equals(Destination.PROPAGATIONTYPE_SERIAL) ||
```
zlamalp: In regards of need to display info in GUI, please add this optional reading of propagation type to ``RICH_DESTINATION_MAPPER`` too and update select statements to retrieve also ``f_s_des_propagation_type`` column.

mvocu: Revisited, but please review this, especially in RPC.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -757,16 +757,24 @@ public Destination call(ApiCaller ac, Deserializer parms) throws PerunException
 		public Destination call(ApiCaller ac, Deserializer parms) throws PerunException {
 			ac.stateChangingCheck();
 
+			Destination destination;
+			
+			if(parms.contains("propagationType")) {
```
zlamalp: Implementation is correct, just please double javadoc -> for each existing javadoc add new with "propagationType" param present.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -787,9 +795,17 @@ public Destination call(ApiCaller ac, Deserializer parms) throws PerunException
 		public List<Destination> call(ApiCaller ac, Deserializer parms) throws PerunException {
 			ac.stateChangingCheck();
 
+			Destination destination;
+
```
zlamalp: Implementation is correct, just please double javadoc -> for each existing javadoc add new with "propagationType" param present.

zlamalp:
Merged with local fixes.

PR #1600, opened 2017-08-24
Change behavior of threading for synchronization of groups

stavamichal:
- new concurent double ended queue for purpose of preparing groups in
   order to be synchronized
 - add some methods to work with this dequeue (putting and removing
   groups) with checking of possible duplicity
 - change definition of thread pool, now keys are threads itself, values
   are actually processing groups (null if thread is waiting for another
   group)
 - now thread is still running except interupting or timeouting, from
   start it always set everything to default and take another group (or
   wait if no such is waiting) from head of the queue with waiting
   groups
 - number of active concurent threads is limited by new property
   perun.group.maxConcurentGroupsToSynchronize with default value=10,
   this property can be override in perun.properties file for every
   instance separate
 - force propagation now does not start new synchronization, instead it
   add forced group at the head of the queue so it will be processed as
   soon as possible and skip other groups in the queue (it also means
   that can be outrun by another forcing different group)
 - there is still process synchronizeGroups running by cron tool every 5
   minutes but behavior of this process is now little different. It
   remove all interrupted or timeouted threads from threadpool and then
   it creates a new threads and starts them (maximum defined by property
   menationed above). After restarting threads it also add new groups to
   the queue (groups where is already time to run defined by interval)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -14,7 +14,7 @@ import java.util.TreeMap;
 
 import cz.metacentrum.perun.core.api.*;
 import cz.metacentrum.perun.core.api.exceptions.*;
-import cz.metacentrum.perun.core.api.exceptions.rt.WrongAttributeAssignmentRuntimeException;
+import cz.metacentrum.perun.core.impl.Utils;
 import cz.metacentrum.perun.core.implApi.ExtSourceApi;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -29,6 +29,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ConcurrentLinkedQueue;
 
 /**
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2578,4 +2585,64 @@ private boolean checkGroupsCycle(PerunSession sess, int resultGroupId, int opera
...
+	 * delayed processing of group
+	 * - method is synchronized so only one thread should add group to the queue at one time (this prevents
+	 * creating duplicities of groups in queue), but it also lasts more time if call at bulk
+	 *
+	 * @param group group to add to the queue
+	 * @param putAsFirst group will be put to the first place in queue (LIFO)
+	 * @return true if group was added, false if group was already in queue on any place, it always return true for synchronizeAsFirst
+	 * @throws InternalErrorException if group is null
+	 */
+	private synchronized boolean putGroupToQueueToBeSynchronized(Group group, boolean putAsFirst) throws InternalErrorException {
```
martin-kuba: The method `putGroupToQueueToBeSynchronized()` should not have the `synchronized` modifier, which synchronizes on "this" object, but it should rather have its content wrapped in `synchronized(queueOfGroupsToBeSynchronized){ ...}` which synchronizes on the queue itself. Otherwise the test for contains and addlast may not prevent duplicities because situatin may change between the test and the call.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -1207,76 +1197,93 @@ public void synchronizeGroups(PerunSession sess) throws InternalErrorException {
...
+				//if exception which produce fail of whole synchronization was thrown
+				boolean failedDueToException = false;
 
-			try {
-				log.debug("Synchronization thread for group {} has started.", group);
-				// Set the start time, so we can check the timeout of the thread
-				startTime = System.currentTimeMillis();
+				try {
+					//Take anouther group from the queue to synchronize it
+					this.group = GroupsManagerBlImpl.this.takeAnotherGroupFromQueueToBeSynchronized();
```
licehammer: Can return null, which causes NullPointerException on following statement.

martin-kuba: I suggest replacing the Deque implementation with LinkedBlockingDeque which provides takeFirst() method that blocks instead of returning null if deque is empty

PR #1603, opened 2017-08-28
Merge member attributes instead of overwrite if needed

stavamichal:
Merge member attributes instead of overwrite if needed

- sometimes we need to merge values of specific member attributes by
  synchronization of groups. For this purpose we can set list of member
  attributes to be merged instead of overwrite (what is default)
- replace existing method for merging user attributes by more general
  one and use this general method for purpose of specific user and
  member variant of this method
 - fix bad indentation in affected existing methods

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -23,6 +23,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Objects;
 
 import cz.metacentrum.perun.core.api.PerunClient;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4508,10 +4517,80 @@ public Attribute mergeAttributeValue(PerunSession sess, User user, Attribute att
...
+				storedAttribute = getPerunBl().getAttributesManagerBl().getAttribute(sess, member, attribute.getName());
+			} catch (AttributeNotExistsException e) {
+				throw new ConsistencyErrorException(e);
+			}
+
+			//if attribute to merge has null value, return the stored attribute
+			convertEmptyAttrValueToNull(attribute);
+			if(attribute.getValue() == null) return storedAttribute;
+
+			if (storedAttribute != null) {
```
licehammer: There is no way where storedAttribute can be null.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4508,10 +4517,80 @@ public Attribute mergeAttributeValue(PerunSession sess, User user, Attribute att
...
+				// If there were someting then find values which haven't been already stored
+				if (updatedList != null) {
+					for (String value : ((ArrayList<String>) attribute.getValue())) {
+						if (!updatedList.contains(value)) {
+							updatedList.add(value);
+						}
+					}
+					attribute.setValue(updatedList);
+				} else {
+					attribute.setValue(attribute.getValue());
```
licehammer: Does this do anything?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4508,10 +4517,80 @@ public Attribute mergeAttributeValue(PerunSession sess, User user, Attribute att
...
+				storedAttribute = getPerunBl().getAttributesManagerBl().getAttribute(sess, member, attribute.getName());
+			} catch (AttributeNotExistsException e) {
+				throw new ConsistencyErrorException(e);
+			}
+
+			//if attribute to merge has null value, return the stored attribute
+			if(attribute.getValue() == null) return storedAttribute;
+
+			// If there were someting then find values which haven't been already stored
+			if (storedAttribute != null) {
```
licehammer: Up to this point the code is exactly same as in the previous if-branch. Consider merging them together and move the branching based on attribute value type to his point.

stavamichal:
Reworked. Please check.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -14,15 +14,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
licehammer: It's better not to use wildcard imports.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4436,82 +4428,119 @@ public void doTheMagic(PerunSession sess, Member member, boolean trueMagic) thro
...
+				}
+				attribute.setValue(updatedList);
+			}
+		// Check type LinkedHashMap
+		} else if (attribute.getType().equals(LinkedHashMap.class.getName())) {
+			//Find values which haven't been already stored
+			LinkedHashMap<String, String> updatedMap = (LinkedHashMap<String, String>) storedAttribute.getValue();
+			if (updatedMap != null) {
+				LinkedHashMap<String, String> receivedMap = (LinkedHashMap<String, String>) attribute.getValue();
+				updatedMap.putAll(receivedMap);
```
licehammer: This means that if old map (updatedMap) and new map (recievedMap) have some same keys, values for such keys will be overwritten with corresponding value from new map.
This behavior is correct, but it's not intuitive. Therefore I would recommend to explicitly mention this in method's javadoc.

stavamichal: Fixed in javadoc of this method.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
stavamichal: Replaced back.

stavamichal:
Please check and merge if possible.

PR #1604, opened 2017-08-29
Added resource and group vomsRoles attributes.

Vojtech-Sassmann:
Added vomsRoles attributes for group and resource.
Also allowed null value in urn_perun_group_attribute_def_def_vomsGroupName.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_vomsRoles.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_vomsRoles.java
@@ -12,46 +12,46 @@ import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.GroupAttributesModuleAbstract;
 import cz.metacentrum.perun.core.implApi.modules.attributes.GroupAttributesModuleImplApi;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
- * Group vomGroupName attribute.
+ * Group vomsRoles attribute.
  *
- * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
  */
-public class urn_perun_group_attribute_def_def_vomsGroupName extends GroupAttributesModuleAbstract implements GroupAttributesModuleImplApi {
+public class urn_perun_group_attribute_def_def_vomsRoles extends GroupAttributesModuleAbstract implements GroupAttributesModuleImplApi {
 
-	private static final Pattern pattern = Pattern.compile("^[^<>&=]*$");
+	private final Pattern pattern = Pattern.compile("^[^<>&]*$");
 
 	@Override
+	@SuppressWarnings("unchecked")
 	public void checkAttributeValue(PerunSessionImpl perunSession, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
-		String vomsGroupName;
-
 		if(attribute.getValue() == null) {
 			return;
 		}
-
-		if(!(attribute.getValue() instanceof String)) {
-			throw new WrongAttributeValueException(attribute, "Wrong type of the attribute. Expected: String");
-		}
-
-		vomsGroupName = (String) attribute.getValue();
-
-		Matcher matcher = pattern.matcher(vomsGroupName);
-
-		if(!matcher.matches()) {
-			throw new WrongAttributeValueException(attribute, "Bad format of attribute vomsGroupName. It should not contain '<>&=' characters.");
+		try {
+			List<String> vomRoles = (List<String>) attribute.getValue();
+			for (String vomRole : vomRoles) {
+				Matcher matcher = pattern.matcher(vomRole);
+				if(!matcher.matches()) {
+					throw new WrongAttributeValueException(attribute, "Bad group vomsRoles value. It should not contain '<>&' characters.");
+				}
+			}
+		} catch (ClassCastException e) {
+			throw new WrongAttributeValueException(attribute, "Value should be a list of Strings.");
 		}
 	}
 
 	@Override
 	public AttributeDefinition getAttributeDefinition() {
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_vomsRoles.java
```diff
@@ -0,0 +1,59 @@
...
+	private final Pattern pattern = Pattern.compile("^[^<>&]*$");
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public void checkAttributeValue(PerunSessionImpl perunSession, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		if(attribute.getValue() == null) {
+			throw new WrongAttributeValueException(attribute, "Attribute value cannot be null.");
+		}
+
+		List<String> vomRoles = (List<String>) attribute.getValue();
```
zlamalp: I guess this is a place, where ``ClassCastException`` might be thrown, but it's outside the try block, opposite to resource version of same attribute module.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_vomsRoles.java
```diff
@@ -0,0 +1,59 @@
...
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class urn_perun_group_attribute_def_def_vomsRoles extends GroupAttributesModuleAbstract implements GroupAttributesModuleImplApi {
+
+	private final Pattern pattern = Pattern.compile("^[^<>&]*$");
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public void checkAttributeValue(PerunSessionImpl perunSession, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		if(attribute.getValue() == null) {
```
zlamalp: @licehammer, do we want to restrict null values for group version? I believe if group roles are not set, values of resource are used. If so, we need to allow null. Redmine issue is not specific about null values. Same question goes for #1605, I already merge it, but now I'm unsure. Not all groups need to have "name" in a VOMS, right ?

zlamalp: I talked with @licehammer, please allow ``null`` values for both attributes (group/resource) and also in attribute module ``urn_perun_group_attribute_def_def_vomsGroupName`` from #1605. Thank you.

Vojtech-Sassmann: It should be all ok now.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_vomsGroupName.java
```diff
@@ -28,9 +28,6 @@
 	public void checkAttributeValue(PerunSessionImpl perunSession, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
 		String vomsGroupName;
 
-		if(attribute.getValue() == null) {
```
zlamalp: Please make it ``if (attribute.getValue() == null) return;``. I'm not sure, that ``matcher.matches()`` would work on null value.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_vomsRoles.java
```diff
@@ -0,0 +1,55 @@
...
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class urn_perun_group_attribute_def_def_vomsRoles extends GroupAttributesModuleAbstract implements GroupAttributesModuleImplApi {
+
+	private final Pattern pattern = Pattern.compile("^[^<>&]*$");
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public void checkAttributeValue(PerunSessionImpl perunSession, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		try {
```
zlamalp: Please make it ``if (attribute.getValue() == null) return;``. I'm not sure, that ``matcher.matches()`` would work on null value, also ``for`` cycle would fail on ``NullPointerException``. Also, we do not usually catch ``ClassCastException``, it's up to you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_vomsRoles.java
```diff
@@ -0,0 +1,55 @@
...
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class urn_perun_resource_attribute_def_def_vomsRoles extends ResourceVirtualAttributesModuleAbstract implements ResourceVirtualAttributesModuleImplApi {
+
+	private final Pattern pattern = Pattern.compile("^[^<>&]*$");
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		try {
```
zlamalp: Please make it ``if (attribute.getValue() == null) return;``. I'm not sure, that ``matcher.matches()`` would work on null value, also ``for`` cycle would fail on ``NullPointerException``. Also, we do not usually catch ``ClassCastException``, it's up to you.

Vojtech-Sassmann: I prefer it this way because I can suppress the unchecked warning and handle the situation that could happen. Now it should be ok.

PR #1606, opened 2017-08-30
checkGroupIsFromTheSameVoLikeResource 

zdenek-strm:
and checkMemberIsFromTheSameVoLikeResource throw WrongAttributeAssignmentException which can be misleading. This exception should be thrown under different circumstances.

- This exception was therefore changed to GroupResourceMismatchException and MemberResourceMismatchException in above mentioned methods.

- Removed serveral instances of WrongAttributeAssignmentException, because it is no longer
thrown by checkGroupIsFromTheSameVoLikeResource and checkMemberIsFromTheSameVoLikeResource.

- A number of methods had to be updated to reflect new exceptions and the removal of WrongAttributeAssignmentException.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -108,7 +108,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 			}
 		} catch(AttributeValueException ex) {
 			throw new ConsistencyErrorException("Member is removed from all groups. There are no required attribute for this member. Member's attributes can be removed without problem.", ex);
-		} catch(WrongAttributeAssignmentException ex) {
+		} catch (MemberResourceMismatchException ex) {
 			throw new InternalErrorException(ex);
 		}
 
@@ -677,7 +677,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 		return getOnlyRichMembersWithAllowedStatuses(sess, this.getCompleteRichMembers(sess, vo, attrsNames), allowedStatuses);
 	}
 
-	public List<RichMember> getCompleteRichMembers(PerunSession sess, Group group, Resource resource, List<String> attrsNames, List<String> allowedStatuses) throws InternalErrorException, AttributeNotExistsException, ParentGroupNotExistsException, WrongAttributeAssignmentException {
+	public List<RichMember> getCompleteRichMembers(PerunSession sess, Group group, Resource resource, List<String> attrsNames, List<String> allowedStatuses) throws InternalErrorException, AttributeNotExistsException, ParentGroupNotExistsException, MemberResourceMismatchException {
 		return getOnlyRichMembersWithAllowedStatuses(sess, this.getRichMembersWithAttributesByNames(sess, group, resource, attrsNames), allowedStatuses);
 	}
 
@@ -779,7 +779,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 		return richMembersWithAttributes;
 	}
 
-	public List<RichMember> getRichMembersWithAttributesByNames(PerunSession sess, Group group, Resource resource, List<String> attrsNames) throws InternalErrorException, WrongAttributeAssignmentException, AttributeNotExistsException {
+	public List<RichMember> getRichMembersWithAttributesByNames(PerunSession sess, Group group, Resource resource, List<String> attrsNames) throws InternalErrorException, AttributeNotExistsException, MemberResourceMismatchException {
 		List<Member> members = new ArrayList<Member>();
 		members.addAll(perunBl.getGroupsManagerBl().getGroupMembers(sess, group));
 		List<RichMember> richMembers = this.convertMembersToRichMembers(sess, members);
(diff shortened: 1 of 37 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -780,7 +780,7 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
 		return richMembersWithAttributes;
 	}
 
-	public List<RichMember> getRichMembersWithAttributesByNames(PerunSession sess, Group group, Resource resource, List<String> attrsNames) throws InternalErrorException, WrongAttributeAssignmentException, AttributeNotExistsException {
+	public List<RichMember> getRichMembersWithAttributesByNames(PerunSession sess, Group group, Resource resource, List<String> attrsNames) throws InternalErrorException, AttributeNotExistsException, MemberResourceMismatchException {
```
zlamalp: I guess checking, that group and resource are from the same VO, would be more interesting. Since when it's true, there is no chance, that member wouldn't be from the same VO as resource. Hence you should catch ``MemberResourceMismatchException`` thrown by conversion method and wrap it as ``ConsistencyErrorException``. Then you can add explicit check on group and resource and throw necessary exception to API layer.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -677,7 +677,7 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
 		return getOnlyRichMembersWithAllowedStatuses(sess, this.getCompleteRichMembers(sess, vo, attrsNames), allowedStatuses);
 	}
 
-	public List<RichMember> getCompleteRichMembers(PerunSession sess, Group group, Resource resource, List<String> attrsNames, List<String> allowedStatuses) throws InternalErrorException, AttributeNotExistsException, ParentGroupNotExistsException, WrongAttributeAssignmentException {
+	public List<RichMember> getCompleteRichMembers(PerunSession sess, Group group, Resource resource, List<String> attrsNames, List<String> allowedStatuses) throws InternalErrorException, AttributeNotExistsException, ParentGroupNotExistsException, MemberResourceMismatchException {
 		return getOnlyRichMembersWithAllowedStatuses(sess, this.getRichMembersWithAttributesByNames(sess, group, resource, attrsNames), allowedStatuses);
```
zlamalp: Check group-resource, catch member-resource and convert - but it should be solved by itself by fixing ``getRichMembersWithAttributesByNames(sess, group, resource, attrNames)`` first (in below comment).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -949,7 +949,7 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
 	 * @throws InternalErrorException
 	 * @throws WrongAttributeAssignmentException
 	 */
-	public List<RichMember> convertMembersToRichMembersWithAttributes(PerunSession sess, Group group, Resource resource, List<RichMember> richMembers, List<AttributeDefinition> attrsDef)  throws InternalErrorException, WrongAttributeAssignmentException {
+	public List<RichMember> convertMembersToRichMembersWithAttributes(PerunSession sess, Group group, Resource resource, List<RichMember> richMembers, List<AttributeDefinition> attrsDef) throws InternalErrorException, MemberResourceMismatchException {
```
zlamalp: Maybe you can add also check on Group-Resource.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -896,7 +896,7 @@ public void synchronizeGroups(PerunSession sess) throws InternalErrorException,
 		return getGroupsManagerBl().getAllMemberGroups(sess, member);
 	}
 
-	public List<RichGroup> getRichGroupsAssignedToResourceWithAttributesByNames(PerunSession sess, Resource resource, List<String> attrNames) throws InternalErrorException, WrongAttributeAssignmentException, ResourceNotExistsException, PrivilegeException {
+	public List<RichGroup> getRichGroupsAssignedToResourceWithAttributesByNames(PerunSession sess, Resource resource, List<String> attrNames) throws InternalErrorException, ResourceNotExistsException, PrivilegeException, GroupResourceMismatchException {
```
zlamalp: Since group is not between params you shouldn't throw ``GroupResourceMismatchException``. It should be solved once ``getRichGroupsWithAttributesAssignedToResource()`` is fixed in GroupsManagerBl as noted in my previous comments.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/ResourceGroupAttributesModuleAbstract.java
```diff
@@ -24,7 +25,7 @@
  */
 public abstract class ResourceGroupAttributesModuleAbstract extends AttributesModuleAbstract implements ResourceGroupAttributesModuleImplApi{
 
-	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+	public void checkAttributeValue(PerunSessionImpl perunSession, Resource resource, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException, GroupResourceMismatchException {
```
zlamalp: Please remove ``GroupResourceMismatchException`` and ``MemberResourceMismatchException`` from all member-resource and group-resource attribute modules (api, abstract). If passed entities mismatch, it's programmers fault.
We want these exceptions mostly to be used in outer api (Api and Entry, sometimes it's in Bl/BlImpl).
Then please catch such exceptions in modules implementation and wrap them to ``InternalErrorException``.
This should result in lowering exceptions usages in normal methods from ``AttributesManager`` and back in normal managers.

PR #1609, opened 2017-08-30
Changed sponsorMember(userId,userId) 

zdenek-strm:
to accept member as a first parameter instead of a user.


martin-kuba:
There is no reason to keep the "Vo vo" parameter, as the member already specifies the VO.

zdenek-strm:
Thank you. I updated the code accordingly.

PR #1611, opened 2017-08-31
Search user optimization.

Vojtech-Sassmann:
Changed the way of searching richMembers with specific attributes.
It used to find all of his attributes and then filter them. Now it
directly search only requested attributes.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -696,8 +696,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	}
 
 	public List<RichMember> findCompleteRichMembers(PerunSession sess, Vo vo, List<String> attrsNames, String searchString) throws InternalErrorException, AttributeNotExistsException {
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1215,6 +1214,20 @@ public Vo getMemberVo(PerunSession sess, Member member) throws InternalErrorExce
...
 
+	public List<RichMember> findRichMembersWithAttributesInVo(PerunSession sess, Vo vo, String searchString,
+															  List<String> attrsNames) throws InternalErrorException {
+		List<RichMember> list = findRichMembersInVo(sess, vo, searchString);
+		List<AttributeDefinition> attrsDefs = new ArrayList<>();
+		for (String attrsName : attrsNames) {
+			try {
+				attrsDefs.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrsName));
+			} catch (AttributeNotExistsException e) {
+				throw new InternalErrorException("Attribute with given name: '" + attrsName + "' does not exist", e);
```
licehammer: Just skip non-existing attributes.

PR #1616, opened 2017-09-05
Revert "Change behavior of threading for synchronization of groups"

martin-kuba:
Reverts CESNET/perun#1600


martin-kuba:
Reverted for fixing some problems.

PR #1639, opened 2017-09-21
REGISTRAR: Prevent multiple application submission

zlamalp:
- When browser lags, it might allow users to send same application
  multiple times. Check on existing application inside might fail,
  since previous attempt is not processed yet (committed in DB).
  So we will use Set with synchronized access and each application
  is stored on processing start and removed on error or finish.
  In case of concurrency, AlreadyProcessingException is thrown
  and will be silently ignored in GUI.


zlamalp:
@martin-kuba can you please quickly look at this ? I just want to be sure it won't cause block for submitting regular applications. Thank you.

PR #1640, opened 2017-09-26
Method moveGroup to move one group under another group

balcirakpeter:
Created method moveGroup to move one group structure under another group in same vo
Created exception GroupMoveNotAllowed for that method
Created tests to check if method works corectly
Modified method updateGroup to allsow update parentGroupId if group was moved

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupMoveNotAllowedException.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupMoveNotAllowedException.java
@@ -1,0 +1,43 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+import cz.metacentrum.perun.core.api.Group;
+
+/**
+ * Exception thrown when the group cannot be moved, because it's not allowed
+ * f.e. Destination group is subGroup of Moving group, Moving group is already in destination group as subGroup...
+ *
+ * @author Peter Balcirak
+ */
+public class GroupMoveNotAllowedException extends PerunException {
+
+	private Group movingGroup;
+	private Group destinationGroup;
+
+	public GroupMoveNotAllowedException() {}
+
+	public GroupMoveNotAllowedException(String message) {
+		super(message);
+	}
+
+	public GroupMoveNotAllowedException(String message, Throwable cause) {
+		super(message, cause);
+	}
+
+	public GroupMoveNotAllowedException(Throwable cause) {
+		super(cause);
+	}
+
+	public GroupMoveNotAllowedException(String message, Group movingGroup, Group destinationGroup) {
+		super(message);
+		this.movingGroup = movingGroup;
+		this.destinationGroup = destinationGroup;
+	}
+
+	public Group getMovingGroup() {
+		return movingGroup;
+	}
+
+	public Group getDestinationGroup() {
+		return destinationGroup;
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
@@ -902,4 +902,17 @@ public interface GroupsManager {
 	 * @throws PrivilegeException
 	 */
 	List<Group> getGroupUnions(PerunSession sess, Group group, boolean reverseDirection) throws InternalErrorException, GroupNotExistsException, PrivilegeException;
+
+	/**
+	 * Move one group structure under another group in same vo or as top level group
+	 *
+	 * @param sess perun session
+	 * @param destinationGroup group to which is moving group moved, if it's null group will be moved as top level group
+	 * @param movingGroup group which is moved to destination group
+	 *
+	 * @throws InternalErrorException
+	 * @throws GroupNotExistsException
+	 * @throws PrivilegeException
+	 */
+	void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupNotExistsException, PrivilegeException, GroupMoveNotAllowedException;
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -408,6 +408,187 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		return group;
 	}
 
+	public Group updateParentGroupId(PerunSession sess, Group group) throws InternalErrorException {
+		if(group == null) throw new InternalErrorException("Group can't be null.");
+
+		//return group with updated parentGroupId
+		group = getGroupsManagerImpl().updateParentGroupId(sess, group);
+
+		return group;
+	}
+
+	public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupMoveNotAllowedException {
+
+		// check if moving group is null
+		if (movingGroup == null) {
+			throw new GroupMoveNotAllowedException("Moving group: " + movingGroup + " cannot be null.", movingGroup, destinationGroup);
+		}
+
+		// check if moving group is members group
+		if (movingGroup.getName().equals(VosManager.MEMBERS_GROUP)) {
+			throw new GroupMoveNotAllowedException("It is not possible to move Members group.", movingGroup, destinationGroup);
+		}
+
+		//if destination group is null, it means group will be moved as top level group
+		if (destinationGroup != null) {
+
+			// check if both groups are from same VO
+			if (destinationGroup.getVoId() != movingGroup.getVoId()) {
+				throw new GroupMoveNotAllowedException("Groups are not from same VO. Moving group: " + movingGroup + " has VO:" + movingGroup.getVoId() + " and destination group: " + destinationGroup + " has VO:" + movingGroup.getVoId() + ".", movingGroup, destinationGroup);
+			}
+
+			// check if moving group is the same as destination group
+			if (destinationGroup.getId() == movingGroup.getId()) {
+				throw new GroupMoveNotAllowedException("Moving group: " + movingGroup + " cannot be the same as destination group: " + destinationGroup + ".", movingGroup, destinationGroup);
+			}
+
+			// check if destination group is members group
+			if (destinationGroup.getName().equals(VosManager.MEMBERS_GROUP)) {
+				throw new GroupMoveNotAllowedException("It is not possible to move group under Members group.", movingGroup, destinationGroup);
+			}
+
+			// check if moving group is already under destination group
+			if (movingGroup.getParentGroupId() != null && destinationGroup.getId() == movingGroup.getParentGroupId()) {
+				throw new GroupMoveNotAllowedException("Moving group: " + movingGroup + " is already under destination group: " + destinationGroup + " as subGroup.", movingGroup, destinationGroup);
+			}
+
+			List<Group> movingGroupAllSubGroups = getAllSubGroups(sess, movingGroup);
+
+			// check if destination group exists as subGroup in moving group subGroups
+			if (movingGroupAllSubGroups.contains(destinationGroup)) {
+				throw new GroupMoveNotAllowedException("Destination group: " + destinationGroup + " is subGroup of Moving group: " + movingGroup + ".", movingGroup, destinationGroup);
+			}
+
+			// check if this operation would create cycle
+			if (checkGroupsCycle(sess, destinationGroup.getId(), movingGroup.getId())) {
+				throw new GroupMoveNotAllowedException("There would be cycle created after moving group: " + movingGroup + " under destination group: " + destinationGroup + ".", movingGroup, destinationGroup);
+			}
+
+			List<Group> destinationGroupSubGroups = getSubGroups(sess, destinationGroup);
+
+			// check if under destination group is group with same short name as Moving group short name
+			for (Group group: destinationGroupSubGroups) {
+				if(movingGroup.getShortName().equals(group.getShortName())){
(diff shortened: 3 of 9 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -2088,6 +2074,152 @@ public void getAssignedGroupsToFacility() throws Exception {
 
 	}
 
+	@Test (expected = GroupMoveNotAllowedException.class)
+	public void testMoveGroupTodiffVo() throws Exception {
+		System.out.println("\ntest moveGroup() to different vo");
```
zlamalp: We usually print test name like this: ``System.out.println(CLASS_NAME + "testMoveGroupTodiffVo");`` Please fix it in a rest of tests.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -232,6 +232,18 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
 			}
 			dbGroup.setDescription(group.getDescription());
 		}
+
```
zlamalp: I would prefer implementing own method, instead of using updateGroup(), since some people use our API directly and they could create inconsistency by changing ``parentGroupId``, but not doing the rest of logic.

balcirakpeter:
I've created new method for updating parentGroupId, changed test names and print.
I've also added check to group cycle in method moveGroup and test if that works correctly, removed wild cards for exceptions in GroupsmanagerBl and GroupsManagerEntryIntegrationTest.

stavamichal:
What about audit messages? How Engine, LDAPc or other connectors get info about this change? How does they process it?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -258,6 +259,31 @@ public Group updateGroupName(PerunSession sess, Group group) throws InternalErro
 		return dbGroup;
 	}
 
+	public Group updateParentGroupId(PerunSession sess, Group group) throws InternalErrorException {
+		Utils.notNull(group.getName(), "group.getName()");
```
zlamalp: Is this check necessary ? Technically, name should always exists, but we do not check other params and you do not work with group name in this method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
zlamalp: Regarding future necessity of audit message about moving group, you should probably create ``updateParentGroupId()`` in BL layer too, where audit message will be logged in a future. Then use BL version from ``moveGroup()``

balcirakpeter:
Method updateParentGroupId() created also in BL layer
Check for group name in updateParentGroupId() removed

perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
```diff
@@ -902,4 +902,17 @@
...
+	 *
+	 * @param sess perun session
+	 * @param destinationGroup group to which is moving group moved
+	 * @param movingGroup group which is moved to destination group
+	 *
+	 * @throws InternalErrorException
+	 * @throws GroupNotExistsException
+	 * @throws PrivilegeException
+	 */
+	void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupNotExistsException, PrivilegeException, GroupMoveNotAllowedException;
```
zlamalp: We should support the case "make group top-level", in such case ``destinationGroup`` would be null. Please allow it in a method implementation and update corresponding check before throwing ``GroupMoveNotAllowedException``.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,94 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+			throw new GroupMoveNotAllowedException("There would be cycle created after moving group: " + movingGroup + " under destination group: " + destinationGroup + ".");
+		}
+
+		//Save old parent id for moving indirect members
+		int oldParentGroupId = movingGroup.getParentGroupId();
+		// We can move whole moving group tree under destination group
+		movingGroup.setParentGroupId(destinationGroup.getId());
+		movingGroup.setName(destinationGroup.getName()+":"+movingGroup.getShortName());
+		//we have to update group name in database
+		getGroupsManagerImpl().updateGroupName(sess, movingGroup);
```
zlamalp: I believe, that in BL of normal updateGroup() or in DB ? we make sure, that on each hierarchy level, group name is unique. By moving group we could create two groups with same name on same level. Please check corresponding code to make sure this doesn't happen. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,94 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+	public Group updateParentGroupId(PerunSession sess, Group group) throws InternalErrorException {
+		//return group with updated parentGroupId
+		group = getGroupsManagerImpl().updateParentGroupId(sess, group);
+
+		return group;
+	}
+
+	public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupMoveNotAllowedException {
+
+		// check if moving group or destination group is null
```
zlamalp: Please make sure, that ``destinationGroup`` is never ``members`` group of a VO, since we do not allow subgroups for members groups. We have name constant for this somewhere. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,94 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+
+		// check if this operation would create cycle
+		if (checkGroupsCycle(sess, destinationGroup.getId(), movingGroup.getId())){
+			throw new GroupMoveNotAllowedException("There would be cycle created after moving group: " + movingGroup + " under destination group: " + destinationGroup + ".");
+		}
+
+		//Save old parent id for moving indirect members
+		int oldParentGroupId = movingGroup.getParentGroupId();
+		// We can move whole moving group tree under destination group
+		movingGroup.setParentGroupId(destinationGroup.getId());
```
zlamalp: In a case of new top-level move, set ``parentGroupId`` to ``null``.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,94 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+		movingGroup.setName(destinationGroup.getName()+":"+movingGroup.getShortName());
```
zlamalp: In a case of new top-level move, just use ``movingGroup.getShortName()``

balcirakpeter:
Added check if destination group is members group
Added check if group move would create conflict with group names
Added logs for audit
Implementation changed to allow move group also as top level group
Added new tests

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,178 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+				levelSubGroups.clear();
+				for (Group gr : subGroups) {
+					try {
+						gr.setName(getParentGroup(sess, gr).getName() + ":" + gr.getShortName());
+					} catch (ParentGroupNotExistsException ex) {
+						//that should never happen
+						throw new InternalErrorException("Parent group does not exists for group " + gr);
+					}
+					//we have to update each subGroup name in database
+					getGroupsManagerImpl().updateGroupName(sess, gr);
```
zlamalp: You are missing audit message about group update ``getPerunBl().getAuditer().log(sess, "{} updated.", gr);`` after this line, since all subgroups were changed, like in a case of moving group to top-level.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,178 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+		return group;
+	}
+
+	public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupMoveNotAllowedException {
+
+		// check if moving group is null
+		if (movingGroup == null) {
+			throw new GroupMoveNotAllowedException("Moving group: " + movingGroup + " cannot be null.");
+		}
+
```
zlamalp: Please add check, that "movingGroup" is NOT "Members" group too, like you do for destinationGroup.

balcirakpeter:
moveGroup() added to RPC layer
added chcek if movingGroup is membersGroup
added test for that check
added audit message about group update

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupMoveNotAllowedException.java
```diff
@@ -0,0 +1,24 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+/**
+ * Exception thrown when the group cannot be moved, because it's not allowed
+ * f.e. Destination group is subGroup of Moving group, Moving group is already in destination group as subGroup...
+ *
+ * @author Peter Balcirak
+ */
+public class GroupMoveNotAllowedException extends PerunException {
```
zlamalp: Can you please update this exception to contain ``movingGroup`` and ``destinationGroup`` objects ? I can later use them in GUI to display nice message and group names. Right now exception message contains only serialized object.

balcirakpeter:
Added movingGroup and destinationGroup objects to GroupMoveNotAllowedException.

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupMoveNotAllowedException.java
```diff
@@ -0,0 +1,35 @@
...
+
+	public GroupMoveNotAllowedException(Throwable cause) {
+		super(cause);
+	}
+
+	public GroupMoveNotAllowedException(String message, Group movingGroup, Group destinationGroup) {
+		super(message);
+		this.movingGroup = movingGroup;
+		this.destinationGroup = destinationGroup;
+	}
```
zlamalp: Please add also getters for ``movingGroup`` and ``destinationGroup``.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -408,6 +408,187 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
...
+			// check if under destination group is group with same short name as Moving group short name
+			for (Group group: destinationGroupSubGroups) {
+				if(movingGroup.getShortName().equals(group.getShortName())){
+					throw new GroupMoveNotAllowedException("Under destination group: " + destinationGroup + " is group with the same name as moving group: " + movingGroup + ".", movingGroup, destinationGroup);
+				}
+			}
+
+			//Save old parent id for moving indirect members
+			int oldParentGroupId = movingGroup.getParentGroupId();
```
stavamichal: This will throw NullPointerException (I tested it) when group has no parent group, because then this attribute is null and you can't save null to int.

balcirakpeter:
I've moved removing/adding indirect members before the process of moving group, so we won't need that Integer. Instead of that we are removing indirect members just in case movingGroup is not top level group.
I've also created some tests for that so I hope its working correctly.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -173,6 +173,28 @@ public Group call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Create union of two groups, where "operandGroup" is technically set as subgroup of "resultGroup".
```
zlamalp: This RPC javadoc is copy/pasted, please update it to represent moveGroup() logic.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -173,6 +173,28 @@ public Group call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @return Group Result group
+	 */
+	moveGroup {
+
+		@Override
+		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
+			ac.stateChangingCheck();
+
+			ac.getGroupsManager().moveGroup(ac.getSession(),
+					ac.getGroupById(parms.readInt("destinationGroup")),
```
zlamalp: Unlike in perun-core, there is no possibility to provide ``null`` for ``destinationGroup``, since it always retrieve group by provided ID. Please make ``destinationGroup`` param optional (check it's presence) to allow moving groups to top-level.

balcirakpeter:
Updated javadoc for processRelationMembers().
Fixed method call() in rpc and method moveGroup() in GroupsManagerEntry.

stavamichal:
Need to be merged after #1696

PR #1648, opened 2017-10-03
Add missing holders for attribute row mapper

stavamichal:
- some missing holders (PerunBeans affected by change of this
   attribute) can process and error or bad behavior while requesting
   attributes with value from Perun DB
 - add missing holders where it is possible
 - add missing group_member attributes to row mapper processing for
   virtual attributes

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
@@ -493,6 +493,18 @@ public class AttributesManagerImpl implements AttributesManagerImplApi {
 					} catch (InternalErrorException ex) {
 						throw new InternalErrorRuntimeException(ex);
 					}
+				} else if(this.attributesManagerImpl.isFromNamespace(sess, attribute, AttributesManager.NS_MEMBER_GROUP_ATTR_VIRT)) {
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -860,7 +872,7 @@ public RichAttributeRowMapper(AttributeRowMapper attributeRowMapper, RowMapper<P
 			return namedParameterJdbcTemplate.query("select " + getAttributeMappingSelectQuery("mem_res") + " from attr_names " +
 							"left join member_resource_attr_values mem_res on id=mem_res.attr_id and member_id=:mId and resource_id=:rId " +
 							"where namespace in ( :nSO,:nSD,:nSV ) and attr_names.attr_name in ( :attrNames )",
-					parameters, new AttributeRowMapper(sess, this, member, resource));
+					parameters, new AttributeRowMapper(sess, this, resource, member));
```
zlamalp: It doesn't have to be part of this pull-request but it would be nice to fix implementation of AttributeRowMapper and attribute modules API so that param order conform namespace name: ``member_resource -> member, resource`` like we do for ``member_group``, ``group_resource`` and ``user_facility``.

PR #1649, opened 2017-10-04
For reading text of message conversion to utf8 added.

zoraseb:


zlamalp:
@licehammer can you please check, it? I'm never sure when it's about input / locale changes - based on perl version, system locale etc. If I recall, it worked for us until now, but when Roman tried it, it caused encoding problems, so this is a reason for change.

licehammer:
Binmode on STDIN is useless in this case, because the scripts don't read anything from it. Binode on the individual file handles is that what you need.

You also need decode ARGV, because you might get UTF8 characters from that as well. You can do that using following code. That should be applied to all scripts where UFT8 might be in arguments of that scripts.
```perl
use Encode qw(decode_utf8);
@ARGV = map { decode_utf8($_, 1) } @ARGV;
```

Before you will apply this, please test it thoroughly. If there is another decoding later on the way, it will mess up the text.

zlamalp:
@licehammer can you please look at it again ? It should be fixed.

licehammer:
Code looks good now. But I would recommend to squash the commits, because the second one is undoing all the work done in the first one.

PR #1654, opened 2017-10-10
GR:D:adName valueCheck edit

Vojtech-Sassmann:
* The attribute module urn:perun:group_resource:attribute-def:def:adName
now checks, if the given value is a String and if it contains only
letters, digits, hyphens or dots.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_adName.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_adName.java
@@ -17,6 +17,7 @@ import cz.metacentrum.perun.core.implApi.modules.attributes.ResourceGroupAttribu
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 /**
  * AD Name module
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_adName.java
```diff
@@ -34,15 +35,26 @@
...
 
 		//Attribute can be null
 		if(attribute.getValue() == null) {
 			return;
 		}
 
+		if(attribute.getValue() instanceof String) {
+			attributeValue = (String) attribute.getValue();
+			if (!pattern.matcher(attributeValue).matches()) {
+				throw new WrongAttributeValueException(attribute, "Invalid attribute adName value. It should contain only letters, digits, hyphens or dots.");
```
zlamalp: Please mention also underscore as allowed character, since its allowed by ``\\w`` in a regex and its correct by redmine task description.

Also, isn't regex too complex ? @licehammer I believe ``\\w`` contains ``\\d`` - https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html so we can use only ``\\w`` ?

Vojtech-Sassmann: @zlamalp 
I will mention the underscore. And I will siplify the regex, I thought the \\w corresponds to [a-z_A-Z].
My bad.

Vojtech-Sassmann:
@zlamalp Now it should be OK.

Vojtech-Sassmann:
Hi @zlamalp,
could you please take a look at this pull request? It should be all fine.

zlamalp:
@licehammer please tell us, if you want to allow spaces and diacritic, since current attribute values doesn't match regex restriction.

zlamalp:
Implementation is correct, existing value will be fixed directly on IDM instance.

PR #1658, opened 2017-10-16
CORE: Check on UES existence must check ID too

zlamalp:
- We initially checked only UES ID, but it caused errors in
  synchronization, so we switched too check on login/extSource.
  But dropping check on ID causes updateUserExtSource to fail,
  since identity might change in posted data but ID prevails.
  So if UES is not found by login/extSource, we try to find it by its
  ID (if not 0) in source object.
- We can now update login and loa form the CLI by updateUserExtSource
  tool.
- Added necessary API call definitions to UsersAgent.pm.
- Correctly align help output in some notif CLI tools.
- Updated UserExtSource definition in CLI (added userId and persistent
  params).

### DIFF
--- a/perun-cli/removeNotifTemplateMessage
+++ b/perun-cli/removeNotifTemplateMessage
@@ -8,12 +8,12 @@ use Perun::Common qw(printMessage);
 
 sub help {
 	return qq{
-Removes NotifTemplateMessage from db by id. Id is required field.
-------------------------------------
-Available options:
- --NotifTemplateMessageId        | -i id of the NotifTemplateMessage
- --batch                         | -b batch
- --help                          | -h prints this help
+	Removes NotifTemplateMessage from db by id. Id is required field.
+	------------------------------------
+	Available options:
+	--NotifTemplateMessageId        | -i id of the NotifTemplateMessage
+	--batch                         | -b batch
+	--help                          | -h prints this help
 
 };
 }
--- a/perun-cli/removeNotifTemplateMessage
+++ b/perun-cli/removeNotifTemplateMessage
@@ -14,7 +14,7 @@ Available options:
  --NotifTemplateMessageId        | -i id of the NotifTemplateMessage
  --batch                         | -b batch
  --help                          | -h prints this help
- 
+
 };
 }
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
@@ -362,9 +362,12 @@ public class UsersManagerEntry implements UsersManager {
 			throw new PrivilegeException(sess, "updateUserExtSource");
 		}
 
-		getUsersManagerBl().checkUserExtSourceExists(sess, userExtSource);
+		if (getUsersManagerBl().checkUserExtSourceExistsById(sess, userExtSource.getId())) {
+			return getUsersManagerBl().updateUserExtSource(sess, userExtSource);
+		} else {
+			throw new UserExtSourceNotExistsException("UserExtSource: " + userExtSource);
+		}
 
-		return getUsersManagerBl().updateUserExtSource(sess, userExtSource);
 	}
 
 	public List<UserExtSource> getUserExtSources(PerunSession sess, User user) throws InternalErrorException, UserNotExistsException, PrivilegeException {
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
@@ -362,11 +362,9 @@ public class UsersManagerEntry implements UsersManager {
 			throw new PrivilegeException(sess, "updateUserExtSource");
 		}
 
-		if (getUsersManagerBl().checkUserExtSourceExistsById(sess, userExtSource.getId())) {
-			return getUsersManagerBl().updateUserExtSource(sess, userExtSource);
-		} else {
-			throw new UserExtSourceNotExistsException("UserExtSource: " + userExtSource);
-		}
+		getUsersManagerBl().checkUserExtSourceExistsById(sess, userExtSource.getId());
+
+		return getUsersManagerBl().updateUserExtSource(sess, userExtSource);
 
 	}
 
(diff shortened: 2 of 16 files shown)


perun-cli/Perun/Agent.pm
```diff
@@ -143,6 +143,10 @@ sub call
 	my $fullUrl = "$self->{_url}/$format/$class/$method";
 
 	my $content = $self->{_jsonXs}->encode( $hash );
+
+	#print $content;
+	#print "\n\n\n";
```
stavamichal: Please remove this debug part, or use it in more suitable way like "DEBUG=0; if(DEBUG) { print something; } if really needed.

perun-cli/removeNotifTemplateMessage
```diff
@@ -8,13 +8,13 @@ use Perun::Common qw(printMessage);
...
- --NotifTemplateMessageId        | -i id of the NotifTemplateMessage
- --batch                         | -b batch
- --help                          | -h prints this help
- 
+	Removes NotifTemplateMessage from db by id. Id is required field.
+	------------------------------------
+	Available options:
+	--NotifTemplateMessageId        | -i id of the NotifTemplateMessage
+	--batch                         | -b batch
+	--help                          | -h prints this help
```
stavamichal: This change (even if it is in commit message) is just totally different from purpose of this commit so I would prefer to separate it from this one.

stavamichal: The same for the other similar changes in this commit which fixes indentation in different scripts.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -939,23 +939,42 @@ public boolean userExtSourceExists(PerunSession sess, UserExtSource userExtSourc
...
+		}
+
+		try {
+			// check by UES ID if present in object
+			if (!exists && userExtSource.getId() > 0) {
+				exists = 1 == jdbc.queryForInt("select 1 from user_ext_sources where id=?",
+						userExtSource.getId());
+			}
+
+			return exists;
```
stavamichal: I am afraid that this specific implementation will return "true" even if there is userExtSources with wrong ID (id of different user ext source) and correct combination of login and extSourceId. Maybe the best solution will be create two implementation of this method: One checking by id of user ext source, one checking by login and other attributes and then use implementation you need to on the higher level.

perun-cli/removeNotifTemplateMessage
licehammer: I agree with @stavamichal 

BTW, do you know this indent will be visible in the output? If you want to prevent that either do not indent here or use proper quoting. Example follows:
```perl
sub help {
    return "Removes NotifTemplateMessage from db by id. Id is required field.\n" .
           "------------------------------------\n";
```

zlamalp: @stavamichal I moved it to own commit. @licehammer yes, I know indentation is visible and it's like this in all our CLI tools (help is then more readable). Unless we decide to change it globally, I would prefer keeping my change to this tool.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
zlamalp: @stavamichal yes, you are right. But then we probably need to talk about purpose of this method and its current usage in a code. I wanted to avoid creating two similar methods, since then we would need to call both in some cases (and who will remember this while implementing). We usually don't check object on exact match of all params with DB. We usually use only ID if present and this check is used in entry or before new object creation. But we heavily use this method when manipulating users UserExtSources to check, if passed identity is already registered or not. ID is not much relevant (most times not present in passed object).
So creating two methods is a possible solution, but we need to find all usages and properly decide what to do and update javadoc.

stavamichal: @licehammer can you please give your opinion on this thing? If you are ok with it, I will approve it at current state.

licehammer: Indeed we are using the check methods to either validate the the ID of the object and also for checking if the object is not in the DB (in cases where the object have to be unique).

Do we have any use-case where you don't want to check other attributes of the object (integrity check) when you have the ID filled?

Anyway, this cannot be merged as it's written right now. The code is very unclear and I'm pretty sure that it can be written in more comprehensive way.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -916,6 +916,16 @@ public void checkUserExtSourceExists(PerunSession sess, UserExtSource userExtSou
 		if(!userExtSourceExists(sess, userExtSource)) throw new UserExtSourceNotExistsException("UserExtSource: " + userExtSource);
 	}
 
+	public boolean checkUserExtSourceExistsById(PerunSession sess, int id) throws InternalErrorException {
```
stavamichal: Check methods do not return boolean. As we are using them, they always end with or without error without any returned output.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/UsersManagerImplApi.java
```diff
@@ -397,6 +397,15 @@
 	 */
 	void checkUserExtSourceExists(PerunSession perunSession, UserExtSource userExtSource) throws InternalErrorException, UserExtSourceNotExistsException;
 
+	/**
+	 * Check if userExtSource exists in underlaying data source by its ID.
+	 *
+	 * @param perunSession
+	 * @param id
+	 * @throws InternalErrorException
```
stavamichal: There should be information about returned value of this method. But check methods should not return any value.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -362,9 +362,12 @@ public UserExtSource updateUserExtSource(PerunSession sess, UserExtSource userEx
 			throw new PrivilegeException(sess, "updateUserExtSource");
 		}
 
-		getUsersManagerBl().checkUserExtSourceExists(sess, userExtSource);
+		if (getUsersManagerBl().checkUserExtSourceExistsById(sess, userExtSource.getId())) {
```
stavamichal: Check methods should throw exception instead of returning value.

zlamalp: Reworked to throw exception instead of returning boolean. Rest of comments is not relevant then.

PR #1659, opened 2017-10-16
checkAttributeValue methods for Entityless and Hosts

balcirakpeter:
- Created classes HostAttributesModuleImplApi and HostAttributesModuleAbstract,
  which are needed for checkAttributeValue method for Hosts.
- Implemented getHostAttributeModule, which is used in method checkAttributeValue for Hosts.
- Implemented checkAttributeValue methods in Impl layer for entityless and hosts,
  because there was no implementation in that layer for these entities and attributeModules
  (specialy for Entityless) didn't check anything.
- Implemented changedAttributeHook for Hosts in Impl layer, which was in TODOs in couple of methods.
  These TODOs were uncommented.
- Implemented fillAttribute for Hosts in Impl layer,
  which was marked as TODO due to missing class HostAttributesModuleAbstract.
- in method setAttributeWithoutCheck for Entityless was TODO comment which was saying,
  that there isn't implemented method changedAttributeHook for Entityless,
  but in fact this method is already implemented, so this TODO was uncommented.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
@@ -3215,7 +3215,7 @@ public interface AttributesManager {
 	 * @throws AttributeNotExistsException if given attribute doesn't exist
 	 * @throws HostNotExistsException if specified host doesn't exist
 	 */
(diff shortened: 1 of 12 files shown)


zlamalp:
@balcirakpeter please also fix implementation of modules for entityless attributes. There is a wrong implementation of max/minUID or GID, where check value is performed against itself.

balcirakpeter:
Fixed entityless attribute modules - minUID, minGID, maxUID, maxGID

PR #1660, opened 2017-10-17
AttributeRowMapper fix

Vojtech-Sassmann:
* AttributeRowMapper changed to abstract and generic class so the order
of attribute holders can be managed using subclasses.
* Created generic subclass SingleBeanAttributeRowMapper which is used
for only one attribute holder
* Created multiple subclasses for pairs of attribute holders, e.g.
UserFacilityAttributeRowMapper


martin-kuba:
Well done.

PR #1661, opened 2017-10-17
Bona fide attributes

Vojtech-Sassmann:
* Creates attribute
urn:perun:user:attribute-def:virt:elixirBonaFideStatus
* This attribute is used for to indicate that a user is a
researcher. If so, it provides URL:
'http://www.ga4gh.org/beacon/bonafide/ver1.0'.
* The decision depends on attribute 'elixirBonaFideStatusREMS',
if it is not empty then the user is a researcher. If it is
empty, then this module searches in user's affiliations. If
any of those affiliations starts with 'faculty@' then the
user is a researcher as well. Otherwise, null value is set.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
@@ -1,0 +1,135 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleImplApi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This module determines if user is a researcher. If so,
+ * it provides URL: 'http://www.ga4gh.org/beacon/bonafide/ver1.0'.
+ *
+ * The decision depends on attribute 'elixirBonaFideStatusREMS', if it
+ * is not empty then the user is a researcher. If it is empty, then this module
+ * searches in user's affiliations. If any of those affiliations starts
+ * with 'faculty@' then the user is a researcher as well. Otherwise,
+ * null value is set.
+ *
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class urn_perun_user_attribute_def_virt_elixirBonaFideStatus extends UserVirtualAttributesModuleAbstract implements UserVirtualAttributesModuleImplApi {
+
+	private final static Logger log = LoggerFactory.getLogger(urn_perun_user_attribute_def_virt_elixirBonaFideStatus.class);
+
+	private final Pattern userSetPattern = Pattern.compile("Attribute:\\[(.*)friendlyName=<elixirBonaFideStatusREMS>(.*)] set for User:\\[(.*)]");
+	private final Pattern userRemovePattern = Pattern.compile("AttributeDefinition:\\[(.*)friendlyName=<elixirBonaFideStatusREMS>(.*)] removed for User:\\[(.*)]");
+	private final Pattern userAllAttrsRemovedPattern = Pattern.compile("All attributes removed for User:\\[(.*)]");
+	private final Pattern uesAllAttrsRemovedPattern = Pattern.compile("All attributes removed for UserExtSource:\\[(.*)]");
+	private final Pattern uesSetAffiliationAttributePattern = Pattern.compile("Attribute:\\[(.*)friendlyName=<affiliation>(.*)] set for UserExtSource:\\[(.*)]");
+	private final Pattern uesRemoveAffiliationAttributePattern = Pattern.compile("AttributeDefinition:\\[(.*)friendlyName=<affiliation>(.*)] removed for UserExtSource:\\[(.*)]");
+
+	private static final String FRIENDLY_NAME = "elixirBonaFideStatus";
+	private static final String URL = "http://www.ga4gh.org/beacon/bonafide/ver1.0";
+
+	private static final String USER_REMS_ATTR_NAME = "elixirBonaFideStatusREMS";
+	private static final String USER_AFFILIATIONS_ATTR_NAME = "eduPersonScopedAffiliations";
+
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDefinition attributeDefinition) throws InternalErrorException {
+		Attribute attribute = new Attribute(attributeDefinition);
+
+		try {
+			Attribute statusAttr = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_DEF + ":" + USER_REMS_ATTR_NAME);
+			String statusAttrValue = (String) statusAttr.getValue();
+			if (statusAttrValue != null && !statusAttrValue.isEmpty()) {
+				attribute.setValue(URL);
+			}
+		} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+			log.error("Cannot read {} from user {}", USER_REMS_ATTR_NAME, user, e);
+		}
+
+		if (attribute.getValue() == null) {
+			try {
+				Attribute userAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_VIRT + ":" + USER_AFFILIATIONS_ATTR_NAME);
+				if (userAttribute.getValue() != null) {
+					List<String> affiliations = (List<String>) userAttribute.getValue();
+					for (String affiliation : affiliations) {
+						if (affiliation != null && affiliation.startsWith("faculty@")) {
+							attribute.setValue(URL);
+							break;
+						}
+					}
+				}
+			} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+				log.error("Cannot read {} from user {}", USER_AFFILIATIONS_ATTR_NAME, user, e);
+			}
+		}
+
+		return attribute;
+	}
+
+	@Override
+	public List<String> resolveVirtualAttributeValueChange(PerunSessionImpl sess, String message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<String> resolvingMessages = new ArrayList<>();
+		if (message == null) return resolvingMessages;
+
+		Matcher userSetMatcher = userSetPattern.matcher(message);
+		Matcher userRemoveMatcher = userRemovePattern.matcher(message);
+		Matcher userAllAttrsRemovedMatcher = userAllAttrsRemovedPattern.matcher(message);
+		Matcher uesAllAttrsRemovedMatcher = uesAllAttrsRemovedPattern.matcher(message);
+		Matcher uesSetAffiliationAttributeMatcher = uesSetAffiliationAttributePattern.matcher(message);
+		Matcher uesRemoveAffiliationAttributeMatcher = uesRemoveAffiliationAttributePattern.matcher(message);
+
+		User user;
+		Attribute attribute;
+
+		if (uesSetAffiliationAttributeMatcher.find() ||	uesRemoveAffiliationAttributeMatcher.find() ||	uesAllAttrsRemovedMatcher.find() ||
+				userSetMatcher.find() || userRemoveMatcher.find() || userAllAttrsRemovedMatcher.find()) {
+
+			user = sess.getPerunBl().getModulesUtilsBl().getUserFromMessage(sess, message);
+			if (user != null) {
+				String messageAttributeSet;
+
+				attribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_VIRT + ":" + FRIENDLY_NAME);
+				String value = (String) attribute.getValue();
+
+				if (value == null || value.isEmpty()) {
+					AttributeDefinition attributeDefinition = new AttributeDefinition(attribute);
+					messageAttributeSet = attributeDefinition.serializeToString() + " removed for " + user.serializeToString() + ".";
+				} else {
+					messageAttributeSet = attribute.serializeToString() + " set for " + user.serializeToString() + ".";
+				}
+				resolvingMessages.add(messageAttributeSet);
+			} else {
+				log.error("Failed to get user from message: {}", message);
+			}
+		}
+		return resolvingMessages;
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_USER_ATTR_VIRT);
+		attr.setFriendlyName(FRIENDLY_NAME);
+		attr.setDisplayName("Bona fide researcher status");
+		attr.setType(String.class.getName());
+		attr.setDescription("Flag if user is qualified researcher. URI ‘http://www.ga4gh.org/beacon/bonafide/ver1.0’ value is provided if person is bona fide researcher. Empty value otherwise.");
+		return attr;
+	}
+}


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatusREMS.java
```diff
@@ -0,0 +1,26 @@
...
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserAttributesModuleImplApi;
+
+/**
+ * This attribute is a flag for that user is a researcher in REMS
+ *
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class urn_perun_user_attribute_def_def_elixirBonaFideStatusREMS extends UserAttributesModuleAbstract implements UserAttributesModuleImplApi {
```
stavamichal: Is there any reason for this module even if there is no implementation in it?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,120 @@
...
+	private final Pattern removeUserExtSource = Pattern.compile("UserExtSource:\\[(.*)] removed from User:\\[(.*)]");
+	private final Pattern extSourceRems = Pattern.compile("cz.metacentrum.perun.core.impl.ExtSourceREMS");
+
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDefinition attributeDefinition) throws InternalErrorException {
+		Attribute attribute = new Attribute(attributeDefinition);
+		String url = "http://www.ga4gh.org/beacon/bonafide/ver1.0";
+
+		try {
+			Attribute attributeDef = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_DEF + ":elixirBonaFideStatusREMS");
```
stavamichal: Please do not use "attributeDef" name, because it looks like "attributeDefinition" and object AttributeDefinition is Attribute template without value.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,120 @@
...
+			String attrDefValue = (String) attributeDef.getValue();
+			if (attrDefValue != null && !attrDefValue.isEmpty()) {
+				attribute.setValue(attrDefValue);
+			} else {
+				Attribute uesAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_VIRT + ":eduPersonScopedAffiliations");
+				List<String> affiliations = (List<String>)uesAttribute.getValue();
+				for (String affiliation : affiliations) {
```
stavamichal: What if value of virtual attribute is null?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	/**
+	 * Get User from message if exists and there is only one. In other case return null instead.
+	 *
+	 * @param message
+	 * @return user or null
+	 * @throws InternalErrorException
+	 */
+	private User getUserFromMessage(String message) throws InternalErrorException {
```
stavamichal: This method is now without duplicities (and with customizations) in ModulesUtilsBl, please use it instead.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,120 @@
...
+
+		Matcher addUserExtSourceMatcher = addUserExtSource.matcher(message);
+		Matcher removeUserExtSourceMatcher = removeUserExtSource.matcher(message);
+		Matcher extSourceRemsMatcher = extSourceRems.matcher(message);
+
+		User user;
+		Attribute attrVirtRemsLogins;
+
+		if(extSourceRemsMatcher.find()) {
+			if (addUserExtSourceMatcher.find() || removeUserExtSourceMatcher.find()) {
```
stavamichal: I am afraid this is not enough. The value of this attribute can be changed by updating of affiliations attribute value (for example if record with "faculty@" was removed).

Vojtech-Sassmann:
Hi @stavamichal,
I have just made changes according to your requests and the Jenkins build is failing and I can not access the datails.

stavamichal:
test this please

stavamichal:
It was just some unexpected error between GIT and Jenkins, it is ok now.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,100 @@
...
+		Matcher addUserExtSourceMatcher = addUserExtSource.matcher(message);
+		Matcher removeUserExtSourceMatcher = removeUserExtSource.matcher(message);
+		Matcher extSourceRemsMatcher = extSourceRems.matcher(message);
+		Matcher affiliationsAttributeMatcher = affiliationsAttribute.matcher(message);
+
+		User user;
+		Attribute attrVirtRemsLogins;
+
+		if (affiliationsAttributeMatcher.find() || (
+				extSourceRemsMatcher.find() && (addUserExtSourceMatcher.find() || removeUserExtSourceMatcher.find()))) {
```
stavamichal: So if I understand it right there are these possible ways:
- attribute with friendly name "..Affiliations" was set for "user
- extSourceREMS was add to user
- extSourceREMS was remove for user

There is still missing part with removing value from affiliations attribute I think.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,100 @@
...
+
+			user = (new ModulesUtilsBlImpl()).getUserFromMessage(sess, message);
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Yes, it is. This method is not implemented in the interface, and in class ModulesUtilsBlImpl it is not marked as a static method. And actually, I think there is a bug inside of it. On line no. 1011 there is 
`user = (User) user` which probably should have been `user = (User) perunBean`.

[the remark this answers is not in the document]
stavamichal: You are right, it is not static method but it is implemented in ModulesUtilsBl on the line 428 as far as I can say from master branch.
And yes, thank you, there is a bug. I will create hot fix for it.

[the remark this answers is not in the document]
stavamichal: This shoudl be the right way how to call it: perunSession.getPerunBl().getModulesUtilsBl().getUserFromMessage(perunSession, message);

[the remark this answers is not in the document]
Vojtech-Sassmann: @stavamichal Thanks, I have found out it already yesterday, I guess that editing comments here is not a good idea since it probably does not let you know about editing the comment.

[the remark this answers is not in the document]
stavamichal: @Vojtech-Sassmann yes, I did not get info about changing comment. We already merged hot fix for mentioned bug to the production branch. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatusREMS.java
```diff
@@ -0,0 +1,40 @@
...
+
+	@Override
+	public void checkAttributeValue(PerunSessionImpl perunSession, User user, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		if (attribute.getValue() != null && !(attribute.getValue() instanceof String)) {
```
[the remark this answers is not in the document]
Vojtech-Sassmann: What happens when in ```urn_perun_user_attribute_def_virt_elixirBonaFideStatus```
is called ```sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_DEF + ":elixirBonaFideStatusREMS");```
and this module would not exist? Just asking, I don't know how this works exactly.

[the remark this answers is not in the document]
stavamichal: Non-virtual attributes use defaults when module not exists. Only for virtual attributes is needed module for their existence.

Vojtech-Sassmann:
@stavamichal I have just made the changes we talked about. And the Jenkins build is again failing.

stavamichal:
I think jenkins is ok with it.

Vojtech-Sassmann:
@stavamichal I made another commit --amend and then it was OK.

stavamichal:
It seems to be ok now. We just need to do some manual tests against devel instance of Perun before merging.

Vojtech-Sassmann:
@stavamichal Ok, I have finally managed to get this working. There were several issues that we will probably discuss tomorrow. I have tested all situations except: removing all user attributes and removing all extSource attribute. But I think it should work for these situations as well.

Vojtech-Sassmann:
@stavamichal It should be all OK. I have just tested it again on devel.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,112 @@
...
+					for (String affiliation : affiliations) {
+						if (affiliation != null && affiliation.startsWith("faculty@")) {
+							attribute.setValue(url);
+							break;
+						}
+					}
+				}
+			}
+		} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+			log.error("Cannot read elixirBonaFideStatusREMS from user " + user.getId(), e);
```
stavamichal: This could be also thrown by getting eduPersonScopedAffilations attribute, you should probably separate it to be sure which one it is.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,112 @@
...
+	private final Pattern userRemovePattern = Pattern.compile("AttributeDefinition:\\[(.*)elixirBonaFideStatusREMS(.*)] removed for User:\\[(.*)]");
+	private final Pattern userAllAttrsRemovedPattern = Pattern.compile("All attributes removed for User:\\[(.*)]");
+	private final Pattern uesAllAttrsRemovedPattern = Pattern.compile("All attributes removed for UserExtSource:\\[(.*)]");
+	private final Pattern uesSetAffiliationAttributePattern = Pattern.compile("Attribute:\\[(.*)affiliation(.*)] set for UserExtSource:\\[(.*)]");
+	private final Pattern uesRemoveAffiliationAttributePattern = Pattern.compile("AttributeDefinition:\\[(.*)affiliation(.*)] removed for UserExtSource:\\[(.*)]");
+
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDefinition attributeDefinition) throws InternalErrorException {
+		Attribute attribute = new Attribute(attributeDefinition);
+		String url = "http://www.ga4gh.org/beacon/bonafide/ver1.0";
```
stavamichal: I think this variable can be used as constant for whole module and it can be final.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,112 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
```
stavamichal: Please describe little more functionality of this specific module.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,112 @@
...
+public class urn_perun_user_attribute_def_virt_elixirBonaFideStatus extends UserVirtualAttributesModuleAbstract implements UserVirtualAttributesModuleImplApi {
+
+	private final static Logger log = LoggerFactory.getLogger(urn_perun_user_attribute_def_virt_elixirBonaFideStatus.class);
+
+	private final Pattern userSetPattern = Pattern.compile("Attribute:\\[(.*)elixirBonaFideStatusREMS(.*)] set for User:\\[(.*)]");
+	private final Pattern userRemovePattern = Pattern.compile("AttributeDefinition:\\[(.*)elixirBonaFideStatusREMS(.*)] removed for User:\\[(.*)]");
+	private final Pattern userAllAttrsRemovedPattern = Pattern.compile("All attributes removed for User:\\[(.*)]");
+	private final Pattern uesAllAttrsRemovedPattern = Pattern.compile("All attributes removed for UserExtSource:\\[(.*)]");
+	private final Pattern uesSetAffiliationAttributePattern = Pattern.compile("Attribute:\\[(.*)affiliation(.*)] set for UserExtSource:\\[(.*)]");
+	private final Pattern uesRemoveAffiliationAttributePattern = Pattern.compile("AttributeDefinition:\\[(.*)affiliation(.*)] removed for UserExtSource:\\[(.*)]");
```
stavamichal: For these attribute messages will be probably the best to use little more specific regex with something like `"...(.*)friendlyName=<elixirBonaFideStatusREMS>(.*)..."` and the same for the affilation attribute. It prevent to be used for any attribute which have for example "affilation" in it's name.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,112 @@
...
+				String value = (String) attribute.getValue();
+
+				if (value == null || value.isEmpty()) {
+					AttributeDefinition attributeDefinition = new AttributeDefinition(attribute);
+					messageAttributeSet = attributeDefinition.serializeToString() + " removed for " + user.serializeToString() + ".";
+				} else {
+					messageAttributeSet = attribute.serializeToString() + " set for " + user.serializeToString() + ".";
+				}
+				resolvingMessages.add(messageAttributeSet);
+			}
```
stavamichal: If user is null, it is probably not totally OK. There should be at least log error about this situation.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,112 @@
...
+	private final Pattern uesSetAffiliationAttributePattern = Pattern.compile("Attribute:\\[(.*)affiliation(.*)] set for UserExtSource:\\[(.*)]");
+	private final Pattern uesRemoveAffiliationAttributePattern = Pattern.compile("AttributeDefinition:\\[(.*)affiliation(.*)] removed for UserExtSource:\\[(.*)]");
+
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDefinition attributeDefinition) throws InternalErrorException {
+		Attribute attribute = new Attribute(attributeDefinition);
+		String url = "http://www.ga4gh.org/beacon/bonafide/ver1.0";
+
+		try {
+			Attribute statusAttr = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_DEF + ":elixirBonaFideStatusREMS");
```
stavamichal: All names of attributes can be also moved as constants to the module itself (not only in methods). It will be more easy to read these values in the future.

Vojtech-Sassmann:
@stavamichal I have made the changes and tested it again.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,135 @@
...
+					List<String> affiliations = (List<String>) userAttribute.getValue();
+					for (String affiliation : affiliations) {
+						if (affiliation != null && affiliation.startsWith("faculty@")) {
+							attribute.setValue(URL);
+							break;
+						}
+					}
+				}
+			} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+				log.error("Cannot read {} from user {}", USER_AFFILIATIONS_ATTR_NAME, user.getId(), e);
```
stavamichal: Please use whole object user instead of just it's ID. (for both "log error" in this file)

stavamichal:
@Vojtech-Sassmann Great work! Just one last change please and I will merge it. There is no need for testing this change on devel, successful compilation of this code will be ok for this little change.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
Vojtech-Sassmann: Done.

stavamichal:
Test this please

Vojtech-Sassmann:
@stavamichal I thought this minor change does not need to be tested.

PR #1663, opened 2017-10-18
User ext source for REMS

Vojtech-Sassmann:
* Created ExtSourceREMS that loads subjects with Sql and filters
subjects that already exists in Perun by checking their login(id) or
additionalues.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
@@ -1,0 +1,209 @@
+package cz.metacentrum.perun.core.impl;
+
+import cz.metacentrum.perun.core.api.ExtSourcesManager;
+import cz.metacentrum.perun.core.api.PerunClient;
+import cz.metacentrum.perun.core.api.PerunPrincipal;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.SubjectNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.UserExtSourceNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
+import cz.metacentrum.perun.core.blImpl.PerunBlImpl;
+import cz.metacentrum.perun.core.implApi.ExtSourceApi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * ExtSource class for REMS, filters subjects that does not have a corresponding user in Perun
+ * by ues REMS or by additionalueses in format: {extSourceName}|{extSourceClass}|{eppn}|0.
+ * The eppn is used as a 'login'.
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class ExtSourceREMS extends ExtSourceSqlComplex implements ExtSourceApi {
+
+	private final static Logger log = LoggerFactory.getLogger(ExtSourceREMS.class);
+
+	private static PerunBlImpl perunBl;
+
+	// filled by spring (perun-core.xml)
+	public static PerunBlImpl setPerunBlImpl(PerunBlImpl perun) {
+		perunBl = perun;
+		return perun;
+	}
+
+	@Override
+	public List<Map<String, String>> findSubjects(String searchString) throws InternalErrorException, ExtSourceUnsupportedOperationException {
+		List<Map<String, String>> subjects = super.findSubjects(searchString);
+		return filterNonExistingUsers(subjects);
+	}
+
+	@Override
+	public List<Map<String, String>> findSubjects(String searchString, int maxResults) throws InternalErrorException, ExtSourceUnsupportedOperationException {
+		List<Map<String, String>> subjects = super.findSubjects(searchString, maxResults);
+		return filterNonExistingUsers(subjects);
+	}
+
+	@Override
+	public Map<String, String> getSubjectByLogin(String login) throws InternalErrorException, SubjectNotExistsException {
+		Map<String, String> subject = super.getSubjectByLogin(login);
+		if (!isExistingUser(subject)) {
+			throw new SubjectNotExistsException("Subject for given login does not exist in Perun");
+		}
+		return subject;
+	}
+
+	@Override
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,157 @@
...
+
+		List<User> usersFromREMS = perunBl.getUsersManagerBl().getUsersByExtSourceTypeAndLogin(sess, ExtSourcesManager.EXTSOURCE_REMS, login);
+		if (usersFromREMS.size() > 0) {
+			return true;
+		}
+
+		String extSource = subject.get("additionalues_1");
+		if (extSource == null) {
+			log.error("Failed to get additionalues_1 from subject {}", subject);
+			throw new InternalErrorException("Failed to get additionalues_1 from subject: " + subject);
```
stavamichal: Because thrown exception is also logged to the perun log file, you don't need to log it twice.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		String[] extSourceSplit = subject.get("additionalues_1").split("\\|", 4);
+		if(extSourceSplit.length != 4) {
+			log.error("Ivalid format of additionalues_1. It should be '{extSourceName}|{extSourceClass}|{eppn}|0'. Actual: {}", extSource);
+			throw new InternalErrorException("Ivalid format of additionalues_1. It should be '{extSourceName}|{extSourceClass}|{eppn}|0', Actual: " + extSource);
```
stavamichal: Because thrown exception is also logged to the perun log file, you don't need to log it twice.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,157 @@
...
+		return false;
+	}
+
+	private PerunSession getSession() throws InternalErrorException {
+		final PerunPrincipal pp = new PerunPrincipal("ExtSourceREMS", ExtSourcesManager.EXTSOURCE_NAME_INTERNAL, ExtSourcesManager.EXTSOURCE_INTERNAL);
+		try {
+			return perunBl.getPerunSession(pp, new PerunClient());
+		} catch (InternalErrorException e) {
+			log.error("Failed to get session for ExtSourceREMS.", e);
+			throw new InternalErrorException("Failed to get session for ExtSourceREMS.", e);
```
stavamichal: Because thrown exception is also logged to the perun log file, you don't need to log it twice.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,157 @@
...
+	 * or by additionalues_1 in format: {extSourceName}|{extSourceClass}|{eppn}|0
+	 *
+	 * @param subject subject
+	 * @return true if the subject has, false otherwise
+	 * @throws InternalErrorException
+	 */
+	private boolean isExistingUser(Map<String, String> subject) throws InternalErrorException {
+		if (subject.isEmpty()) {
+			return false;
+		}
```
stavamichal: Subject can be also null if something really nasty happened in code. I would prefer to throw an exception in both cases - null or emptiness (there should always be at least login of the subject).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,157 @@
...
+		String login = subject.get("login");
+
+		PerunSession sess = getSession();
+
+		List<User> usersFromREMS = perunBl.getUsersManagerBl().getUsersByExtSourceTypeAndLogin(sess, ExtSourcesManager.EXTSOURCE_REMS, login);
+		if (usersFromREMS.size() > 0) {
+			return true;
+		}
+
+		String extSource = subject.get("additionalues_1");
```
stavamichal: This seems to be very strict setting. You should try all "additionalues_x" and if there is none, you can log it and return false instead of throwing and exception. What do you think?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,157 @@
...
+			return false;
+		}
+		String login = subject.get("login");
+
+		PerunSession sess = getSession();
+
+		List<User> usersFromREMS = perunBl.getUsersManagerBl().getUsersByExtSourceTypeAndLogin(sess, ExtSourcesManager.EXTSOURCE_REMS, login);
+		if (usersFromREMS.size() > 0) {
+			return true;
+		}
```
stavamichal: Please describe in comment why is there this part of code. It is little hard to understand it just from code although it is very simple.

Vojtech-Sassmann:
@stavamichal I have just made some changes according to your advice. Now it should work that way, that it tries to find existing user by maximum of 10 additionalueses.

Vojtech-Sassmann:
@stavamichal Ok, now it should use all additionalueses.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,201 @@
...
+	 * @param subject subject
+	 * @return true if the subject has, false otherwise
+	 * @throws InternalErrorException internalError
+	 */
+	private boolean isExistingUser(Map<String, String> subject) throws InternalErrorException {
+		if (subject == null || subject.isEmpty()) {
+			throw new InternalErrorException("Subject can not be empty or null: " + subject);
+		}
+
+		String login = subject.get("login");
```
stavamichal: What if login is not there?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
```diff
@@ -0,0 +1,201 @@
...
+				return true;
+			}
+		}
+
+		return false;
+	}
+
+	/**
+	 * Finds out for given ues and login exits user in Perun.
+	 * Format of ues is {extSourceName}|{extSourceClass}|{eppn}|0
```
stavamichal: Is good to say, that eppn=login there.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
Vojtech-Sassmann: I have added validation for it.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceREMS.java
Vojtech-Sassmann: It is now mentioned in the JavaDoc.

stavamichal:
For me it looks fine. @zlamalp can you please check it too? Thank you.

PR #1666, opened 2017-10-19
Perun RPC enabled for OAuth2 JavaScript clients

martin-kuba:
- removed perun-oidc module, no longer needed
- changed Api servlet to use OIDC "sub" claim as user identification
- added two configuration properties for specifying extSource name and type used in translating the user identification to User object
- added check that scope perun_api was enabled by user
- added handling of CORS HTTP headers
- modified initialization and reload of PerunSession to remove PERUN_ADMIN role for OAuth clients


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -64,47 +112,50 @@ public PerunBlImpl() {
...
-		log.debug("creating PerunSession for user {}",principal.getActor());
-		if (principal.getUser() == null && usersManagerBl!=null && !dontLookupUsersForLogins.contains(principal.getActor())) {
+		log.debug("creating PerunSession for user {}", principal.getActor());
+		if (principal.getUser() == null && usersManagerBl != null && !dontLookupUsersForLogins.contains(principal.getActor())) {
 			// Get the user if we are completely initialized
 			try {
 				PerunSession internalSession = getPerunSession();
 				User user = usersManagerBl.getUserByExtSourceNameAndExtLogin(internalSession, principal.getExtSourceName(), principal.getActor());
 				principal.setUser(user);
-				// Try to update LoA for userExtSource
```
martin-kuba: Updates to loa and userExtSource attrbutes should not be done for OAuth clients

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
```diff
@@ -210,10 +206,11 @@ private static PerunPrincipal setupPerunPrincipal(HttpServletRequest req, Deseri
...
 		// If OIDC_CLAIM_sub header is present, it means user authenticated via OAuth2 with MITRE.
 		else if (isNotEmpty(req.getHeader(OIDC_CLAIM_SUB))) {
-			extSourceName = req.getHeader(OIDC_CLAIM_EXTSOURCE_NAME);
-			extSourceType = req.getHeader(OIDC_CLAIM_EXTSOURCE_TYPE);
-			extSourceLoaString = req.getHeader(OIDC_CLAIM_EXTSOURCE_LOA);
-			extLogin = remoteUser;
+			//this is configurable, as the OIDC server has the sub claim also configurable
+			extSourceName = config.getOidcExtsourceName();
+			extSourceType = config.getOidcExtsourceType();
+			extSourceLoaString = "N/A";
```
zlamalp: On many places we expect LoA to be a number (even when it's retrieved as a string). If unsure about trust of source identity I suggest to use "0".

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -34,34 +34,10 @@
...
-		if (sess.getPerunPrincipal().getUser() != null) {
-			sess.getPerunPrincipal().getRoles().putAuthzRole(Role.SELF, sess.getPerunPrincipal().getUser());
-
-			// Add service user role
-			if (sess.getPerunPrincipal().getUser().isServiceUser()) {
-				sess.getPerunPrincipal().getRoles().putAuthzRole(Role.SERVICEUSER);
-			}
-		}
-		sess.getPerunPrincipal().setAuthzInitialized(true);
+		refreshAuthz(sess);
```
zlamalp: Please fix ``refreshAuthz(sess)`` implementation, it's not 1:1 replacement of above removed code right now.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1198,6 +1174,7 @@ public static synchronized void refreshAuthz(PerunSession sess) throws InternalE
 		log.debug("Refreshing authz roles for session {}.", sess);
 
 		sess.getPerunPrincipal().setRoles(AuthzResolverBlImpl.getRoles(sess));
```
zlamalp: This should be:
```java
// Prepare first users rights on all subgroups of groups where user is GroupAdmin and add them to AuthzRoles of the user
AuthzRoles authzRoles = addAllSubgroupsToAuthzRoles(sess, authzResolverImpl.getRoles(sess.getPerunPrincipal().getUser()));
```
like in ``init(sess)`` method.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1198,6 +1174,7 @@ public static synchronized void refreshAuthz(PerunSession sess) throws InternalE
...
+
 		prepareServiceRoles(sess);
```
zlamalp: Should be before "setting roles" to session, since when account has service role, we do not load the rest, like in ``init(sess)`` method.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1210,6 +1187,11 @@ public static synchronized void refreshAuthz(PerunSession sess) throws InternalE
 			}
 		}
 
+		//for OAuth clients, do not allow delegating PERUNADMIN role
```
zlamalp: Regarding fixing above impl, this should be right after calling ``prepareServiceRoles(sess);`` and before check if roles are non-empty and considered initialized. Otherwise you would probably get perun_admin even for OAUTH client.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
martin-kuba: It is, the AuthzResolverBlImpl.getRoles() method calls addAllSubgroupsToAuthzRoles()

zlamalp: Sorry, I overlooked you call bl method and not impl.

martin-kuba:
refreshAuthz was reimplemented, now it is like the old init(), but it clears roles as the first step

PR #1677, opened 2017-11-02
Fix auditing of removing attribute

Vojtech-Sassmann:
* When the attribute is now removed by setting empty value, it is
correctly audited as 'AttributeDefinition {} removed for {}'.
* Redmine issue: https://homeproj.cesnet.cz/issues/3778


Vojtech-Sassmann:
I have searched the code and none of the modules was using the incorrect format so it should be safe.

PR #1682, opened 2017-11-08
Added JavaDoc for methods created while adding sponsored member functionality

VMecko:
Added JavaDoc for all methods created while adding functionality for
sponsored members

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AuthzResolverImplApi.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AuthzResolverImplApi.java
@@ -323,7 +323,7 @@ public interface AuthzResolverImplApi {
 	void addVoRole(PerunSession sess, Role role, Vo vo, Group group) throws InternalErrorException, AlreadyAdminException;
 
 	/**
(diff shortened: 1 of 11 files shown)


VMecko:
I didn't give description to all exceptions (i think it's unnecessary) but only to those I think are not so clear from their name or those used first time in their class.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AuthzResolverImplApi.java
```diff
@@ -299,43 +299,88 @@
...
+	 * @param vo virtual organization
+	 * @param group group
+	 * @throws InternalErrorException
+	 * @throws AlreadyAdminException
 	 */
 	void addVoRole(PerunSession sess, Role role, Vo vo, Group group) throws InternalErrorException, AlreadyAdminException;
 
 	/**
-	 * Removes role from user in a VO.
+	 * Removes role from user in a VO
```
martin-kuba: Missing dot - full stop.

martin-kuba:
OK, let's merge it.

PR #1684, opened 2017-11-13
GetCompleteCandidates feature

Vojtech-Sassmann:
* Created two new methods in Entry:
  * vosManager.getCompleteCandidates(vo, attrNames, searchString)
  * vosManager.getCompleteCandidates(vo, group, attrNames, searchString)
* Those methods are used to find MemberCandidate entities that consists
of Candidate, RichUser and Member entities.
* Those entities are used to decide if some candidate has appropriate
user in perun, member in a Vo or member in a group.
* The first method can be called only by VO_ADMIN and it searches for
candidates in all VO's ExtSources.
* The second method can be called by VO_ADMIN or GROUP_ADMIN. When it is
called by GROUP_ADMIN, it searches for candidates only in ExtSources
assigned to given group. Otherwise, it searches in all VO's ExtSources.

* Also were created appropriate methods in BL layer and method for
searching candidates - "getCandidates()" - was made more general. Now it
searches for candidates in given List of ExtSources and it can be
decided if it should return candidates for those users, who are already
in given VO.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberCandidate.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberCandidate.java
@@ -1,0 +1,76 @@
+package cz.metacentrum.perun.core.api;
+
+
+/**
+ * Candidate with member and rich user of a Virtual Organization.
+ *
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class MemberCandidate {
+
+	private Candidate candidate;
+
+	private Member member;
+
+	private RichUser richUser;
+
+	public Candidate getCandidate() {
+		return candidate;
+	}
+
+	public void setCandidate(Candidate candidate) {
+		this.candidate = candidate;
+	}
+
+	public Member getMember() {
+		return member;
+	}
+
+	public void setMember(Member member) {
+		this.member = member;
+	}
+
+	public RichUser getRichUser() {
+		return richUser;
+	}
+
+	public void setRichUser(RichUser richUser) {
+		this.richUser = richUser;
+	}
+
+	/**
+	 *  Returns bean name like VO, Member, Resource,...
+	 */
+	public String getBeanName() {
+		return this.getClass().getSimpleName();
+	}
+
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
+
+		MemberCandidate that = (MemberCandidate) o;
+
+		if (candidate != null ? !candidate.equals(that.candidate) : that.candidate != null) return false;
+		if (member != null ? !member.equals(that.member) : that.member != null) return false;
+		return richUser != null ? richUser.equals(that.richUser) : that.richUser == null;
+	}
+
+	@Override
+	public int hashCode() {
+		int result = candidate != null ? candidate.hashCode() : 0;
+		result = 31 * result + (member != null ? member.hashCode() : 0);
+		result = 31 * result + (richUser != null ? richUser.hashCode() : 0);
+		return result;
+	}
+
+	@Override
+	public String toString() {
+		return getClass().getSimpleName()+ ":[" +
+				"candidate='" + candidate +
+				"', member='" + member +
+				"', richUser='" + richUser +
+				"']";
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
@@ -13,6 +13,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * VosManager business logic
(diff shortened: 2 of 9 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberCandidate.java
```diff
@@ -0,0 +1,69 @@
...
+	public int hashCode() {
+		int result = candidate != null ? candidate.hashCode() : 0;
+		result = 31 * result + (member != null ? member.hashCode() : 0);
+		result = 31 * result + (richUser != null ? richUser.hashCode() : 0);
+		return result;
+	}
+
+	@Override
+	public String toString() {
+		return "MemberCandidate{" +
```
zlamalp: Even if it's not PerunBean, please use similar toString: ``MemberCandidate:[content]``.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -385,6 +400,52 @@ public Vo getVoById(PerunSession sess, int id) throws InternalErrorException, Vo
...
+		if (group != null) {
+			candidates = findCandidates(sess, group, searchString, extSources, false);
+		} else {
+			candidates = findCandidates(sess, vo, searchString, 0, extSources, false);
+		}
+
+		List<RichUser> users;
+
+		try {
+			users = getPerunBl().getUsersManagerBl().findRichUsersWithAttributes(sess, searchString, attrNames);
```
zlamalp: Based on RPC API javadoc for group manager it should search only through members of VO, but this method search through whole perun.

perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberCandidate.java
Vojtech-Sassmann: Ok, I have changed it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Vojtech-Sassmann: Now it should search only users in Vo, if it is called by group admin.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -604,7 +665,75 @@ private void removeSponsorFromSponsoredMembers(PerunSession sess, Vo vo, User us
...
+		List<MemberCandidate> memberCandidates = new ArrayList<>();
+
+		// try to find RichUser for candidates
+		for (Candidate candidate : candidates) {
+			MemberCandidate mc = new MemberCandidate();
+			mc.setCandidate(candidate);
+
+			try {
+				User user = getPerunBl().getUsersManagerBl().getUserByUserExtSources(sess, candidate.getUserExtSources());
+				RichUser richUser = getPerunBl().getUsersManagerBl().getRichUser(sess, user);
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Created method for this purpose.

zlamalp:
Can you please also update ``parseRpcMethods.pl`` in ``perun-utils/rpc-methods-javadoc-generator/``? At the file start, there are ``$objectExamples{}`` for each object. Please add example object for ``MemberCandidate``. Thank you.

Vojtech-Sassmann:
@zlamalp Hi, I have added the entity examples and changed the way of dealing with failures during searching candidates in ExtSource.

perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberCandidate.java
```diff
@@ -0,0 +1,69 @@
+package cz.metacentrum.perun.core.api;
+
+
+/**
+ * Candidate with member and rich user of a Virtual Organization.
+ *
+ * @author Vojtech Sassmann &lt;vojtech.sassmann@gmail.com&gt;
+ */
+public class MemberCandidate {
```
zlamalp: Please add ``getBeanName()`` method like it's in normal PerunBeans, it is helpful in GUI, since otherwise we don't know object type. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -385,6 +408,76 @@ public Vo getVoById(PerunSession sess, int id) throws InternalErrorException, Vo
...
+	 */
+	private List<MemberCandidate> getMemberCandidates(PerunSession sess, Vo vo, Group group, List<String> attrNames, String searchString, List<ExtSource> extSources) throws InternalErrorException {
+		List<Candidate> candidates;
+		List<RichUser> richUsers;
+
+		if (vo == null && group == null) {
+			throw new InternalErrorException("Vo and group can not be both null");
+		}
+
+		if (vo != null) {
```
zlamalp: This should be ``group != null`` and then you can merge below IF on group inside this one. Right now it doesn't work correctly for vo_manager (it searches members of VO instead of whole Perun).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -385,6 +408,76 @@ public Vo getVoById(PerunSession sess, int id) throws InternalErrorException, Vo
...
+		} else {
+			// vo is null, find users it whole perun
+
+			try {
+				richUsers = getPerunBl().getUsersManagerBl().findRichUsersWithAttributes(sess, searchString, attrNames);
+			} catch (UserNotExistsException e) {
+				richUsers = new ArrayList<>();
+			}
+
+			vo = getPerunBl().getGroupsManagerBl().getVo(sess, group);
```
zlamalp: Regarding above comment, this should be moved to if branch where group != null.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Vojtech-Sassmann: Looking at he code again, I think that it is ok. When the method in entry is called by vo_admin, then the vo is set to null. So then in the getMemberCandidates() method is called the 'else' branch on line 455 and all users from perun are searched.

Vojtech-Sassmann: On the other hand, the javadoc is not written very well, I will change it so it is easier to understand the methor.

zlamalp: Fair point, but this apply only, when you use ``group,searchString,attrNames`` in RPC API, but when VO manager tries to add member to VO, ``vo,searchString,attrNames`` are used in RPC API, so you get non-null VO with VO ext sources here and then it searches for users only inside VO, which is pointless, since admin wants to add new people to the VO.

Then you should fix ``getCompleteCandidates(PerunSession sess, Vo vo, List<String> attrNames, String searchString)`` to not pass VO, but it would end-up with both objects null, which is not allowed.

When I think about it, my fix wasn't ok either, since it always returned only vo members when adding to group. Maybe you should just use some boolean flag instead of null/non-null in vo and group params.

Vojtech-Sassmann: Yeah, you are right. I will probably change the whole logic of it, maybe with those boolean flags. I just dont like methods with a lot of arguments but on the other hand a don't want to create duplicate code.

Vojtech-Sassmann: Ok, I have changed it. I think it is better now.

Vojtech-Sassmann:
@zlamalp Hi, what do you think of the changes I have made?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -385,6 +408,73 @@ public Vo getVoById(PerunSession sess, int id) throws InternalErrorException, Vo
...
+		List<Candidate> candidates = findCandidates(sess, vo, searchString, 0, extSources, false);
+
+		return createMemberCandidates(sess, richUsers, vo, candidates, attrNames);
+	}
+
+	@Override
+	public List<MemberCandidate> getCompleteCandidates(PerunSession sess, Vo vo, Group group, List<String> attrNames, String searchString, List<ExtSource> extSources) throws InternalErrorException {
+		List<RichUser> richUsers = getRichUsersForMemberCandidates(sess, vo, attrNames, searchString);
+		List<Candidate> candidates = findCandidates(sess, group, searchString, extSources, false);
+
```
zlamalp: When adding members to group, VO is null and call ``createMemberCandidates()`` ends with null pointer in members manager impl layer. You should re-get VO from a Group and pass it to ``createMemberCandidates()`` since its behavior doesn't differ on VO and is expected non-null.

Vojtech-Sassmann: Ok, I have fixed it.

PR #1688, opened 2017-11-20
implemented modules for M:D:o365EmailAddresses:mu and GR:D:o365EmailA…

martin-kuba:
…ddresses:mu

Partial implementation for redmine issue 3755 without checking global uniqueness. 
Attribute modules and their unit tests are implemented.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_o365EmailAddresses_mu.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_o365EmailAddresses_mu.java
@@ -1,0 +1,120 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.Resource;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.bl.AttributesManagerBl;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.impl.Utils;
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_o365EmailAddresses_mu.java
```diff
@@ -0,0 +1,120 @@
...
+	private static final String NAMESPACE = AttributesManager.NS_GROUP_RESOURCE_ATTR_DEF;
+	static final String ADNAME_ATTRIBUTE = NAMESPACE + ":adName";
+
+	@Override
+	public void checkAttributeValue(PerunSessionImpl sess, Resource resource, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		List<String> emails;
+		//get values
+		Object value = attribute.getValue();
+		if (value == null) {
+			throw new WrongAttributeValueException(attribute, resource, group, "can't be null.");
```
zlamalp: Shouldn't this be checked after check on adName value existence ? What do you think @licehammer ? I'm not sure, what is desired workflow with adName itself - regarding later manual changes and fixes, etc.
Also we translate empty list to null when storing (not sure on retrieving), so you might never get "empty list" on line 73.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_o365EmailAddresses_mu.java
```diff
@@ -0,0 +1,120 @@
...
+
+	/**
+	 * Prefills value created by joining value of urn:perun:group_resource:attribute-def:def:adName with "@group.muni.cz"
+	 */
+	@Override
+	public Attribute fillAttribute(PerunSessionImpl sess, Resource resource, Group group, AttributeDefinition attrDef) throws InternalErrorException, WrongAttributeAssignmentException {
+		if (!NAMESPACE.equals(attrDef.getNamespace())) throw new WrongAttributeAssignmentException(attrDef);
+		try {
+			Attribute result = new Attribute(attrDef);
+			Object adName =
```
zlamalp: Not sure, why splitted to 2 lines ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_o365EmailAddresses_mu.java
martin-kuba: OK, I will allow null value representing empty list.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_resource_attribute_def_def_o365EmailAddresses_mu.java
martin-kuba: just to make it shorter, but can be put on one line.

martin-kuba:
@zlamalp changes implemented

PR #1689, opened 2017-11-21
New attributes with GID ranges for specific namespace

stavamichal:
- new GIDRanges entityless attribute to set ranges of gids managed by
 system Perun and replace logic of min and max gid entityless attributes
- new GIDRanges virtual facility attribute to replace logic of min and
max gid virtual attributes defined for facility
- these attributes will be used in the future instead minGID and maxGID
attributes, because there is only one interval between min and max
gid and we need to choose some specific gids to manage them too
- method to check and transfer data from gid ranges attribute is
located in ModulsUtils, because we will need it for further work in
other funcionality working with GID ranges
- added tests for this method

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_namespace_GIDRanges.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_namespace_GIDRanges.java
@@ -1,0 +1,37 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.EntitylessAttributesModuleImplApi;
+import cz.metacentrum.perun.core.implApi.modules.attributes.EntitylessAttributesModuleAbstract;
+import java.util.LinkedHashMap;
+
+/**
+ * GID Ranges for specific namespace which can be managed by Perun
+ *
+ * @author Michal Stava <stavamichal@gmail.com>
+ */
+public class urn_perun_entityless_attribute_def_def_namespace_GIDRanges extends EntitylessAttributesModuleAbstract implements EntitylessAttributesModuleImplApi {
+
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_namespace_GIDRanges.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_ENTITYLESS_ATTR_DEF);
+		attr.setFriendlyName("namespace_GIDRanges");
+		attr.setDisplayName("GID ranges in namespace");
+		attr.setType(List.class.getName());
+		attr.setDescription("Manageable GID ranges in namespace.");
```
licehammer: Please add note to description how correct value look like (syntax).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_facility_attribute_def_virt_GIDRanges.java
```diff
@@ -0,0 +1,96 @@
...
+	public Attribute getAttributeValue(PerunSessionImpl sess, Facility facility, AttributeDefinition attributeDefinition) throws InternalErrorException {
+		Attribute attribute = new Attribute(attributeDefinition);
+		Attribute gidNamespaceAttribute = getUnixGIDNamespaceAttribute(sess, facility);
+		if(gidNamespaceAttribute.getValue() == null) return attribute;
+		Attribute namespaceGIDRangesAttribute = getNamespaceGIDRangesAttribute(sess, (String) gidNamespaceAttribute.getValue());
+		attribute = Utils.copyAttributeToVirtualAttributeWithValue(namespaceGIDRangesAttribute, attribute);
+		return attribute;
+	}
+
+	public boolean setAttributeValue(PerunSessionImpl sess, Facility facility, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException {
```
licehammer: I would rather not implement this method. This can be called only by perun admin and I would recommend to require to set entityless attribute directly (to avoid possible mistake, when one could thing he only changes settings for this facility).

zlamalp: We discussed this with @stavamichal and this implementation differ from module default also in original GID module. I believe it throws exception, if you are trying to set different value, than is already stored and return false otherwise. In both cases it doesn't store/change anything. We are not sure, if exception is not expected somewhere in peruns logic. Like some automatic process wants to set required attributes, it sets value inside the object and call setAttribute() method. If it would only return false, process could continue with wrong value in object causing problems to other attributes. Like this it kill the process. But this code might not be relevant anymore. What do you think @licehammer ?

licehammer: You are right Pavel. I just saw the implementation of the method itself and I didn't check what is does exactly. My bad. This method can stay as it is.

To your question: This method doesn't need to be implemented, because when you don't have the "fill" method, no automatic process should try to set this attribute. On the other hand, user might do that for example by modifying some attributes from GUI and then select and save all (including this one). In that case might be handy to have this method implemented as it is now.

stavamichal:
I fixed description and reworked the commit to have method separate from code of this specific module (we will need it for further work with gid ranges in other functionality). I also added tests for this method. It is prepared for review.

PR #1691, opened 2017-11-22
Added functionality for DNS/TLD to state conversion.

VMecko:
added new functionality which do mapping of TLD/DNS addresses of
user to specific state name. DNS addresses are taken from userExtSource
attribute of user.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_dnsStateMapping.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_dnsStateMapping.java
@@ -27,19 +27,19 @@ public class urn_perun_entityless_attribute_def_def_dnsStateMapping extends Enti
 	static String KEY = "config";
 
 	public void checkAttributeValue(PerunSessionImpl perunSession, String key, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		if(!KEY.equals(key)) {
+			throw new WrongAttributeValueException(attribute, key, "entityless key for this attribute must be " + KEY);
+		}
+
 		if(attribute.getValue() == null) return;
 		LinkedHashMap<String, String> map = (LinkedHashMap<String, String>) attribute.getValue();
 		if(map.isEmpty()) return;
 
-		if(!KEY.equals(key)) {
-			throw new WrongAttributeValueException(attribute, key, "entityless key for this attribute must be " + KEY);
-		}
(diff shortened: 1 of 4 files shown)


VMecko:
After conversation with martin kuba, we changed the definition of modifyValue function in UserVirtualAttributeCollectedFromUserExtSource class. We created ModifyValueContext to suit our needs for passing dnsStateMapping argument into institutionsCountry attribute. Changing of modifyValue function doesn't affect functionality of other classes.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_dnsStateMapping.java
```diff
@@ -0,0 +1,59 @@
...
+
+		if(!KEY.equals(key)) {
+			throw new WrongAttributeValueException(attribute, key, "entityless key for this attribute must be " + KEY);
+		}
+		//check keys and values
+		Set<String> mapKeys = map.keySet();
+		for(String mapKey : mapKeys) {
+			//check keys
+			if(mapKey == null) throw new WrongAttributeValueException(attribute, key, "Key in dnsStateMapping map can not be null.");
+			if(!mapKey.startsWith(".")) throw new WrongAttributeValueException(attribute, key, "Key in dnsStateMapping map must start with dot symbol ('.').");
```
martin-kuba: The requirement for domain names to start with a dot is not in the specification in redmine issue 3798. Actually it may be harmful. Please remove the check.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_dnsStateMapping.java
```diff
@@ -0,0 +1,59 @@
...
+public class urn_perun_entityless_attribute_def_def_dnsStateMapping extends EntitylessAttributesModuleAbstract implements EntitylessAttributesModuleImplApi {
+
+	static String KEY = "config";
+
+	public void checkAttributeValue(PerunSessionImpl perunSession, String key, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		if(attribute.getValue() == null) return;
+		LinkedHashMap<String, String> map = (LinkedHashMap<String, String>) attribute.getValue();
+		if(map.isEmpty()) return;
+
+		if(!KEY.equals(key)) {
```
martin-kuba: This check for key value should be before the check for null value and empty list. Please move the check to the first place among checks.

VMecko:
Thanks for feedback, fixed.

PR #1696, opened 2017-11-29
Fix behavior of removing some LDAPc attributes

stavamichal:
- there was problem with removing LDAP attributes with null value from
   LDAP. In code for some cases as Group Attribute Description was
   created and empty attribute without any description.
 - because this behavior was the same for more than 1 method, we merged
   it to the one method converting map of operations and associated
   attributes and their values to list of modification items needed for
   LDAP processing behavior
 - add method to test if some group attribute in ldap is set for the
   group or not
 - replace attribute context mappers for different objects by one
   attribute context mapper, because the code is the same here
 - add test for some group attributes to be checked if they exists in
   LDAP before removing them (if not, just to skip this removing at all)


stavamichal:
Need to be merged before #1640

PR #1697, opened 2017-11-29
Changed regex pattern for checking unix-like file paths.

VMecko:
I changed the regex pattern so it now also accept special symbols stated
between these quotation marks ". ? * + $ %". I also added one new test
testing if file doesn't end with slash '/' (because that way if would be
directory and not file path).


VMecko:
commit message updated with reasons for each change

PR #1700, opened 2017-11-30
New concept of quotas in Perun

stavamichal:
- adding two new attributes maxUserDataQuotas and maxUserFileQuotas.
   These attributes will add a way how to limit setting of
   defaultData|FileQuotas by attribute with the maximum of available
   values for softQuota and hardQuota for specific volume and specific
   resource by FacilityManagers. Then we can allow setting of
   defaultData|FileQuota attribute by VoManagers, because they will
   be limited by the maximum.
 - adding two new attribute dataQuotaOverride and fileQuotaOverride.
   These attributes will add a way how to set special override values
   for specific member, specific resource and specific volumes. These
   overrides will always replace defaults or member-specific settings.
 - defaultQuotas and member-resource specific quotas now can't exceed
   limits of maxUserQuotas
 - change behavior of counting member-resource quotas. Priority has
   override, if override on volume not exists, use the old way "bigger
   is better" from default and member-specific settings.
 - add tests for all ModuleUtils methods used in this quotas
   functionality

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
@@ -384,7 +384,7 @@ public interface ModulesUtilsBl {
 	 *
 	 * @return map with path in key and pair with <softQuota, hardQuota> in big decimal
 	 *
-	 * @throws InternalErrorException if attribute or his value is null
+	 * @throws InternalErrorException if first mandatory placeholder is null
 	 * @throws WrongAttributeValueException if something is wrong in format of attribute
 	 */
 	Map<String, Pair<BigDecimal, BigDecimal>> checkAndTransferQuotas(Attribute quotasAttribute, PerunBean firstPlaceholder, PerunBean secondPlaceholder, boolean withMetrics) throws InternalErrorException, WrongAttributeValueException;
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
@@ -10,6 +10,7 @@ import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.QuotaNotInAllowedLimitException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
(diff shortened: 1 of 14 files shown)


stavamichal:
Test mentioned in commit message are still missing. They are empty at this moment. But it can be also reviewed meanwhile. I will add them as soon as possible.

stavamichal:
All necessary tests added! It is ready for full review.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
```diff
@@ -376,16 +405,19 @@
 	Map<String, String> transferQuotasBackToAttributeValue(Map<String, Pair<BigDecimal, BigDecimal>> transferedQuotasMap, boolean withMetrics) throws InternalErrorException;
 
 	/**
-	 * Merge two transfered Quotas together.
+	 * Merge two transferred Quotas together.
 	 * Paths are always unique, quotas are merged. (soft together and hard together)
 	 * Together means = bigger is better, 0 means unlimited so it is always bigger
```
licehammer: Now it should be according to fixed priorities. First which have the value set from following list will be used: MR:D:quotaOverride, MR:D:quota, R:D:defaultQuota.

Please update the implementation accordingly.

stavamichal:
Test this please.

stavamichal:
Fixed. Please check again. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataQuotas.java
```diff
@@ -22,16 +30,52 @@
...
+			maxUserDataQuotasAttribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, resource, A_R_maxUserDataQuotas);
+		} catch (AttributeNotExistsException ex) {
+			throw new ConsistencyErrorException(ex);
+		}
+
+		//Check and transfer maxUserDataQuotasForResource
+		Map<String, Pair<BigDecimal, BigDecimal>> maxUserDataQuotasForResource;
+		try {
+			maxUserDataQuotasForResource = perunSession.getPerunBl().getModulesUtilsBl().checkAndTransferQuotas(maxUserDataQuotasAttribute, resource, null, true);
+		} catch (WrongAttributeValueException | InternalErrorException ex) {
```
licehammer: InternalErrorException means that maxUserQuota is not in correct format?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_maxUserDataQuotas.java
```diff
@@ -0,0 +1,50 @@
...
+		//And also check if every quota part has right settings (softQuota<=hardQuota)
+		perunSession.getPerunBl().getModulesUtilsBl().checkAndTransferQuotas(attribute, resource, null, true);
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_RESOURCE_ATTR_DEF);
+		attr.setFriendlyName("maxUserDataQuotas");
+		attr.setDisplayName("Maximum of data quotas of User on any volumes.");
```
licehammer: user with lowercase 'u'

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_maxUserDataQuotas.java
```diff
@@ -0,0 +1,50 @@
...
+		attr.setType(LinkedHashMap.class.getName());
+		attr.setDescription("This is Facility manager maximum set for Resource on defined volumes. " +
```
licehammer: I would replace beginning of the text with "Maximum data quota for each user on this resource."

"Facility manager maximum" don't have to be clear for everyone. Also, I wouldn't mentioned explicit role here, because that might change.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataQuotas.java
stavamichal: Yes, because the only way how InternalError can be thrown there is by problem with "null" value in quotas attribute.

stavamichal: And of course if place holder is null.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
```diff
@@ -835,48 +868,31 @@ protected void checkPerunNamespacesMap() {
...
-				newValue = firstQuotasValue;
-				mergedTransferedQuotas.put(path, newValue);
-			//if there are values to merge, merge them
+
+		//first go through member-resource quotas values
+		for(String path: memberResourceQuotas.keySet()) {
+			//override has the highest priority
+			if (quotasOverride.containsKey(path)) {
+				mergedTransferedQuotas.put(path, quotasOverride.get(path));
+			//if override not exists, take the original value
```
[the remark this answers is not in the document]
stavamichal: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_maxUserDataQuotas.java
stavamichal: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_maxUserDataQuotas.java
stavamichal: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataQuotas.java
licehammer: The problem with null value in quotas attributes should be handled in that attribute check module.

Not sure, what you mean by "place holder is null".

stavamichal: Ok, I will let throw InternalErrorException where it is expected in modules.

PR #1703, opened 2017-12-04
reimplemented institutionsCountries to react to dnsStateMapping changes

martin-kuba:
- changed type of entityless attribute dnsStateMapping from Map to String, multiple value are now stored using keys
- implemented getEntitylessStringAttributeMapping() in AttributeManagerBl for getting map from keys to values for an entityless attribute
- reimplemented module institutionsCountries to load DNS-to-country mapping from dnsStateMapping of the new type
- reimplemented resolveVirtualAttributeValueChange() in institutionsCountries user attribute to react to changes both entityless dnsStateMapping attribute and userExtSource schacHomeOrganization attribute
- implemented unit tests


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -1481,6 +1482,22 @@ public Attribute getAttribute(PerunSession sess, String key, String attributeNam
 		}
 	}
 
+	@Override
+	public Map<String,String> getEntitylessStringAttributeMapping(PerunSession sess, String attributeName) throws InternalErrorException, AttributeNotExistsException {
```
zlamalp: Maybe we could retrieve ``Map<String,Object>`` instead, to support also other types of attributes.

PR #1704, opened 2017-12-05
CORE: Initialize AttributesManagerBl after AuthzResolverImpl

zlamalp:
- When DB is empty, there are no roles stored in DB and then
  we can't insert attribute authz with null role.
- To make it work we must break spring dependency cycle between
  perun and authzResolverImpl.
  Since perun bean was used in init method of authzResolverImpl only
  to determine "read_only" status, we removed it from it and now it
  calls directly BeansUtils.isPerunReadOnly().


zlamalp:
Was tested on both local and production environment. We should consider using ``BeansUtils.isPerunReadOnly()`` everywhere instead of ``perunBl.isPerunReadOnly()``. I'm merging this pull-request.

PR #1710, opened 2017-12-12
CORE: Fixed group deletion logic to remove member-group attributes

zlamalp:
- When deleting group, we muse also remove all member-group attributes.
- Fixed usage of parent group inside subgroups cycle. Until now,
  reserved logins, facility contacts groups and ext sources were
  repeatedly removed from the parent group and not the subgroup.
- Renamed variable g->subGroup for clarity.


stavamichal:
Please fix the commit message "Until now".

PR #1728, opened 2018-01-08
Add two new modules to count member-resource quotas

stavamichal:
- before this change, only user-facility quotas was computed from all
   other affected attributes. Now we are computing member-resource quotas
   and from them we are computing user-facility quotas.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_dataQuotas.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_dataQuotas.java
@@ -1,0 +1,134 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Member;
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_fileQuotas.java
```diff
@@ -0,0 +1,129 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.*;
```
licehammer: Please replace this with individual imports.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_dataQuotas.java
```diff
@@ -0,0 +1,129 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.*;
```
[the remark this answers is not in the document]
stavamichal: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_fileQuotas.java
stavamichal: Done.

PR #1731, opened 2018-01-10
CORE, REGISTRAR: Save UES attributes from application

zlamalp:
- When we create new user through registration, UES attributes are now
  lost. This change saves UES attributes in application itself.
  On approval, UES attributes are set to newly created UES.
- Added wrapper method attributeValueToString() to BeansUtils.
  It allows us to convert Object (attribute value) to it's expected
  String representation. Current method required whole Attribute object.
- Logic of saving UES attributes from session in PerunBlImpl moved
  to own public method. It's now called by getPerunSession() (during
  session initiation) and by approveApplication() (from registrar).

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/BeansUtils.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/BeansUtils.java
@@ -187,6 +187,25 @@ public class BeansUtils {
 		return attrNew.toString();
 	}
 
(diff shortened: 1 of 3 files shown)


zlamalp:
We should think about, if we want to store also UES attributes from applications, where user is known. There might be a problem, when approving old applications, where UES attributes are actually older (invalid) than the one already stored in perun.

PR #1732, opened 2018-01-12
Move strongDependencies under virtual attribute api

stavamichal:
- move method getStrongDependencies only under virtual attributes api
   (non-virtual attributes can't have strong dependencies). Fix all
   places where moving this method to the new place will occure any
   error.
 - add better javadoc for method getDependencies, getStrongDependencies
   and checkAttributeDependencies to be more understandable


stavamichal:
@licehammer I think it is ok, because it is already done in VirtualAttributesModuleImplApi.java interface which is implemented for every specific virtual attributes module.

licehammer:
Sorry, I overlooked the part with VirtualAttributesModuleImplApi. Pull request is all right than.

PR #1739, opened 2018-01-17
Implement new gidRanges attribute to the existing code

stavamichal:
- all methods in ModulesUtils which works with gid was reworked to use
   new attribute gidRanges intstead old attributes minGID and maxGID
 - fix javadocs where minGID and maxGID were mentioned
 - remove unused methods from ModulesUtilsBlImpl
 - add both new attributes to the initialization of attributes
 - add info about deprycated classes of old modules minGID and maxGID
   (entityless attributes and also virtual attributes for facility)
 - replace old depepndencies where needed
 - fix typo in name of attribute in module namespace-GIDRanges
 - fix tests where old behavior was used
 - remove not used test class


stavamichal:
This code can be merged only if new attributes with gid ranges will be set on all instances with GID namespaces.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -6091,6 +6093,26 @@ protected void initialize() throws InternalErrorException {
...
+		//set attribute rights (with dummy id of attribute - not known yet)
+		rights = new ArrayList<>();
+		rights.add(new AttributeRights(-1, Role.SELF, Collections.singletonList(ActionType.READ)));
+		rights.add(new AttributeRights(-1, Role.VOADMIN, Collections.singletonList(ActionType.READ)));
+		rights.add(new AttributeRights(-1, Role.GROUPADMIN, Collections.singletonList(ActionType.READ)));
+		rights.add(new AttributeRights(-1, Role.FACILITYADMIN, Collections.singletonList(ActionType.READ)));
+		attributes.put(attr, rights);
+
+		//urn_perun_entityless_attribute_def_def_namespace_GIDRanges
+		attr = new AttributeDefinition( (new urn_perun_entityless_attribute_def_def_namespace_GIDRanges()).getAttributeDefinition() );
```
zlamalp: I'm not sure, that user or vo/group admin need to read gidRanges entityless attribute. This is usually managed only by perun admin anyway.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -6091,6 +6093,26 @@ protected void initialize() throws InternalErrorException {
 		rights.add(new AttributeRights(-1, Role.FACILITYADMIN, Collections.singletonList(ActionType.READ)));
 		attributes.put(attr, rights);
 
+		//urn_perun_facility_attribute_def_virt_GIDRanges
+		attr = new AttributeDefinition( (new urn_perun_facility_attribute_def_virt_GIDRanges()).getAttributeDefinition() );
+		//set attribute rights (with dummy id of attribute - not known yet)
+		rights = new ArrayList<>();
+		rights.add(new AttributeRights(-1, Role.SELF, Collections.singletonList(ActionType.READ)));
```
zlamalp: Not sure, that user need to read this, it's probably interesting for vo/group manager for support use-cases.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ModulesUtilsEntryIntegrationTest.java
```diff
@@ -40,6 +40,8 @@
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.bl.ModulesUtilsBl;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import sun.awt.image.ImageWatched;
```
zlamalp: Probably bad autoimport by IDEA and it shouldn't be here.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
stavamichal: We don't want to create any restriction on rights to read value of this attribute.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
stavamichal: We don't want to create any restriction on rights to read value of this attribute.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ModulesUtilsEntryIntegrationTest.java
stavamichal: Removed.

PR #1742, opened 2018-01-18
Method getMemberByUserExtSources improved

Vojtech-Sassmann:
* Added better javadoc to this method so it is clear what it can be
used for.

* Now when the method finds more members it throws an exception.

* Also removed unused method from Entry.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -12,10 +12,12 @@ import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -588,14 +590,28 @@ public Member getMemberByUserExtSource(PerunSession sess, Vo vo, UserExtSource u
...
+
+		Set<Member> foundMembers = new HashSet<>();
+
 		for (UserExtSource ues: ueas) {
 			try {
-				return getMembersManagerImpl().getMemberByUserExtSource(sess, vo, ues);
+				foundMembers.add(getMembersManagerImpl().getMemberByUserExtSource(sess, vo, ues));
 			} catch (MemberNotExistsException e) {
-				// Ignore
+				throw new MemberNotExistsException("Member with userExtSources " + ueas + " doesn't exists.");
```
zlamalp: I think you should ignore this exception, since you get list of UES, but only some of them might be registered in Perun, meaning you might find user only for some of them a those new UES you will add to the user.

Then you should throw ``MemberNotExistsException``, if ``foundMembers.size() == 0``.

Vojtech-Sassmann: Ok, done.

zlamalp:
Tested locally, seems to be working.

PR #1749, opened 2018-01-29
Edited synchronization and group arithmetic.

VMecko:
- created 2 new methods: (add/remove)RelationMembers from the old
  processRelationMembers to better reflect between usage of add/remove
  in code, also to differentiate between (possible) thrown exceptions.
- Splitting GroupOperationsExceptions to all possibilities of thrown
  exceptions respectively for (add/remove)RelationMembers to get better
  view of what is wrong when the exception is thrown.


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2771,19 +2784,25 @@ public Group createGroupUnion(PerunSession sess, Group resultGroup, Group operan
...
 
 		// do the operation logic
-		processRelationMembers(sess, resultGroup, getGroupMembers(sess, operandGroup), operandGroup.getId(), true);
+		try {
+			addRelationMembers(sess, resultGroup, getGroupMembers(sess, operandGroup), operandGroup.getId());
+		} catch (WrongReferenceAttributeValueException | AlreadyMemberException | WrongAttributeValueException e ) {
+			throw new InternalErrorException("Some operation was not allowed during processing members relations", e);
+		} catch (GroupNotExistsException e) {
+			throw new InternalErrorException("Some group does not exists...", e);
+		}
```
stavamichal: Is there any reason to throw InternalErrorException instead of other exceptions there?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1458,170 +1460,170 @@ public boolean canExtendMembershipWithReason(PerunSession sess, Member member) t
...
+	  }
+
+	  // Reset hours, minutes and seconds to 0
+	  calendar.set(Calendar.HOUR, 0);
+	  calendar.set(Calendar.MINUTE, 0);
+	  calendar.set(Calendar.SECOND, 0);
+	  calendar.set(Calendar.MILLISECOND, 0);
+	}
+
+	return calendar.getTime();
```
stavamichal: This fixing of code indentation is not need for this pull request. Please remove it.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -52,7 +52,7 @@ public Group createGroup(PerunSession sess, Vo vo, Group group) throws GroupExis
 
 		// Authorization
 		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, vo) &&
-		    !AuthzResolver.isAuthorized(sess, Role.TOPGROUPCREATOR, vo)) {
+			!AuthzResolver.isAuthorized(sess, Role.TOPGROUPCREATOR, vo)) {
```
stavamichal: This fixing of code indentation is not need for this pull request. Please remove it.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/GroupsManagerBl.java
```diff
@@ -89,11 +88,14 @@
...
-	 * @throws GroupOperationsException
 	 * @throws GroupRelationNotAllowed
 	 * @throws GroupRelationAlreadyExists
+	 * @throws GroupNotExistsException
+	 * @throws AlreadyMemberException
+	 * @throws WrongReferenceAttributeValueException
+	 * @throws WrongAttributeValueException
 	 */
-	Group createGroup(PerunSession perunSession, Group parentGroup, Group group) throws GroupExistsException, InternalErrorException, GroupOperationsException, GroupRelationNotAllowed, GroupRelationAlreadyExists;
+	Group createGroup(PerunSession perunSession, Group parentGroup, Group group) throws GroupExistsException, InternalErrorException, GroupRelationNotAllowed, GroupRelationAlreadyExists, GroupNotExistsException, AlreadyMemberException, WrongReferenceAttributeValueException, WrongAttributeValueException;
```
zlamalp: This method shouldn't throw ``GroupNotExistsException, AlreadyMemberException, WrongReferenceAttributeValueException, WrongAttributeValueException``. By it's implementation, it's thrown by ``createGroupUnion()`` which is fine, but in this case of creating new sub-group, there is no chance, there are any members (not already member / attributes exceptions) and also, if GroupNotExistsException would be thrown, it's some DB consistency error, so it should be wrapped as ``ConsistencyErrorException``. So please wrap it or silently skip such exceptions and log them.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/GroupsManagerBl.java
```diff
@@ -117,12 +119,15 @@
...
 	 * @throws GroupNotExistsException
 	 * @throws GroupRelationDoesNotExist
 	 * @throws GroupRelationCannotBeRemoved
+	 * @throws NotGroupMemberException
+	 * @throws AlreadyMemberException
+	 * @throws WrongReferenceAttributeValueException
+	 * @throws WrongAttributeValueException
 	 */
-	void deleteGroup(PerunSession perunSession, Group group, boolean forceDelete) throws InternalErrorException, RelationExistsException, GroupAlreadyRemovedException, GroupAlreadyRemovedFromResourceException, GroupOperationsException, GroupNotExistsException, GroupRelationDoesNotExist, GroupRelationCannotBeRemoved;
+	void deleteGroup(PerunSession perunSession, Group group, boolean forceDelete) throws InternalErrorException, RelationExistsException, GroupAlreadyRemovedException, GroupAlreadyRemovedFromResourceException, GroupNotExistsException, GroupRelationDoesNotExist, GroupRelationCannotBeRemoved, NotGroupMemberException, AlreadyMemberException, WrongReferenceAttributeValueException, WrongAttributeValueException;
```
zlamalp: ``AlreadyMemberException`` doesn't make sense here. It is transitively caused by ``removeIndirectMembers()``, where it shouldn't be -> I don't see anything causing it and it probably just stayed here from previous implementation.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -36,9 +36,11 @@
 	 * @throws MemberNotExistsException
 	 * @throws PrivilegeException
 	 * @throws MemberAlreadyRemovedException
-	 * @throws GroupOperationsException
+	 * @throws WrongAttributeValueException
+	 * @throws AlreadyMemberException
+	 * @throws WrongReferenceAttributeValueException
 	 */
-	void deleteMember(PerunSession sess, Member member) throws InternalErrorException, MemberNotExistsException, PrivilegeException, MemberAlreadyRemovedException, GroupOperationsException;
+	void deleteMember(PerunSession sess, Member member) throws InternalErrorException, MemberNotExistsException, PrivilegeException, MemberAlreadyRemovedException, WrongAttributeValueException, AlreadyMemberException, WrongReferenceAttributeValueException;
```
zlamalp: Shouldn't throw ``AlreadyMemberException``, gets down to ``removeDirectMember()`` in GroupsManagerBlImpl, which doesn't actually throws it inside, just method definition says it, please remove it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
VMecko: I was inspired by method moveGroup where it was solved the same way for the old processRelationMembers method, so i thought we followed this logic. What would be the best solution here? Not to catch the exceptions at all (as in old code over here) or to throw them but not mask them behind InternalErrorException?

stavamichal: I think (if it is possible) to throw it out without change.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
VMecko: It doesn't stop in `removeDirectMember` it gets more down to `removeRelationMembers` which calls `removeIndirectMembers()`, that is the source of that exception. Does` AlreadyMemberException` make sense in `removeIndirectMembers()` method? If so, i can't delete it.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -49,9 +50,11 @@
 	 * @throws VoNotExistsException
 	 * @throws PrivilegeException
 	 * @throws MemberAlreadyRemovedException
-	 * @throws GroupOperationsException
+	 * @throws WrongAttributeValueException
+	 * @throws AlreadyMemberException
```
zlamalp: Again, deleting member (all members) shouldn't throw stuff like ``AlreadyMemberException``, try/catch it in Bl and throw ``ConsistencyErrorException``.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -278,9 +280,11 @@
 	 * @throws MemberAlreadyRemovedException
 	 * @throws UserAlreadyRemovedException
 	 * @throws SpecificUserAlreadyRemovedException
-	 * @throws GroupOperationsException
+	 * @throws WrongAttributeValueException
+	 * @throws AlreadyMemberException
+	 * @throws WrongReferenceAttributeValueException
 	 */
-	void deleteUser(PerunSession perunSession, User user, boolean forceDelete) throws InternalErrorException, UserNotExistsException, PrivilegeException, RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException, GroupOperationsException;
+	void deleteUser(PerunSession perunSession, User user, boolean forceDelete) throws InternalErrorException, UserNotExistsException, PrivilegeException, RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException, WrongAttributeValueException, AlreadyMemberException, WrongReferenceAttributeValueException;
```
zlamalp: Same as comment on:
```java
void deleteUser(PerunSession perunSession, User user)
```

VMecko:
Commit is almost fixed, I'm waiting for @stavamichal 's comment about those two attribute exceptions.

stavamichal:
I think that you can throw InternalErrorException where you don't expect of throwing Attribute*Exception from the bottom layer like @zlamalp said.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -1228,7 +1246,7 @@ protected boolean hasCandidateExistingMember(Candidate candidate, RichMember ric
 	/**
 	 * This method run in separate transaction.
 	 */
-	public List<String> synchronizeGroup(PerunSession sess, Group group) throws InternalErrorException, MemberAlreadyRemovedException, AttributeNotExistsException, WrongAttributeAssignmentException, ExtSourceNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException, GroupOperationsException, GroupNotExistsException {
+	public List<String> synchronizeGroup(PerunSession sess, Group group) throws InternalErrorException, MemberAlreadyRemovedException, AttributeNotExistsException, WrongAttributeAssignmentException, ExtSourceNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException, GroupNotExistsException, AlreadyMemberException {
```
zlamalp: Please check ``removeFormerMembersWhileSynchronization()`` used in this method. Its defined to throw ``AlreadyMemberException`` and ``MemberAlreadyRemovedException``, but they are not actually thrown and can be removed from it. Then it can be transitively removed from here and interface class too.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -1461,7 +1479,7 @@ public void run() {
 				log.debug("Synchronization thread for group {} has finished in {} ms.", group, System.currentTimeMillis()-startTime);
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException | InternalErrorException |
 					WrongAttributeAssignmentException | MemberAlreadyRemovedException | GroupNotExistsException |
-					GroupOperationsException | AttributeNotExistsException | ExtSourceNotExistsException e) {
+					AttributeNotExistsException | AlreadyMemberException | ExtSourceNotExistsException e) {
```
zlamalp: Once my previous comment is resolved, you can remove ``AlreadyMemberException`` and ``MemberAlreadyRemovedException`` from catch clause.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -23,9 +23,10 @@
 	 * @param member
 	 * @throws InternalErrorException
 	 * @throws MemberAlreadyRemovedException
-	 * @throws GroupOperationsException
+	 * @throws WrongAttributeValueException
+	 * @throws WrongReferenceAttributeValueException
 	 */
-	void deleteMember(PerunSession sess, Member member) throws InternalErrorException, MemberAlreadyRemovedException, GroupOperationsException;
+	void deleteMember(PerunSession sess, Member member) throws InternalErrorException, MemberAlreadyRemovedException, WrongAttributeValueException, WrongReferenceAttributeValueException;
```
zlamalp: Please remove ``WrongAttributeValueException`` and ``WrongReferenceAttributeValueException``, they are not actually thrown in Bl of this method. After removal, you can remove it also from ``deleteAllMembers()`` and also ``deleteUser()``

PR #1761, opened 2018-02-14
Implementation of reCAPTCHA V2

Vojtech-Sassmann:
* Method 'verifyCaptcha' is now implementing the reCAPTCHA V2.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
@@ -2,6 +2,7 @@ package cz.metacentrum.perun.rpc.methods;
 
 import java.util.*;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import cz.metacentrum.perun.core.api.*;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.registrar.model.*;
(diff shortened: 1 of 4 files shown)


martin-kuba:
I suggest to use the RestTemplate class which is available from the package spring-web.

Vojtech-Sassmann:
I have implemented the verification using the RestTemplate class.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
```diff
@@ -1126,25 +1130,25 @@ public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			HttpHeaders headers = new HttpHeaders();
+			headers.setContentType(MediaType.APPLICATION_FORM_URLENCODED);
 
+			MultiValueMap<String, String> map = new LinkedMultiValueMap<>();
+			map.add("response", response);
+			map.add("secret", secret);
 
+			HttpEntity<MultiValueMap<String, String>> request = new HttpEntity<>(map, headers);
 
+			JSONObject answer = new JSONObject(restTemplate.postForObject(verificationUrl, request , String.class));
```
martin-kuba: It can be written as 
`return restTemplate.postForObject(verificationUrl, request , JsonNode.class).path("success").asBoolean()`

Vojtech-Sassmann: Ok, I will change it.

martin-kuba:
As far as I know, if the Jackson library is in classpath, Spring will find and use it automatically. Just add the package jackson-databind.

Vojtech-Sassmann:
Now it should be ok.

PR #1778, opened 2018-02-26
Upload LDAP scripts to utils scripts - perun git project

stavamichal:
- remove old perun.ldif file from perun-ldapc (it is old and
   deprecated)
 - upload all needed script files for ldapc and ldapc-initializer to
   perun-utils


perun-utils/ldapc-scripts/initialize-ldap.sh
```diff
@@ -0,0 +1,182 @@
...
+    t)
+        TEST=1
+				echo "Initialization executed in the test mode - without any changes!"
+        ;;
+    esac
+done
+
+if [ ! ${TEST} ]; then
+  #If LDAPc is running, inform us and exit without processing
+  LDAPC_PID_FILE=/home/perun/perun-ldapc/pid
```
zlamalp: In ldapc start/stop scripts you also check ``/proc/pid`` if ldapc process is actually running, it would be better to use it here too, since if killed, pid file remains.

zlamalp: We agreed not to check it.

PR #1780, opened 2018-02-27
Fix relations in moveGroup method

balcirakpeter:
I'm not sure how to handle GroupNotExistsException, WrongAttributeValueException and WrongReferenceAttributeValueException while creating and deleting group relations. Could you give me a hint so i can rewrite it?

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -452,7 +452,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		return group;
 	}
 
-	public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupMoveNotAllowedException {
+	public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGroup) throws InternalErrorException, GroupMoveNotAllowedException, WrongAttributeValueException, WrongReferenceAttributeValueException {
 
 		// check if moving group is null
 		if (movingGroup == null) {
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -508,21 +508,44 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
...
+			} catch (GroupRelationNotAllowed e) {
+				//that should never happened
+				throw new InternalErrorException("Group relation cannot be created between destination group "  + destinationGroup + " and moving group " + movingGroup + ".");
 			} catch (GroupNotExistsException e) {
-				throw new InternalErrorException("Some group does not exists...", e);
+				throw new InternalErrorException(e);
+			} catch (WrongAttributeValueException e) {
+				throw new InternalErrorException(e);
+			} catch (WrongReferenceAttributeValueException e) {
+				throw new InternalErrorException(e);
```
stavamichal: If you have the same throw from more than one exception, you should merge them together (if there is no difference).

stavamichal:
> I'm not sure how to handle GroupNotExistsException, WrongAttributeValueException and WrongReferenceAttributeValueException while creating and deleting group relations. Could you give me a hint so i can rewrite it?

GroupNotExistsException should be probably ConstistencyErrorException.
WrongReference and WrongAttributeValue you can throw higher without change.

What do you thikn @zlamalp ?

zlamalp:
I agree with @stavamichal, GroupNotExists would be consistency and attributes error should be probably thrown higher.

balcirakpeter:
Exceptions are fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2857,4 +2838,37 @@ private boolean checkGroupsCycle(PerunSession sess, int resultGroupId, int opera
...
+	 * @param sess perun session
+	 * @param subGroups groups with same parent group
+	 * @throws InternalErrorException
+	 */
+	private void setSubGroupsNames(PerunSession sess, List<Group> subGroups) throws InternalErrorException {
+		List<Group> levelSubGroups = new ArrayList<>();
+
+		while (!subGroups.isEmpty()) {
+			levelSubGroups.clear();
+			for (Group gr : subGroups) {
```
stavamichal: Please use 'subgroup' instead of 'gr'. It is much better for reading the code.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2857,4 +2838,37 @@ private boolean checkGroupsCycle(PerunSession sess, int resultGroupId, int opera
...
+
+	/**
+	 * Method sets subGroups names by their parent group
+	 * Private method for moving groups
+	 *
+	 * @param sess perun session
+	 * @param subGroups groups with same parent group
+	 * @throws InternalErrorException
+	 */
+	private void setSubGroupsNames(PerunSession sess, List<Group> subGroups) throws InternalErrorException {
```
stavamichal: I think this method can be processed recursively. On every subgroup you can call this method for all it's subgroups until any subgroups exists. What do you think? Or you can use method 'getAllSubGroups' and then order them by name from the shortest to the longest and for every such group change the name, because you can be sure, that parent group has always shorter name then it's subgroup.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -508,21 +508,40 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
 				}
 			}
 
-			//First we have to move indirect members
-			List<Member> members = getGroupMembers(sess, movingGroup);
+			//check if there is union between destination group and moving group
+			if(groupsManagerImpl.isOneWayRelationBetweenGroups(destinationGroup, movingGroup)) {
```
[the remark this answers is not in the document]
balcirakpeter: Cycle is checked before this one, so both cases should be covered.

balcirakpeter:
setSubGroupsNames is now using recursion and better name of group attribute.
Duplicity is erased by creating method processRelationsWhileMovingGroup.
Also use of methods updateGroupName, setSubGroupsNames and updateParentGroupId is moved outside of the if/else conditions in moveGroup method.

Vojtech-Sassmann:
Hi @balcirakpeter, don't forget also to edit the method moveGroup in `GroupsManagerMethod`.

balcirakpeter:
moveGroup in GroupsManagerMethod edited.

stavamichal:
I read it and also tested it manually with @Vojtech-Sassmann and it seems to be ok now. I approve this PR.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -575,53 +535,23 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
 				}
 			}
 
-			//After that, we have to remove indirect members from old position
-			List<Member> members = getGroupMembers(sess, movingGroup);
+			processRelationsWhileMovingGroup(sess, destinationGroup, movingGroup);
```
zlamalp: Since I can't comment on unchanged lines...at line 531 there is a comment ``// check if under destination group is group with same short name as Moving group short name`` which is a little confusing, since this branch is resolving moving to the top-level.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
zlamalp: I believe it already uses recursion. We probably could process all sub group at once, but I'm not sure now `just sorting` would work (maybe yes, since we limit possible characters for group names).
But goal here is to modify part of name for all subgroups, shared prefix actually. If you would get old and new prefix on method input, you could perform string replace on all subgroups name prefixes at once (without recursion) and then store them to DB.
But current implementation is clear/readable, which is also benefit.
I would just optimize this and pass also "parentGroup" to the method input, so you don't have to call ``"getParentGroup(sess, subGroup)`` for each subgroup you are renaming. Because you already fixed the name of parent group in previous method run.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2857,4 +2787,77 @@ private boolean checkGroupsCycle(PerunSession sess, int resultGroupId, int opera
...
+
+			setSubGroupsNames(sess, getSubGroups(sess, subGroup));
+		}
+	}
+
+	/**
+	 * Method remove old relation between moving group and its parent,
+	 * also create new relation between destination and moving group if destination is not null
+	 * Private method for moving groups
+	 *
```
zlamalp: Maybe add note to the javadoc, that after calling this method, proper name and parentGroupId must be set to the moving group for keeping DB consistency !

balcirakpeter:
Comments and javaDoc fixed.
parentGroup added to method setSubGroupsNames as an input parameter.

PR #1792, opened 2018-03-08
Uniqueness of attributes to CLI + some modifications

zoraseb:
- Beans Attribute.pm and AttributeDefinition.pm have been enriched with methods isUnique and setUnique.
- CommonArrayRepresentation and CommonArrayRepresentationHeading have been extened with isUnique parameter to.
- Tool getAttribute displays parameter isUnique in Attribute Definition now.
- Tool createAttribute has newly possibility to enter parameter isUnique.
- Newly creaated tool convertAttributeToUnique, which converst existing attribute to unique attribute.
- Into AttributesAgent.pm has been added new method convertAttributeToUnique.
- Perun/AttributesAgent.pm is added too.

### DIFF
--- a/perun-cli/Perun/beans/Attribute.pm
+++ b/perun-cli/Perun/beans/Attribute.pm
@@ -18,6 +18,7 @@ sub toString {
 	my $namespace = $self->getNamespace;
 	my $description = $self->getDescription;
 	my $type = $self->getType;
+	my $unique = $self->isUnique;
 
 	my $str = 'Attribute (';
 	$str .= "id: $id, " if (defined($id));
@@ -27,6 +28,7 @@ sub toString {
 	$str .= "namespace: $namespace, " if (defined($namespace));
 	$str .= "description: $description, " if (defined($description));
 	$str .= "type: $type, " if (defined($type));
+	$str .= "unique: $unique, " if (defined($unique));
 	$str .= ")";
 
 	return $str;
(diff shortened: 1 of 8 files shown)


perun-cli/Perun/beans/Attribute.pm
```diff
@@ -332,13 +342,37 @@ sub setValueFromArray {
...
+	my $self = shift;
+
+	return ($self->{_unique}) ? 'true' : 'false';
+}
+
+sub setUnique
+{
+	my $self = shift;
+	my $val = shift;
+        if (ref $val eq "JSON::XS::Boolean")
```
stavamichal: Wrong indentation. Not just on this line. Please check all lines. Thank you.

perun-cli/Perun/beans/Attribute.pm
```diff
@@ -332,13 +342,37 @@ sub setValueFromArray {
 	}
 }
 
+sub isUnique
+{
+	my $self = shift;
+
+	return ($self->{_unique}) ? 'true' : 'false';
```
stavamichal: We should return "0" instead false and "1" instead true. Both Strings "true" and "false" are in perl evaluate as boolean "true". Also you can't just evaluate $self->{_unique} because if there is object 'JSON::XS::false' it will be evaluate as boolean 'true'.

perun-cli/createAttribute
```diff
@@ -47,6 +49,7 @@ $attribute->setNamespace( $attributeNameSpace );
 $attribute->setDisplayName( $attributeDisplayName );
 $attribute->setType( $attributeType );
 $attribute->setDescription( $attributeDsc ) if (defined($attributeDsc));
+$attribute->setUnique( $unique ) if (defined($unique));
```
stavamichal: If unique is set in Options, you should set "true" there directly.

perun-cli/Perun/beans/Attribute.pm
stavamichal: The JSON::XS::false is really false but only if you have "use JSON::XS" in libs. So you can skip this problem, it is evaluated correctly there.

PR #1793, opened 2018-03-08
copyResource method

VMecko:
created base sketch of copyResource method for further discussions.

-Created copyResource method for easier and automated
 copying of some existing resource when needed.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
@@ -66,18 +66,19 @@ public interface ResourcesManager {
 	Resource createResource(PerunSession perunSession, Resource resource, Vo vo, Facility facility) throws InternalErrorException, PrivilegeException, VoNotExistsException, FacilityNotExistsException, ResourceExistsException;
 
 	/**
-	 * Copy "template" settings(attributes,services,..) from user's another existing resource and create new resource with this template.
+	 * Copy "template" settings(attributes,services,tags (if exists),...) from user's another existing resource and create new resource with this template.
 	 * Resource can be from any of user's facilities.
 	 *
 	 * @param perunSession
-	 * @param resource template resource to copy
-	 * @param newResourcename name of the new Resource
-	 * @param destinationVo virtual organization where the new resource will be created
-	 * @param destinationFacility facility where the new resource will be created.
+	 * @param templateResource template resource to copy
+	 * @param destinationResource destination resource containing destination facility, VO & resource name.
+	 * @param withGroups if set to true and resources ARE from the same VO we also copy all group-resource and member-resource attributes,
+	 *                   if set to true and resources ARE NOT from the same VO we copy resource only attributes,
+	 *                   if set to false we copy resource only attributes.
 	 * @throws InternalErrorException
 	 * @throws ResourceNotExistsException
 	 */
-	Resource copyResource(PerunSession perunSession, Resource resource, String newResourcename, Vo destinationVo, Facility destinationFacility) throws InternalErrorException, ResourceNotExistsException, FacilityNotExistsException, PrivilegeException, VoNotExistsException, ResourceExistsException;
+	Resource copyResource(PerunSession perunSession, Resource templateResource, Resource destinationResource, boolean withGroups) throws InternalErrorException, ResourceNotExistsException, FacilityNotExistsException, PrivilegeException, VoNotExistsException, ResourceExistsException;
 
 	/**
 	 *  Deletes resource by id.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
@@ -72,10 +72,10 @@ public interface ResourcesManager {
 	 *
 	 * @param perunSession
 	 * @param templateResource template resource to copy
-	 * @param destinationResource destination resource containing destination facility, VO and resource name.
-	 * @param withGroups if set to true and resources ARE from the same VO we also - copy all group-resource and member-resource attributes
-	 *                   														   - assign all groups same as on templateResource
-	 *	                 if set to true and resources ARE NOT from the same VO InternalErrorException is thrown,
+	 * @param destinationResource destination resource containing IDs of destination facility, VO and resource name.
+	 * @param withGroups if set to true and resources ARE from the same VO we also
+	 *                      copy all group-resource and member-resource attributes and assign all groups same as on templateResource
+	 *                   if set to true and resources ARE NOT from the same VO InternalErrorException is thrown,
 	 *                   if set to false we will NOT copy groups and group related attributes.
 	 * @throws InternalErrorException
 	 * @throws ResourceNotExistsException
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
@@ -66,6 +66,20 @@ public interface ResourcesManager {
 	Resource createResource(PerunSession perunSession, Resource resource, Vo vo, Facility facility) throws InternalErrorException, PrivilegeException, VoNotExistsException, FacilityNotExistsException, ResourceExistsException;
 
 	/**
+	 * Copy "template" settings(attributes,services,..) from user's another existing resource and create new resource with this template.
+	 * Resource can be from any of user's facilities.
+	 *
+	 * @param perunSession
+	 * @param resource template resource to copy
+	 * @param newResourcename name of the new Resource
+	 * @param destinationVo virtual organization where the new resource will be created
+	 * @param destinationFacility facility where the new resource will be created.
+	 * @throws InternalErrorException
+	 * @throws ResourceNotExistsException
+	 */
+	Resource copyResource(PerunSession perunSession, Resource resource, String newResourcename, Vo destinationVo, Facility destinationFacility) throws InternalErrorException, ResourceNotExistsException, FacilityNotExistsException, PrivilegeException, VoNotExistsException, ResourceExistsException;
+
+	/**
 	 *  Deletes resource by id.
 	 *
 	 * @param perunSession
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
@@ -66,15 +66,17 @@ public interface ResourcesManager {
 	Resource createResource(PerunSession perunSession, Resource resource, Vo vo, Facility facility) throws InternalErrorException, PrivilegeException, VoNotExistsException, FacilityNotExistsException, ResourceExistsException;
 
 	/**
-	 * Copy "template" settings(attributes,services,tags (if exists),...) from user's another existing resource and create new resource with this template.
-	 * Resource can be from any of user's facilities.
+	 * Copy "template" settings from user's another existing resource and create new resource with this template.
+	 * The settings are attributes, services, tags (if exists), groups and their members (if the resources are from the same VO and withGroups is true)
+	 * Template Resource can be from any of user's facilities.
 	 *
 	 * @param perunSession
 	 * @param templateResource template resource to copy
-	 * @param destinationResource destination resource containing destination facility, VO & resource name.
-	 * @param withGroups if set to true and resources ARE from the same VO we also copy all group-resource and member-resource attributes,
-	 *                   if set to true and resources ARE NOT from the same VO we copy resource only attributes,
-	 *                   if set to false we copy resource only attributes.
+	 * @param destinationResource destination resource containing destination facility, VO and resource name.
+	 * @param withGroups if set to true and resources ARE from the same VO we also - copy all group-resource and member-resource attributes
+	 *                   														   - assign all groups same as on templateResource
+	 *	                 if set to true and resources ARE NOT from the same VO InternalErrorException is thrown,
+	 *                   if set to false we will NOT copy groups and group related attributes.
 	 * @throws InternalErrorException
 	 * @throws ResourceNotExistsException
 	 */
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
@@ -58,6 +58,18 @@ public interface ResourcesManagerBl {
 	Resource createResource(PerunSession perunSession, Resource resource, Vo vo, Facility facility) throws InternalErrorException, FacilityNotExistsException, ResourceExistsException;
 
 	/**
+	 * Copy "template" settings from user's another existing resource and create new resource with this template.
+	 * Resource can be from any of user's facilities.
+	 *
+	 * @param perunSession
+	 * @param resource "template" resource
+	 * @param newResourceName name of the new Resource
+	 * @param destinationVo
+	 * @param destinationFacility
+	 */
+	Resource copyResource(PerunSession perunSession, Resource resource, String newResourceName, Vo destinationVo, Facility destinationFacility) throws ResourceExistsException, InternalErrorException;
+
+	/**
 	 *  Deletes resource by id.
 	 *
 	 * @param perunSession
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
@@ -69,10 +69,11 @@ public interface ResourcesManagerBl {
 	 *                      copy all group-resource and member-resource attributes and assign all groups same as on templateResource,
 	 *                   if set to true and resources ARE NOT from the same VO InternalErrorException is thrown,
 	 *                   if set to false we will NOT copy groups and group related attributes.
+	 * @throws ResourceExistsException
 	 * @throws InternalErrorException
-	 * @throws ResourceNotExistsException
+	 * @throws FacilityNotExistsException
 	 */
-	Resource copyResource(PerunSession perunSession, Resource templateResource, Resource destinationResource, boolean withGroups) throws ResourceExistsException, InternalErrorException;
+	Resource copyResource(PerunSession perunSession, Resource templateResource, Resource destinationResource, boolean withGroups) throws ResourceExistsException, InternalErrorException, FacilityNotExistsException;
 
 	/**
 	 *  Deletes resource by id.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
@@ -62,12 +62,13 @@ public interface ResourcesManagerBl {
 	 * Resource can be from any of user's facilities.
 	 *
 	 * @param perunSession
-	 * @param resource "template" resource
-	 * @param newResourceName name of the new Resource
-	 * @param destinationVo
-	 * @param destinationFacility
+	 * @param templateResource "template" resource
+	 * @param destinationResource destination resource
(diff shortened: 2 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -51,6 +51,44 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+		Resource newResource = new Resource();
+		try{
+			Resource existingResource = getResourcesManagerImpl().getResourceByName(sess, destinationVo, destinationFacility, newResourceName);
+			throw new ResourceExistsException(existingResource);
+		} catch (ResourceNotExistsException e){
+			//if it is not taken we create new Resource in DB, then copy everything possible from template resource to our new one.
+			newResource.setName(newResourceName);
+			newResource = getResourcesManagerImpl().createResource(sess, destinationVo, newResource, destinationFacility);
+
+			// We don't copy group_resource/member_resource attributes because they are different for each VO and needs to be set manually.
```
VMecko: What else can we copy except resource attributes and services?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -51,6 +51,44 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+			List<Attribute> templateResourceAttributes = perunBl.getAttributesManagerBl().getAttributes(sess,resource);
+			for (Attribute resourceAttribute : templateResourceAttributes) {
+				try {
+					if (!resourceAttribute.getNamespace().startsWith(AttributesManager.NS_RESOURCE_ATTR_VIRT)) {
```
VMecko: I'm not sure if i can copy virtual resource attributes, if yes then i will remove this if.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -117,6 +117,49 @@ public void createResourceWhenFacilityNotExists() throws Exception {
...
+
+		String newResourceName = "TestResource";
+
+		resource1 = resourcesManager.createResource(sess, resource1, vo, facility);
+		assertNotNull("unable to create resource1 before copying",resource1);
+
+		resourcesManager.copyResource(sess, resource1, newResourceName, vo, facility);
+	}
+
+	@Ignore //copyResource method is not completed yet.
```
VMecko: Test is ignored for now because i don't know if i can PR when test fails. The problem occurs when we call setAttribute in ResourcesManagerBlImpl (70th line). The method tries to copy ID of resource:att-def:core attribute and throws WrongAttributeAssignment exception. How do we solve this issue of copying IDs?

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,20 @@
...
+	 *
+	 * @param perunSession
+	 * @param resource template resource to copy
+	 * @param newResourcename name of the new Resource
+	 * @param destinationVo virtual organization where the new resource will be created
+	 * @param destinationFacility facility where the new resource will be created.
+	 * @throws InternalErrorException
+	 * @throws ResourceNotExistsException
+	 */
+	Resource copyResource(PerunSession perunSession, Resource resource, String newResourcename, Vo destinationVo, Facility destinationFacility) throws InternalErrorException, ResourceNotExistsException, FacilityNotExistsException, PrivilegeException, VoNotExistsException, ResourceExistsException;
```
stavamichal: I think you should use "destinationResource" instead of VO + resourceName + destinationFacility because destinationResource object will have all these information. I am also missing the parameter "withGroups" if both resources are from the same Vo.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
stavamichal: Groups structure  - if both resources are from the same Vo and optional parameter is set to true. In this case you should copy also all group-resource and member-resource attributes from the old setting
Tags - if exists.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
stavamichal: You can't copy virtual attributes. They are computed from other values in Perun. You can copy just attributes with set value (def and opt).

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
stavamichal: You can't copy core attributes. Core attribute is for example 'id' of Object which is get by method "getId" on such Object, but it is returned as core attribute.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -51,31 +51,67 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+							}
+						}
+					}
+				} catch (GroupResourceMismatchException | WrongAttributeValueException | GroupAlreadyAssignedException |
+					WrongAttributeAssignmentException | WrongReferenceAttributeValueException ex) {
+					throw new ConsistencyErrorException("DB inconsistency while copying group-resource attributes. Cause:{}", ex);
+				}
+
+				List<Member> templateResourceMembers = perunBl.getResourcesManagerBl().getAssignedMembers(sess, templateResource);
+				try {
```
VMecko: I'm not sure if this try/catch block is correct because we are assigning member-resource attributes to non-existing members in new Resource. Should i copy all members from template resource to the new one (same idea as with groups above)? If so, should i assign them into corresponding groups?

stavamichal: It is little complicated. Members are always part of existing group. If you are copying groups from resource A to resource B, it means that all members on resource A will be also members on resource B. So for those members you can set member-resource attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,21 @@
 	 */
 	Resource createResource(PerunSession perunSession, Resource resource, Vo vo, Facility facility) throws InternalErrorException, PrivilegeException, VoNotExistsException, FacilityNotExistsException, ResourceExistsException;
 
+	/**
+	 * Copy "template" settings(attributes,services,tags (if exists),...) from user's another existing resource and create new resource with this template.
```
stavamichal: List of copied setting is very important. User need to know what settings will be copied and what is needed to be set manually. Please be more specific about it.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,21 @@
...
+	 * Resource can be from any of user's facilities.
```
stavamichal: Template Resource. Not just any resource.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,21 @@
...
+	 *
+	 * @param perunSession
+	 * @param templateResource template resource to copy
+	 * @param destinationResource destination resource containing destination facility, VO & resource name.
```
stavamichal: It looks like & is used for something different in javadoc. Use "and" instead please.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,21 @@
...
+	 * @param withGroups if set to true and resources ARE from the same VO we also copy all group-resource and member-resource attributes,
+	 *                   if set to true and resources ARE NOT from the same VO we copy resource only attributes,
+	 *                   if set to false we copy resource only attributes.
```
stavamichal: If set to true and resources ARE from the same VO => we will copy not only attributes but also assign all groups same as on the templateResource.
If set to true and resources ARE NOT from the same VO => I would prefer exception to be thrown, because this should be forbidden state.
If set to false => We will not copy groups and group-related attributes. There are still tags, services etc.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
```diff
@@ -57,6 +57,19 @@
...
+	 * Copy "template" settings from user's another existing resource and create new resource with this template.
+	 * Resource can be from any of user's facilities.
+	 *
+	 * @param perunSession
+	 * @param templateResource "template" resource
+	 * @param destinationResource destination resource
+	 * @param withGroups if set to true and resources ARE from the same VO we also copy all group-resource and member-resource attributes,
+	 *                   if set to true and resources ARE NOT from the same VO we copy resource only attributes,
+	 *                   if set to false we copy resource only attributes.
+	 */
```
stavamichal: The same comments as in API javadoc of this method.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,32 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
 		return getResourcesManagerBl().createResource(sess, resource, vo, facility);
 	}
 
+	public Resource copyResource(PerunSession sess, Resource templateResource, Resource destinationResource, boolean withGroups) throws InternalErrorException, ResourceNotExistsException, FacilityNotExistsException, PrivilegeException, VoNotExistsException, ResourceExistsException {
+		Utils.checkPerunSession(sess);
```
stavamichal: You also should check not-null value for both Resource objects.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,32 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+		Vo destinationVo = getPerunBl().getVosManagerBl().getVoById(sess, destinationResource.getVoId());
+		Facility destinationFacility = getPerunBl().getFacilitiesManagerBl().getFacilityById(sess, destinationResource.getFacilityId());
+		getPerunBl().getVosManagerBl().checkVoExists(sess, destinationVo);
+		getPerunBl().getFacilitiesManagerBl().checkFacilityExists(sess, destinationFacility);
```
stavamichal: This is not needed. The Vo was selected from DB by it's ID so it should exist. Same for the Facility.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,32 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+
+		Facility templateResourceFacility = getPerunBl().getFacilitiesManagerBl().getFacilityById(sess, templateResource.getFacilityId());
+
+		//check if user is part of template resource's facility
+		if(!AuthzResolver.isAuthorized(sess, Role.VOADMIN, templateResource) &&
+				!AuthzResolver.isAuthorized(sess,Role.FACILITYADMIN, templateResourceFacility)){
```
stavamichal: VOADMIN is not allowed to see all settings of the Resource. For this reason, probably only FACILITYADMIN should be able to call this. VOADMIN is needed just for copying groups and group related attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,32 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+			throw new PrivilegeException(sess, "Problem with template Resource's facility in copyResource.");
+		}
+
+		//check if user is part of destination's facility
+		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, destinationVo) &&
+				!AuthzResolver.isAuthorized(sess, Role.FACILITYADMIN, destinationFacility)) {
```
stavamichal: Same as for template resource check.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,32 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+			throw new PrivilegeException(sess, "Problem with new Resource's facility in copyResource.");
+		}
+
+		getResourcesManagerBl().checkResourceExists(sess, templateResource);
```
stavamichal: This check should be right after session check.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,32 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+
+		return getResourcesManagerBl().copyResource(sess, templateResource, destinationResource, withGroups);
```
stavamichal: I am still missing a part of code checking if variable "withGroup" is true or false.
False => we don't need to check if resources are from the same VO and user has VOADMIN role in this VO
True => because we want to copy groups, we need to check user's role in VO (it should be VOADMIN) and also both resources should be from the same VO (throw exception if not)

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,23 @@
...
 	Resource createResource(PerunSession perunSession, Resource resource, Vo vo, Facility facility) throws InternalErrorException, PrivilegeException, VoNotExistsException, FacilityNotExistsException, ResourceExistsException;
 
+	/**
+	 * Copy "template" settings from user's another existing resource and create new resource with this template.
+	 * The settings are attributes, services, tags (if exists), groups and their members (if the resources are from the same VO and withGroups is true)
+	 * Template Resource can be from any of user's facilities.
+	 *
+	 * @param perunSession
+	 * @param templateResource template resource to copy
+	 * @param destinationResource destination resource containing destination facility, VO and resource name.
```
stavamichal: Object resource do not contain the whole object facility or vo, just their IDs.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -65,6 +65,23 @@
...
+	 * @param withGroups if set to true and resources ARE from the same VO we also - copy all group-resource and member-resource attributes
+	 *                   														   - assign all groups same as on templateResource
+	 *	                 if set to true and resources ARE NOT from the same VO InternalErrorException is thrown,
+	 *                   if set to false we will NOT copy groups and group related attributes.
```
stavamichal: Weird indentation.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -99,6 +99,41 @@ public Resource createResource(PerunSession sess, Resource resource, Vo vo, Faci
...
+		//check if user is facility admin of destination's facility
+		if (!AuthzResolver.isAuthorized(sess, Role.FACILITYADMIN, destinationFacility)) {
+			throw new PrivilegeException(sess, "User is not facility admin of destination Resource's facility.");
+		}
+
+		if(withGroups) {
+			Vo destinationVo = getPerunBl().getVosManagerBl().getVoById(sess, destinationResource.getVoId());
+			Vo templateVo = getPerunBl().getVosManagerBl().getVoById(sess, templateResource.getVoId());
+
+			if(!destinationVo.equals(templateVo)) {
```
stavamichal: You don't need to get whole VO objects from Perun. You can just check ids of them there and for authorization can be used object resource instead of VO.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -179,6 +179,57 @@ public void copyResourceSameVO() throws Exception{
...
+		Resource createdResource = resourcesManager.getResourceByName(sess, vo, facility, destinationResource.getName());
+		assertNotNull("Resource was not created.", createdResource);
+
+		// group, service and resource tags copy check
+		assertFalse("Group assigned to original resource not copied to destination resource.", resourcesManager.getAssignedGroups(sess, createdResource).isEmpty());
+		assertFalse("Service not copied to destination resource.", resourcesManager.getAssignedServices(sess, createdResource).isEmpty());
+		assertFalse("Resource tag not created for destination resource.", resourcesManager.getAllResourcesTagsForResource(sess, createdResource).isEmpty());
+
+		// group and resource-member attributes check
+		assertFalse("Created resource does not contain group attributes.", perun.getAttributesManagerBl().getAttributes(sess, group).isEmpty());
```
stavamichal: Group's attributes don't have any real connection with resource. So they will exists no matter on any Resource. Also you are not setting any "Group" attribute in the test directly.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -179,6 +179,57 @@ public void copyResourceSameVO() throws Exception{
...
+		assertFalse("Created resource does not contain resource-member attributes.", perun.getAttributesManagerBl().getAttributes(sess, createdResource, member).isEmpty());
```
stavamichal: You can also check if this attribute for "member-resource" is the same and with the same value.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -225,9 +225,10 @@ public void copyResourceWithEverythingFilled() throws Exception{
 		assertFalse("Service not copied to destination resource.", resourcesManager.getAssignedServices(sess, createdResource).isEmpty());
 		assertFalse("Resource tag not created for destination resource.", resourcesManager.getAllResourcesTagsForResource(sess, createdResource).isEmpty());
 
-		// group and resource-member attributes check
-		assertFalse("Created resource does not contain group attributes.", perun.getAttributesManagerBl().getAttributes(sess, group).isEmpty());
-		assertFalse("Created resource does not contain resource-member attributes.", perun.getAttributesManagerBl().getAttributes(sess, createdResource, member).isEmpty());
+		// resource-member attributes check
+		List<Attribute> resMembAttrs = perun.getAttributesManagerBl().getAttributes(sess, createdResource, member);
+		assertFalse("Created resource does not contain any resource-member attributes.", resMembAttrs.isEmpty());
+		assertTrue("Created resource does not contain template resource-member attribute (or copied value of attribute is wrong).",resMembAttrs.contains(resourceMemberAttr));
```
VMecko: I used contains because I think its better option than checking for example name of attributes to see if the attribute was copied and then checking their value to see if some of value is corrupted (it's basically same option, but 'contains' wraps it into one test)

stavamichal: I agree.

PR #1804, opened 2018-03-19
Tool getAttribute has been updated

zoraseb:
New possibility to display attribute definition has been added
(by entering only attribute without other entities)
Entityless attributes can be entered
Combinations member-resource,group-resource,member-group,user-facility are tested and they are not permited in case that combination does not exist.
Testing prints have been deleted, indentation has been corrected.

Method getAllowedMembers was replaced by method getAssignedMembers.
New method isGroupMember is used for check of existing of combination between group and member.


perun-cli/getAttribute
```diff
@@ -116,7 +118,45 @@ if (defined $entitylessSubj) {
 	defined ($uesId) and $parameters{userExtSource} = $uesId;
 }
 
-my @attributes = $attributesAgent->getAttribute( %parameters );
+# find if exists combination of entered parameters
+my $is=0;
+if (exists($parameters{'resource'}) and exists($parameters{'member'})) {
+	my $resourcesAgent = $agent->getResourcesAgent;
+	my @members = $resourcesAgent->getAllowedMembers( resource => $resourceId );
```
zlamalp: I would prefer getting assigned members and not allowed (as you state on your own in error message ``member is not assigned to resource``). Allowed means member is also valid, but assigned represents connection between them.

perun-cli/getAttribute
```diff
@@ -127,8 +167,15 @@ unless (@attributes) {
 my $table = Text::ASCIITable->new( { reportErrors => 0, utf8 => 0 } );
 $table->setCols( 'ID', 'Name', 'Type', 'Value' );
 
+my $aa=0;
 foreach my $attribute (sort $sortingFunction @attributes) {
-	$table->addRow( $attribute->getId, $attribute->getName, $attribute->getType, $attribute->getValueAsScalar );
```
zlamalp: You have conflict on this row, please rebase your commit on current master. You are now missing ``$attribute->isUniqueToPrint`` in table print, which causes this conflict.

perun-cli/getAttribute
```diff
@@ -116,7 +118,45 @@ if (defined $entitylessSubj) {
...
+	my @groups = $resourcesAgent->getAssignedGroups( resource => $resourceId );		             	
+	foreach my $group ( @groups) {
+		if ($group->getId == $groupId) { $is=1;}
+	}
+	if ($is == 0) { die "ERROR: group is not assigned to resource \n";}
+}
+
+if (exists($parameters{'member'}) and exists($parameters{'group'})) {
+	my $groupsAgent = $agent->getGroupsAgent;
+	my @members = $groupsAgent->getGroupMembers( group => $groupId);;
```
zlamalp: I would suggest to get member by its ID and then it's group. Since member usually won't be inside thousands of Groups, but Group can have thousands of Members. Also we have ``isGroupMember(group,member)`` in ``GroupsManagerBlImpl``, So if you would implement Entry and RPC layer for this method, you could use it in a CLI and it would be 1 single select with boolean result.

PR #1808, opened 2018-03-21
Auditer debug logs

Vojtech-Sassmann:
* Added certain debug logs auditer's flush method to discover a bug
regarding invalid auditer's behaviour.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
@@ -395,6 +395,8 @@ public class Auditer {
 
 		for(AuditerMessage message: messages) {
 			for(VirtualAttributesModuleImplApi virtAttrModuleImplApi : registeredAttributesModules) {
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -158,6 +158,7 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
 	@Override
 	public List<String> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, String message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
 		List<String> resolvingMessages = new ArrayList<>();
+		log.debug("Received message {}", message);
```
stavamichal: This log is too frequent. Please remove it and add only "info" log to the Auditer class itself where we will have better control on using these debug messages.

PR #1811, opened 2018-03-23
LDAPC: Added init files to manage LDAPc as service

zlamalp:
- Added init.d script for LDAPc based on engine service.
- Added systemd service wrapper just like for engine service,
  it still requires start/stop scripts inside perun-ldapc folder.

### DIFF
--- a/perun-utils/systemd/perun-ldapc.service
+++ b/perun-utils/systemd/perun-ldapc.service
@@ -1,0 +1,19 @@
+# Systemd unit file for perun-ldapc
+#
+#
+
+[Unit]
+Description=Perun-LDAPc
+After=perun.service
+
+[Service]
+Type=forking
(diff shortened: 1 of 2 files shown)


perun-utils/systemd/perun-ldapc.service
```diff
@@ -0,0 +1,19 @@
+# Systemd unit file for perun-ldapc
```
martin-kuba: This file seems to be unnecessary. The first invocation of "systemctl start perun-ldapc" generates file /run/systemd/generator.late/perun-ldapc.service using systemd-sysv-generator

zlamalp: Does it work like this on RHEL too ? I'm not sure it works like this on RHEL, since they ditched init.d scripts completely and I installed tomcat and engine systemd services manually.

zlamalp:
We need custom systemd unit for RHEL systems, its not used on debian. Merging.

PR #1814, opened 2018-03-29
Extend and improve behavior of virtual attribute value change

stavamichal:
- Logic, which is checking if any virtual attribute module would like to
   generate new auditer messages processing other auditer messages,
   was moved from method "flush" to the method "storeMessagesToDB"
   because method "flush" is not used for all possible scenarios of
   storing messages to auditer log, but method "storeMessagesToDB" is.
 - Newly the method "storeMessageToDb" prepares list of messages with exactly
   one message and then calls "storeMessagesToDB" to prevent unwanted
   duplicity in code.
 - New recursive method "checkRegisteredAttributesModules" was created
   to prepare all messages generated by registered attribute modules
   (we are calling them "resolved messages"). This method is able
   to generate messages by processing resolved messages. This
   funcionality is adding way how to connect two virtual modules by auditer
   messages. It is also preventing cycles, but in the case of existing
   cycle, the list of resolving messages can be incomplete.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
@@ -38,6 +38,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -512,14 +513,22 @@ public class Auditer {
 	 * @param auditerMessages list of AuditerMessages
 	 */
 	public void storeMessagesToDb(final List<AuditerMessage> auditerMessages) {
+		//Avoid working with empty list of auditer-messages
+		if(auditerMessages == null || auditerMessages.isEmpty()) {
+			log.trace("Trying to store empty list of messages to DB!");
+			return;
+		}
+
 		synchronized (LOCK_DB_TABLE_AUDITER_LOG) {
 
 			//Add all possible resolving messages to the bulk
 			try {
 				List<String> messages = new ArrayList<>();
-				PerunSessionImpl session = null;
+
+				//Get perun session from the first message (all sessions should be same from the same principal)
+				PerunSessionImpl session = (PerunSessionImpl) auditerMessages.get(0).getOriginaterPerunSession();
+
 				for (AuditerMessage auditerMessage : auditerMessages) {
-					if (session == null) session = (PerunSessionImpl) auditerMessage.getOriginaterPerunSession();
 					messages.add(auditerMessage.getMessage());
 				}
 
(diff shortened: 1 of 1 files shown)


stavamichal:
It was manually tested without any problem.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -791,6 +777,71 @@ public void initialize() throws InternalErrorException {
...
+		//We still have new resolving messages, so we need to detect if there isn't cycle and if not, continue the recursion
+		if(!addedResolvedMessages.isEmpty()) {
+			//Cycle detection
+			for(String addedResolvedMessage: addedResolvedMessages) {
+				//if message is already present in the list of resolving messages, we are probably in the cycle between two or more registered modules
+				if(alreadyResolvedMessages.contains(addedResolvedMessage)) {
+					//do not continue, just return already resolved messages and inform in the log about this unexpected situation
+					log.error("There is a cycle for resolving message {}. Recursion was stopped forcibly and more resolving messages was not generated! " +
+							"For this reason, some of them may be missing!", addedResolvedMessage);
+					return alreadyResolvedMessages;
```
licehammer: I would only remove repeating message from addedResolvedMessages and continue as usual.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -533,18 +504,45 @@ public int getAuditerMessagesCount(PerunSession perunSession) throws InternalErr
...
-	public void storeMessagesToDb(final List<AuditerMessage> messages) {
+	public void storeMessagesToDb(final List<AuditerMessage> auditerMessages) {
 		synchronized (LOCK_DB_TABLE_AUDITER_LOG) {
+
+			//Add all possible resolving messages to the bulk
+			try {
+				List<String> messages = new ArrayList<>();
+				PerunSessionImpl session = null;
+				for (AuditerMessage auditerMessage : auditerMessages) {
+					if (session == null) session = (PerunSessionImpl) auditerMessage.getOriginaterPerunSession();
```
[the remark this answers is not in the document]
stavamichal: Code was changed to avoid this behavior.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
stavamichal: You are right, this is much better behavior. I changed that.

stavamichal:
Please check it again.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -791,6 +785,70 @@ public void initialize() throws InternalErrorException {
...
+
+		//We still have new resolving messages, so we need to detect if there isn't cycle and if not, continue the recursion
+		if(!addedResolvedMessages.isEmpty()) {
+			//Cycle detection
+			Iterator<String> msgIterator = addedResolvedMessages.iterator();
+			while(msgIterator.hasNext()) {
+				String addedResolvedMessage = msgIterator.next();
+				//If message is already present in the list of resovling messages, remove it from the list of added resolved messages, log it and continue
+				if(alreadyResolvedMessages.contains(addedResolvedMessage)) {
+					log.error("There is a cycle for resolving message {}. This message won't be processed more than once!");
```
licehammer: The variable with message is missing.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -533,18 +505,52 @@ public int getAuditerMessagesCount(PerunSession perunSession) throws InternalErr
...
+	 * Store these resolved messages too.
 	 *
-	 * @param messages list of AuditerMessages
+	 * @param auditerMessages list of AuditerMessages
 	 */
-	public void storeMessagesToDb(final List<AuditerMessage> messages) {
+	public void storeMessagesToDb(final List<AuditerMessage> auditerMessages) {
+		//Avoid working with empty list of auditer-messages
+		if(auditerMessages == null || auditerMessages.isEmpty()) {
+			log.error("Trying to store empty list of messages to DB!");
```
licehammer: Error level seems to be exaggerated. I would use trace level, because calling this method with empty list seems to be common. For example during commit of transaction with no associated audit messages.

stavamichal: Ok.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
stavamichal: Right. Fixed. Thank you.

stavamichal:
Can be checked again.

stavamichal:
In this case the cycle would be between two or more modules and for this reason we have detection of cycles there - attributesDependenciesCheck. So I think this should not happen.

Vojtech-Sassmann:
I have just manually tested the changes and everything seems to be working fine.

PR #1817, opened 2018-04-04
Add new attribute institutionsCountries to LDAPc

stavamichal:
- add processing of the new attribute to the LDAPc event processor
   (updating and removing)
 - add new attribute to ldapc initializer
 - add new attribute to perun ldap schema
 - rearange order of code in ldapc initializer to be more eaiser for
   editing and adding new funcionality in the future


stavamichal:
Manually tested.

PR #1820, opened 2018-04-10
Upgrade to Spring 5

martin-kuba:
- the Spring Platform was upgraded to Cairo-RELEASE which contains Spring 5
- however Spring Platform is going to be discontinued in April 2019, it is recommended to replace it with Spring Boot Starter Parent which is used as the base for Spring Platform
- so Spring Platform was replaced with Spring Boot Starter Parent 2.0.1.RELEASE
- some small fixes were needed, mainly deprecated annotation @TransactionConfiguration was removed


perun-core/src/test/java/cz/metacentrum/perun/core/AbstractPerunIntegrationTest.java
```diff
@@ -50,7 +49,6 @@
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @ContextConfiguration(locations = { "classpath:perun-base.xml", "classpath:perun-core.xml" })
-@Rollback
```
zlamalp: Is now rollback feature default in ``@Transactional(transactionManager = "springTransactionManager")`` ? We usually used ``@TransactionConfiguration(transactionManager = "springTransactionManager", defaultRollback = true)`` in old code. Because transactional is generic annotation, does Spring behave different for "code" and "tests" ?

PR #1824, opened 2018-04-16
ENGINE: Reimplemented Task processing

zlamalp:
- Do not use BlockingBoundedMap to store running Task Futures.
- Related logic was moved directly to BlockingExecutorCompletionService.
- Removed unnecessary logic from SchedulingPool - we gathered Futures
  and Tasks there. Now we only count running/remaining SendTasks
  and on completion Task is removed from pool.

- BlockingExecutorCompletionService now provides blocking on its
  own using same logic like BlockingBoundedMap (using Semaphore).
  Semaphore is released whenever Task / Future finishes (successfully
  or not).
- We can now handle CancellationException, since we get related Task
  for failed Future from running tasks map inside CompletionService.
- Updated setting Task status and times in Collector threads to be more
  close to blocking methods in CompletionService.

- Updated PropagationMaintainerImpl. If any running task (Future) is
  considered as stuck, we try to cancel it first, then Collector
  thread should take it and remove it from the pool.
  If on next check Future is still present and cancelled, then
  we assume collector thread is not working and we report status
  and remove Task from the pool.

### DIFF
--- a/perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/BlockingGenExecutorCompletionService.java
+++ b/perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/BlockingGenExecutorCompletionService.java
@@ -30,6 +30,11 @@ public class BlockingGenExecutorCompletionService implements BlockingCompletionS
 	private final static Logger log = LoggerFactory.getLogger(BlockingGenExecutorCompletionService.class);
 	private CompletionService<Task> completionService;
 	private ConcurrentMap<Future<Task>, Task> executingGenTasks = new ConcurrentHashMap<>();
+	/**
+	 * Provide blocking-waiting behavior to GEN Tasks, which are not started, until semaphore is acquired.
+	 * When job is cancelled or done, semaphore is released. Semaphore shares limit for concurrently running
+	 * GEN Tasks with javas ExecutorCompletionService.
+	 */
 	private Semaphore semaphore;
 
 	/**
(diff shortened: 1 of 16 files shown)


perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/BlockingGenExecutorCompletionService.java
```diff
@@ -25,65 +23,107 @@
...
 public class BlockingGenExecutorCompletionService implements BlockingCompletionService<Task> {
 
 	private final static Logger log = LoggerFactory.getLogger(BlockingGenExecutorCompletionService.class);
 	private CompletionService<Task> completionService;
-	@Autowired
-	@Qualifier("generatingTasks")
-	private BlockingBoundedMap<Integer, Task> executingTasks;
-	private int limit;
+	private ConcurrentMap<Future<Task>, Task> executingGenTasks = new ConcurrentHashMap<>();
+	private Semaphore semaphore;
```
licehammer: Could you add a comment here, describing how the semaphore should work? It will be a huge help to anyone who will try to understand your code.

perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/BlockingGenExecutorCompletionService.java
```diff
@@ -25,65 +23,107 @@
...
 	 */
 	public BlockingGenExecutorCompletionService(int limit) {
-		this.limit = limit;
 		completionService = new ExecutorCompletionService<Task>(Executors.newFixedThreadPool(limit), new LinkedBlockingQueue<Future<Task>>());
+		this.semaphore = new Semaphore(limit);
 	}
 
 	@Override
 	public Future<Task> blockingSubmit(EngineWorker<Task> taskWorker) throws InterruptedException {
+		semaphore.acquire();
```
licehammer: You should expect the following code in this method could throw an exception, in such case you might to release the semaphore.

PR #1826, opened 2018-04-17
Updated resolving value of elixirBonaFideStatus

zlamalp:
- We now consider user a "researcher" also when he has publications
  reported in Perun with thanks to owner "ELIXIR".
  So now attribute module check changes of user:def:publications
  attribute and when getting value its used as a last resort after
  previous processing.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
@@ -1,6 +1,5 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
-
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
@@ -16,6 +15,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -27,8 +27,10 @@
  * The decision depends on attribute 'elixirBonaFideStatusREMS', if it
  * is not empty then the user is a researcher. If it is empty, then this module
  * searches in user's affiliations. If any of those affiliations starts
- * with 'faculty@' then the user is a researcher as well. Otherwise,
- * null value is set.
+ * with 'faculty@' then the user is a researcher as well.
+ * If anything of above fails, attribute looks into user:def:publications and if its value (map)
```
stavamichal: I think this is not correct. Better will be "If none of above success".

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -99,7 +134,8 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
 		Attribute attribute;
 
 		if (uesSetAffiliationAttributeMatcher.find() ||	uesRemoveAffiliationAttributeMatcher.find() ||	uesAllAttrsRemovedMatcher.find() ||
-				userSetMatcher.find() || userRemoveMatcher.find() || userAllAttrsRemovedMatcher.find()) {
+				userSetMatcher.find() || userRemoveMatcher.find() || userAllAttrsRemovedMatcher.find() || userPublicationsSetMatcher.find() ||
+				userPublicationsRemoveMatcher.find()) {
```
stavamichal: Please change the old names of variables "userSetMatcher" and "userRemoveMatcher". Thank you.

PR #1830, opened 2018-04-20
Tool for moving groups

zoraseb:
Moves Group within structure of groups in VO with all it's subgroups. Group id or group name together with vo id or vo short name are required fields. Destination group is optional. If destination group is not defined, moving group (and it's subgroups) is moved as first level group in VO.

### DIFF
--- a/perun-cli/moveGroup
+++ b/perun-cli/moveGroup
@@ -1,0 +1,76 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Getopt::Long qw(:config no_ignore_case);
+use Perun::Agent;
+use Perun::Common qw(printMessage);
+#use Data::Dumper;
+
+sub help {
+	return qq{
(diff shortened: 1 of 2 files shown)


perun-cli/moveGroup
```diff
@@ -0,0 +1,76 @@
...
+	return qq{
+	Moves Group within structure of groups in VO with all it's subgroups. Group id or group name together with vo id or vo short name are required fields. Destination group is optional. If destination group is not defined, moving group (and it's subgroups) is moved as first level group in VO. 
+	------------------------------------
+	Available options:
+	--movingGroupId   | -g moving group Id
+	--movingGroupName | -G moving group name
+	--voId            | -v vo id
+	--voShortName     | -V vo short name
+	--destGroupId     | -d destination group Id 
+        --destGroupName   | -D destination group name
```
zlamalp: Please fix indentation.

PR #1831, opened 2018-04-20
Check assignment of attribute and it's handlers

stavamichal:
- new method checkAttributeAssignment for one or two handlers
 - add this check to method isAuthorizedForAttribute
 - reson for this change is that we want to know if there was wrong check for non-existent types of attributes

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
@@ -15,6 +15,7 @@ import cz.metacentrum.perun.core.api.exceptions.RoleNotSupportedException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.blImpl.AuthzResolverBlImpl;
 import cz.metacentrum.perun.core.impl.Utils;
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -4026,5 +4026,29 @@ public void mergeAttributesValues(PerunSession sess, Member member, List<Attribu
...
+	/**
+	 * Check if attribute is from the same namespace as combination of perunBean handlers.
+	 * Order of handlers does not matter.
+	 *
+	 * @param sess perun session
+	 * @param attributeDefinition the attribute definition
+	 * @param handler1 one of attribute's handlers
+	 * @param handler2 one of attribute's handlers
+	 *
+	 * @throws WrongAttributeAssignmentException if assignment of attribute is correct
```
licehammer: You probably wanted to say when the assignment is **not** correct.

stavamichal: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -49,12 +50,30 @@ public static boolean isAuthorized(PerunSession sess, Role role, PerunBean compl
...
 			return AuthzResolverBlImpl.isAuthorizedForAttribute(sess, actionType, attrDef, primaryHolder, secondaryHolder);
-		} catch (AttributeNotExistsException | ActionTypeNotExistsException ex) {
+		} catch (AttributeNotExistsException | WrongAttributeAssignmentException ex) {
 			throw new InternalErrorException(ex);
 		}
 	}
 
 	/**
+<<<<<<< HEAD
+=======
```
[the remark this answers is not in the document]
stavamichal: Removed.

stavamichal:
Fixed, rebased. Ready for another review or merge.

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -49,11 +50,26 @@ public static boolean isAuthorized(PerunSession sess, Role role, PerunBean compl
...
+	 *
+	 * @param sess perun session
+	 * @param actionType type of action on attribute (ex.: write, read, etc...)
+	 * @param attrDef attribute what principal want to work with
+	 * @return list of roles
+	 * @throws InternalErrorException
+	 * @throws AttributeNotExistsException
+	 * @throws ActionTypeNotExistsException
+	 */
+	public static List<Role> getRolesWhichCanWorkWithAttribute(PerunSession sess, ActionType actionType, AttributeDefinition attrDef) throws InternalErrorException, AttributeNotExistsException, ActionTypeNotExistsException {
```
zlamalp: I don't see any usage of this. Is it a planned feature ? Because we have also #1970, which modify this method in Bl and would probably need a rebase after merging.

stavamichal: You are right. Removed.

PR #1839, opened 2018-05-11
Reinitialize session when user is not recognized

licehammer:
When user register himself to Perun for the firs time (i.e. new User
object is created), he have the session without User object inside,
because there were no User during the registration time. After
registration is approved and user object created, the user can still
have the original session without the User object inside.

With this patch, new session is created each time when there is no User
object in session which fixes described problem.


zlamalp:
Fix proposed in #1840, needs to be tested. This solution would cause session init on each call of anonymous or unrecognized users. Hence two callbacks through /non/ would have different actor (login) generated in API. If the use-case is lazy init once user visit GUI, it's better to be solved by explicit call to getPerunPrincipal() with user+authz refresh.
We probably don't have use-case where some unregistered identity creates an account and calls API directly with new identity (like some service account). This change would solve this case, otherwise implementors have to add explicit call to getPerunPrincipal().

PR #1844, opened 2018-05-15
CLI tool listOfUserRoles for listing of all roles of user

zoraseb:
Tool lists all user's roles similar to tool listOfPrincipalRoles,
but for entered userId.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -1016,6 +1016,18 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1015,6 +1015,18 @@ public static boolean isPerunAdmin(PerunSession sess) {
...
 	}
 
+	/**
+	 * Get all User's roles. Role is defined as a name, translation table is in Role class.
+	 *
+	 * @param sess perun session
+	 * @param user User
+	 * @return list of integers, which represents role from enum Role.
+	 */
+	public static List<String> getUserRoles(PerunSession sess,User user) throws InternalErrorException {
```
zlamalp: There is a mismatch in method name. In above layer ``getUserRoleNames()`` is used, but here it's ``getUserRoles()``. Please use same name everywhere.

PR #1848, opened 2018-05-17
Possibility to remove UserExternalSource in case of persistence is set.

zoraseb:
Tool remoceUserExtSource has new parameter -forceRemove for removing
UserExteSources which are defined as persistant.

### DIFF
--- a/perun-cli/removeUserExtSource
+++ b/perun-cli/removeUserExtSource
@@ -11,44 +11,59 @@ sub help {
 	Removes an external source from the user. User id and external login and external source id or name are required fields
 	------------------------------------
 	Available options:
-	--userId         | -u user id
-	--extSourceId    | -e external source id
-	--extSourceName  | -E external source name
-	--extSourceLogin | -l external login
-	--batch          | -b batch
-	--help           | -h prints this help
+	--userId          | -u user id
+	--extSourceId     | -e external source id
+	--extSourceName   | -E external source name
+	--extSourceLogin  | -l external login
+	--userExtSourceId | -x user external source id
+	--forceRemove     | -f remove persistent external source too
+	--batch           | -b batch
+	--help            | -h prints this help
 
 	};
 }
 
-my ($userId, $extSourceId, $extSourceName, $extSourceLogin, $batch);
+my ($userId, $extSourceId, $extSourceName, $extSourceLogin, $batch, $force,$userExtSourceId);
 GetOptions ("help|h"     => sub {
 		print help();
 		exit 0;
 	}, "userId|u=i"      => \$userId,
-	"extSourceId|e=i"    => \$extSourceId, "extSourceName|E=s" => \$extSourceName,
-	"extSourceLogin|l=s" => \$extSourceLogin, "batch|b" => \$batch) || die help();
+	"extSourceId|e=i"    => \$extSourceId, 
+	"extSourceName|E=s" => \$extSourceName,
+	"extSourceLogin|l=s" => \$extSourceLogin, 
+	"userExtSourceId|x=i" => \$userExtSourceId,
+	"forceRemove|f" => \$force,
+	"batch|b" => \$batch) || die help();
 
 # Check options
 unless (defined($userId)) { die "ERROR: userId is required \n";}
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -454,6 +454,23 @@ public void removeUserExtSource(PerunSession sess, User user, UserExtSource user
 		getUsersManagerBl().removeUserExtSource(sess, user, userExtSource);
 	}
 
+	public void removeUserExtSource(PerunSession sess, User user, UserExtSource userExtSource, boolean forceDelete) throws InternalErrorException, UserNotExistsException, UserExtSourceNotExistsException, PrivilegeException, UserExtSourceAlreadyRemovedException {
```
zlamalp: Implementation should use and check param ``forceDelete``. If true, it should continue, on false I suggest to simply call above method (without forced flag). Also I believe user shouldn't be able to force delete own UES (even if its available in API only), it should be available to Perun admins.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -478,16 +478,24 @@ public UserExtSource call(ApiCaller ac, Deserializer parms) throws PerunExceptio
 	 * Remove user's external sources.
 	 * @param user int User <code>id</code>
 	 * @param userExtSource int UserExtSource <code>id</code>
+	 * @param force boolean. If true, use force deletion.
```
zlamalp: Please remove dot after boolean keyword, it's wrongly parsed by our docs generating tool.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -374,6 +374,21 @@
 	 */
 	void removeUserExtSource(PerunSession perunSession, User user, UserExtSource userExtSource) throws InternalErrorException, UserNotExistsException, UserExtSourceNotExistsException, PrivilegeException, UserExtSourceAlreadyRemovedException;
 
+	/**
+	 * Removes user's external sources.
```
zlamalp: Please use singular, it removes single UES at the time.

perun-cli/removeUserExtSource
```diff
@@ -11,44 +11,59 @@ sub help {
...
+	"extSourceLogin|l=s" => \$extSourceLogin, 
+	"userExtSourceId|x=i" => \$userExtSourceId,
+	"forceRemove|f" => \$force,
+	"batch|b" => \$batch) || die help();
 
 # Check options
 unless (defined($userId)) { die "ERROR: userId is required \n";}
-unless (defined($extSourceId) or defined($extSourceName)) { die "ERROR: extSourceId or extSourceName is required \n";}
+unless (defined $userExtSourceId) {
+	if ((!defined($extSourceId) and !defined($extSourceName)) or !defined($extSourceLogin)) { die "ERROR: extSourceId or extSourceName and extSourceLogin is required \n";}
```
zlamalp: Based on exception text, shouldn't it be like this ? (switched brackets to enclosing OR instead of AND) 
```
if (!defined($extSourceId) and (!defined($extSourceName) or !defined($extSourceLogin)))
```

perun-cli/removeUserExtSource
```diff
@@ -11,44 +11,59 @@ sub help {
...
-}
-if ($extSourceName) {
-	$extSource = $extSourcesAgent->getExtSourceByName( name => $extSourceName );
-	$extSourceId = $extSource->getId;
-}
+unless (defined $userExtSourceId) {
+	if ($extSourceId) {
+		$extSource = $extSourcesAgent->getExtSourceById( id => $extSourceId );
+	}
+	if ($extSourceName) {
```
zlamalp: I would prefer ``elsif`` since if anybody passes both params (id and name), then name takes preference and it overrides extSourceId value. And such behavior is not documented in tool help. We could also die in else branch, but presence of one of params is checked before, so it's not necessary.

PR #1850, opened 2018-05-25
Methods for setting and unseting specific users and sponsored members

stavamichal:
- now we can set or unset specific user (SPONSOR or SERVICE) for any
   existing user
 - now we can set or unset sponsored member for any existing member
 - add new exception to be able to work with already sponsored members
 - add "FALSE" value to DB Compatibility class
 - add simple tests of these new methods

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1101,6 +1101,39 @@ public class MembersManagerEntry implements MembersManager {
 	}
 
 	@Override
+	public RichMember setSponsorshipForMember(PerunSession session, Member sponsoredMember, User sponsor) throws InternalErrorException, MemberNotExistsException, AlreadySponsoredMemberException, UserNotInRoleException, PrivilegeException {
+		Utils.checkPerunSession(session);
+		getPerunBl().getMembersManagerBl().checkMemberExists(session, sponsoredMember);
+
+		if (sponsor == null) {
+			//sponsor is the caller
+			sponsor = session.getPerunPrincipal().getUser();
+		}
+
+		//only Perun Admin should has rights to do this operation
+		if (!AuthzResolver.isAuthorized(session, Role.PERUNADMIN)) {
+			throw new PrivilegeException(session, "Only PerunAdmin should have rights to call this method.");
+		}
+
+		//set member to be sponsored
+		return membersManagerBl.getRichMember(session, membersManagerBl.setSponsorshipForMember(session, sponsoredMember, sponsor));
+	}
+
+	@Override
+	public RichMember unsetSponsorshipForMember(PerunSession session, Member sponsoredMember) throws MemberNotExistsException, MemberNotSponsoredException, InternalErrorException, PrivilegeException {
+		Utils.checkPerunSession(session);
+		getPerunBl().getMembersManagerBl().checkMemberExists(session, sponsoredMember);
+
+		//only Perun Admin should has rights to do this operation
+		if (!AuthzResolver.isAuthorized(session, Role.PERUNADMIN)) {
+			throw new PrivilegeException(session, "Only PerunAdmin should have rights to call this method.");
+		}
+
+		//unset sponsorship for member
+		return membersManagerBl.getRichMember(session, membersManagerBl.unsetSponsorshipForMember(session, sponsoredMember));
+	}
+
+	@Override
 	public RichMember sponsorMember(PerunSession session, Member sponsored, User sponsor) throws InternalErrorException, PrivilegeException, MemberNotSponsoredException, AlreadySponsorException, UserNotInRoleException {
 		Utils.checkPerunSession(session);
 		Utils.notNull(sponsored, "sponsored");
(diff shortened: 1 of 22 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -284,6 +301,15 @@ public void removeSponsor(PerunSession session, Member sponsoredMember, User spo
 		}
 	}
 
+	@Override
+	public void deleteAllSponsors(PerunSession session, Member sponsoredMember) throws InternalErrorException {
+		try {
+			jdbc.update("DELETE FROM members_sponsored WHERE sponsored_id=?", sponsoredMember.getId() );
```
martin-kuba: This bulk deletion of all sponsors in the database will not generate needed auditlog messages, so I think that this method should not be implemented on the database level, it should be implemented on the service level where it will find all sponsors and remove them one by one using the method MembesManagerBl.removeSponsor(), so that an auditlog message will be generated for each one and expiration would be set correctly.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/MembersManagerImplApi.java
```diff
@@ -215,6 +236,15 @@
...
 	void removeSponsor(PerunSession sess, Member sponsoredMember, User sponsor) throws InternalErrorException;
 
+	/**
+	 * Delete all existing sponsorships for defined member. This method will delete them, not just marked.
+	 *
+	 * @param session perun session
+	 * @param sponsoredMember member which is sponsored
+	 * @throws InternalErrorException
+	 */
+	void deleteAllSponsors(PerunSession session, Member sponsoredMember) throws InternalErrorException;
```
martin-kuba: This method should not exist on the database level, vis. my comment for its implementation.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
stavamichal: This method is only in "impl" layer and need to be used with caution. You can see it is used only in method unsetSponsorshipForMember and this method takes care of all necessary messages.
Also we don't want to change expiration when removing sponsored state of a Member.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/MembersManagerImplApi.java
stavamichal: The only purpose of this method is to delete all existing connections of member to his sponsors. It is a problem of methods from business logic to take care of needed audit messages.

stavamichal:
We don't want to change expiration at all in these cases:
1] sponsored member -> normal member
2] normal member -> sponsored member
3] normal user -> sponsored user | service user
4] service user -> normal user
5] sponsored user -> normal user

So there is no such functionality in this logic.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -1380,6 +1380,31 @@ public void createSponsoredMember() throws Exception {
 		assertTrue("sponsored member without sponsors must be expired",sponsoredMember.getStatus()==Status.EXPIRED);
 	}
 
+	@Test
+	public void setAndUnsetSponsorshipForMember() throws Exception {
+		System.out.println(CLASS_NAME + "setSponsorshipForMember");
```
Vojtech-Sassmann: Invalid test name.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -155,6 +155,46 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Set specific user type for specific user and set ownership of this user for the owner.
+	 *
+	 * @param specificUser int User <code>id</code>
+	 * @param specificUserType String specific user type
+	 * @param user int User <code>id</code>
```
Vojtech-Sassmann: Probably should be `owner` instead of `user`.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1036,6 +1036,39 @@ public RichMember createSponsoredMember(PerunSession session, Vo vo, String name
...
+	public RichMember unsetSponsorshipForMember(PerunSession session, Member sponsoredMember) throws MemberNotExistsException, MemberNotSponsoredException, InternalErrorException, PrivilegeException {
+		Utils.checkPerunSession(session);
+		getPerunBl().getMembersManagerBl().checkMemberExists(session, sponsoredMember);
+
+		//only Perun Admin should has rights to do this operation
+		if (!AuthzResolver.isAuthorized(session, Role.PERUNADMIN)) {
+			throw new PrivilegeException(session, "Only PerunAdmin should have rights to call this method.");
+		}
+
+		//set member to be sponsored
```
martin-kuba: comment should be "unset sponsorship for member"

stavamichal: fixed

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
stavamichal: fixed

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
stavamichal: fixed

stavamichal:
@martin-kuba If it is ok now, please allow this PR to be merged. Thank you.

PR #1873, opened 2018-06-13
CORE,GUI,CLI: Use CSRF protection in Perun API

zlamalp:
- CSRF protection is implemented using custom filter.
- We use cookie based tokens, since JS client need to access them
  from the browser and also CLI will need to send the cookie value
  in the header.
- Token is valid as long as http session.
- For state changing requests (like POST/PUT) we check, that token
  value is the same for all Cookie, Header and server side Session.

CORE: Read allowed CORS domains from perun config

- Allow CORS only from trusted domains from perun config.

CLI: Support CSRF protection in our API

- Since creation of Agent instance calls API, we will get
  CSRF cookie and we can use it later for all requests made by
  our Agent instance (set as default header).
- For "localhost" we change domain to "localhost.local" since
  its stored like this in a cookie_jar.
- Configuration of cookie file is respected, new cookie is stored
  along side PERUNSESSION cookie.

GUI,RPC-LIB: Support for CSRF (send header if cookie is present).

!! After merge it must be deployed also with update to WUI and any third-party application calling our API !!

### DIFF
--- a/perun-base/src/main/resources/perun-base.xml
+++ b/perun-base/src/main/resources/perun-base.xml
@@ -46,6 +46,7 @@
 		<property name="attributesForUpdateX509" value="#{'${perun.attributesForUpdate.x509}'.split('\s*,\s*')}"/>
 		<property name="oidcIssuers" value="#{'${perun.oidc.issuers}'.split('\s*,\s*')}"/>
 		<property name="allowedCorsDomains" value="#{'${perun.allowedCorsDomains}'.split('\s*,\s*')}" />
+		<property name="csrfEnabled" value="${perun.csrfEnabled}" />
 	</bean>
 
 
@@ -101,6 +102,7 @@
 				<prop key="perun.instanceId">AOJ26J3D9DCK3OA7</prop>
 				<prop key="perun.instanceName">LOCAL</prop>
 				<prop key="perun.allowedCorsDomains"></prop>
+				<prop key="perun.csrfEnabled">false</prop>
 				<!--
 				   this creates a map from OIDC issuer to user extsources that are used for looking up a user identified by "sub" claim
 				-->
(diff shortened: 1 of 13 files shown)


perun-base/src/main/resources/perun-base.xml
```diff
@@ -45,6 +45,8 @@
 		<property name="attributesForUpdateIdP" value="#{'${perun.attributesForUpdate.idp}'.split('\s*,\s*')}"/>
 		<property name="attributesForUpdateX509" value="#{'${perun.attributesForUpdate.x509}'.split('\s*,\s*')}"/>
 		<property name="oidcIssuers" value="#{'${perun.oidc.issuers}'.split('\s*,\s*')}"/>
+		<property name="allowedCorsDomains" value="#{'${perun.allowedCorsDomains}'.split('\s*,\s*')}" />
+		<property name="csrfEnabled" value="${perun.csrfEnabled}" />
```
martin-kuba: I do think that the purpose of this flag is to prevent CSRF attacks, not to enable them :-) Thus the name should be the opposite, something like csrfProtectionEnabled.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
```diff
@@ -435,9 +435,18 @@ protected void doOptions(HttpServletRequest req, HttpServletResponse resp) throw
 
 	private void mirrorOriginHeader(HttpServletRequest req, HttpServletResponse resp) {
```
martin-kuba: The method no longer mirrors the header, so it should be renamed.

perun-base/src/main/resources/perun-base.xml
zlamalp: Renamed to csrfProtectionEnabled.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
zlamalp: Changed to ```checkOriginHeader()```.

zlamalp:
CSRF protection would block OAuth2 clients, also on tests it seems, that browsers provides necessary protection. Closing this pull-request.

PR #1886, opened 2018-06-27
Updated eduPersonScopedAffiliations attribute module

dBucik:
Attribute eduPersonScopedAffiliations is now a combination of values from UserExtSources and manually assigned affiliations in the eduPersonScopedAffiliationsManuallyAssigned attribute.

Changes to getAttributeValue(...):
    - attribute now checks also value of urn:perun:user:attribute-def:def:eduPersonScopedAffiliationsManuallyAssigned attribute
 Changes to resolveVirtualAttributeValueChange(...)
    - attribute now reacts also to changes in value of urn:perun:user:attribute-def:def:eduPersonScopedAffiliationsManuallyAssigned attribute

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
@@ -1,18 +1,59 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributeCollectedFromUserExtSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Pattern;
 
 /**
- * All affiliations collected from UserExtSources attributes.
+ * All affiliations collected from UserExtSources attributes and eduPersonScopedAffiliationsManuallyAssigned.
  *
  * @author Martin Kuba makub@ics.muni.cz
+ * @author Dominik Frantisek Bucik <bucik@ics.muni.cz>
  */
 @SuppressWarnings("unused")
 public class urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations extends UserVirtualAttributeCollectedFromUserExtSource {
 
+	private final Logger log = LoggerFactory.getLogger(this.getClass());
+
+	private final Pattern userAllAttrsRemovedPattern = Pattern.compile("All attributes removed for User:\\[(.*)]");
+	private final Pattern userEPSAMASetPattern = Pattern.compile("Attribute:\\[(.*)friendlyName=<" + getSecondarySourceAttributeFriendlyName() +">(.*)] set for User:\\[(.*)]");
+	private final Pattern userEPSAMARemovePattern = Pattern.compile("AttributeDefinition:\\[(.*)friendlyName=<" + getSecondarySourceAttributeFriendlyName() + ">(.*)] removed for User:\\[(.*)]");
+
+
 	@Override
 	public String getSourceAttributeFriendlyName() {
 		return "affiliation";
+	}
+
+	/**
+	 * Get friendly name of secondary source attribute
+	 * @return friendly name of secondary source attribute
+	 */
+	public String getSecondarySourceAttributeFriendlyName() {
+		return "eduPersonScopedAffiliationsManuallyAssigned";
+	}
+
+	/**
+	 * Get name of secondary source attribute
+	 * @return name of secondary source attribute
+	 */
+	public String getSecondarySourceAttributeName() {
+		return AttributesManager.NS_USER_ATTR_DEF + ":" + getSecondarySourceAttributeFriendlyName();
 	}
 
 	@Override
@@ -20,4 +61,46 @@ public class urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations exten
 		return "eduPersonScopedAffiliations";
 	}
 
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDefinition destinationAttributeDefinition) throws InternalErrorException {
+		//get already filled value obtained from UserExtSources
+		Attribute attribute = super.getAttributeValue(sess, user, destinationAttributeDefinition);
+
+		Attribute destinationAttribute = new Attribute(destinationAttributeDefinition);
+		//get values previously obtained and add them to Set representing final value
+		//for values use set because of avoiding duplicities
+		Set<String> valuesWithoutDuplicities = new HashSet<>(attribute.valueAsList());
+
+		Attribute manualEPSAAttr = null;
+		try {
+			//get value from urn:perun:user:attribute-def:def:eduPersonScopedAffiliationsManuallyAssigned
+			manualEPSAAttr = sess.getPerunBl().getAttributesManagerBl()
+					.getAttribute(sess, user, getSecondarySourceAttributeName());
+		} catch (WrongAttributeAssignmentException e) {
+			throw new InternalErrorException("Wrong assignment of " + getSecondarySourceAttributeFriendlyName() + " for user " + user.getId(), e);
+		} catch (AttributeNotExistsException e) {
+			log.debug("Attribute " + getSecondarySourceAttributeFriendlyName() + " of user " + user.getId() + "does not exist, values will be skipped", e);
+		}
+
+		if (manualEPSAAttr != null) {
+			Map<String, String> value = manualEPSAAttr.valueAsMap();
+			if (value != null) {
+				valuesWithoutDuplicities.addAll(value.keySet());
+			}
+		}
+
+		//convert set to list (values in list will be without duplicities)
+		destinationAttribute.setValue(new ArrayList<>(valuesWithoutDuplicities));
+		return destinationAttribute;
+	}
+
+	@Override
+	public List<Pattern> getPatternsForMatch() {
+		List<Pattern> patterns = super.getPatternsForMatch();
+		patterns.add(userAllAttrsRemovedPattern);
+		patterns.add(userEPSAMARemovePattern);
+		patterns.add(userEPSAMASetPattern);
+
+		return patterns;
+	}
 }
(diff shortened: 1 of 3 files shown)


stavamichal:
There is an error in test: 
`urn_perun_user_attribute_def_virt_eduPersonScopedAffiliationsTest.getAttributeValue:67 collected values are incorrect expected:<[affiliate@company.com, library-walk-in@company.com, member@somewhere.edu]> but was:<[]>`

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -1,15 +1,40 @@
...
+ * @author Dominik Frantisek Bucik <bucik@ics.muni.cz>
  */
 @SuppressWarnings("unused")
 public class urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations extends UserVirtualAttributeCollectedFromUserExtSource {
 
+	private final Logger log = LoggerFactory.getLogger(this.getClass());
+
+	//eduPersonScopedAffiliationsManuallyAssigned
+	private final String SOURCE_ATTRIBUTE_FRIENDLY_NAME_EPSA_MA = "eduPersonScopedAffiliationsManuallyAssigned";
+	private final String SOURCE_ATTRIBUTE_FRIENDLY_NAME_EPSA_MA_URN = "urn:perun:user:attribute-def:def:eduPersonScopedAffiliationsManuallyAssigned";
```
stavamichal: You can use
`AttributesManager.NS_USER_ATTR_DEF + ":" + SOURCE_ATTRIBUTE_FRIENDLY_NAME_EPSA_MA `

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -20,4 +45,73 @@ public String getDestinationAttributeFriendlyName() {
...
+			Object value = attribute.getValue();
+			if (value != null && value instanceof Set) {
+				Object[] valArray = ((Set) value).toArray();
+				if (valArray instanceof String[]) {
+					for (String val: (String[]) valArray) {
+						valuesWithoutDuplicities.add(val);
+					}
+				}
+			}
+		}
```
stavamichal: This is really complicated code. I would prefer something more human-readable. You can work with situation, that super implementation of method getAttributeValue never return null instead of Attribute and also the value (if it is not-null) is always instance of ArrayList (not Set).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -20,4 +45,73 @@ public String getDestinationAttributeFriendlyName() {
...
+
+		Attribute attribute1 = null;
```
stavamichal: Please use more suitable name for this attribute than this one like "manualEPSAAttr" or something like that.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -20,4 +45,73 @@ public String getDestinationAttributeFriendlyName() {
...
+		try {
+			//get value from urn:perun:user:attribute-def:def:eduPersonScopedAffiliationsManuallyAssigned
+			attribute1 = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, SOURCE_ATTRIBUTE_FRIENDLY_NAME_EPSA_MA_URN);
+		} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+			log.error("cannot read " + SOURCE_ATTRIBUTE_FRIENDLY_NAME_EPSA_MA + " of user " + user.getId(), e);
```
stavamichal: This is not the right way how to catch and process WrongAttributeAssignment exception. Wrong assignment is probably error in the code itself so it can be thrown as InternalErrorException if needed.
For AttributeNotExists exception you need to choose if it is mandatory attribute or optional attribute for the purpose of getting final value.
If it is mandatory, then you can throw ConsistencyErrorException instead.
If it is optional, you can use more precise message like "Attribute .... not exists so it's values will be skipped" and log it as debug, not as error.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -20,4 +45,73 @@ public String getDestinationAttributeFriendlyName() {
...
+		}
+		if (attribute1 != null) {
+			Object value = attribute1.getValue();
+			if (value != null && value instanceof Map) {
```
stavamichal: If this attribute's type is HashMap, then it will be always HashMap or null so you don't need to test it there.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -20,4 +45,73 @@ public String getDestinationAttributeFriendlyName() {
...
+				String messageAttributeSet;
+				if (attributeValue == null || attributeValue.isEmpty()) {
+					AttributeDefinition attributeDefinition = new AttributeDefinition(attribute);
+					messageAttributeSet = attributeDefinition.serializeToString() + " removed for " + user.serializeToString() + ".";
+				} else {
+					messageAttributeSet = attribute.serializeToString() + " set for " + user.serializeToString() + ".";
+				}
+				resolvingMessages.add(messageAttributeSet);
+			}
+			if(!resolvingMessages.isEmpty()) log.debug("These new messages will be generated: " + resolvingMessages);
```
stavamichal: I think this method can be fully override, because it looks very similar and there is no really advantage of using the code from the super method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
martin-kuba: The Attribute class has the method valueAsList() which returns its value as ArrayList<String>.

dBucik:
Requested changes were made. I have also updated the tests to test more specific situations.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -1,23 +1,100 @@
...
+import cz.metacentrum.perun.core.api.*;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributeCollectedFromUserExtSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.*;
```
stavamichal: If possible, evade wildcards.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -1,23 +1,100 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
+import cz.metacentrum.perun.core.api.*;
```
stavamichal: If possible, evade wildcards.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -1,23 +1,100 @@
...
+		//for values use set because of avoiding duplicities
+		Set<String> valuesWithoutDuplicities = new HashSet<>(attribute.valueAsList());
+
+		Attribute manualEPSAAttr = null;
+		try {
+			//get value from urn:perun:user:attribute-def:def:eduPersonScopedAffiliationsManuallyAssigned
+			manualEPSAAttr = sess.getPerunBl().getAttributesManagerBl()
+					.getAttribute(sess, user, getSecondarySourceAttributeName());
+		} catch (WrongAttributeAssignmentException e) {
+			log.error("Wrong assignment of " + getSecondarySourceAttributeFriendlyName() + " for user " + user.getId(), e);
```
stavamichal: Throwing an exception is similar to "loggin" so you don't need to use both.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -154,16 +154,26 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
 	@Override
 	public List<String> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, String message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
 		List<String> resolvingMessages = new ArrayList<>();
 		if (message == null) return resolvingMessages;
-
-		if (allAttributesRemovedForUserExtSource.matcher(message).find() ||
-				removeUserExtSourceAttribute.matcher(message).find() ||
-				setUserExtSourceAttribute.matcher(message).find()) {
-			log.debug("Resolving virtual attribute value change for attribute " + this.getSourceAttributeFriendlyName() + " and message: " + message);
+		System.out.println(getPatternsForMatch().size());
```
stavamichal: You can remove this line. It is probably just for debugging.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliationsTest.java
```diff
@@ -1,20 +1,11 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
-import cz.metacentrum.perun.core.api.Attribute;
-import cz.metacentrum.perun.core.api.ExtSource;
-import cz.metacentrum.perun.core.api.User;
-import cz.metacentrum.perun.core.api.UserExtSource;
-import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.*;
```
stavamichal: Evade of using wildcards.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliationsTest.java
```diff
@@ -1,20 +1,11 @@
...
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import org.junit.Before;
 import org.junit.Test;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
+import java.util.*;
```
stavamichal: Evade of using wildcards.

PR #1889, opened 2018-06-29
introduced versioning, removed profiles

martin-kuba:
Cancels continuous development, introduces versioning.
- removed spring profile "devel", so only "default" and "production" exist, default uses in-memory database for tests and default properties, production reads perun.properties and uses real database
-  removed all existing Maven profiles, introduced profile "oracle" to perun-core, perun-ldap-initializer, and perun-auditer-exporter, which adds Oracle drivers (during tests in perun-code, to executable JARs in the two other cases)
- changed version to 3.1.0-SNAPSHOT in anticipation of release 3.1.0

### DIFF
--- a/perun-auditer-exporter/pom.xml
+++ b/perun-auditer-exporter/pom.xml
@@ -11,7 +11,6 @@
 
 	<groupId>cz.metacentrum.perun</groupId>
 	<artifactId>perun-auditer-exporter</artifactId>
-	<version>3.1.0-SNAPSHOT</version>
 	<packaging>jar</packaging>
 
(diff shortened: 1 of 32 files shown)


martin-kuba:
Two changes added:
- tests of perun-core can now be run against Oracle DB using 
```
mvn -Dspring.profiles.active=production -Doracle=True test
```
- perun-auditer-exporter.jar and perun-ldapc-initializer.jar are created containing Oracle drivers if the build is run using
```
mvn -Doracle=True install
```
where the value of the property `oracle` can be specified by Ansible script to be either True or False.

PR #1893, opened 2018-07-09
def version of elixirBonaFideStatus attribute

dBucik:
Attribute module for urn:perun:user:attribute:def:def:elixirBonaFideStatus attribute. This attribute replaces the VIRT version with the same name.
Tests for new attribute module
Marked module for VIRT version of attribute as deprecated (will be deleted in future)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatus.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatus.java
@@ -1,0 +1,117 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserAttributesModuleImplApi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.Map;
+
+/**
+ * This module determines if user is a researcher. If so,
+ * it provides URL: 'http://www.ga4gh.org/beacon/bonafide/ver1.0'.
+ *
+ * The decision depends on attributes 'elixirBonaFideStatusREMS', 'eduPersonScopedAffiliations'
+ * and 'user:def:publications'.
+ * If 'elixirBonaFideStatusREMS' is not empty, user is a researcher.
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	@Override
+	public List<String> getDependencies() {
+		List<String> dependencies = new ArrayList<>();
+		dependencies.add(A_U_D_userBonaFideStatusRems);
+		dependencies.add(A_U_V_userAffiliations);
+		dependencies.add(A_U_D_userPublications);
+		return dependencies;
+	}
```
stavamichal: There are no dependencies, because dependencies are strictly bound to getAttributeValue or checkAttributeValue methods.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,145 @@
...
+				String oldAttrValue = (String) oldAttr.getValue();
+				if (!oldAttrValue.isEmpty()) {
+					return oldAttr;
+				}
+			}
+		} catch (WrongAttributeAssignmentException e) {
+			log.error("Cannot read {} from user {}", getAttributeDefinition().getFriendlyName(), user, e);
+		} catch (AttributeNotExistsException e) {
+			//do nothing - if this attribute does not exist it is OK, we will create it
+		}
```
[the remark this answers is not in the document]
dBucik: What will happen in case when all "dependencies" will get removed? Will the value of this attribute persist or will it be deleted as well?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,115 @@
...
+
+	@Override
+	public Attribute fillAttribute(PerunSessionImpl sess, User user, AttributeDefinition attributeDefinition) throws InternalErrorException, WrongAttributeAssignmentException {
+		Attribute attribute = new Attribute(attributeDefinition);
+
+		//try to get value from 'elixirBonaFideStatusREMS': if not empty, we have bona_fide
+		try {
+			Attribute statusAttr = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, A_U_D_userBonaFideStatusRems);
+			if (statusAttr != null && statusAttr.getValue() != null) {
+				attribute.setValue(URL);
```
stavamichal: There should be a return in my opinion.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_elixirBonaFideStatus.java
```diff
@@ -0,0 +1,115 @@
...
+		try {
+			Attribute publicationsAttr = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, A_U_D_userPublications);
+			if (publicationsAttr != null && publicationsAttr.getValue() != null) {
+				Map<String, String> publications = publicationsAttr.valueAsMap();
+				if (publications.containsKey(ELIXIR_KEY)) {
+					String value = publications.get(ELIXIR_KEY);
+					try {
+						int count = Integer.parseInt(value);
+						if (count > 0) {
+							attribute.setValue(URL);
```
stavamichal: Will be more clear if there will be also return after setting value. Even if global return is right after this place.

dBucik:
Changed the return values as requested

PR #1898, opened 2018-07-10
replaced hardcoded perun version with version read from pom.xml

martin-kuba:
The version of Perun reported by RPC Api servlet was hardcoded as "3.0.0". Now it is read dynamically from pom.xml which is automatically packaged by Maven.


martin-kuba:
We agreed that the version reported by /gui/ should be the version of RPC, and the CLI tools should match just the major version of it. CLI should fail if the major version is different.
The admin API is part of the main Perun git repo, so its version is always in sync with the version of RPC.

PR #1915, opened 2018-07-23
Fix wrong reading of jobs in synchronizer

stavamichal:
- methods "Collections.unmodifiableList|unmodifiableSet" do not create
   a new container, they are just views and they still working with
   these concurent containers so runtime exception can be thrown this
   way
 - for both calls we will create a new set and new array to prevent this
   behavior


stavamichal:
I agree. It was fixed.

PR #1922, opened 2018-07-30
Add new attribute modules for lifescience hostel needs

xjancus:
Attribute module has functionality to generate login from first name and last name


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_lifescience_hostel_persistent_shadow.java
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Class for checking logins uniqueness in the namespace and filling lifescience-hostel-persistent id.
+ * It is only storage! Use module login lifescience-hostel_persistent for access the value.
+ *
+ * @author Peter Jancus <p.jancus1996@gmail.com>
+ */
+public class urn_perun_user_attribute_def_def_login_namespace_lifescience_hostel_persistent_shadow extends urn_perun_user_attribute_def_def_login_namespace {
+
+	private final static Logger log = LoggerFactory.getLogger(urn_perun_user_attribute_def_def_login_namespace_bbmri_test_persistent_shadow.class);
```
zlamalp: We should log under the class name of the module. Probably copy/paste.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_lifescience_hostel_persistent_shadow.java
```diff
@@ -0,0 +1,137 @@
...
+	 * @throws cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException
+	 */
+	@Override
+	public void changedAttributeHook(PerunSessionImpl session, User user, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException {
+		try {
+			String userNamespace = attribute.getFriendlyNameParameter();
+
+			if(userNamespace.equals("lifescience-hostel-persistent-shadow") && attribute.getValue() != null){
+				ExtSource extSource = session.getPerunBl().getExtSourcesManagerBl().getExtSourceByName(session, extSourceNameLSH);
+				UserExtSource userExtSource = new UserExtSource(extSource, 0, attribute.getValue().toString());
```
zlamalp: We usually change type of resulting attribute value to String ``((String)attribute.getValue())`` instead of calling ``toString()`` on object.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_lifescience_hostel_persistent_shadow.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+public class urn_perun_user_attribute_def_def_login_namespace_lifescience_hostel_persistent_shadow extends urn_perun_user_attribute_def_def_login_namespace {
+
+	private final static Logger log = LoggerFactory.getLogger(urn_perun_user_attribute_def_def_login_namespace_bbmri_test_persistent_shadow.class);
+	private final static String extSourceNameLSH = "https://login.bbmri-eric.eu/idp/";
+
+	/**
+	 * Filling implemented for login:namespace:lifescience-hostel-persistent attribute
+	 *
+	 * Resulting format/rules:
```
zlamalp: Implementation also crops resulting login to 20 chars, since its limit of MS Active Directory. But there is no indication, if it will be used/stored in AD, also javadoc doesn't specify it. So please make sure, what it supposed to generate and how it will be used. Then change the code or javadoc respectively.

Also we recently find out a problem, when user name is filled with eg. Chinese characters. Then even if name is not empty, it's not used to generate the login and this logic fails. We should make sure, that on empty result we get something for backup.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_lifescience_hostel_persistent_shadow.java
```diff
@@ -0,0 +1,94 @@
...
+				session.getPerunBl().getUsersManagerBl().addUserExtSource(session, user, userExtSource);
+			}
+		} catch (UserExtSourceExistsException ex) {
+			log.warn("LIFESCIENCE HOSTEL IdP external source already exists for the user.", ex);
+		} catch (ExtSourceNotExistsException ex) {
+			throw new InternalErrorException("IdP external source for LIFESCIENCE HOSTEL doesn't exist.", ex);
+		}
+	}
+
+	public AttributeDefinition getAttributeDefinition() {
```
Vojtech-Sassmann: Missing Override annotation.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_login_namespace_lifescience_hostel_persistent.java
```diff
@@ -0,0 +1,68 @@
...
+			throw new InternalErrorException(e);
+		}
+	}
+
+	@Override
+	public List<String> getStrongDependencies() {
+		return Collections.singletonList(SHADOW);
+	}
+
+	public AttributeDefinition getAttributeDefinition() {
```
Vojtech-Sassmann: Missing Override annotation.

PR #1927, opened 2018-08-07
Tools for get and set attribute rights

zoraseb:
Tool getAttributeRights lists permissions READ, WRITE or both for
roles SELF,VOADMIN,GROUPADMIN and FACILITYADMIN
By tool setAttributeRights can be set permissions READ, WRITE or both
to selected role.

### DIFF
--- a/perun-cli/getAttributeRights
+++ b/perun-cli/getAttributeRights
@@ -1,0 +1,66 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Getopt::Long qw(:config no_ignore_case);
+use Text::ASCIITable;
+use Perun::Agent;
+use Perun::Common qw(printMessage tableToPrint getSortingFunction);
+#use Data::Dumper;
+
+sub help {
+	return qq{
+	Prints attribute rights of roles
+	--------------------------------
+	Available options:
+	--attributeId     | -a  attribute identifier
+	--attributeName   | -A  attribute name (including namespace)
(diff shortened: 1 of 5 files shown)


perun-cli/getAttributeRights
```diff
@@ -0,0 +1,66 @@
...
+use Perun::Common qw(printMessage tableToPrint getSortingFunction);
+#use Data::Dumper;
+
+sub help {
+	return qq{
+	Prints attribute rights of roles
+	--------------------------------
+	Available options:
+	--attributeId     | -a  attribute identifier
+	--attributeName   | -A attribute name (including namespace)
```
zlamalp: Please fix description alignment.

perun-cli/setAttributeRights
```diff
@@ -0,0 +1,65 @@
...
+
+my ($attributeId, $attributeName, $reqRole, @rights, $batch);
+GetOptions("help|h"   => sub {
+		print help;
+		exit 0;
+	},
+	"attributeId|a=i"       => \$attributeId,
+	"attributeName|A=s"     => \$attributeName,
+	"userRole|R=s"          => \$reqRole,
+	'rolesRights|r=s@{1,}'  => \@rights,
```
zlamalp: Right now tool can't remove attribute rights, since ``-r`` param is required to have value by the CLI, but empty list / null should be allowed in perun API. Please test it and allow removal either by another CLI tool or allowing empty list of rights.

PR #1928, opened 2018-08-07
Fix behavior of equals method on objects Member and RichMember

stavamichal:
- we should be able to compare objects Member and RichMember if they
   are equal, before this change they were always compared as not equal


stavamichal:
Relation for method equals is not transition, so we can't change it this way.

PR #1929, opened 2018-08-07
Change equals and hashCode methods on specific rich objects in Perun

stavamichal:
- RichMember, RichUser and Candidate will use equals and hashCode methods
   from parent class (Member and User) instead of using it's own
 - this change will allow to compare instance of parent class with
   instance of child class for these mentioned classes and also use this
   behavior correctly for all types of Collections

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/Member.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/Member.java
@@ -131,7 +131,7 @@ public class Member extends Auditable {
 		if (obj == null) {
 			return false;
 		}
-		if (!getClass().equals(obj.getClass())) {
+		if (!(obj instanceof Member)) {
 			return false;
 		}
 		Member other = (Member) obj;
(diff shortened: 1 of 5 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/Member.java
```diff
@@ -128,7 +128,7 @@ public boolean equals(Object obj) {
 		if (this == obj) {
 			return true;
 		}
-		if (obj == null) {
+		if (!(obj instanceof Member)) {
```
zlamalp: Shouldn't this be at line 134 ? Replacing ``.getClass()`` check instead of ``obj == null`` check ? Like in ``User`` object ? I know it will return ``false`` for ``null`` anyway, but this will also return false if you compare ``Member`` and ``RichMember`` which we wanted to be comparable. In ``User`` object you just continue with params comparison when you compare ``User`` and ``Candidate`` or ``RichUser``. I believe implementation should be at least equivalent in both ``Member`` and ``User`` objects.

stavamichal: You are right. It should be fixed now.

PR #1950, opened 2018-08-30
Added new functions to create* methods in RPC.

metodej:
create* methods now accept parameters to create new object and not just JSON object.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
@@ -1275,16 +1275,43 @@ public enum AttributesManagerMethod implements ManagerMethod {
 	 * @param attribute AttributeDefinition object
 	 * @return AttributeDefinition Created AttributeDefinition
 	 */
+	/*#
+	 * Creates AttributeDefinition
+	 *
+	 * @param friendlyName String friendlyName
+	 * @param namespace String namespace
+	 * @param description String description
+	 * @param type String type which is one of the: "java.lang.String", "java.lang.Integer", "java.lang.Boolean", "java.util.ArrayList",
+	 *                                              "java.util.LinkedHashMap", "java.lang.LargeString" or "java.util.LargeArrayList"
+	 * @param displayName String displayName
+	 * @param unique Boolean unique
+	 * @return AttributeDefinition Created AttributeDefinition
+	 * @exampleParam friendlyName [ "kerberosLogins" ]
+	 * @exampleParam namespace [ "urn:perun:user:attribute-def:def" ]
+	 */
 	createAttribute {
 
 		@Override
 		public AttributeDefinition call(ApiCaller ac, Deserializer parms) throws PerunException {
 			ac.stateChangingCheck();
 
-			AttributeDefinition attribute = parms.read("attribute", AttributeDefinition.class);
+			if (parms.contains("attribute")) {
+				return ac.getAttributesManager().createAttribute(ac.getSession(),
+						parms.read("attribute", AttributeDefinition.class));
+			} else if (parms.contains("friendlyName") && parms.contains("namespace") && parms.contains("description") && parms.contains("type")
+					&& parms.contains("displayName") && parms.contains("unique")) {
 
-			return ac.getAttributesManager().createAttribute(ac.getSession(), attribute);
-
+				AttributeDefinition attribute = new AttributeDefinition();
+				attribute.setFriendlyName(parms.readString("friendlyName"));
+				attribute.setNamespace(parms.readString("namespace"));
+				attribute.setDescription(parms.readString("description"));
+				attribute.setType(parms.readString("type"));
+				attribute.setDisplayName(parms.readString("displayName"));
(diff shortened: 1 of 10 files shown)


licehammer:
Do you have any rule when you split command to several lines and when you keep it on one long line?
Now you have some long commands on a single line, but shorter commands split between two or more lines.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -1275,16 +1275,42 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 	 */
+	/*#
+	 * Creates AttributeDefinition
+	 *
+	 * @param friendlyName String friendlyName
+	 * @param namespace String namespace
+	 * @param description String description
+	 * @param type String type
+	 * @param displayName String displayName
+	 * @param writable Boolean writable
```
zlamalp: Writable flag is actually set on attribute retrieval based on users rights for specified attribute. It might be confusing having this option in API - user could think its some generic option. We should add it between ignored properties on read by deserializer instead (project ``perun-base``, package ``...rpc.deserializer``).

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -36,24 +44,60 @@
...
+				if (parms.contains("group")) {
+					Group group = parms.read("group", Group.class);
+					if (group.getParentGroupId() == null) {
+						return ac.getGroupsManager().createGroup(ac.getSession(),
+								ac.getVoById(parms.readInt("vo")),
+								group);
+					} else {
+						throw new RpcException(RpcException.Type.WRONG_PARAMETER, "Top-level groups can't have parentGroupId set!");
+					}
+				} else if (parms.contains("parentGroupId") && parms.contains("name") && parms.contains("description")) {
```
zlamalp: With ``parentGroupId`` present in params you will never get into this branch, since it's first condition on method start. I would rework this and retrieve/construct Group on method start based on ``group / name,description`` presence. Than you either have ``parentGroupId`` or ``vo`` to determine, if new group is a sub-group or top-level.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -36,24 +44,60 @@
 	 * @param group Group JSON Group class
 	 * @return Group Newly created group
 	 */
+	/*#
+	 * Creates a new group in the specific VO defined by object vo in parameter.
+	 *
+	 * @param vo int Parent VO <code>id</code>
+	 * @param parentGroupId Integer parentGroupId
```
zlamalp: Note existing javadoc ``Important: voId in object group is ignored.``. You should equally omit ``parentGroupId`` param for case when you are creating top-level group and passing ``vo`` param.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -23,14 +24,30 @@
 	 * @param service Service JSON object
 	 * @return Service Created Service
 	 */
+	/*#
+	 * Creates a new service.
```
zlamalp: Service object has more properties than name and description, you are free to create new constructor for Service object or use setters. At least script param should be taken on input since it doesn't have default value. You can note in a javadoc, that it should be constructed like ``./service_name`` (where anything else than ``[a-z,A-Z]`` is converted to undrescore ``_``.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -278,12 +278,27 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 	 * @param facility Facility JSON object
 	 * @return Facility Created Facility object
 	 */
+	/*#
+	 * Creates a facility.
```
zlamalp: You might want to note, that caller is made facility manager (same like its for createVo).

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -1275,16 +1275,42 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 	 * @param attribute AttributeDefinition object
 	 * @return AttributeDefinition Created AttributeDefinition
 	 */
+	/*#
+	 * Creates AttributeDefinition
```
zlamalp: Especially for attribute it would be nice to have example param values or specify allowed values in docs explicitly. I'm not sure its documented anywhere and sadly it's no represented as enum either. 

This goes for ``type`` param. Also stating format for ``namespace`` and ``friendlyName`` would be nice.

zlamalp:
Merged locally with small changes.

PR #1952, opened 2018-09-04
ElixirBonaFideStatus implementation

dBucik:
* Implemented logic of the elixirBonaFideStatus - implemented as registration module.
* Attribute module for urn:perun:user:attribute-def:def:elixirBonaFideStatus marked @Deprecated
* tests of attribute module for urn:perun:user:attribute-def:def:elixirBonaFideStatus marked @Deprecated

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/ElixirBonaFideStatus.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/ElixirBonaFideStatus.java
@@ -1,0 +1,146 @@
+package cz.metacentrum.perun.registrar.modules;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.PerunException;
+import cz.metacentrum.perun.registrar.RegistrarManager;
+import cz.metacentrum.perun.registrar.RegistrarModule;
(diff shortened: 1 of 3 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/ElixirBonaFideStatus.java
```diff
@@ -0,0 +1,141 @@
...
+
+	@Override
+	public Application rejectApplication(PerunSession session, Application app, String reason) throws PerunException {
+		return app;
+	}
+
+	@Override
+	public Application beforeApprove(PerunSession session, Application app) throws PerunException {
+		User user = app.getUser();
+		Group group = app.getGroup();
```
zlamalp: Group is present only for group type applications. Are you sure, that nobody will use this registration module on VO ? I would prefer checking group on null before passing it to getAttribute() and not to rely on its check. Maybe skip all logic if group is not present (depends on your use-case). Right now it will throw ugly exception and user/admin will see only something like: "something went wrong - please report".

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/ElixirBonaFideStatus.java
```diff
@@ -0,0 +1,141 @@
...
+	public void canBeApproved(PerunSession session, Application app) throws PerunException {
+
+	}
+
+	/**
+	 * Validate if the user meets criteria for applying to group.
+	 */
+	@Override
+	public void canBeSubmitted(PerunSession session, Map<String, String> params) throws PerunException {
+		User user = session.getPerunPrincipal().getUser();
```
zlamalp: Same as ``beforeApprove()``, but now with ``User`` object. It might not be present, if module is used on VO. There is always a way to wrap all implementation and catch any exception and translate it to generic ``CantBeSubmittedException`` so user will see nicer message and not report error box.

PR #1953, opened 2018-09-04
LDAP: Regex matching must support multi-line messages

zlamalp:
- Since we don't restrict newlines in object descriptions and other
  params, our regexes must match multi-line audit messages.
- Every object content in regex ``(.*)`` was replaced by ``(.|\\s)*``.


zlamalp:
I tested two of regexes locally against auditer messages, which were previously ignored on production instance causing incosistency. It worked and I believe should be generic for any object content in auditer message.

PR #1958, opened 2018-09-10
Attribute modules Override annotations

Vojtech-Sassmann:
* Inserted @Override annotations where missing or moved directly above
 the method declaration if it was above JavaDoc.


Vojtech-Sassmann:
Since these changes are regarding only changes of `@Override` annotations which are checked by the compiler, I don't think this PR needs a complex review.

PR #1959, opened 2018-09-10
Core managers Override annotations

Vojtech-Sassmann:
* Inserted @Override annotations where missing.


Vojtech-Sassmann:
Again, this PR does not require a complex review. Once the build passes, it should be ok.

PR #1972, opened 2018-09-26
create attributes status in vos

kubazak:
create attributes  Status member in Vos

### DIFF
--- a/perun-core/src/main/resources/perun-core.xml
+++ b/perun-core/src/main/resources/perun-core.xml
@@ -143,7 +143,7 @@ http://www.springframework.org/schema/context http://www.springframework.org/sch
 		</tx:attributes>
 	</tx:advice>
 
-	<!--FIXME   hack ktery donuti spring zavolat statickou metodu -->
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_statusesInVos.java
```diff
@@ -0,0 +1,46 @@
...
+        return attribute;
+    }
+
+    @Override
+    public AttributeDefinition getAttributeDefinition() {
+        AttributeDefinition attr = new AttributeDefinition();
+        attr.setNamespace(AttributesManager.NS_USER_ATTR_VIRT);
+        attr.setFriendlyName("statusInVos");
+        attr.setDisplayName("Status member In Vos");
+        attr.setType(Map.class.getName());
```
stavamichal: There has to be 'LinkedHashMap' instead of 'Map'.

PR #1973, opened 2018-09-27
CORE: Fixed passing password in XML in MU pwd-manager

zlamalp:
- We must escape & < > in XML element content (password in this
  case) to be sure we pass valid XML document to the API.
- Fixed logging of password element to be present, but with
  faked password value.


licehammer:
This works. What about creating some XML escape function and use it for all XML fields? I'm pretty sure, that there is almost no risk to have such special characters in fields like name, but still is better to be ready for it.

zlamalp:
Good point @licehammer, I reworked it.

PR #1985, opened 2018-10-04
Reform log

kubazak:
I have modified some logs to meet conventions. In class perun.core.impl.AttributesManagerImpl.
I created new info logs in class perun.core.AttributesManagerBlImpl.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -248,7 +248,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} else if (attrName.startsWith(AttributesManager.NS_MEMBER_GROUP_ATTR)) {
 				memberGroupAttributeNames.add(attrName);
 			} else {
-				log.error("Attribute defined by " + attrName + " is not in supported namespace. Skip it there!");
+				log.warn("Attribute defined by {} is not in supported namespace. Skip it there!", attrName);
 			}
 		}
 
@@ -301,7 +301,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} else if (attributeName.startsWith(AttributesManager.NS_USER_FACILITY_ATTR)) {
 				userFacilityAttirbuteNames.add(attributeName);
 			} else {
-				log.error("Attribute defined by " + attributeName + " is not in supported namespace. Skip it there!");
+				log.warn("Attribute defined by {} is not in supported namespace. Skip it there!", attributeName);
 			}
 		}
 
@@ -565,7 +565,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} else if (attributeName.startsWith(AttributesManager.NS_GROUP_RESOURCE_ATTR)) {
 				groupResourceAttributeNames.add(attributeName);
 			} else {
-				log.error("Attribute defined by " + attributeName + " is not in supported namespace. Skip it there!");
+				log.warn("Attribute defined by {} is not in supported namespace. Skip it there!", attributeName);
 			}
 		}
 
@@ -1381,7 +1381,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			else if (entity instanceof UserExtSource) ues = (UserExtSource) entity;
 				//Else skip not identified entity (log it)
 			else
-				log.debug("In method GetAttributesDefinitionWithRights there are entity which is not identified correctly and will be skipped: " + entity);
+				log.warn("In method GetAttributesDefinitionWithRights there is entity which is not identified correctly and will be skipped: {}", entity);
 		}
 
 		//Iterate through all attributesDefinitions and remove those which are not in the possible namespace or user in session has no rights to read them
@@ -3689,6 +3689,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex) {
 				throw new InternalErrorException(ex);
 			}
+			log.info("{} removed attribute: {} by key {}.",sess.getLogId(), attribute.getName(), key);
 			getPerunBl().getAuditer().log(sess, "{} removed for {}.", new AttributeDefinition(attribute), key);
 		}
 		return changed;
@@ -3730,6 +3731,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex) {
 				throw new InternalErrorException(ex);
 			}
+			log.info("{} removed attribute: {} from facility {}.",sess.getLogId(), attribute.getName(), facility.getId());
 			getPerunBl().getAuditer().log(sess, "{} removed for {}.", new AttributeDefinition(attribute), facility);
 		}
 		return changed;
@@ -3850,6 +3852,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 	public void removeAllAttributes(PerunSession sess, Facility facility) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException {
 		List<Attribute> attributes = getAttributes(sess, facility);
 		getAttributesManagerImpl().removeAllAttributes(sess, facility);
+		log.info("{} removed all attributes from facility {}.", sess.getLogId(), facility.getId());
 		getPerunBl().getAuditer().log(sess, "All attributes removed for {}.", facility);
 
 		for (Attribute attribute : attributes) attribute.setValue(null);
@@ -3882,6 +3885,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		if (removeAlsoUserFacilityAttributes) {
 			List<Attribute> userFacilityAttributes = getUserFacilityAttributesForAnyUser(sess, facility);
 			getAttributesManagerImpl().removeAllUserFacilityAttributesForAnyUser(sess, facility);
+			log.info("{} removed all attributes from any user on facility {}.",sess.getLogId(), facility.getId());
 			getPerunBl().getAuditer().log(sess, "All user-facility attributes removed for {} for any user.", facility);
 
 			for (Attribute attribute : userFacilityAttributes) attribute.setValue(null);
@@ -3929,7 +3933,8 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex) {
 				throw new InternalErrorException(ex);
 			}
-			getPerunBl().getAuditer().log(sess, "{} removed for {}.", new AttributeDefinition(attribute), host);
+			log.info("{} removed attribute {} from host {}.", sess.getLogId(), attribute.getName(), host.getId());
+			getPerunBl().getAuditer().log(sess, "{} removed for {}", new AttributeDefinition(attribute), host);
 		}
 		return changed;
 	}
@@ -3956,7 +3961,8 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 	public void removeAllAttributes(PerunSession sess, Host host) throws InternalErrorException, WrongAttributeValueException {
 		List<Attribute> attributes = getAttributes(sess, host);
 		getAttributesManagerImpl().removeAllAttributes(sess, host);
-		getPerunBl().getAuditer().log(sess, "All attributes removed for {}.", host);
+		log.info("{} removed all attributes from host {}.", sess.getLogId(), host.getId());
+		getPerunBl().getAuditer().log(sess, "All attributes removed for {}", host);
 
 		for (Attribute attribute : attributes) attribute.setValue(null);
 		try {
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -3689,6 +3689,7 @@ public boolean removeAttributeWithoutCheck(PerunSession sess, String key, Attrib
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex) {
 				throw new InternalErrorException(ex);
 			}
+			log.info("{} removed attribute: {} by key = {}",sess.getPerunClient().getId(), attribute.getName(), key);
```
zlamalp: ``PerunClient`` is valid only for OAUTH2 clients. Otherwise its half empty object, especially client ID is not present. I would prefer logging whole ``PerunSession``, but I understand, that this message should be rather short. But neither ``PerunSession`` or ``PerunPrincipal`` within it have a short ``toString()``. Please add method in ``PerunPrincipal`` like: ``getLogId()`` which will return value of ``actor`` property, so we will see user identity. If it won't be sufficient in a future, we can safely change such method to return also extSourceName or other properties. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4131,6 +4141,7 @@ public boolean removeAttributeWithoutCheck(PerunSession sess, Resource resource,
 			throw new InternalErrorException(ex);
 		}
 		if (changed)
+			log.info("{} removed attribute({}) from esource. Id resource = {}",sess.getPerunClient().getId(), attribute.getName(), resource.getId());
```
zlamalp: ``esource`` -> ``resource``
Also you probably don't want () around {} which is replaced for attr name, since in other messages you don't do that.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4335,15 +4335,15 @@ public void convertAttributeValuesToUnique(PerunSession session, AttributeDefini
 								break;
 						}
 						int c = counter.addAndGet(1);
-						if(c%1000==0) log.info(""+c+" values of "+attrDef.getName()+" were converted");
+						if(c%1000==0) log.debug(""+c+" values of "+attrDef.getName()+" were converted");
```
martin-kuba: Do not use string concatenation for constructing log messages, use message formatting with {} as placeholders and variable arguments for values instead. Message formatting is more efficient if the active loglevel is less detailed than the level of this statement, because no new objects need to be created in that case.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -3451,7 +3451,7 @@ public boolean removeVirtualAttribute(PerunSession sess, Resource resource, Grou
 	public boolean removeAttribute(PerunSession sess, User user, AttributeDefinition attribute) throws InternalErrorException {
 		try {
 			if (0 < jdbc.update("DELETE FROM user_attr_values WHERE attr_id=? AND user_id=?", attribute.getId(), user.getId())) {
-				log.info("Attribute (its value) was removed from user. Attribute={}, user={}", attribute, user);
+				log.debug("Attribute (its value) was removed from user. Attribute={}, user={}", attribute, user);
```
martin-kuba: "Attribute value was removed" is more concise than "Attribute (its value) was removed"

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4516,6 +4534,7 @@ public void removeAllUserFacilityAttributes(PerunSession sess, User user) throws
 
 		//remove all non-virtual attributes
 		getAttributesManagerImpl().removeAllUserFacilityAttributes(sess, user);
+		log.info("{} removed all attributes from user on all facilities. User id = ", user.getId());
```
martin-kuba: Should be 
```
log.info("{} removed all attributes from user {} on all facilities", sess.getLogId(), user.getId());
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -4293,6 +4306,7 @@ public boolean removeAttributeWithoutCheck(PerunSession sess, Member member, Gro
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex) {
 				throw new InternalErrorException(ex);
 			}
+			log.info("{} removed this attribute({}) from member in group. Id member = {}, id group = {}",sess.getPerunClient().getId(), attribute.getName(), member.getId(), group.getId());
```
martin-kuba: I would prefer the format to use id next to the entity type, like
```
log.info("{} removed  attribute {} from member {} in group {}",...
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -3929,6 +3933,7 @@ public boolean removeAttributeWithoutCheck(PerunSession sess, Host host, Attribu
 			} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex) {
 				throw new InternalErrorException(ex);
 			}
+			log.info("{} removed this attribute {} from  host {}", sess.getLogId(), attribute.getName(), host.getId());
```
zlamalp: I would remove unnecessary "this" from the message (and other similar messages). Also there is double space before "host" word.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -2110,7 +2110,7 @@ public void deleteAttribute(PerunSession sess, AttributeDefinition attribute) th
 
 			jdbc.update("DELETE FROM "+ attributeToTablePrefix(attribute)+"_attr_values WHERE attr_id=?", attribute.getId());
 			jdbc.update("DELETE FROM attr_names WHERE id=?", attribute.getId());
-			log.info("Attribute deleted [{}]", attribute);
+			log.debug("Attribute deleted [{}]", attribute);
```
zlamalp: Attribute object is unncessary enclosed in ``[ ]``, opposite to createAttribute() method/message. I suggest ``Attribute created: {}`` and ``Attribute deleted: {}``.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -2731,7 +2731,7 @@ public Attribute fillAttribute(PerunSession sess, Resource resource, Attribute a
 		//Use attributes module
 		ResourceAttributesModuleImplApi attributeModule = getResourceAttributeModule(sess, attribute);
 		if (attributeModule == null) {
-			log.debug("fillAttribute - There's no attribute module for this attribute. Attribute wasn't filled. Attribute={}", attribute);
+			log.debug("fillAttribute - There's no attribute module for this attribute. Attribute {} wasn't filled.", attribute);
```
zlamalp: Original message is not ok, but having placeholder for attribute object at the end has meaning. Since ``toString()`` on Attribute if pretty long, you won't see "wasn't filled." on first sight. Maybe rephrase whole message, like: ``Attribute wasn't filled. There is no module for: {}`` and place attribute object at the end. Also I think that attribute name is sufficient in this message, so you don't have to pass the whole object.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -3317,7 +3317,7 @@ public void removeAllAttributes(PerunSession sess, Resource resource) throws Int
 	public boolean removeAttribute(PerunSession sess, Resource resource, Member member, AttributeDefinition attribute) throws InternalErrorException {
 		try {
 			if (0 < jdbc.update("DELETE FROM member_resource_attr_values WHERE attr_id=? AND member_id=? AND resource_id=?", attribute.getId(), member.getId(), resource.getId())) {
-				log.info("Attribute (its value) was removed from member on resource. Attribute={}, member={}, resource=" + resource, attribute, member);
+				log.debug("Attribute value {} was removed from member {} on resource {}" + resource, attribute, member);
```
zlamalp: Technically its usefull to have full objects in the message, but also in this case we can't relay on previous (correct) value to be present in passed ``AttributeDefinition`` (which, in a fact, lacks value). So maybe rephrasing all message to something like this would help: ``Attribute value for {} was removed from member {} on resource {}.`` where you again pass attribute name and then member and resource objects.
Also there is an inconsistency with ending dot between this kind of messages.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -1381,7 +1381,7 @@ public AttributeDefinition getAttributeDefinition(PerunSession sess, String attr
 			else if (entity instanceof UserExtSource) ues = (UserExtSource) entity;
 				//Else skip not identified entity (log it)
 			else
-				log.debug("In method GetAttributesDefinitionWithRights there are entity which is not identified correctly and will be skipped: " + entity);
+				log.warn("In method GetAttributesDefinitionWithRights there are entity which is not identified correctly and will be skipped: {}", entity);
```
zlamalp: there are entity -> there is entity

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -3882,6 +3885,7 @@ public void removeAllAttributes(PerunSession sess, Facility facility, boolean re
 		if (removeAlsoUserFacilityAttributes) {
 			List<Attribute> userFacilityAttributes = getUserFacilityAttributesForAnyUser(sess, facility);
 			getAttributesManagerImpl().removeAllUserFacilityAttributesForAnyUser(sess, facility);
+			log.info("{} removed all attributes from any user facility {}.",sess.getLogId(), facility.getId());
```
zlamalp: from any user facility {} -> from any user on facility {}

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -6934,7 +6959,7 @@ private void updateInverseDependenciesForAttribute(Map<AttributeDefinition, Set<
 					AttributeDefinition attrDef = getAttributeDefinition(sess, s);
 					strongDepSet.add(attrDef);
 				} catch (AttributeNotExistsException ex) {
-					log.error("For attribute dependency name " + s + "can't be found attributeDefinition.");
+					log.error("For attribute dependency name {} can't be found attributeDefinition.", s);
```
zlamalp: I would prefer something like: ``AttributeDefinition can't be found for dependency {}.`` Same for the else branch.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -7081,7 +7106,7 @@ public void convertAttributeToUnique(PerunSession session, int attrId) throws In
 		long startTime = System.currentTimeMillis();
 		attributesManagerImpl.convertAttributeValuesToUnique(session, attrDef);
 		long endTime = System.currentTimeMillis();
-		log.info("attribute {} was converted to unique in {} ms",attrDef.getName(),(endTime-startTime));
+		log.debug("attribute {} was converted to unique in {} ms",attrDef.getName(),(endTime-startTime));
```
zlamalp: Start with big A, ends with dot.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -2089,7 +2089,7 @@ public void deleteAttribute(PerunSession sess, AttributeDefinition attribute) th
 	public void deleteAllAttributeAuthz(PerunSession sess, AttributeDefinition attribute) throws InternalErrorException {
 		try {
 			if (0 < jdbc.update("DELETE FROM attributes_authz WHERE attr_id=?", attribute.getId())) {
-				log.info("All attribute_authz were deleted for Attribute={}.", attribute);
+				log.debug("All attribute_authz were deleted for Attribute={}.", attribute);
```
zlamalp: Since you log the whole object you can change ending to ``deleted for {}.``, since attributes ``toString()`` is like ``Attribute:[content]``.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -3394,7 +3394,7 @@ public void removeAllUserFacilityAttributesForAnyUser(PerunSession sess, Facilit
 	public void removeAllUserFacilityAttributes(PerunSession sess, User user) throws InternalErrorException {
 		try {
 			if (0 < jdbc.update("DELETE FROM user_facility_attr_values WHERE user_id=?", user.getId())) {
-				log.info("All attributes (theirs values) were removed from user on  all facilities. User={}", user);
+				log.debug("All attributes s values were removed from user {} on  all facilities.", user);
```
zlamalp: All attributes s values were -> All attributes values were

PR #2003, opened 2018-10-15
SCIM protocol support

licehammer:


perun-scim/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+
+    <parent>
+        <artifactId>perun</artifactId>
+        <groupId>cz.metacentrum</groupId>
+        <version>3.0.1-SNAPSHOT</version>
+    </parent>
+
+    <groupId>cz.metacentrum.perun</groupId>
+    <artifactId>perun-scim</artifactId>
+    <version>3.0.1-SNAPSHOT-${perun.build.type}</version>
```
zlamalp: Please remove version here.

perun-scim/pom.xml
```diff
@@ -0,0 +1,162 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>perun</artifactId>
+        <groupId>cz.metacentrum</groupId>
+        <version>3.0.1-SNAPSHOT</version>
```
zlamalp: Please change version to current master ``3.2.0-SNAPSHOT``

perun-scim/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-common</artifactId>
+            <version>2.23.2</version>
+        </dependency>
+    </dependencies>
+
+    <!-- PROFILES -->
+
+    <profiles>
```
zlamalp: We do not use profiles ``devel`` and ``production`` anymore. Unless you require oracle drivers for this module, please remove them. Otherwise see perun-core module for "oracle" profile.

licehammer:
Thanks for comments - there are fixed now.

perun-scim/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+            <version>2.0.1</version>
```
martin-kuba: Do not specify versions of dependencies in subprojects.  Use the <dependencyManagement>/<dependencies> in the parent pom.xml, and specify the version using a property, so that all version information is available at one place in the properties at the parent pom.xml.

perun-scim/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>2.7.6</version>
+            <type>jar</type>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.jersey.core</groupId>
+            <artifactId>jersey-common</artifactId>
+            <version>2.23.2</version>
```
martin-kuba: The same, do not specify version here, but in the parent pom.xml.

perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/GroupSCIM.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Group resource type for SCIM protocol.
+ *
+ * @author Sona Mastrakova <sona.mastrakova@gmail.com>
+ * @date 08.10.2016
+ */
+
+@JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
+public class GroupSCIM extends Resource {
```
balcirakpeter: If it is possible, it would be better to use composition instead of inheritance. It may cause problems (e.g. implementation of equals and hashCode methods).

perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/UserSCIM.java
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	public UserSCIM(Long id, Long externalId, Meta meta) {
+		super(id, externalId, meta);
+	}
+
+	public UserSCIM() {
+	}
+
+	public List<String> getSchemas() {
```
balcirakpeter: Getters which return collections should do that as unmodifiable list or immutable collection.

perun-scim/src/test/java/cz/metacentrum/perun/scim/AbstractSCIMTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import cz.metacentrum.perun.core.api.ExtSourcesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.PerunClient;
+import cz.metacentrum.perun.core.api.PerunPrincipal;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.exceptions.*;
```
balcirakpeter: We don't want to use wildcards in imports.

perun-scim/src/main/java/cz/metacentrum/perun/scim/api/endpoints/ResourceTypesEndpointController.java
```diff
@@ -0,0 +1,61 @@
...
+			result.addAll(getAllResourceTypes());
+
+			ObjectMapper mapper = new ObjectMapper();
+			return Response.ok(mapper.writeValueAsString(result)).build();
+		} catch (IOException ex) {
+			throw new SCIMException("Cannot convert resource types to json string", ex);
+		}
+	}
+
+	private List<ResourceTypeSCIM> getAllResourceTypes() {
```
balcirakpeter: This method can be static

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
```diff
@@ -698,6 +702,17 @@ private void serve(HttpServletRequest req, HttpServletResponse resp, boolean isG
...
+				// Process SCIM protocol
+				result = caller.getSCIMManager().process(caller.getSession(), method, des.readAll());
+				if (perunRequest != null) perunRequest.setResult(result);
+				if (!(result instanceof Response)) throw new InternalErrorException("SCIM manager returned unexpected result: " + result);
+				resp.setStatus(((Response) result).getStatus());
+				String response = (String) ((Response) result).getEntity();
+				PrintWriter printWriter = new PrintWriter(resp.getOutputStream());
+				printWriter.println(response);
+				printWriter.flush();
+				printWriter.close();
```
balcirakpeter: This is not safe way to close the PrintWriter. There should be used try-with-resources method or close() should be called in finally block.

perun-scim/src/main/java/cz/metacentrum/perun/scim/api/endpoints/UserResourceEndpointController.java
```diff
@@ -0,0 +1,110 @@
...
+
+		try {
+			preferredEmailAttribute = perunBl.getAttributesManagerBl().getAttribute(session, perunUser, AttributesManager.NS_USER_ATTR_DEF + ":preferredMail");
+			if (preferredEmailAttribute.getValue() != null) {
+				email.setValue(preferredEmailAttribute.getValue().toString());
+				email.setPrimary(true);
+				email.setType("preferred email");
+				return email;
+			}
+		} catch (InternalErrorException | WrongAttributeAssignmentException ex) {
```
balcirakpeter: This should be split into two separate catch blocks, so the error log will contains different message for different exception.

zlamalp:
Reworked in #2038. Some issues were not addressed, but I made task in redmine for them. It doesn't prevent merge itself. Closing this pull-request.

PR #2004, opened 2018-10-16
listOfExpiredMembers

zoraseb:
Member's Id added to list of expired members


zlamalp:
Failing test build has nothing to do with CLI change. Merging.

PR #2005, opened 2018-10-16
Added check for synchronizationEnabled module.

metodej:
Method checkAttributeValue now checks whether all needed attributes for group synchronization (synchronizationInterval, groupMemberQuery and groupExtSource) are filled in.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_synchronizationEnabled.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_synchronizationEnabled.java
@@ -4,6 +4,8 @@ import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_synchronizationEnabled.java
```diff
@@ -33,6 +39,16 @@ public void checkAttributeValue(PerunSessionImpl sess, Group group, Attribute at
 		if(!attrValue.equals("true") && !attrValue.equals("false")) {
 			throw new WrongAttributeValueException(attribute, group, "If attribute is not null, only string 'true' or 'false' is correct format.");
 		}
+		try {
+			if (attrValue.equals("true")
+			        && (sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROINTERVAL_ATTRNAME).getValue() == null
+			        || sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPMEMBERSQUERY_ATTRNAME).getValue() == null
+			        || sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPEXTSOURCE_ATTRNAME).getValue() == null)) {
+				throw new WrongReferenceAttributeValueException("At least one of required attributes " +  GROUPSYNCHROINTERVAL_ATTRNAME + ", " + GROUPMEMBERSQUERY_ATTRNAME + " or " + GROUPEXTSOURCE_ATTRNAME + " is not set.");
```
zlamalp: This exception can hold actual attributes, which is most convenient in GUI, where we state which attributes are wrong and what is their value. Can you please split this throwing into 3 with proper attributes passed to the exception. Thank you.

PR #2014, opened 2018-10-23
sendPasswordResetLinkEmail extended with choice of language and email address

DanoFecko:
* membersManager/sendPasswordResetLinkEmail
* both frontend and backend changed

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -2194,7 +2194,8 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	}
 
 	@Override
-	public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String namespace, String url) throws InternalErrorException {
+	public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String namespace, String url,
+										   String mailAddress, String language) throws InternalErrorException {
 
 		User user = perunBl.getUsersManagerBl().getUserByMember(sess, member);
 
@@ -2203,22 +2204,39 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 		for (Attribute a : logins) {
 			if (a.getFriendlyNameParameter().equals(namespace)) found = true;
 		}
-		if (!found) throw new InternalErrorException(user.toString()+" doesn't have login in namespace: "+namespace);
+		if (!found)
+			throw new InternalErrorException(user.toString() + " doesn't have login in namespace: " + namespace);
 
-		String email = "";
+		String subject;
 		try {
-			Attribute a = perunBl.getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_DEF+":preferredMail");
-			if (a != null && a.getValue() != null) {
-				email = (String)a.getValue();
+			Attribute subjectTemplateAttribute = perunBl.getAttributesManagerBl().getAttribute(sess, language,
+					AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":nonAuthzPwdResetMailSubject:" + namespace);
+			subject = (String) subjectTemplateAttribute.getValue();
+			if (subject == null) {
+				subjectTemplateAttribute = perunBl.getAttributesManagerBl().getAttribute(sess, "en",
+						AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":nonAuthzPwdResetMailSubject:" + namespace);
+				subject = (String) subjectTemplateAttribute.getValue();
 			}
-		} catch (WrongAttributeAssignmentException ex) {
+		} catch (AttributeNotExistsException | WrongAttributeAssignmentException ex) {
 			throw new InternalErrorException(ex);
-		} catch (AttributeNotExistsException ex) {
+		}
+
+		String message;
+		try {
+			Attribute messageTemplateAttribute = perunBl.getAttributesManagerBl().getAttribute(sess, language,
+					AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":nonAuthzPwdResetMailTemplate:" + namespace);
+			message = (String) messageTemplateAttribute.getValue();
+			if (message == null) {
+				messageTemplateAttribute = perunBl.getAttributesManagerBl().getAttribute(sess, "en",
+						AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":nonAuthzPwdResetMailTemplate:" + namespace);
+				message = (String) messageTemplateAttribute.getValue();
+			}
+		} catch (AttributeNotExistsException | WrongAttributeAssignmentException ex) {
 			throw new InternalErrorException(ex);
 		}
 
 		int id = getMembersManagerImpl().storePasswordResetRequest(sess, user, namespace);
-		Utils.sendPasswordResetEmail(user, email, namespace, url, id);
+		Utils.sendPasswordResetEmail(user, mailAddress, namespace, url, id, message, subject);
 
 	}
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_nonAuthzPwdResetMailTemplate_namespace.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_nonAuthzPwdResetMailTemplate_namespace.java
@@ -1,0 +1,32 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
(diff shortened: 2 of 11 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2215,8 +2215,24 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
 			throw new InternalErrorException(ex);
 		}
 
+		String subject;
+		try {
+			Attribute a = perunBl.getAttributesManager().getAttribute(sess, language, "urn_perun_entityless_attribute_def_def_nonAuthzPwdResetMailSubject_namespace");
```
Vojtech-Sassmann: `a` does not look like a good variable name.

Vojtech-Sassmann: Also, you are using attriubte urn wrongly here. Proper user:`"urn:perun:entityless:attribute-def:def:nonAuthzPwdResetMailSubject:" + namespace`

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2215,8 +2215,24 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
...
+			subject = a.getValue().toString();
+		} catch (PrivilegeException | AttributeNotExistsException | WrongAttributeAssignmentException ex) {
+			throw new InternalErrorException(ex);
+		}
+
+		String message;
+		try {
+			Attribute a = perunBl.getAttributesManager().getAttribute(sess, language, "urn_perun_entityless_attribute_def_def_nonAuthzPwdResetMailTemplate_namespace");
```
Vojtech-Sassmann: Same here.

DanoFecko:
@Vojtech-Sassmann I have made requested changes

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2201,22 +2201,41 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
...
+
+		String message;
+		try {
+			Attribute subjectTemplateAttribute = perunBl.getAttributesManager().getAttribute(sess, language,
+					"urn:perun:entityless:attribute-def:def:nonAuthzPwdResetMailTemplate:" + namespace);
+			message = subjectTemplateAttribute.getValue().toString();
+		} catch (PrivilegeException | AttributeNotExistsException | WrongAttributeAssignmentException ex) {
+			throw new InternalErrorException(ex);
+		}
+
```
zlamalp: Please add fallback for both message and subject. When filling with passed namespace (eg. "cs") fails (attribute or its value is null), then you retry this with generic "en" value for the namespace. So we have configurable fallback in attribute with "en" namespace. Right now following implementation in Utils fallback to hardcoded text only when passed value is null.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2201,22 +2201,41 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
...
 			if (a.getFriendlyNameParameter().equals(namespace)) found = true;
 		}
-		if (!found) throw new InternalErrorException(user.toString()+" doesn't have login in namespace: "+namespace);
+		if (!found)
+			throw new InternalErrorException(user.toString() + " doesn't have login in namespace: " + namespace);
 
 		String email = "";
 		try {
-			Attribute a = perunBl.getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_DEF+":preferredMail");
+			Attribute a = perunBl.getAttributesManagerBl().getAttribute(sess, user, mailAttributeUrn);
```
Vojtech-Sassmann: The attribute does not have to be associated only with User entity. E.g. if you have urn `urn:perun:member:attribute-def:def:mail` this will fail because of `WrongAttributeAssignmentException`.  You have to check if the urn is for Member or User and then call `getAttribute` with either member or user.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2201,22 +2201,41 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
...
+			subject = messageTemplateAttribute.getValue().toString();
+		} catch (PrivilegeException | AttributeNotExistsException | WrongAttributeAssignmentException ex) {
+			throw new InternalErrorException(ex);
+		}
+
+		String message;
+		try {
+			Attribute subjectTemplateAttribute = perunBl.getAttributesManager().getAttribute(sess, language,
+					"urn:perun:entityless:attribute-def:def:nonAuthzPwdResetMailTemplate:" + namespace);
+			message = subjectTemplateAttribute.getValue().toString();
```
zlamalp: Since attribute value can be null its not safe to call ``.toString()`` on it. You should use ``message = (String) subjectTemplateAttribute.getValue()`` when assigning. I know IDE then complains on unsafe cast, but its safe since we know which attribute we use here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_nonAuthzPwdResetMailTemplate_namespace.java
```diff
@@ -0,0 +1,42 @@
...
+		if (!value.contains("{link}")) {
+			throw new WrongAttributeValueException(attribute, key, "value doesn't contain contain tag {link}");
+		}
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition aD = new AttributeDefinition();
+		aD.setNamespace(AttributesManager.NS_ENTITYLESS_ATTR_DEF);
+		aD.setFriendlyName("nonAuthzPwdResetMailTemplate_namespace");
```
zlamalp: There is no ``_namespace`` in the friendlyName of attribute, since specific namespace is used on attribute creation. Attribute without namespace doesn't exist. So you can delete whole ``getAttributeDefinition()`` since in this contex it doesn't make sense (its similar to login-namespace attribute).

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/client/resources/Utils.java
```diff
@@ -478,6 +506,23 @@ public static String logoUrl() {
...
+	 * @return string representing native language
+	 */
+	public static String getNativeLanguageShortcut() {
+		String string = null;
+		if (PerunWebSession.getInstance().getConfiguration() != null) {
+			String value = PerunWebSession.getInstance().getConfiguration().getCustomProperty("nativeLanguage");
+			if (value != null && !value.isEmpty()) {
+				String[] parts = value.split(",");
+				string = parts[0];
+				}
```
zlamalp: indentation

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/memberstabs/SendPasswordResetRequestTabItem.java
```diff
@@ -83,12 +83,35 @@ public Widget draw() {
...
 		final ListBox namespaces = new ListBox();
 		layout.setWidget(0, 1, namespaces);
 
+		final ListBox languages = new ListBox();
+		layout.setWidget(1, 1, languages);
+
+		final ListBox emails = new ListBox();
+		layout.setWidget(2, 1, emails);
+
+		languages.addItem("en");
```
zlamalp: It would be nice for user to see full language names, you should be able to get it from native language property too. So the add it to the ListBox like ``box.add(key,value)`` if I remember correctly. So one value is displayed and other is used as backend value.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/memberstabs/SendPasswordResetRequestTabItem.java
```diff
@@ -134,7 +157,8 @@ public void onChange(ChangeEvent event) {
 			@Override
 			public void onClick(ClickEvent clickEvent) {
 				SendPasswordResetLinkEmail request = new SendPasswordResetLinkEmail(JsonCallbackEvents.closeTabDisableButtonEvents(changeButton, tab));
-				request.sendEmail(member, namespaces.getItemText(namespaces.getSelectedIndex()).toLowerCase());
+				request.sendEmail(member, namespaces.getItemText(namespaces.getSelectedIndex()).toLowerCase(),
+						urns.get(emails.getItemText(emails.getSelectedIndex())), languages.getItemText(languages.getSelectedIndex()));
```
zlamalp: Once you modify the box to have different displayed and backend value, you must use ``emails.getItemValue()`` or something like this instead of getItemText() which I believe returns only displayed value.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: If you want, you can then suppress the warning and write comment why this check is ok so when anyone else sees it, he knows its fine.

DanoFecko:
@Vojtech-Sassmann i have made requested changes

PR #2021, opened 2018-10-25
RPC: added check of max length of user titles

Vojtech-Sassmann:
* Problem: when calling method updateUserTitles, there is no check of
given values in titleAfter and titleBefore but the DB limit is 40. So,
when longer titles are given, the InternalErrorException is risen
without reasonable message.

* Solution: added check that in case of titleAfter/titleBefore being
longer than 40 characters rise an MaxSizeExceededException with proper
message.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -164,27 +164,27 @@ public static void notNull(Object e, String name) throws InternalErrorException
...
-	 * @param name name of entity
+	 * @param propertyName name of checked property
 	 * @param minLength
 	 * @throws MinSizeExceededException
 	 */
-	public static void checkMinLength(String name, int minLength) throws MinSizeExceededException{
+	public static void checkMinLength(String propertyName, String actualValue, int minLength) throws MinSizeExceededException{
 
-		if(name.length()<minLength) throw new MinSizeExceededException("Length of name is too short! MinLength=" + minLength + ", ActualLength=" + name.length());
+		if(actualValue.length() < minLength) throw new MinSizeExceededException("Length of '" + propertyName + "' is too short! MinLength=" + minLength + ", ActualLength=" + actualValue.length());
```
zlamalp: You should cover also the case, when passed value is ``null``, since ``Candidate`` params titleBefore/titleAfter can be null or empty.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -164,27 +164,27 @@ public static void notNull(Object e, String name) throws InternalErrorException
...
-	 * @param name name of entity
+	 * @param propertyName name of checked property
 	 * @param maxLength
 	 * @throws
 	 */
-	public static void checkMaxLength(String name, int maxLength) throws MaxSizeExceededException{
+	public static void checkMaxLength(String propertyName, String actualValue, int maxLength) throws MaxSizeExceededException{
 
-		if(name.length()<maxLength) throw new MaxSizeExceededException("Length of name is too long! MaxLength=" + maxLength + ", ActualLength=" + name.length());
+		if(actualValue.length() > maxLength) throw new MaxSizeExceededException("Length of '" + propertyName + "' is too long! MaxLength=" + maxLength + ", ActualLength=" + actualValue.length());
```
zlamalp: You should cover also the case, when passed value is ``null``, since ``Candidate`` params titleBefore/titleAfter can be null or empty.

Vojtech-Sassmann:
Hi @zlamalp,
I have made requested changes. Also, the method createUser is used only in createMember and in createSponsoredMember where the length of titles is already limited.

PR #2030, opened 2018-11-02
Moved propt text for map attrribute

dBucik:
* The original prompt has been set as a "value" attribute of input. Now the prompt is in the "placeholder" attribute.
* Modification allows to start filling the value without need to select all of the original "prompt" content and replacing it with key/value strings


zlamalp:
Just a note: we need to make sure, that user enters key of map attribute first. Solution is to block editing on value cell until key is entered.

PR #2045, opened 2018-11-12
Added new attribute module representing reason to suspend a member

metodej:
Also all attribute modules now listen to auditer messages.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -6178,7 +6178,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		//Load all attributes modules
 		ServiceLoader<AttributesModuleImplApi> attributeModulesLoader = ServiceLoader.load(AttributesModuleImplApi.class);
 		getAttributesManagerImpl().initAttributeModules(attributeModulesLoader);
-		getAttributesManagerImpl().registerVirtAttributeModules(attributeModulesLoader);
+		getAttributesManagerImpl().registerAttributeModules(attributeModulesLoader);
 
 		//Check if all core attributes exists, create if doesn't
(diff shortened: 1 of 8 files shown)


Vojtech-Sassmann:
Hi @metodej,
this solutions will not work as expected. The problem you are trying to solve here cannot be solved with only implementing one module. It is a little bit more complex. You need to either have one def attribute for storing value and other virtual for checking the auditer events. Another solution is to edit the Auditer so it can notify all of the attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_attribute_def_def_reasonToSuspend.java
```diff
@@ -0,0 +1,92 @@
...
+		Matcher statusChangeMatcher = statusChange.matcher(message);
+
+		if (statusChangeMatcher.find() && (statusChangeMatcher.group(2).equals("expired") || statusChangeMatcher.group(2).equals("disabled") || statusChangeMatcher.group(2).equals("validated"))) {
+			try {
+				int memberId = Integer.valueOf(statusChangeMatcher.group(1));
+				Member member = session.getPerunBl().getMembersManager().getMemberById(session, memberId);
+				Attribute attribute = session.getPerunBl().getAttributesManagerBl().getAttribute(session, member, AttributesManager.NS_MEMBER_ATTR_DEF + ":reasonToSuspend");
+				attribute.setValue(new HashMap<String, String>());
+				session.getPerunBl().getAttributesManagerBl().setAttribute(session, member, attribute);
+			} catch (Exception e) {
```
Vojtech-Sassmann: Dont catch all exceptions like that. First, catch AttributeNotExistsException if this exception is thrown, it means that member's status has been changed but he did not have 'reasonToSuspend' attribute set and that is not required. So, just catch that exception and put a comment inside of this catch block.

Then catch all other excpetions by their specific names and log them the way you already do.

PR #2048, opened 2018-11-14
Fixed deleting group when the group was admin of some entity

metodej:
Added new methods to find all facilities/groups/resources/security teams/vos where group is admin.
deleteAnyGroup method now unsets admin roles of group before deleting it.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -380,6 +380,52 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 			}
 		}
 
+		// remove admin roles of group
+		List<Facility> facilitiesWhereGroupIsAdmin = getGroupsManagerImpl().getFacilitiesWhereGroupIsAdmin(sess, group);
+		for (Facility facility : facilitiesWhereGroupIsAdmin) {
+			try {
+				perunBl.getFacilitiesManagerBl().removeAdmin(sess, facility, group);
+			} catch (GroupNotAdminException e) {
+				log.warn("Can't unset group {} as admin of facility {} due to group not admin exception {}.", group, facility, e);
+			}
+		}
+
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -380,6 +380,52 @@ public int compare(Group groupToCompare,Group groupToCompareWith) {
 			}
 		}
 
+		// remove admin roles of group
+		List<Facility> facilitiesWhereGroupIsAdmin = getGroupsManagerImpl().getFacilitiesWhereGroupIsAdmin(sess, group);
+		for (Facility facility : facilitiesWhereGroupIsAdmin) {
+			try {
+				perunBl.getFacilitiesManager().removeAdmin(sess, facility, group);
```
zlamalp: Please call ``perunBl..getFacilitiesMangerBl()`` to get rid of ``PrivilegeException``. Same for other removeAdmin() calls in this.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -380,6 +380,52 @@ public int compare(Group groupToCompare,Group groupToCompareWith) {
 			}
 		}
 
+		// remove admin roles of group
+		List<Facility> facilitiesWhereGroupIsAdmin = getGroupsManagerImpl().getFacilitiesWhereGroupIsAdmin(sess, group);
+		for (Facility facility : facilitiesWhereGroupIsAdmin) {
+			try {
+				perunBl.getFacilitiesManagerBl().removeAdmin(sess, facility, group);
+			} catch (GroupNotAdminException e) {
+				throw new ConsistencyErrorException(e);
```
zlamalp: Since we delete the group here, consistency in rights is not so relevant. It basically means, somebody removed relation during your delete operation. Please catch the exception and just log in as a WARN level instead of re-thorwing it as ConsistencyErrorException.

PR #2056, opened 2018-11-19
CORE: Fixed select for allowed users, members, resources

zlamalp:
- Select didn't respect new group member status, since it
  compared value in groups_members.status column, which is always 0.
  Member expiration in stored in direct/indirect membership relation
  in source_group_status, since for each group its calculated later.
  Now we limit selected members/users or resources (for user/facility)
  with respect to groups status too.
  This means, if user is in single group assigned to resource and is
  expired in such group, it is not considered as allowed anymore.

  WARNING: this have strong implication to many attribute modules,
  which will no longer process "not allowed" members. Like data quotas,
  shell, homeMountPoint, basicDefaultGID etc. Also processes like
  assignGroup will skip such members.


zlamalp:
We agreed, that we don't want to limit returned "allowed" members, since the term means, that member is globally not DISABLED and not INVALID. Since the logic is shared for getData() and resolving members to set/check required attributes, it would make problems. We will create new virtual attributes instead and each service can decide to respect them - they will reflect member status in a group and on resource (regarding all groups statuses).
Closing this for now.

PR #2088, opened 2018-12-03
Added new version of method resolveVirtualAttributeValueChange

metodej:
resolveVirtualAttributeValueChange now accepts and returns messages as AuditEvents instead of Strings
attribute urn_perun_user_attribute_def_virt_elixirBonaFideStatus is not used anymore and the class was deleted


Vojtech-Sassmann:
Hi @metodej,
I was thinking about the problem I am describing above and I think a better way how to do this is to implement the equals and hashcode methods only in the abstract class AuditEvent and change the type of the 'List<AuditerMessage> alreadyResolvedMessages' to 'List<AuditEvent> alreadyResolvedEvents'.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
```diff
@@ -73,58 +81,148 @@ public Attribute getAttributeValue(PerunSessionImpl sess, Member member, Resourc
...
+
+		return resolvingMessages;
+	}
+
+
+	private AuditerMessage resolveBanSetForResource(PerunSessionImpl perunSession, BanSetForResource banSetForResource) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		try {
+			Member member = perunSession.getPerunBl().getMembersManagerBl().getMemberById(perunSession, banSetForResource.getMemberId());
+			Resource resource = perunSession.getPerunBl().getResourcesManagerBl().getResourceById(perunSession, banSetForResource.getResourceId());
+			Attribute attribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, member, resource, AttributesManager.NS_MEMBER_RESOURCE_ATTR_VIRT + ":isBanned");
```
stavamichal: This attribute name should be set and use as constant. Then there will be much lower chance to typo.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -107,12 +108,46 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
-	public List<Pattern> getPatternsForMatch() {
-		List<Pattern> patterns = super.getPatternsForMatch();
-		patterns.add(userAllAttrsRemovedPattern);
-		patterns.add(userEPSAMARemovePattern);
-		patterns.add(userEPSAMASetPattern);
+	public List<AuditerMessage> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, AuditerMessage message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<AuditerMessage> resolvingMessages = super.resolveVirtualAttributeValueChange(perunSession, message);
+
+		if (message.getEvent() instanceof AllAttributesRemovedForUser) {
+			User user = ((AllAttributesRemovedForUser) message.getEvent()).getUser();
```
stavamichal: I would like to see this parsed the same way as in urn_perun_member_resource_attribute_def_virt_isBanned.java module to private methods. The same for all other modules.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
```diff
@@ -131,47 +131,47 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
-					messageAttributeSet = attribute.serializeToString() + " set for " + user.serializeToString() + ".";
-				}
-				resolvingMessages.add(messageAttributeSet);
-			} else {
-				log.error("Failed to get user from message: {}", message);
-			}
-		}
-		return resolvingMessages;
-	}
+//	@Override
```
stavamichal: Please, add the reason of this to the commit message if not yet there.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
```diff
@@ -73,58 +81,148 @@ public Attribute getAttributeValue(PerunSessionImpl sess, Member member, Resourc
...
+
+		} else if (message.getEvent() instanceof BanUpdatedForFacility) {
+			return resolveBanUpdatedForFacility(perunSession, (BanUpdatedForFacility) message.getEvent());
+		}
+
+		return resolvingMessages;
+	}
+
+
+	private AuditerMessage resolveBanSetForResource(PerunSessionImpl perunSession, BanSetForResource banSetForResource) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
```
Vojtech-Sassmann: Please, change the return type of these 3 methods to ```List<AuditerMessage>```. If no message needs to be generated, you can simply return empty List. Also, it is more consistent to have all of these 6 methods returning the same type.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/AttributesModuleAbstract.java
```diff
@@ -33,7 +34,7 @@ public AttributeDefinition getAttributeDefinition() {
 		return attr;
 	}
 
-	public List<String> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, String message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
+	public List<AuditerMessage> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, AuditerMessage message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
```
Vojtech-Sassmann: I whink it would be better if this method would work with AuditEvents instead of AuditerMessage. The information about session is redundant.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_attribute_def_def_suspensionInfo.java
```diff
@@ -30,10 +31,9 @@
  */
 public class urn_perun_member_attribute_def_def_suspensionInfo extends MemberAttributesModuleAbstract {
 
+	private final static String SUSPENSION_INFO = AttributesManager.NS_MEMBER_ATTR_DEF + ":suspensionInfo";
```
Vojtech-Sassmann: For attribute urns we usually use names like "A_M_D_suspensionInfo"

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
```diff
@@ -36,13 +44,9 @@
  */
 public class urn_perun_member_resource_attribute_def_virt_isBanned extends MemberResourceVirtualAttributesModuleAbstract implements MemberResourceVirtualAttributesModuleImplApi {
 
+	private final static String IS_BANNED = AttributesManager.NS_MEMBER_RESOURCE_ATTR_VIRT + ":isBanned";
```
Vojtech-Sassmann: A_MR_D_isBanned

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_elixirBonaFideStatus.java
Vojtech-Sassmann: @metodej I think it is save to just completely delete this class.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -107,12 +107,61 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
 	}
 
 	@Override
-	public List<Pattern> getPatternsForMatch() {
-		List<Pattern> patterns = super.getPatternsForMatch();
-		patterns.add(userAllAttrsRemovedPattern);
-		patterns.add(userEPSAMARemovePattern);
-		patterns.add(userEPSAMASetPattern);
+	public List<AuditEvent> resolveVirtualAttributeValueChange(PerunSessionImpl perunSession, AuditEvent message) throws InternalErrorException, WrongReferenceAttributeValueException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<AuditEvent> resolvingMessages = super.resolveVirtualAttributeValueChange(perunSession, message);
```
Vojtech-Sassmann: If you implemet all of the logic for checking events, you do not need to call super here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -107,12 +107,61 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+			log.debug("These new messages will be generated: " + resolvingMessages);
+		}
+		return resolvingMessages;
+	}
+
+	private List<AuditEvent> resolveAllAttributesRemovedForUser(PerunSessionImpl perunSession, AllAttributesRemovedForUser allAttributesRemovedForUser) throws WrongAttributeAssignmentException, InternalErrorException, AttributeNotExistsException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
+
+		User user = allAttributesRemovedForUser.getUser();
+		Attribute attribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, getDestinationAttributeFriendlyName());
```
Vojtech-Sassmann: I think this is not implemented correctly. This module should check if one of the three events that occured has attribute with friendly name `eduPersonScopedAffiliations` (the event for all attributes removed should always be processed).
If so, then it should try to get the current value of this virtual attribute. If the value is null, then you should return event AttributeRemovedForUser. If the value is non-null, you should return AttributeSetForUser.

Also, `getDestinationAttributeFriendlyName` doesn't return the whole urn, only the friendly name. All of those calls would fail. You need to pass the whole `urn:perun:user:......:eduPersonScopedAffiliations.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -60,36 +58,35 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
 	@Override
-	public List<String> resolveVirtualAttributeValueChange(PerunSessionImpl sess, String message) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
-		List<String> resolvingMessages = new ArrayList<>();
+	public List<AuditEvent> resolveVirtualAttributeValueChange(PerunSessionImpl sess, AuditEvent message) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
 		if (message == null) return resolvingMessages;
 
-		Matcher memberAddedToMatcher = memberAddedToPattern.matcher(message);
-		Matcher memberTotallyRemovedFromMatcher = memberTotallyRemovedFromPattern.matcher(message);
+		if (message instanceof DirectMemberAddedToGroup) {
```
Vojtech-Sassmann: You should also add here case for event IndirectMemberAddedToGroup.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -60,36 +58,35 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
 
-		User user;
-		Attribute attribute;
+	private List<AuditEvent> resolveDirectMemberAddedToGroup (PerunSessionImpl sess, DirectMemberAddedToGroup directMemberAddedToGroup) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
 
-		if (memberAddedToMatcher.find() || memberTotallyRemovedFromMatcher.find()) {
+		User user = sess.getPerunBl().getUsersManagerBl().getUserByMember(sess, directMemberAddedToGroup.getMember());
+		Attribute attribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, WHOLE_NAME);
+		resolvingMessages.add(new AttributeSetForUser(attribute, user));
```
Vojtech-Sassmann: Again the same problem. When you receive the attribute, you should check its value. If the value is null or empty, you should return event AttributeRemovedForMember. If the value is non-null and not empty, you should return AttributeSetForUser.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -60,36 +58,35 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+	}
 
-				attribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, AttributesManager.NS_USER_ATTR_VIRT + ":" + FRIENDLY_NAME);
-				List<String> value = attribute.valueAsList();
+	private List<AuditEvent> resolveMemberRemovedFromGroupTotally(PerunSessionImpl sess, MemberRemovedFromGroupTotally memberRemovedFromGroupTotally) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
+
+		User user = sess.getPerunBl().getUsersManagerBl().getUserByMember(sess, memberRemovedFromGroupTotally.getMember());
+		Attribute attribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, user, WHOLE_NAME);
+		resolvingMessages.add(new AttributeRemovedForUser(attribute, user));
```
Vojtech-Sassmann: Same here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_institutionsCountries.java
```diff
@@ -107,44 +108,69 @@ public String modifyValue(PerunSession session, DnsMapCtx ctx, UserExtSource ues
...
-				Attribute thisAttribute = am.getAttribute(sess, user, getDestinationAttributeName());
-				messages.add(thisAttribute.serializeToString() + " set for " + user.serializeToString() + ".");
+		return messages;
+	}
+
+	private List<AuditEvent> resolveAttributeSetForKey(PerunSessionImpl sess, AttributeSetForKey attributeSetForKey) throws WrongAttributeAssignmentException, InternalErrorException, AttributeNotExistsException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
+
+		String key = attributeSetForKey.getKey();
+		List<String> longerDomains =  new ArrayList<>();
```
Vojtech-Sassmann: You can extract the code from this method into another private method since there is exactly the same code in the method resolveAttributeRemovedForKey.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_institutionsCountries.java
```diff
@@ -107,44 +108,69 @@ public String modifyValue(PerunSession session, DnsMapCtx ctx, UserExtSource ues
...
+			if (dnsDomain.endsWith(key) && dnsDomain.length() > key.length()) {
+				longerDomains.add(dnsDomain);
+			}
+		}
+		log.debug("DNS domain '{}' changed, found longer domains: {}", key, longerDomains);
+		//find users that are affected by the change - have schacHomeOrganization value ending in key but not ending with longerDomains
+		List<User> affectedUsers = sess.getPerunBl().getUsersManagerBl().findUsersWithExtSourceAttributeValueEnding(sess,getSourceAttributeName(), key, longerDomains);
+		for (User user : affectedUsers) {
+			Attribute attribute = am.getAttribute(sess, user, getDestinationAttributeName());
+			resolvingMessages.add(new AttributeRemovedForUser(attribute, user));
```
Vojtech-Sassmann: Here should be AttributeSetForUse.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_kerberosLogins.java
```diff
@@ -64,27 +65,31 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+			resolvingMessages.addAll(resolveEvent(perunSession, ((UserExtSourceRemovedFromUser) message).getUser()));
 		}
+
+		return resolvingMessages;
+	}
+
+	private List<AuditEvent> resolveEvent(PerunSessionImpl perunSession, User user) throws WrongAttributeAssignmentException, InternalErrorException, AttributeNotExistsException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
+
+		Attribute attrVirtKerberosLogins = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, A_U_V_KERBEROS_LOGINS);
```
Vojtech-Sassmann: Please, add here this comment: `//FIXME: if the attribute value is null or empty, this method should return AttributeRemovedForUser instead`

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertDNs.java
```diff
@@ -61,26 +59,38 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+		return resolvingMessages;
+	}
+
+	private List<AuditEvent> resolveUserExtSourceAddedToUser(PerunSessionImpl perunSession, UserExtSourceAddedToUser userExtSourceAddedToUser) throws WrongAttributeAssignmentException, InternalErrorException, AttributeNotExistsException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
+
+		if (userExtSourceAddedToUser.getUserExtSource().getExtSource() instanceof ExtSourceX509) {
+			User user = userExtSourceAddedToUser.getUser();
+			Attribute attrVirtUserCertDNs = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, A_U_V_USER_CERT_DNS);
+			resolvingMessages.add(new AttributeSetForUser(attrVirtUserCertDNs, user));
```
Vojtech-Sassmann: Please, add here the same comment. `//FIXME: if the attribute value is null or empty, this method should return AttributeRemovedForUser instead`

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertDNs.java
```diff
@@ -61,26 +59,38 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+
+		return resolvingMessages;
+	}
+
+	private List<AuditEvent> resolveUserExtSourceRemovedFromUser(PerunSessionImpl perunSession, UserExtSourceRemovedFromUser userExtSourceRemovedFromUser) throws WrongAttributeAssignmentException, InternalErrorException, AttributeNotExistsException {
+		List<AuditEvent> resolvingMessages = new ArrayList<>();
+
+		if (userExtSourceRemovedFromUser.getUserExtSource().getExtSource() instanceof ExtSourceX509) {
+			User user = userExtSourceRemovedFromUser.getUser();
+			Attribute attrVirtUserCertDNs = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, A_U_V_USER_CERT_DNS);
```
Vojtech-Sassmann: Same here: `//FIXME: if the attribute value is null or empty, this method should return AttributeRemovedForUser instead`

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -157,39 +160,66 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
 	}
 
 	/**
-	 * Get list of message patterns used for resolving attribute value change.
-	 * @return List of Patterns
+	 * Functional interface controlling AuditEvents
```
Vojtech-Sassmann: This JavaDoc should be more verbose. You should state here, what it should do and what it is used for. Also, you should add javadoc to the `check` method. You should state there, that the method is used to determine whether some auditevent should be handled. And that is represented by returning user id. If it should not be handled, null is returned.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -157,39 +160,66 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
-		patterns.add(removeUserExtSourceAttribute);
+	@FunctionalInterface
+	public interface AttributeHandleIdenfier {
+		Integer check(AuditEvent ae);
+	}
+
 
-		return patterns;
+	/**
+	 * Returns userId if AuditEvent is correct and null if not.
```
Vojtech-Sassmann: You don't need to use javadoc in method implementations, only in theri declarations -> in the functional interface.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -157,39 +160,66 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+	 * Functional interface controlling AuditEvents
 	 */
-	public List<Pattern> getPatternsForMatch() {
-		List<Pattern> patterns = new ArrayList<>();
-		patterns.add(allAttributesRemovedForUserExtSource);
-		patterns.add(setUserExtSourceAttribute);
-		patterns.add(removeUserExtSourceAttribute);
+	@FunctionalInterface
+	public interface AttributeHandleIdenfier {
+		Integer check(AuditEvent ae);
```
Vojtech-Sassmann: I would change the name of this method to `shouldBeEventHandled`

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -157,39 +160,66 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
-			if (user != null) {
-				Attribute attribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, getDestinationAttributeName());
-				@SuppressWarnings("unchecked") List<String> attributeValue = (ArrayList<String>) attribute.getValue();
-				String messageAttributeSet;
-				if (attributeValue == null || attributeValue.isEmpty()) {
-					AttributeDefinition attributeDefinition = new AttributeDefinition(attribute);
-					messageAttributeSet = attributeDefinition.serializeToString() + " removed for " + user.serializeToString() + ".";
-				} else {
-					messageAttributeSet = attribute.serializeToString() + " set for " + user.serializeToString() + ".";
+		List<AttributeHandleIdenfier> list = getChecks();
```
Vojtech-Sassmann: `getChecks` -> `getHandleIdentifiers` or something like that.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserVirtualAttributeCollectedFromUserExtSource.java
```diff
@@ -157,39 +160,66 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+		for (AttributeHandleIdenfier attributeHandleIdenfier : list) {
+			if (attributeHandleIdenfier.check(message) != null) {
```
Vojtech-Sassmann:
```suggestion
			Long userId = attributeHandleIdenfier.check(message);
			if (userId != null) {
```
This reduces the need of calling the check method twice.

PR #2094, opened 2018-12-11
Delete log from PerunException and PerunRuntimeException

kubazak:
Problem:
Some methods are use Exception as return value and every Exception extends from PerunException where is in constructor logs error message.
We don't want to log Exception that is processed in the code.

Fix:
I removed log from PerunException and PerunRuntimeException and add log exception to Api.java. Here is catched every exception that going out.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
@@ -738,15 +738,18 @@ public class Api extends HttpServlet {
 			if (!isJsonp) {
 				resp.setStatus(400);
 			}
+			log.warn("Perun exception {}: {}.", pex.getErrorId(), pex);
 			ser.writePerunException(pex);
 		} catch (PerunRuntimeException prex) {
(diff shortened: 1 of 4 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
```diff
@@ -723,6 +723,7 @@ private void serve(HttpServletRequest req, HttpServletResponse resp, boolean isG
 					result = caller.call(manager, method, des);
 					if (perunRequest != null) perunRequest.setResult(result);
 				} catch (Exception ex) {
+					log.warn("Exception {}: {}.", Long.toHexString(System.currentTimeMillis()), this);
```
martin-kuba: This line is not correct, the id logged is different than the id of the exception, because the id logged is the hexadecimal number of milliseconds in the time of catching the exception, while the id of the exception is the time when the exception was created.

kubazak: Done.

PR #2095, opened 2018-12-11
There is an error in the process of extending grace period

stavamichal:
- when member is in the grace period, he will not get the date for it
 correctly!
 - the year for extensionCalendar there has to be the current year, not
 the year of possible expiration


stavamichal:
There is better way how to process this problem in #2096

PR #2102, opened 2019-01-15
Added force flag to deleteFacility

metodej:
Method deleteFacility has new parameter boolean "force". If "force" is true, method deletes all constrains that would prevent the facility being deleted.


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -303,10 +310,21 @@ public Facility createFacility(PerunSession sess, Facility facility) throws Inte
...
-		if (getFacilitiesManagerImpl().getAssignedResources(sess, facility).size() > 0) {
-			throw new RelationExistsException("Facility is still used as a resource");
+		if (force) {
+			List<Resource> resources = this.getAssignedResources(sess, facility);
+			for (Resource resource : resources) {
+				getPerunBl().getResourcesManagerBl().deleteResource(sess, resource);
+			}
+			List<Task> tasks = taskManager.listAllTasksForFacility(facility.getId());
+			for (Task task : tasks) {
+				taskManager.removeTask(task.getId());
```
zlamalp: Ok, so we need to delete TaskResults first, because ``removeTask()`` doesn't do it by itself. You can call ``clearByTask(task.getId())`` before actual Task deletion. Thank you.

PR #2105, opened 2019-01-21
Change logic of working with values in o365 email addresses module

stavamichal:
- fill method will not generate any default value from now
 - check method will accept empty value as valid one
 - check method will not expected that there must be always
 value UCO@muni.cz, it is not mandatory value any more
 - check method will still prevent duplicities in the value and between
 other attributes of members and groups (the same attribute exists also
 for group in group_resource namespace)


stavamichal:
This PR was replaced by PR #2197.

PR #2109, opened 2019-01-22
new version of perun-ldapc

mvocu:
Rewrite of perun-ldapc (and perun-ldapc-initializer):
  - broke up the code into separate classes and hierarchies
  - for every major perun ldap objectclass there is a Java model class responsible for attribute mapping
  - ldap attributes are described by PerunAttributeDesc objects, which can be instantiated and injected using Spring IoC
  - auditer events are processed by separate processors registered with dispatcher
  - dispatch conditions for each type of processor are described by DispatchCondition objects also managed by Spring IoC container
  - during startup, perun-ldapc may be asked to re-initialize the ldap directory, doing the work of perun-ldapc-initializer (the re-initialization can be done any time, eg. when conf files are modified - not implemented yet)
  - during the re-initialization existing entries are checked and modified only when required


stavamichal:
Please, do a whole new module, something like perun-ldapc-new. We need to run both LDAP connectors together for some time. Thank you.

mvocu:
Done.

zlamalp:
I made necessary fixes during merge. There is no reason to postpone it. We can continue on master now.

PR #2114, opened 2019-01-29
Option to add reason for member suspension

DanoFecko:
* During suspending member (when "SUSPENDED" is selected in popup), you can add reason for suspension.
* Required member attributes suspensinInfo is created during startup.
* Extended RPC API to handle suspension info message.


DanoFecko:
@zlamalp I mede ale the changes, beside the thing (passed user ID) in last comment, that i could not find.

zlamalp:
I didn't find issue with wrongly passed user ID either. Probably some re-compilation error on GWT side. Thanks for the work.

PR #2118, opened 2019-01-30
Support for post-processing paging for method returning list of objects

stavamichal:
- new utility method for rpc getSublist, which is able to return
 subList of existing list as the view defined by it's lower index
 (fromIndex) and it's higher index (toIndex), both included
 - use this new method for processing of existing method
 getAllRichGroupsWithAttributesByNames to be able to get just defined
 part of the output

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
@@ -475,4 +475,41 @@ public class ApiCaller {
 	public Object call(String managerName, String methodName, Deserializer parms) throws PerunException {
 		return PerunManager.call(managerName, methodName, this, parms);
 	}
+
+	/**
+	 * Returns a view of the portion of this list between the specified fromIndex, inclusive, and toIndex, inclusive.
+	 *
+	 * If list of objects is null or empty, it will return empty list as sublist.
+	 * If fromIndex is lower than 0, it will be set to 0.
+	 * If toIndex is bigger than size of an input array, it will be set to the size of an array.
+	 * If fromIndex and toIndex are same, it will return array with exactly 1 object on this position.
+	 * If fromIndex is bigger than toIndex, it will return empty array.
+	 *
+	 * @param listOfObjects original list of objects from which we will get the sublist view by indexes
+	 * @param fromIndex index of the object from original list, which will be the first object in the sublist view (included)
+	 * @param toIndex index of the object from original list, which will be the last object in the sublist view (included)
+	 *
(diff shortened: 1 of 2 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
```diff
@@ -475,4 +475,40 @@ public void stateChangingCheck() throws RpcException {
...
 		return PerunManager.call(managerName, methodName, this, parms);
 	}
+
+	/**
+	 * Returns a view of the portion of this list between the specified fromIndex, inclusive, and toIndex, inclusive.
+	 *
+	 * If list of objects is null or empty, it will return empty list as sublist.
+	 * If fromIndex is lower than 0, it will be set to 0.
+	 * If toIndex is bigger than size of an input array, it will be set to the size of an array.
+	 * If fromIndex and toIndex are same or fromIndex is bigger than toIndex, it will return empty array.
```
zlamalp: First part of the sentence is not true. If I pass eg. 5/5 and sourcing list contains 6+ groups, it will return group at the index 5 (6th group).

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
```diff
@@ -475,4 +475,40 @@ public void stateChangingCheck() throws RpcException {
...
+	 *
+	 * @param listOfObjects original list of objects from which we will get the sublist view by indexes
+	 * @param fromIndex index of the object from original list, which will be the first object in the sublist view (included)
+	 * @param toIndex index of the object from original list, which will be the last object in the sublist view (included)
+	 *
+	 * @return a view of the portion of the listOfObjects between fromIndex and toIndex (both inclusive), for example
+	 * if fromIndex=0 and toIndex=10 it will return view on first 11 objects of the input list if they are exist, if
```
zlamalp: ``if they are exist,`` -> ``if they exist,``

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -930,7 +930,41 @@ public Vo call(ApiCaller ac, Deserializer parms) throws PerunException {
 	},
 
 	/*#
-	 * Returns all RichGroups containing selected attributes
+	 * Returns list of all RichGroups containing selected attributes as sublist of original list starting at fromIndex
```
zlamalp: I would use something like this: 
``Returns sub-list of all RichGroups, each containing selected attributes, starting at fromIndex....``

PR #2129, opened 2019-02-13
DISPATCHER: Ignore some not-relevant Auditer messages

zlamalp:
- Since we don't have pub-sub mechanism for auditer messages right
  now, we will get all messages.
  This creates overhead, since some messages are not relevant for
  service propagation (like "member will expire" or "group sync
  finished"). These actions either have no impact on data or other
  related changes make own audit message.
- This change will simply skip resolving related service/facility
  combination for objects from these non-relevant messages.


zlamalp:
Logic replaced by #2132. Closing this pull-request.

PR #2134, opened 2019-02-19
Refactorization of isMemberInGracePeriod

metodej:


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2155,28 +2155,23 @@ private boolean isMemberInGracePeriod(Map<String, String> membershipExpirationRu
 
 		int amount = Integer.valueOf(m.group(1));
 
-		int field;
+		TemporalUnit field;
```
Vojtech-Sassmann: Maybe I would use a different name for this variable. Maybe `gracePeriodTimeUnit` ?

metodej: Okay, that is more accurate, thanks.

PR #2136, opened 2019-02-20
Refactorization of getActiveUserExtSources

metodej:

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
@@ -17,6 +17,7 @@ import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 import javax.sql.DataSource;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.time.LocalDate;
 import java.util.*;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -552,17 +553,14 @@ public UserExtSource getUserExtSourceByExtLogin(PerunSession sess, ExtSource sou
...
-		String compareDate = BeansUtils.getDateFormatterWithoutTime().format(date.getTime());
+		LocalDate date = LocalDate.now();
+		date = date.minusMonths(MAX_OLD_OF_ACTIVE_USER_EXTSOURCE);
 
 		try {
 			String query = "select " + userExtSourceMappingSelectQuery + ", " + ExtSourcesManagerImpl.extSourceMappingSelectQuery +
 					" from user_ext_sources left join ext_sources on user_ext_sources.ext_sources_id=ext_sources.id where " +
 					" user_ext_sources.user_id=? and " +
-					" user_ext_sources.last_access > " + Compatibility.toDate("'" + compareDate + "'", "'YYYY-MM-DD'");
+					" user_ext_sources.last_access > " + Compatibility.toDate("'" + date.toString() + "'", "'YYYY-MM-DD'");
```
Vojtech-Sassmann: I think you do not need to explicitly call the `toString` method.

Vojtech-Sassmann: Also, the previous version produced String containing time only. Are you sure this cast to String will produce the same result?

metodej: Yes, that `toString` was redundant, it is fixed now. Concerning the `compareDate`, I am pretty sure it is the same String as `date.toString()` returns, that is just date, no time.

PR #2141, opened 2019-02-26
Refactorization of getMembersByExpiration and getMembersByGroupExpiration

metodej:
Time in these methods was updated to use Java 8 api. Also other classes using this method needed to be updated.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
@@ -35,9 +35,9 @@ import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.JdbcPerunTemplate;
 
 import javax.sql.DataSource;
+import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Calendar;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -435,25 +433,21 @@ private void auditOldGroupExpirations(List<Status> allowedStatuses, Group group)
 	 * @throws InternalErrorException internal error
 	 */
 	private void auditIncomingExpirations(List<Status> allowedStatuses, Map<Integer, Vo> vosMap) throws InternalErrorException {
-		Calendar monthBefore = Calendar.getInstance();
-		monthBefore.add(Calendar.MONTH, 1);
+		LocalDate monthBefore = LocalDate.now().plusMonths(1);
```
Vojtech-Sassmann: I would change the variable name. `monthBefore` is missleading. It should be something like `nextMonth`.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -465,25 +459,21 @@ private void auditIncomingExpirations(List<Status> allowedStatuses, Map<Integer,
 	 * @throws InternalErrorException internal error
 	 */
 	private void auditIncomingGroupExpirations(List<Status> allowedStatuses, Group group) throws InternalErrorException {
-		Calendar monthBefore = Calendar.getInstance();
-		monthBefore.add(Calendar.MONTH, 1);
+		LocalDate monthBefore = LocalDate.now().plusMonths(1);
```
Vojtech-Sassmann: Same here

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
metodej: Ok, it is fixed now, thank you.

PR #2154, opened 2019-03-05
Refactorization of methods beforeApprove, canBeApproved and canBeSubm…

metodej:
…itted

Time in these methods was updated to use Java 8 api.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Metacentrum.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Metacentrum.java
@@ -16,7 +16,8 @@ import org.slf4j.LoggerFactory;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.Calendar;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Metacentrum.java
```diff
@@ -142,17 +143,10 @@ public void canBeApproved(PerunSession session, Application app) throws PerunExc
...
 
-				Calendar c = Calendar.getInstance();
-				c.setTime(eligibleDate);
-				c.add(Calendar.YEAR, 1);
-				Date eligibleDatePlusYear = c.getTime();
-
-				// get now
-				Calendar cal = Calendar.getInstance();
-				Date now = cal.getTime();
+				LocalDateTime timeInOneYear = LocalDateTime.ofInstant(eligibleDate.toInstant(), ZoneId.systemDefault()).plusYears(1);
```
zlamalp: Can you please verify, that this logic doesn't add the value in hours ? Since sourcing value is a date/time without the zone, but was generated in our zone (+1/+2 in summer). So if you again apply zone specific offset to the value, then it can add an hour or two, which doesn't correspond with the logic, that we hold user eligible 1 year after the timestamp.

metodej: Yes, from what I tried, it does not change hours during the conversions. I think that since the SimpleDateFormat uses the same default time zone to create Date as LocalDateTime does, there should not be any problems.

zlamalp: Ok

PR #2169, opened 2019-03-14
CORE: Compare also large attributes before updating in DB

zlamalp:
- Do not always overwrite (and mark as changed) large
  attributes, which were not really changed.
- We previously checked value for normal attributes,
  now we check previous value for LargeString, LargeArray
  and LinkedHashMap attributes too.
- This prevents unnecessary attribute checks and
  audit messages processing.


stavamichal:
I would prefer to test it manually before merging.

zlamalp:
> I would prefer to test it manually before merging.

I did tests on postgres and oracle. Works as expected. No audit message about stored map attribute, if value was not changed.

PR #2174, opened 2019-03-19
port changes from LDAPc to LDAPc-ada

mvocu:
Add expiration of group members to LDAPc-ada

### DIFF
--- a/perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/beans/GroupSynchronizer.java
+++ b/perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/beans/GroupSynchronizer.java
@@ -13,13 +13,14 @@ import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.Perun;
 import cz.metacentrum.perun.core.api.Resource;
+import cz.metacentrum.perun.core.api.Status;
 import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.ldapc.model.PerunGroup;
 import cz.metacentrum.perun.ldapc.service.LdapcManager;
-import cz.metacentrum.perun.rpclib.Rpc;
+import cz.metacentrum.perun.core.bl.PerunBl;
 
 @Component
 public class GroupSynchronizer extends AbstractSynchronizer {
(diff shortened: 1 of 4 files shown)


perun-ldapc-ada/src/main/resources/perun-ldapc.xml
```diff
@@ -272,6 +272,36 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
...
+				<bean class="cz.metacentrum.perun.ldapc.processor.impl.RegexpDispatchEventCondition">
+					<property name="beansCondition">
+						<list>
+							<value>cz.metacentrum.perun.core.api.Group</value>
+							<value>cz.metacentrum.perun.core.api.Member</value>
+						</list>
+					</property>
+					<property name="pattern" value="expired.$|disabled.$|invalidated.$|suspended #" />
+					<property name="handlerMethodName" value="processMemberRemoved" />
+				</bean>
```
stavamichal: These regulars were used in the original configuration, but they are new in this pull request. How did you manage removing members from VO without them before this change?

perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/GroupEventProcessor.java
```diff
@@ -36,6 +36,18 @@ public void processMemberAdded(String msg, MessageBeans beans) {
...
+			return;
+		}
+		try {
+			log.debug("Removing member {} from group {}", beans.getMember(), beans.getGroup());
+			perunGroup.removeMemberFromGroup(beans.getMember(), beans.getGroup());
+		} catch (NamingException | InternalErrorException e) {
+			log.error("Error removing member {} from group {}: {}", beans.getMember().getId(), beans.getGroup().getId(), e.getMessage());;
+		}
+	}
+
```
stavamichal: This is the first time you need method to remove member from group? So in the basic implementation you are not removing members from group or there is different logic in it? What about "remove Member from Group totally" message?

perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/beans/GroupSynchronizer.java
```diff
@@ -58,7 +58,7 @@ public void synchronizeGroups() {
 
 							log.debug("Getting list of members for group {}", group.getId());
 							// List<Member> members = ldapcManager.getRpcCaller().call("groupsManager",  "getGroupMembers", params).readList(Member.class);
-							List<Member> members = perun.getGroupsManager().getGroupMembers(ldapcManager.getPerunSession(), group);
+							List<Member> members = perun.getGroupsManager().getActiveGroupMembers(ldapcManager.getPerunSession(), group);
```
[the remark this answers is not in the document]
mvocu: Not sure, you are the core API expert here. However, this interface does not support three parameter version of getActiveGroupMembers(), so I have to change this to *Bl version.

perun-ldapc-ada/src/main/resources/perun-ldapc.xml
mvocu: It is a bug in the older version.

perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/GroupEventProcessor.java
mvocu: Yes to the first two, "totally" was added too.

PR #2176, opened 2019-03-19
New method getSponsoredMembers with specific attributes

stavamichal:
- we want to return list of RichMembers (as sponsored members) for
 specific sponsor in VO, but not just members but also some of their
 attributes defined by array list of their names
 - method was added to RPC

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1251,6 +1251,31 @@ public class MembersManagerEntry implements MembersManager {
 	}
 
 	@Override
+	public List<RichMember> getSponsoredMembers(PerunSession sess, Vo vo, User user, List<String> attrNames) throws InternalErrorException, AttributeNotExistsException, PrivilegeException, VoNotExistsException, UserNotExistsException {
+		Utils.checkPerunSession(sess);
+		perunBl.getVosManagerBl().checkVoExists(sess, vo);
+		perunBl.getUsersManagerBl().checkUserExists(sess, user);
+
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1204,6 +1204,31 @@ public RichMember sponsorMember(PerunSession session, Member sponsored, User spo
...
+		for (String attrName : attrNames) {
+			attributeDefinitions.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		//Basic rich Members without attributes
+		List<RichMember> richMembers = membersManagerBl.convertMembersToRichMembers(sess, membersManagerBl.getSponsoredMembers(sess, vo, user));
+		//Enriched rich members with attributes by list of attributes
+		richMembers = membersManagerBl.convertMembersToRichMembersWithAttributes(sess, richMembers, attributeDefinitions);
+		//RichMembers with filtered attributes by rights from session
+		richMembers = membersManagerBl.filterOnlyAllowedAttributes(sess, richMembers);
```
zlamalp: I would prefer using ``filterOnlyAllowedAttributes(PerunSession sess, List<RichMember> richMembers, Group group, boolean useContext)`` with group = NULL and useContext = TRUE.

Its optimized to check rights for first member and store results based on attribute name. If members are from the same VO, you can apply same rights to their member/user attributes. If they would be from different VO, exception is thrown.
Group is optional and not useful for this use-case.
It saves up "init" part of attribute rights check method, where it checks if attribute definition exists etc. which adds overhead for each member and his attribute returned.

stavamichal: Fixed. Thank you.

PR #2177, opened 2019-03-19
The great refactorization

Vojtech-Sassmann:
* Java version migration
** Removed unnecessary explicit types
** Removed unnecessary field inicialization
** Try-final replaced with try-with-resources
** Annonymous methods replaced with lambdas

* Multiple else-if replaced with switch

* Access static memver via instance reference
* Added final declarations
* Duplicate throws
* Removed redundant throws declarations
* Stream API call chain simplified
* C-style array declarations replaced with Java-style
* String 'UTF-8' replaced by Standard UTF-8 Charset object
* Unnecessary conversion to String removed
* Removed unnecessary interface modifiers
* Removed unnecessary semicolons
* Removed duplicate condition in 'if' statements
* Loop statement that doesnt loop replaced with 'if'
* Condition statements simplified
* Removed redundant local variables
* Simplified JUnit assertions


Vojtech-Sassmann:
Is there anybody brave enough to merge this PR?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -144,54 +143,46 @@ public AuditMessage mapRow(ResultSet rs, int i) throws SQLException {
 		}
 	};
 
-	protected static final RowMapper<AuditMessage> AUDITMESSAGE_MAPPER_FOR_PARSER = new RowMapper<AuditMessage>() {
-		public AuditMessage mapRow(ResultSet rs, int i) throws SQLException {
+	protected static final RowMapper<AuditMessage> AUDITMESSAGE_MAPPER_FOR_PARSER = (rs, i) -> {
```
zlamalp: I have a little problem with this kind of replacements. Its not easily recognizable, what method is actually implemented. In Intellij IDEA you can get it displayed only by hovering/clicking on brackets. It would be better, if we had at least full name of parameters passed, like ``(resultSet, rowNumber)`` or whatever instead of ``(rs, i)``.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -1,30 +1,5 @@
 package cz.metacentrum.perun.core.blImpl;
 
-import cz.metacentrum.perun.core.api.Attribute;
```
zlamalp: I believe we wanted to omit wildcard imports, because of clarity for non IDE like text editors. Even thou I believe we are not consistent in it.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -291,11 +290,10 @@
 	 *
 	 * @return non-sponsored member
 	 *
-	 * @throws MemberNotExistsException If member not exists by defined id in system Perun
 	 * @throws MemberNotSponsoredException If member was not set as sponsored before calling this method.
 	 * @throws InternalErrorException if something unexpected happend
 	 */
-	Member unsetSponsorshipForMember(PerunSession session, Member sponsoredMember) throws MemberNotExistsException, MemberNotSponsoredException, InternalErrorException;
```
zlamalp: This kind of removal of not-thrown Exceptions (while I like it in principle) worries me. Since if implementation fails in this case, it throws InternalErrorException with SQL Exception inside. Its OK, but I'm not sure if explicit check on member existence was not intended and is actually missing in implementation! Probably depends on what we want to see in logs/GUI - MemberNotExistsException or InternalErrorException.

Vojtech-Sassmann:
Thanks @zlamalp for reviewing this PR! I agree that It would be better to split the changes. When doing this refactorization, I was already thinking about the changes that they should not at least change the logic of code. But I agree with you that removing unnused excepsions in throws clauses is debatable since they could had been forgotten. But there are really a lot of those unthrown exceptions in the code. With those unnecessary exceptions sthrows there are also a lot of unnecessary catch clauses. We can discuss this tomorrow.

licehammer:
Regarding to "Removed redundant throws declarations". Sometimes, this redundant throws were there on purpose, because we foresaw it might be required in the future if we decided to use different implementation or extend current implementation.

Vojtech-Sassmann:
Ok. I think we should suppress those warnings with comment explaining why there is the exception declared.

Vojtech-Sassmann:
I have opened new request https://github.com/CESNET/perun/pull/2183 where each refactorization step is made in separate commit. Also, some discuitable refactorization were omitted.

PR #2179, opened 2019-03-20
Added new wrappers to getData methods

metodej:
- the methods are getHierarchicalData, getFlatData, getDataWithGroups and getDataWithVos
- if filterExpiredMembers is true methods does not return members that are expired in groups
- if member is expired in all groups that associated him to a resource, he is no longer returned in ChildNodes of that resource
- if user is expired in all groups that associated to any resource that connected him to a facility, he is no longer returned in ChildNodes of that facility


stavamichal:
Rebase it please.

stavamichal:
The commit message is little confusing. What exactly means "expired in group/resource/facility"?

metodej:
Thank you, it should be fixed now.

PR #2180, opened 2019-03-20
Refactorization of method clearOld

metodej:

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/controller/model/ServiceState.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/controller/model/ServiceState.java
@@ -4,7 +4,7 @@ import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Service;
 import cz.metacentrum.perun.taskslib.model.Task;
 
-import java.util.Date;
+import java.time.LocalDateTime;
 
 /**
  * ServiceState is object containing information about one service propagated on one facility.
@@ -90,7 +90,7 @@ public class ServiceState {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public Date getScheduled() {
+	public LocalDateTime getScheduled() {
 		return (task != null) ? task.getSchedule() : null;
 	}
 
@@ -99,7 +99,7 @@ public class ServiceState {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public Date getStartTime() {
+	public LocalDateTime getStartTime() {
 		return (task != null) ? task.getStartTime() : null;
 	}
 
@@ -108,7 +108,7 @@ public class ServiceState {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public Date getEndTime() {
+	public LocalDateTime getEndTime() {
 		return (task != null) ? task.getEndTime() : null;
 	}
 
--- a/perun-base/src/main/java/cz/metacentrum/perun/controller/model/ServiceState.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/controller/model/ServiceState.java
@@ -90,8 +90,8 @@ public class ServiceState {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public LocalDateTime getScheduled() {
-		return (task != null) ? task.getSchedule() : null;
+	public Long getScheduled() {
+		return (task != null) ? task.getScheduleAsLong() : null;
 	}
 
 	/**
@@ -99,8 +99,8 @@ public class ServiceState {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public LocalDateTime getStartTime() {
-		return (task != null) ? task.getStartTime() : null;
+	public Long getStartTime() {
+		return (task != null) ? task.getStartTimeAsLong() : null;
 	}
 
 	/**
@@ -108,8 +108,8 @@ public class ServiceState {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public LocalDateTime getEndTime() {
-		return (task != null) ? task.getEndTime() : null;
+	public Long getEndTime() {
+		return (task != null) ? task.getEndTimeAsLong() : null;
 	}
 
 	/**
--- a/perun-base/src/main/java/cz/metacentrum/perun/rpc/serializer/JsonSerializerGWT.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/rpc/serializer/JsonSerializerGWT.java
@@ -5,17 +5,22 @@ import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.core.api.exceptions.rt.PerunRuntimeException;
 import cz.metacentrum.perun.core.api.exceptions.RpcException;
+import cz.metacentrum.perun.taskslib.model.Task;
 import org.codehaus.jackson.JsonEncoding;
 import org.codehaus.jackson.JsonFactory;
 import org.codehaus.jackson.JsonGenerator;
 import org.codehaus.jackson.JsonProcessingException;
+import org.codehaus.jackson.annotate.JsonIgnore;
 import org.codehaus.jackson.annotate.JsonIgnoreProperties;
+import org.codehaus.jackson.annotate.JsonProperty;
 import org.codehaus.jackson.map.ObjectMapper;
+import org.codehaus.jackson.map.annotate.JsonSerialize;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.time.LocalDateTime;
 
 /**
  * JSON serializer.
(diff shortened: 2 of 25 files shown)


perun-core/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskDaoJdbc.java
```diff
@@ -51,6 +53,18 @@ public static SimpleDateFormat getDateFormatter() {
...
 
+	/**
+	 * Method create formatter with default settings for perun timestamps and set ResolverStyle to STRICT
+	 * Timestamp format:  "dd-MM-yyyy HH:mm:ss" - "ex. 01-01-2014 10:10:10"
+	 *
+	 * ResolverStyle.STRICT means that formatter will be more strict to creating timestamp from string
+	 *
+	 * @return date formatter
+	 */
+	public static DateTimeFormatter getDateTimeFormatter() {
```
zlamalp: If this suppose to replace former ``getDateFormatter()`` from above, you should replace all usages, not just a single usage in ``TaskResultDaoJdbc``. There is a plenty of usages from within ``TaskDaoJdbc`` class itself.

zlamalp:
Works on Oracle DB too, but I realized, that different value is returned from RPC inside Task object, since we changed ``Date`` to ``LocalDateTime`` which serialize to complex JSON object instead of simple long number (seconds from epoch start). I managed to fix GUI, but I'm not sure if its used by any of our API consumer.

zlamalp:
Can you please rebase it. After it we should be able to fix (de)serialization feature by re-adding old ``getX`` methods to the Task object, which will return ``long`` (unix timestamp) computed from the ``LocalDateTime`` and they will be used by our serializer instead of normal methods.

You can see examples of how to handle specific method ignoring and usage during (de)serialization in ``perun-base`` module in ``MemberMixIn`` (private interface) inside ``JsonDeserializer``. You should mark methods with ``LocalDateTime`` as ignored and others as (de)serializable.

I will keep it up to you, if you wish to add also setters to the Task object (they should take unix timestamp as input). I don't think that we currently "receive" Task object anywhere in the API, so its not neccessary (nor we want to store such timestamp in the object itself). But it might simplify your (de)serialization config just by marking ``LodalDateTime`` methods as ignored.

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -88,22 +90,38 @@ public void setRecurrence(int recurrence) {
...
 	}
 
-	public Date getStartTime() {
+	public LocalDateTime getStartTime() {
 		return startTime;
 	}
 
-	public void setStartTime(Date startTime) {
+	public long getStartTimeAsLong() {
+		return startTime.atZone(ZoneId.systemDefault()).toEpochSecond();
```
zlamalp: Please do check on ``null`` for all such getters. I believe we previously returned ``null`` in such case, because object was a ``Date``, now we return primitive ``long`` which isn't an object. We should probably return ``Long`` so its serialized either as ``null`` or ``long`` value.

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -129,54 +147,102 @@ public int getFacilityId() {
...
+	public long getEndTimeAsLong() {
+		return endTime.atZone(ZoneId.systemDefault()).toEpochSecond();
+	}
+
+	public void setEndTime(LocalDateTime endTime) {
 		this.endTime = endTime;
 	}
 
+	public void setEndTime(long endTime) {
+		this.genEndTime = Instant.ofEpochMilli(endTime).atZone(ZoneId.systemDefault()).toLocalDateTime();
```
zlamalp: ``genEndTime`` -> ``endTime``

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -129,54 +147,102 @@ public int getFacilityId() {
...
+		this.genEndTime = Instant.ofEpochMilli(sentToEngine).atZone(ZoneId.systemDefault()).toLocalDateTime();
+	}
+
+	public void setSendStartTime(LocalDateTime sendStartTime) {
 		this.sendStartTime = sendStartTime;
 	}
 
-	public Date getEndTime() {
+	public void setSendStartTime(long sendStartTime) {
+		this.genEndTime = Instant.ofEpochMilli(sendStartTime).atZone(ZoneId.systemDefault()).toLocalDateTime();
```
zlamalp: ``genEndTime`` -> ``sendStartTime``

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -129,54 +147,102 @@ public int getFacilityId() {
...
+		return sentToEngine.atZone(ZoneId.systemDefault()).toEpochSecond();
+	}
+
+	public void setSentToEngine(LocalDateTime sentToEngine) {
 		this.sentToEngine = sentToEngine;
 	}
 
-	public void setSendStartTime(Date sendStartTime) {
+	public void setSentToEngine(long sentToEngine) {
+		this.genEndTime = Instant.ofEpochMilli(sentToEngine).atZone(ZoneId.systemDefault()).toLocalDateTime();
```
zlamalp: ``genEndTime`` -> ``sentToEngine``

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -129,54 +147,102 @@ public int getFacilityId() {
...
+		return genStartTime.atZone(ZoneId.systemDefault()).toEpochSecond();
+	}
+
+	public void setGenStartTime(LocalDateTime genStartTime) {
 		this.genStartTime = genStartTime;
 	}
 
-	public Date getSentToEngine() {
+	public void setGenStartTime(long genStartTime) {
+		this.genEndTime = Instant.ofEpochMilli(genStartTime).atZone(ZoneId.systemDefault()).toLocalDateTime();
```
zlamalp: ``genEndTime`` -> ``genStartTime``

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -129,54 +147,102 @@ public int getFacilityId() {
...
+		return sendEndTime.atZone(ZoneId.systemDefault()).toEpochSecond();
+	}
+
+	public void setSendEndTime(LocalDateTime sendEndTime) {
 		this.sendEndTime = sendEndTime;
 	}
 
-	public Date getSendStartTime() {
+	public void setSendEndTime(long sendEndTime) {
+		this.genEndTime = Instant.ofEpochMilli(sendEndTime).atZone(ZoneId.systemDefault()).toLocalDateTime();
```
zlamalp: ``genEndTime`` -> ``sendEndTime``

perun-base/src/main/java/cz/metacentrum/perun/rpc/serializer/JsonSerializerGWT.java
```diff
@@ -41,6 +45,57 @@
...
+	@SuppressWarnings("unused")
+	private interface TaskMixIn {
+		@JsonSerialize
+		long getStartTimeAsLong();
+
+		@JsonIgnore
+		LocalDateTime getStartTime();
+
+		@JsonSerialize
+		long getScheduleAsLong();
```
zlamalp: To prevent necessity of renaming all properties on client side (GUI, CLI) you can specify name of JSON property using annotations like ``@JsonProperty(value = "schedule")``. It should be name of original property containing LocalDateTime.

perun-base/src/main/java/cz/metacentrum/perun/controller/model/ServiceState.java
```diff
@@ -90,7 +90,7 @@ public int getTaskId() {
 	 *
 	 * @return Time when was last task scheduled.
 	 */
-	public Date getScheduled() {
+	public LocalDateTime getScheduled() {
```
zlamalp: Please change it to internally use ``task.getScheduleAsLong()``. Otherwise you need to implement same supplementary methods as in ``Task`` object, including changes in serializer.
Same goes for ``getStartTime`` and ``getEndTime()``.

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
zlamalp: This comment was not resolved last time. Please do check on ``null`` before you call method on the object.

perun-base/src/main/java/cz/metacentrum/perun/taskslib/model/Task.java
```diff
@@ -129,54 +147,102 @@ public int getFacilityId() {
...
 	}
 
-	public Date getGenEndTime() {
+	public LocalDateTime getGenEndTime() {
 		return genEndTime;
 	}
 
-	public void setGenEndTime(Date genEndTime) {
+	public Long getGenEndTimeAsLong() {
+		return (genEndTime == null) ? null : genEndTime.atZone(ZoneId.systemDefault()).toEpochSecond();
```
zlamalp: We need this to return epoch in milliseconds (java Date interpretation) and not unix epoch (in seconds), so GUI will interpret this correctly. Try it like this: ``genEndTime.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli();`` with all getters.

PR #2183, opened 2019-03-21
Minor refactorizations

Vojtech-Sassmann:
* Each refactorization type is seperated in its own commit. It should be more clear what each refactorization means.


perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
```diff
@@ -22,30 +22,30 @@
 
 	// Attributes related to the external groups
 	// Contains query need to get the subject groups
-	public static final String GROUPSQUERY_ATTRNAME = AttributesManager.NS_GROUP_ATTR_DEF + ":groupsQuery";
+	String GROUPSQUERY_ATTRNAME = AttributesManager.NS_GROUP_ATTR_DEF + ":groupsQuery";
```
zlamalp: I believe constants should remain at least ``final``. I'm not saying that this is used outside, but it should be available to the rest of the code, hence reason for ``static`` too.

zlamalp: Ok, @martin-kuba pointed out, that in case of interface variables are final and static by default, hence it can be all omitted.

PR #2184, opened 2019-03-22
Optimized imports in perun-core

Vojtech-Sassmann:
* Optimized imports in perun-core module.


Vojtech-Sassmann:
These changes were made by running `optimize imports` on the whole perun-core module.

PR #2185, opened 2019-03-22
More refactorization

Vojtech-Sassmann:
* Anonymous classes replaced by lambdas. I have tried to use better variables names like `resultSet` instead of `rs`.

* For loop replaced with streamAPI or foreach loops

* StringBuffer replaced with StringBuilder

* Removed unnecessary boxing


zlamalp:
Can you please rebase it ? Thank you.

PR #2190, opened 2019-03-28
GUI: Changed Date to LocalDateTime in Task object

zlamalp:
- We changed Task property data types from Date to LocalDateTime, which
  is not present in GWT and must be replaced with our own class
  extending JavaScriptObject.
- Read date/time properties in Task using this new object.
- This change is related to pull-request #2180 and shouldn't be merged
  without it.


zlamalp:
We resolved backward compatibility in #2180. Closing this pull-request.

PR #2193, opened 2019-04-03
Fixing probable bugs in core module

metodej:
First waves of fixing probable bugs that includes
- deleting unused or redundant assignments
- fixing passing wrong or null object
- removing unnecessary conditions
- fixing conditions that were buggy (wrong operator or empty body)
- fixing some cases where NullPointerException could be thrown
- fixing unchecked conversions of attribute.getValue()
- removing Collections.emptyList()
- removing unused statement

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
@@ -131,7 +131,7 @@ public class PropagationStatsReaderImpl implements PropagationStatsReader {
 		FacilityState state = new FacilityState();
 		state.setFacility(facility);
 		// if no tasks we can't determine facility state
-		if (tasks.isEmpty()) {
+		if (tasks == null || tasks.isEmpty()) {
 			state.setState(FacilityPropagationState.NOT_DETERMINED);
 			return state;
 		} else {
--- a/perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
@@ -131,7 +131,7 @@ public class PropagationStatsReaderImpl implements PropagationStatsReader {
 		FacilityState state = new FacilityState();
 		state.setFacility(facility);
 		// if no tasks we can't determine facility state
-		if (tasks.isEmpty() || tasks == null) {
+		if (tasks.isEmpty()) {
 			state.setState(FacilityPropagationState.NOT_DETERMINED);
 			return state;
 		} else {
--- a/perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
@@ -210,10 +210,9 @@ public class PropagationStatsReaderImpl implements PropagationStatsReader {
 	@Override
 	public List<FacilityState> getAllFacilitiesStates(PerunSession session) throws InternalErrorException, PrivilegeException, FacilityNotExistsException {
 		List<FacilityState> list = new ArrayList<>();
-		List<Facility> facs = new ArrayList<>();
 
 		// return facilities where user is admin or all if perun admin
-		facs = perun.getFacilitiesManager().getFacilities(session);
+		List<Facility> facs = perun.getFacilitiesManager().getFacilities(session);
 		Collections.sort(facs);
 		for (Facility facility : facs) {
 			list.add(getFacilityState(session, facility));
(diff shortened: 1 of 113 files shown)


stavamichal:
In commit message there should be at least a list of bugs resolved in this wave.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_unixGroupName_namespace.java
```diff
@@ -113,9 +113,7 @@ public void changedAttributeHook(PerunSessionImpl session, Resource resource, At
 		String groupNameNamespace = attribute.getFriendlyNameParameter();
 
 		try {
-			if(attribute.getValue() == null) {
-				//This is ok, for now no changes for removing some GroupName of this Resource
```
Vojtech-Sassmann: I would suggest keeping the comment.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_userPreferredCertDN.java
```diff
@@ -40,7 +40,7 @@ public void checkAttributeValue(PerunSessionImpl sess, User user, Attribute attr
 			else return;
 		}
 		if(attribute.getValue() == null) {
-			if(certDNsValue != null || !certDNsValue.isEmpty()) throw new WrongAttributeValueException(attribute, user, "This attribute value can't be null because of notNull attribute userCertDNs");
```
Vojtech-Sassmann: I think there should be `&&` instead of `||` and the condition should not be removed.

perun-core/src/main/java/cz/metacentrum/perun/controller/service/impl/PropagationStatsReaderImpl.java
```diff
@@ -134,7 +134,7 @@ public FacilityState getFacilityState(PerunSession session, Facility facility) t
 		FacilityState state = new FacilityState();
 		state.setFacility(facility);
 		// if no tasks we can't determine facility state
-		if (tasks.isEmpty() || tasks == null) {
```
Vojtech-Sassmann: This should probably be `tasks == null || tasks.isEmpty()`

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/SecurityTeamsManagerBlImpl.java
```diff
@@ -111,19 +111,23 @@ public void deleteSecurityTeam(PerunSession sess, SecurityTeam securityTeam, boo
 		if (blacklist != null && !blacklist.isEmpty() && !forceDelete) {
 			throw new RelationExistsException("SecurityTeam has blacklisted users.");
 		}
-		for (User blacklistedUser : blacklist) {
```
Vojtech-Sassmann: In this case, the blackList cannot be null. The method above never returns null. Actually, the problem is in the if statement above. `blacklist != null` should be removed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/SecurityTeamsManagerBlImpl.java
```diff
@@ -111,19 +111,23 @@ public void deleteSecurityTeam(PerunSession sess, SecurityTeam securityTeam, boo
...
 
 		// remove security team from all facilities
 		List<Facility> facilities = getPerunBl().getFacilitiesManagerBl().getAssignedFacilities(sess, securityTeam);
 		if (facilities != null && !facilities.isEmpty() && !forceDelete) {
 			throw new RelationExistsException("SecurityTeam is assigned to some facilities.");
 		}
-		for (Facility facility : facilities) {
-			// calling BL will make auditer message about facility to appear.
-			getPerunBl().getFacilitiesManagerBl().removeSecurityTeam(sess, facility, securityTeam);
+		if (facilities != null) {
```
Vojtech-Sassmann: Same here. This check is redundant. You should remove the check `facilities != null` from the condition above.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1167,9 +1167,11 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
 			throw new InternalErrorException(ex);
 		}
 
-		String mailAddress = (String) mailAttribute.getValue();
+		if (mailAttribute != null) {
```
Vojtech-Sassmann: This is not a correct fix. The mailAttribute can be null if the given attributeDefinition is not `user` or `member` attribute. Instead, if the value is null, you should throw some Exception. I would suggest throwing InternalErrorException since we don't have an appropriate Exception for this use case.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupStructureSynchronizationIntegrationTest.java
```diff
@@ -93,7 +93,6 @@ public void setUpBeforeEveryMethod() throws Exception {
 		MockitoAnnotations.initMocks(this);
 
 		doReturn(essa).when(extSourceManagerBl).getExtSourceByName(any(PerunSession.class), any(String.class));
-		doReturn(EXT_SOURCE_NAME).when((ExtSourceLdap)essa).getName();
```
Vojtech-Sassmann: This should probably be `doReturn(EXT_SOURCE_NAME).when(((ExtSourceLdap)essa).getName());`

Vojtech-Sassmann:
Also, please resolve the conflicts. You should probably keep all changes from master.

PR #2195, opened 2019-04-08
fix regular expressions matching audit messages 

mvocu:
fix regular expressions matching audit messages with exponential complexity
  - this caused apparent hangups of perun-ldapc-ada on longer audit messages
add $ to RE matching audit messages
do not throw exception when processing attribute not describe in LDAP defs
  - audit messages concerning attributes not reflected in LDAP are ignored just with log message

### DIFF
--- a/perun-ldapc-ada/src/main/resources/perun-ldapc.xml
+++ b/perun-ldapc-ada/src/main/resources/perun-ldapc.xml
@@ -177,7 +177,7 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 							<value>cz.metacentrum.perun.core.api.Attribute</value>
 						</list>
 					</property>
-					<property name="pattern" value=" set for User:\[(.|\s)*\]" />
+					<property name="pattern" value=" set for User:\[.*\]$" />
 					<property name="handlerMethodName" value="processAttributeSet" />
 				</bean>
 				<bean class="cz.metacentrum.perun.ldapc.processor.impl.RegexpDispatchEventCondition">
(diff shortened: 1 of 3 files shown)


perun-ldapc-ada/src/main/resources/perun-ldapc.xml
```diff
@@ -177,7 +177,7 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 							<value>cz.metacentrum.perun.core.api.Attribute</value>
 						</list>
 					</property>
-					<property name="pattern" value=" set for User:\[(.|\s)*\]" />
+					<property name="pattern" value=" set for User:\[.*\]$" />
```
zlamalp: While its not part of changed code I noticed, that you construct regex like this ```Pattern.compile(regexp, Pattern.MULTILINE | Pattern.DOTALL);``` while old LDAPc uses only ```Pattern.DOTALL```.

I don't say, that its wrong, but ``Pattern.DOTALL`` tells that dot matches also all newlines, while ``Pattern.MULTILINE`` tells that ``^`` and ``$`` match on each line of input. And we use ``.*`` to represent any PerunBean object content.

Since you don't match the start of the string (each line) and I believe that present dot matches and consumes everything before end of line is matched, then it probably works.

perun-ldapc-ada/src/main/resources/perun-ldapc.xml
```diff
@@ -337,7 +337,7 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 							<value>cz.metacentrum.perun.core.api.Group</value>
 						</list>
 					</property>
-					<property name="pattern" value="Group:\[(.|\s)*\] was moved" />
+					<property name="pattern" value="Group:\[.*\] was moved$" />
```
zlamalp: In old LDAPc we match it like this ```^Group:\\[(.*)\\] was moved\\.$``` (with ending dot, which is a part of auditer message). I'm not sure, that it will work here since you match end of string right after ``moved``.

Generally this is same for most of messages, but they are not consistent whether they have ending "dot" or not, so in most cases we don't match end of string and don't use ``Pattern.MULTILINE``, but ``Pattern.DOTALL``.

mvocu:
I removed the MULTILINE option and also removed the matching of end of strings. 

I think that matching of messages remains still rather fragile, as we rely on greediness of .* RE. However, RE's are not used here to parse the message, just to indicate if it should be handled, so it is probably safe.

PR #2196, opened 2019-04-08
CORE: Allow Facility managers to read/write User attributes

zlamalp:
- Previously facility managers couldn't read or write any of user
  attributes.
- Now we allow facility manager to read and write User attributes if
  User is assigned to any facility of such manager and if attribute
  itself is set to allow facility manager role.


zlamalp:
We decided not to allow this behavior after all. Closing.

PR #2198, opened 2019-04-10
Added new virtual attribute module userOrganizations

metodej:
- value of the attribute is map from short names of vos to value of member:def:organization

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userOrganizations.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userOrganizations.java
@@ -1,0 +1,63 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userOrganizations.java
```diff
@@ -0,0 +1,64 @@
...
+	public Attribute getAttributeValue(PerunSessionImpl perunSession, User user, AttributeDefinition attributeDefinition) throws InternalErrorException {
+		Map<String, String> userOrganizations = new LinkedHashMap<>();
+		List<Vo> vos = perunSession.getPerunBl().getUsersManagerBl().getVosWhereUserIsMember(perunSession, user);
+
+		for (Vo vo : vos) {
+			try {
+				Member member = perunSession.getPerunBl().getMembersManagerBl().getMemberByUser(perunSession, vo, user);
+				String value = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, member, A_M_organization).valueAsString();
+
+				userOrganizations.put(vo.getShortName(), value);
```
zlamalp: Please do not add key-value pair to the result if value is null -> user doesn't have organization set for such VO.

metodej: Ok, it should be fixed now.

PR #2207, opened 2019-04-23
Fixed cutting of utf8 chars in TaskLib

Vojtech-Sassmann:
* The old implementation of method `clearZeroBytesFromString` was
incorrect. If the cut was made inside a multi byte UTF8 character, the character
would be unreadable.

* The new implementation checks if the cut was made inside a multi byte
character and if so, the rest of the needed bytes are added.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskResultDaoJdbc.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskResultDaoJdbc.java
@@ -18,13 +18,13 @@ import org.springframework.transaction.annotation.Transactional;
 import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 import java.time.LocalDateTime;
-import java.util.ArrayList;
 import java.util.List;
 
 @Transactional
 public class TaskResultDaoJdbc extends JdbcDaoSupport implements TaskResultDao {
 
 	private static final Logger log = LoggerFactory.getLogger(TaskResultDaoJdbc.class);
+	private static final int MAX_NUMBER_OF_UTF8_BYTES = 4;
 	private NamedParameterJdbcTemplate  namedParameterJdbcTemplate;
 
 	public final static String taskResultMappingSelectQuery = " tasks_results.id as tasks_results_id, tasks_results.task_id as tasks_results_task_id," +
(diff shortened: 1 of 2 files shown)


stavamichal:
Please rebase this PR.

Vojtech-Sassmann:
@stavamichal done.

perun-core/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskResultDaoJdbc.java
```diff
@@ -293,20 +292,52 @@ public int clearOld(int engineID, int numDays) {
...
-			}
-			if(data[limit-1] >= 0xF0 ) {
-				dataOut.write(data[limit+2]);
+			// check if we have to add some bytes in case we have split an UTF-8 character that is longer than 1 byte.
+			if (!isASingleByteUTF8Char(data[limit-1])) {
+				int i = 0;
+				while (!isAStartingByteUTF8Char(data[limit + i])) {
+					dataOut.write(data[limit + i]);
+					i++;
+				}
```
stavamichal: I would add some testing of the maximal length of UTF8 character. It will prevent problems with wrong formatted input. What do you think?

Vojtech-Sassmann: That's a good point. I will add it.

Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskResultDaoJdbc.java
```diff
@@ -287,26 +287,61 @@ public int clearOld(int engineID, int numDays) {
 	private static byte[] clearZeroBytesFromString(byte[] data, int maxLength) {
 		if (data == null) return null;
 		ByteArrayOutputStream dataOut = new ByteArrayOutputStream() ;
-		int limit = (maxLength < data.length) ? maxLength - 4 : data.length;
+		int limit = (maxLength < data.length) ? maxLength - MAX_NUMBER_OF_UTF8_BYTES : data.length;
 		for (int i = 0; i < limit; i++) {
 			if (data[i] != 0x00)
 				dataOut.write(data[i]);
 		}
 		if(maxLength < data.length) {
```
stavamichal: This boolean value is there twice. I would add a variable with good description in name. Something like:
`boolean dataExceedMaxLength = maxLength < data.length;`
And then use this variable in both if clause.

perun-core/src/main/java/cz/metacentrum/perun/taskslib/dao/jdbc/TaskResultDaoJdbc.java
```diff
@@ -287,26 +287,61 @@ public int clearOld(int engineID, int numDays) {
...
-			}
-			if(data[limit-1] >= 0xF0 ) {
-				dataOut.write(data[limit+2]);
+			// check if we have to add some bytes in case we have split an UTF-8 character that is longer than 1 byte.
+			if (!isASingleByteUTF8Char(data[limit-1])) {
+				int i = 0;
+				while (!isAStartingByteUTF8Char(data[limit + i])) {
+					if (i == MAX_NUMBER_OF_UTF8_BYTES) {
+						break;
+					}
```
stavamichal: Question is:
Do we want to continue when input is corrupted (not correct UTF-8) and save it anyway?

This is probably question for @zlamalp

stavamichal: Add there at least "log.error" please before break. It should be enough for now. This situation should not happen.

Vojtech-Sassmann: I have added logging of the message with appropriate information.

PR #2209, opened 2019-05-06
add LoA and isCesnetEligible attributes to LDAPc

mvocu:
- adds two new user attributes to both ldapc and ldapc-ada
 - cleans up forgotten definition of preferredMail from ldapc-ada (was not used anyway)


perun-ldapc-initializer/src/main/java/cz/metacentrum/perun/ldapc/initializer/impl/Utils.java
```diff
@@ -482,6 +482,30 @@ public void generateAllUsersToWriter(PerunInitializer perunInitializer) throws I
 			else bonaFideStatus+= (String) attrBonaFideStatus.getValue();
 			if(bonaFideStatus != null) writer.write(bonaFideStatus + '\n');
 
+			//isCesnetEligible
+			String isCesnetEligible = "bonaFideStatus: ";
```
zlamalp: I believe it should be ``isCesnetEligible`` and not ``bonaFideStatus``.

zlamalp:
Must be merged with #2225 in order to update LDAP with current attribute value for user:virt:loa.

PR #2210, opened 2019-05-06
EduPersonEntitlement virtual attribute module

dBucik:
**PR Content:**
- implemented attribute module for 'urn:perun:user:attribute-def:virt:eduPersonEntitlement'
- implemented tests for new attribute module 

**What needs to be done on instances:**
- create attribute urn:perun:user:attribute-def:virt:eduPersonEntitlement with type Array
- create attribute urn:perun:ues:attribute-def:def:eduPersonEntitlement with type String/LargeString
- attribute urn:perun:user:attribute-def:virt:groupNames should exist (type Array)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonEntitlement.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonEntitlement.java
@@ -1,0 +1,130 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AllAttributesRemovedForUser;
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeRemovedForUser;
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeSetForUser;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributeCollectedFromUserExtSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonEntitlement.java
```diff
@@ -0,0 +1,133 @@
...
+@SuppressWarnings("unused")
+public class urn_perun_user_attribute_def_virt_eduPersonEntitlement extends UserVirtualAttributeCollectedFromUserExtSource {
+
+	private final Logger log = LoggerFactory.getLogger(this.getClass());
+
+	private final Pattern userAllAttrsRemovedPattern = Pattern.compile("All attributes removed for User:\\[(.*)]", Pattern.DOTALL);
+
+	@Override
+	public String getSourceAttributeFriendlyName() {
+		return "eduPersonEntitlement";
```
zlamalp:
```suggestion
		return "entitlement";
```
By default SP translate attribute ``eduPersonEntitlement`` to ``entitlement`` and such name must match perun attribute friendly name.

dBucik: For some weird reason when I change the proposed name, the tests start failing. Especially the resolveAttributeValueChangeTest() is problematic. Could you please take a look at it what might be wrong?

zlamalp: I'm not sure, tests are working for me when I change it. But I see following auditer message printed in stdout ``[null set for User:[id=<1>, titleBefore=<\0>, firstName=<\0>, lastName=<\0>, middleName=<\0>, titleAfter=<\0>, serviceAccount=<false>, sponsoredAccount=<false>].]`` which is weird, but might be ok for test purpose.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonEntitlement.java
```diff
@@ -0,0 +1,133 @@
...
+ *  - urn:perun:user:attribute-def:virt:groupNames
+ *
+ * @author Dominik Frantisek Bucik <bucik@ics.muni.cz>
+ */
+@SuppressWarnings("unused")
+public class urn_perun_user_attribute_def_virt_eduPersonEntitlement extends UserVirtualAttributeCollectedFromUserExtSource {
+
+	private final Logger log = LoggerFactory.getLogger(this.getClass());
+
+	private final Pattern userAllAttrsRemovedPattern = Pattern.compile("All attributes removed for User:\\[(.*)]", Pattern.DOTALL);
```
zlamalp: This seems to be an unused variable, since we use message classes now.

dBucik:
made the requested changes, all tests are passing now (at least for me)

PR #2212, opened 2019-05-07
BASE: We will collect also eduPersonEntitlement attributes

zlamalp:
- Its provided by SP under "entitlement" name.
- Add it to the default collected attributes for IdP.
- Since it might contain a lot of data, its large string by default.


zlamalp:
This needs to follow #2210.

PR #2223, opened 2019-05-17
Add new method to get rich groups by resource and member

stavamichal:
- we already have method to get richGroups with attributes by object
 Resource (assigned to the group) but we also want to filter these
 groups by specific member (return groups where member has membership,
 remove others from the list). For this reason, new method with also
 Member object in the parameters was added.
 - add this new method to RPC, Entry layer and BL layer
 - add test for this new method
 - as side effect, method filterOnlyAllowedAttributes now also can
 filter member-group attributes. There are also small changes in the
 names of variable and order of processing this filtering method, but
 these changes are only cosmetic.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -2326,27 +2326,36 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 	@Override
 	public List<RichGroup> convertGroupsToRichGroupsWithAttributes(PerunSession sess, Member member, Resource resource, List<Group> groups, List<String> attrNames) throws InternalErrorException, GroupResourceMismatchException, MemberResourceMismatchException {
 		List<RichGroup> richGroups = new ArrayList<>();
+
+		//filter attr names for different namespaces (we need to process them separately)
+		List<String> groupAndGroupResourceAttrNames = new ArrayList<>();
+		List<String> memberGroupAttrNames = new ArrayList<>();
+		if(attrNames != null && !attrNames.isEmpty()) {
+			groupAndGroupResourceAttrNames = attrNames.stream().filter(attrName ->
+				attrName.startsWith(AttributesManager.NS_GROUP_RESOURCE_ATTR) || attrName.startsWith(AttributesManager.NS_GROUP_ATTR)).collect(Collectors.toList());
+			memberGroupAttrNames = attrNames.stream().filter(attrName -> attrName.startsWith(AttributesManager.NS_MEMBER_GROUP_ATTR)).collect(Collectors.toList());
+		}
+
 		for(Group group: groups) {
-			List<Attribute> attributes = new ArrayList<>();
-
-			//Add all group and group-resource attributes (empty list means all possible attributes there)
-			List<String> supportedNames = attrNames.stream().filter(attrName ->
-				attrName.startsWith(AttributesManager.NS_GROUP_RESOURCE_ATTR) || attrName.startsWith(AttributesManager.NS_GROUP_ATTR)).collect(Collectors.toList()
-			);
-			attributes.addAll(getPerunBl().getAttributesManagerBl().getAttributes(sess, resource, group, supportedNames, true));
-
-			//Add all member-group attributes (empty list means all possible attributes there)
-			//WARNING - there is different behavior of method getAttributes for member and group by list of attr names, empty list means "no attributes to return"
-			supportedNames = attrNames.stream().filter(attrName ->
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2313,6 +2323,34 @@ public RichGroup convertGroupToRichGroupWithAttributesByName(PerunSession sess,
...
+				attrName.startsWith(AttributesManager.NS_GROUP_RESOURCE_ATTR) || attrName.startsWith(AttributesManager.NS_GROUP_ATTR)).collect(Collectors.toList()
+			);
+			attributes.addAll(getPerunBl().getAttributesManagerBl().getAttributes(sess, resource, group, supportedNames, true));
+
+			//Add all member-group attributes (empty list means all possible attributes there)
+			//WARNING - there is different behavior of method getAttributes for member and group by list of attr names, empty list means "no attributes to return"
+			supportedNames = attrNames.stream().filter(attrName ->
+				attrName.startsWith(AttributesManager.NS_MEMBER_GROUP_ATTR)).collect(Collectors.toList()
+			);
+			if(supportedNames.isEmpty()) {
```
zlamalp: I'm not sure this is consistent with the rest of our API. Like this you always return all ``member-group`` attributes, even if caller specified eg. single ``group`` attribute in the outer API call.
So you should probably check emptiness only on outer ``attrNames`` param and not on filtered values. But maybe you have different expectations for this specific method.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -682,6 +682,20 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * Allowed namespaces of attributes are group, group-resource and member-group.
+	 *
+	 * Last step is filtration of attributes:
+	 * Attributes are filtered by rights of user in session. User get only those selected attributes he has rights to read.
+	 *
+	 * @param member int Member <code>id</code>
+	 * @param resource int Resource <code>id</code>
+	 * @param attrNames List<String> names of attributes
+	 * @return List<RichGroup> groups with attributes
+	 *
```
zlamalp: Unnecessary newline.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -682,6 +682,20 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 **
```
zlamalp: I'm not sure how our perl parser will handle this. Please make it a single ``*``.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -682,6 +682,20 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * Get list of all richGroups with selected attributes assigned to the resource filtered by specific member.
+	 * Allowed namespaces of attributes are group, group-resource and member-group.
+	 *
+	 * Last step is filtration of attributes:
```
zlamalp: We should probably rewrite this message to by more RPC API like.

You will get only attributes which you are authorized to read. You must specify names of requested attributes by their URNs in ``attrNames``. You can pass empty list to return all possible attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: I changed this a bit. Now it should work fine.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
stavamichal: Removed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
stavamichal: Removed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
stavamichal: Javadoc and the processing of code was changed.

PR #2225, opened 2019-05-22
CORE: Support resolving value change in user:virt:loa attribute

zlamalp:
- Since we will push user:virt:loa attribute to the LDAP, we must
  resolve its value changes from other auditer messages
  (changes to the user ext source).
- Update of lastSeen timestamp for UserExtSource must also emit
  UserExtSourceUpdated auditer message, since attribute module takes
  value only from UESes user used in a last year.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -618,14 +618,15 @@ public class UsersManagerBlImpl implements UsersManagerBl {
 
 	@Override
 	public UserExtSource updateUserExtSource(PerunSession sess, UserExtSource userExtSource) throws InternalErrorException, UserExtSourceExistsException {
+		UserExtSource updatedUes = getUsersManagerImpl().updateUserExtSource(sess, userExtSource);
 		getPerunBl().getAuditer().log(sess, new UserExtSourceUpdated(userExtSource));
-		return getUsersManagerImpl().updateUserExtSource(sess, userExtSource);
+		return updatedUes;
 	}
 
 	@Override
 	public void updateUserExtSourceLastAccess(PerunSession sess, UserExtSource userExtSource) throws InternalErrorException {
-		getPerunBl().getAuditer().log(sess, new UserExtSourceUpdated(userExtSource));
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_loa.java
```diff
@@ -51,4 +64,51 @@ public AttributeDefinition getAttributeDefinition() {
...
+	 * Resolve and create new auditer messages about LOA attribute change based on current attribute value.
+	 *
+	 * @param sess PerunSession
+	 * @param user User to resolve LoA messages
+	 * @return List of new messages or empty list
+	 * @throws InternalErrorException
+	 * @throws AttributeNotExistsException
+	 * @throws WrongAttributeAssignmentException
+	 */
+	private List<AuditEvent> resolveEvent(PerunSessionImpl sess, User user) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
```
stavamichal: Why is output of this method the list of audit events if only one message is always generated?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -624,6 +624,7 @@ public UserExtSource updateUserExtSource(PerunSession sess, UserExtSource userEx
 
 	@Override
 	public void updateUserExtSourceLastAccess(PerunSession sess, UserExtSource userExtSource) throws InternalErrorException {
+		getPerunBl().getAuditer().log(sess, new UserExtSourceUpdated(userExtSource));
```
stavamichal: You should always generate message after the process is done. You can prevent unexpected issues.

zlamalp: Fixed, also in standard ``updateUserExtSource()`` method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_loa.java
zlamalp: Fixed, I do return single ``AuditEvent`` now.

PR #2228, opened 2019-05-22
CORE: Allow reporting to RT to be redirected to mail

zlamalp:
* Added new config option "perun.rt.sendToMail" into perun.properties.
  If set, it will cause all RT messages (error reporting) to be
  redirected to its value (mail address).
* Updated RTMessagesManager to allow sending of report to mail
  addresses instead of contacting RT API.


zlamalp:
Tested locally on devel, works.

PR #2241, opened 2019-06-03
Modified user:virt:eduPersonEntitlement

dBucik:
- Removed collecting value from urn:perun:user:attribute-def:virt:groupNames
- Modified tests according to change


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonEntitlement.java
```diff
@@ -111,20 +64,6 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
 				return null;
 			}
 		});
-		handleIdentifiers.add(auditEvent -> {
```
zlamalp: Since attribute no longer depends on user attributes, we should emmit audit message about value change if all user attributes are removed. So remove whole ```getHandleIdentifiers()``` and let super implementation to do its work on UserExtSource attributes only.

PR #2246, opened 2019-06-04
Expiration of member need to check validity of attributes

stavamichal:
- if member was in invalid or disabled state, his attributes need to be
 checked, because expired member in Perun is defined as "allowed" and
 his attributes need to be valid for services

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -3071,7 +3071,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 									perunBl.getMembersManagerBl().expireMember(sess, richMember);
 									log.info("Switching member id {} to EXPIRE state, due to expiration {}.", richMember.getId(), membershipExpiration.getValue());
 									log.debug("Switching member to EXPIRE state, additional info: membership expiration date='{}', system now date='{}'", currentMembershipExpirationDate, now);
-								} catch (MemberNotValidYetException e) {
+								} catch (MemberNotValidYetException | WrongReferenceAttributeValueException | WrongAttributeValueException e) {
 									log.error("Consistency error while trying to expire member id {}, exception {}", richMember.getId(), e);
 								}
 							}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -155,6 +155,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 	private static final String A_G_D_EXPIRATION_RULES = AttributesManager.NS_GROUP_ATTR_DEF + ":groupMembershipExpirationRules";
 	private static final String A_MG_D_MEMBERSHIP_EXPIRATION = AttributesManager.NS_MEMBER_GROUP_ATTR_DEF + ":groupMembershipExpiration";
 	private static final String A_M_V_LOA = AttributesManager.NS_MEMBER_ATTR_VIRT + ":loa";
+	private static final List<Status> statusesAffectedBySynchronization = Arrays.asList(Status.DISABLED, Status.EXPIRED, Status.INVALID);
 
 	private final Integer maxConcurrentGroupsStructuresToSynchronize;
 	private final PerunBeanProcessingPool<Group> poolOfGroupsStructuresToBeSynchronized;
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3071,7 +3071,7 @@ private void updateExistingMembersWhileSynchronization(PerunSession sess, Group
 									perunBl.getMembersManagerBl().expireMember(sess, richMember);
 									log.info("Switching member id {} to EXPIRE state, due to expiration {}.", richMember.getId(), membershipExpiration.getValue());
 									log.debug("Switching member to EXPIRE state, additional info: membership expiration date='{}', system now date='{}'", currentMembershipExpirationDate, now);
-								} catch (MemberNotValidYetException e) {
+								} catch (MemberNotValidYetException | WrongReferenceAttributeValueException | WrongAttributeValueException e) {
```
zlamalp: There is a minor problem outside the scope of your change. If the expiration of member fails on attributes, then he is kept in disabled state. Later in this method it tries to validate any still disabled member. Hence we call validation twice for no reason and second one will switch him to valid (disregarding expiration date). We should merge / rework following "IFs" into single one, so that only one branch is used on member.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1541,16 +1541,23 @@ public Member suspendMember(PerunSession sess, Member member, String message) th
...
 		}
 
 		if(this.haveStatus(sess, member, Status.INVALID)) throw new MemberNotValidYetException(member);
+		Status oldStatus = member.getStatus();
 		getMembersManagerImpl().setStatus(sess, member, Status.EXPIRED);
 		member.setStatus(Status.EXPIRED);
 		getPerunBl().getAuditer().log(sess, new MemberExpired(member));
+
+		//We need to check validity of attributes first (expired member has to have valid attributes)
+		if(oldStatus.equals(Status.INVALID) || oldStatus.equals(Status.DISABLED)) {
```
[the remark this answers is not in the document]
stavamichal: Fixed. Change from INVALID to EXPIRED is now correct and supported way. Validity of attributes is checked in this case.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: This par of code was reworked.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3057,56 +3057,51 @@ private void updateExistingMembersWhileSynchronization(PerunSession sess, Group
 			}
 
 			//Set correct member Status
-			// If the member has expired or disabled status, try to expire/validate him (depending on expiration date)
-			if (richMember.getStatus().equals(Status.DISABLED) || richMember.getStatus().equals(Status.EXPIRED)) {
+			List<Status> affectedStatuses = new ArrayList<>();
```
zlamalp: We can make this list ``final static`` variable of the class. Like this its constructed for each member in the cycle.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3057,56 +3057,53 @@ private void updateExistingMembersWhileSynchronization(PerunSession sess, Group
...
 
 			//Set correct member Status
-			// If the member has expired or disabled status, try to expire/validate him (depending on expiration date)
 			if (richMember.getStatus().equals(Status.DISABLED) || richMember.getStatus().equals(Status.EXPIRED)) {
+				//prepare variables with information about member's expiration
+				boolean memberHasExpiration;
+				boolean memberExpiredInPast = false;
+
 				Date now = new Date();
+				Date currentMembershipExpirationDate = now;
```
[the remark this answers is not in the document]
stavamichal: This variable need to be used for logging, so I can't really move it any further.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: Moved as static list for the whole class.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3057,56 +3058,47 @@ private void updateExistingMembersWhileSynchronization(PerunSession sess, Group
...
-			}
 
-			// If the member has still DISABLED status, try to validate the member
-			try {
-				if (richMember.getStatus().equals(Status.DISABLED)) {
-					getPerunBl().getMembersManagerBl().validateMember(sess, richMember);
+				if ((Status.DISABLED.equals(memberStatus) || Status.INVALID.equals(memberStatus)) && memberHasExpiration && memberExpiredInPast) {
+					//If member has expiration in the past (should be expired now) and is in other state than expired, expire him
+					try {
+						perunBl.getMembersManagerBl().expireMember(sess, richMember);
```
licehammer:
```suggestion
						richMember = (RichMember) perunBl.getMembersManagerBl().expireMember(sess, richMember);
```
expireMember method change status of object as a side effect. So either put it into documentation of expireMember or rather not reply on it.

stavamichal: Added to javadoc an information about side effect.

PR #2248, opened 2019-06-06
CORE, GUI: Split form item`s perun attribute to source and destination

zlamalp:
- Commit contains DB version update.
- Relation of form item to perun attribute is now split between
  source and destination. First is used to pre-fill form item and
  second to store value of form item into perun.
- For existing form items, source and destination attribute is set
  to the same value by DB update to make sure, that former behavior
  is kept.
- For new form items user must explicitly set source and destination
  attribute or item is not pre-filled/stored respectively.
- For source attribute we can pick either VO and group attributes too
  (if form is for group). Like this you can fill default value
  from VO/Group settings.
  Also by splitting source and destination you can use virtual
  attributes of user/member/group/vo to dynamically calculate
  pre-filled value.
- Updated GUI so we can actually set both attribute relations.
- Removed unused code.


perun-core/src/main/resources/hsqldbChangelog.txt
```diff
@@ -7,10 +7,15 @@
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
 -- this update is not supported on hsql since its used only as in-memory db
+
+3.1.53
+alter table application_form_items ADD src_attr varchar(384);
+update application_form_items set src_attr = application_form_items.dst_attr where id=application_form_items.id;
```
stavamichal: I think where clause can be omitted.

perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,11 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.53
+alter table application_form_items ADD src_attr varchar(384);
+update application_form_items set src_attr = application_form_items.dst_attr where id=application_form_items.id;
```
stavamichal: I think where clause can be omitted.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/EditFormItemTabItem.java
```diff
@@ -628,13 +675,25 @@ public void onChange(ChangeEvent event) {
 
 			row++;
 
+			Label srcAttrLabel = new Label("Source attribute:");
+			ft.setWidget(row, 0, srcAttrLabel);
+			ft.setWidget(row, 1, perunSourceAttributeListBox);
+			ftf.setColSpan(row, 1, 2);
+
+			row++;
+			ft.setHTML(row, 1, "Select attribute, which will be used to pre-fill form value. You can select also VO "+(forGroup ? "and group" : "")+" attributes.");
```
stavamichal: If forGroup is false, then there will be two spaces between words "VO" and "attributes". I would replace it by:
```
ft.setHTML(row, 1, "Select attribute, which will be used to pre-fill form value. You can select also VO "+(forGroup ? "and group " : "")+"attributes.");
```

perun-core/src/main/resources/hsqldbChangelog.txt
zlamalp: Fixed in new commit.

perun-core/src/main/resources/postgresChangelog.txt
zlamalp: Fixed in new commit.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/EditFormItemTabItem.java
zlamalp: Fixed in new commit.

PR #2251, opened 2019-06-07
Create a new way how to work with suspension of Member in the VO

stavamichal:
Reason: Member's status 'Suspend' is very unclear state and for this reason
 must be removed from the member's lifecycle in the future. Instead of
 status, there will be new information on object Member 'suspendedTo'
 with a Date which means to which Date will be specific Member considered
 as suspended in the VO (member has only 1 VO). This commit adds new
 funcionality, which will be used for replacing:

 - add new attribute 'suspendedTo' to object Member which is a Date to
 which member will be considered as suspended
 - add this attribute to toString method and also serializeToString
 method to respect existing auditer funcionality. The same change was
 made also in object RichMember
 - fix behavior of auditer parser to respect these changes
 - add new method 'suspendMemberTo' to all layers of MembersManager
 (Impl, BlImpl, Entry, Rpc), which allow to set suspension of member by
 Vo Managers of member's Vo
 - add these changes to all database, schema and changelog files
 - add simple test


perun-core/src/main/resources/hsqldbChangelog.txt
```diff
@@ -8,6 +8,10 @@
 
 -- this update is not supported on hsql since its used only as in-memory db
 
+3.1.54
+update configurations set value='3.1.54' where property='DATABASE VERSION';
```
zlamalp: Version update in sql changelog should be last command (all three files).

perun-base/src/main/java/cz/metacentrum/perun/core/api/Member.java
```diff
@@ -119,6 +122,20 @@ public void setSponsored(boolean sponsored) {
...
 
+	public Date getSuspendedTo() {
+		return suspendedTo;
+	}
+
+	public void setSuspendedTo(Date suspendedTo) { this.suspendedTo = suspendedTo; }
+
+	public boolean isSuspended() {
+		if(getSuspendedTo() != null) {
+			return suspendedTo.after(Calendar.getInstance().getTime());
```
zlamalp: We should probably use new Java Date API (LocalDateTime.now()) instead of Calendar.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1032,6 +1032,21 @@ public Member setStatus(PerunSession sess, Member member, Status status, String
 		return getMembersManagerBl().setStatus(sess, member, status, message);
 	}
 
+	@Override
+	public void suspendMemberTo(PerunSession sess, Member member, Date suspendedTo) throws InternalErrorException, MemberNotExistsException, PrivilegeException {
+		Utils.checkPerunSession(sess);
+		Utils.notNull(suspendedTo, "suspendedTo");
```
zlamalp: There is no way to "clear" suspended state otherwise than setting date in past. We should either allow null and properly clear db column or add method to un-suspend member.

perun-core/src/main/resources/hsqldbChangelog.txt
stavamichal: Order was fixed.

perun-base/src/main/java/cz/metacentrum/perun/core/api/Member.java
stavamichal: Of course, fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
stavamichal: Method "unsuspendMember" was added. Good point!

stavamichal:
Rebased. Now it should be alright.

Also "suspended" was added to mixins.

PR #2255, opened 2019-06-11
Disable custom FileSerializer for GraphText methods

Vojtech-Sassmann:
* The methods used for generating graphs and returning them as a String
now doesn't return the results as a file.

* Now for these methods, common serializers can be used.

* These changes are made so the output can be processed in frontend
applications.


Vojtech-Sassmann:
@zlamalp I have removed the serializer.

PR #2256, opened 2019-06-11
LDAPC-ADA: Use directly BL layer of perun-core

zlamalp:
- We should call BL layer directly, otherwise nested transaction
  is created for each call to perun-core (through API layer).
- Removed unused imports.


zlamalp:
Closing this one, replaced by #2261.

PR #2270, opened 2019-06-20
Added two new methods for listing roles

metodej:
- getUserRoles(int userId) returns all roles for a given user
- getGroupRoles(int groupId) returns all roles for a given group


metodej:
Thank you for review, the authorization for methods was added as requested.

PR #2277, opened 2019-06-25
CORE: Removed dependency on opencsv, we use jackson library instead

zlamalp:
- Jackson library for CSV is curated in spring-boot-starter-parent
  and by using it we don't have to manually updated version in
  dependencies, like we do for opencsv.
- Our ExtSourceCSV now uses jackson library to parse CSV files.
  Functionality is the same, just implementation changed.
  Basically we still expect CSV files to have first row as a header,
  then we read all lines and check each row if it matches query
  and if expected result limit was not passed.

### DIFF
--- a/perun-core/pom.xml
+++ b/perun-core/pom.xml
@@ -193,17 +193,6 @@
 		</dependency>
 
 		<dependency>
-			<groupId>com.opencsv</groupId>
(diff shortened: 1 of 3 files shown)


zlamalp:
Manually tested and works, docs updated on wiki.

PR #2278, opened 2019-06-25
remove old entries from LDAP during synchronization in ldapc

mvocu:
search and remove old (not in Perun anymore) entries 
wait for workers end before shutting down pool

### DIFF
--- a/perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunVOImpl.java
+++ b/perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunVOImpl.java
@@ -152,4 +152,18 @@ public class PerunVOImpl extends AbstractPerunEntry<Vo> implements PerunVO {
 				getNameMapper());
 	}
 
+	@Override
+	public void deleteEntry(Name dn) throws InternalErrorException {
+		// first find and remove entries in subtree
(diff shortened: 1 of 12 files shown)


mvocu:
Added removing all subentries of VO when removing VO, added the RE for Resource deletion audit message.

perun-ldapc-ada/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunVOImpl.java
```diff
@@ -144,4 +145,25 @@ public Name getEntryDN(String ...voId) {
...
+	public List<Name> listEntries() throws InternalErrorException {
+		return ldapTemplate.search(query().
+				where("objectclass").is(PerunAttribute.PerunAttributeNames.objectClassPerunVO),
+				getNameMapper());
+	}
+
+	@Override
+	public void deleteEntry(Name dn) throws InternalErrorException {
+		// first find and remove entries in subtree
+		List<Name> subentries = ldapTemplate.search(query()
```
zlamalp: This search returns also base itself, which can't be deleted. Either use different ``search()`` method with specified SearchScope or simply check returned ``Name`` on orignal ``dn`` param and skip it in the cycle below.

mvocu:
Excluded base entry from subentry list in the latest commit.

PR #2282, opened 2019-06-26
CORE: Fixed user resolving in module for user:virt:loa

zlamalp:
- When user was deleted, message about removed UserExtSources was
  processed by module user:virt:loa and logged at the end. But it is
  logically after user deletion and LDAPc fails on it, since it tries
  to update attribute of non-existing user.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_loa.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_loa.java
@@ -12,7 +12,6 @@ import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.UserExtSource;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_loa.java
```diff
@@ -78,18 +78,24 @@ public AttributeDefinition getAttributeDefinition() {
...
+				resolvingMessages.add(resolveEvent(sess, user));
+			} else if (message instanceof UserExtSourceUpdated) {
 				resolvingMessages.add(resolveEvent(sess, sess.getPerunBl().getUsersManagerBl().getUserById(
-						sess, ((UserExtSourceUpdated) message).getUserExtSource().getUserId())));
-			} catch (UserNotExistsException e) {
-				throw new ConsistencyErrorException("User associated with updated UserExtSource no longer exists while resolving virtual attribute value change.", e);
+							sess, ((UserExtSourceUpdated) message).getUserExtSource().getUserId())));
 			}
+		} catch (UserNotExistsException e) {
+			throw new ConsistencyErrorException("User "+user+" associated with "+message.getName()+" event no longer exists while resolving virtual attribute value change for LoA.", e);
```
stavamichal: We can ignore this situation, just log it instead of throwing an exception.

zlamalp: Fixed, now message is logged and no exception is thrown.

PR #2283, opened 2019-06-26
CORE,RPC: Fixed repetitive session initiation for Kerberos/BA

zlamalp:
- When session is constructed, we properly resolve actor and ext source
  name. But supplementary methods to check actor and ext source against
  current session retrieved wrong value, if certificate was provided
  to the server along with kerberos/ba authz.

  So generally, apache always fills properties with cert data, if
  cert is provided. We must then first check all other authentication
  options, then cert and then authz delegation.


zlamalp:
Authz delegation was tested on devel with old LDAPc and works.

PR #2286, opened 2019-07-01
Remove members method

Vojtech-Sassmann:
* Created new method allowing removal of multiple memebers at once.

* Created method in entry layer.


perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -58,6 +58,18 @@
 	 */
 	void deleteMember(PerunSession sess, Member member) throws InternalErrorException, MemberNotExistsException, PrivilegeException, MemberAlreadyRemovedException;
 
+	/**
+	 *  Delete given members.
```
stavamichal: Two spaces instead of one. And you should probably inform about ability of this method to removing members from different virtual organizations at once.

Vojtech-Sassmann: I have updated the Javadocs.

PR #2288, opened 2019-07-01
Remove status SUSPEND from member in core

stavamichal:
- status suspend was removed from object Status
 - sql with needed changes were added to changelogs for all databases
 - object Status.SUSPENDED was removed from all parts where it was used
 before (expect services which need to be changed separately), methods
 used for suspension this way were also removed
 - some logic was removed from registrar
 - some logic was removed from gui
 - new method "suspendMemberTo" with reason of suspension was created
 and used also in entry and RPC
 - new event for unsuspending member was added

WARNING: this part of change has impact on services and other
funcionality which is using status SUSPEND and should be merged only
together with all other changes in separate commits.


stavamichal:
Need to be merged together with https://github.com/CESNET/perun-services/pull/414

zlamalp (changes requested):
Please rebase this pull-request (db changelog and probably members mapper changes). Then we can review a merge it.

zlamalp:
It was replaced by #2848, closing this one.

PR #2291, opened 2019-07-03
Added method to get rich user with all attributes

metodej:
- added auxiliary method to get all (even with null value) attributes for user
- added auxiliary method to convert rich users to rich users with all attributes
- also added tests for these methods


balcirakpeter:
The functionality will not be used for now so I am closing the pull request, but we can reopen it in the future.

PR #2296, opened 2019-07-09
Update tools using member status SUSPENDED

zoraseb:


zlamalp:
Merge only after #2288

zlamalp:
Since this needs to be updated, we will solve this directly in #2848 and this pull-request can be closed.

PR #2297, opened 2019-07-09
New module to collect additionalIdentifiers for User

stavamichal:
- userExtSource can have list of additionalIdentifiers and we want to
 collect these identifiers from all user ext sources to one attribute
 (virtual user additionalIdentifiers)


zlamalp:
Just a note, if you with to have specific attribute displayName or description, just override ``getAttributeDefinition()`` otherwise generic description and friendlyName is used.

PR #2301, opened 2019-07-11
Adding methods to remove member(s) from group(s)

oernst9:
* Adding new methods to make removing of more members from one group and one member from more groups easier
* Added:
removeMember(PerunSession sess, Member member, List<Group> groups);
removeMembers(PerunSession sess, Group group, List<Member> members);

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -990,6 +990,41 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 	}
 
 	@Override
+	public void removeMembers(PerunSession sess, Group group, List<Member> members) throws InternalErrorException, NotGroupMemberException, GroupNotExistsException {
+		// Check if the group is NOT members or administrators group
+		if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
+			throw new InternalErrorException("Cannot remove member directly from the members group.");
+		} else {
+			Collections.sort(members);
+			for (Member member : members) {
+				try {
+					this.removeDirectMember(sess, group, member);
+				} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex){
+					throw new InternalErrorException(ex);
+				}
+			}
+		}
+	}
+
+
+	@Override
+	public void removeMember(PerunSession sess, List<Group> groups, Member member) throws InternalErrorException, NotGroupMemberException, GroupNotExistsException {
+		Collections.sort(groups, Collections.reverseOrder());
+		for (Group group : groups) {
+			// Check if the group is NOT members or administrators group
+			if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
+				throw new InternalErrorException("Cannot remove member directly from the members group.");
+			} else {
+				try {
+					this.removeDirectMember(sess, group, member);
+				} catch (WrongAttributeValueException | WrongReferenceAttributeValueException ex){
+					throw new InternalErrorException(ex);
+				}
+			}
+		}
+	}
+
+	@Override
 	public void removeMemberFromMembersOrAdministratorsGroup(PerunSession sess, Group group, Member member) throws InternalErrorException, NotGroupMemberException, GroupNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException {
 		// Check if the group IS memebers or administrators group
 		if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
@@ -8,11 +8,13 @@ import java.util.Map;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.MemberGroupStatus;
+import cz.metacentrum.perun.core.api.MembershipType;
 import cz.metacentrum.perun.core.api.RichGroup;
 import cz.metacentrum.perun.core.api.RichMember;
 import cz.metacentrum.perun.core.api.RichUser;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.exceptions.NotGroupMemberException;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.rpc.ApiCaller;
 import cz.metacentrum.perun.rpc.ManagerMethod;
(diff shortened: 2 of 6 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -314,21 +314,61 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 			ac.stateChangingCheck();
-
-			ac.getGroupsManager().removeMember(ac.getSession(),
+			if (parms.contains("group")) {
+				ac.getGroupsManager().removeMember(ac.getSession(),
 					ac.getGroupById(parms.readInt("group")),
 					ac.getMemberById(parms.readInt("member")));
+			} else {
+				List<Integer> groupsInts = parms.readList("groups", Integer.class);
```
stavamichal: The same problem like with list of members. Also there is a new problem with situation where "groups" parameter is also missing. You can prevent it by 3 branches of if clause where the third will throw an RpcException.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -989,6 +989,42 @@ public void removeMember(PerunSession sess, Group group, Member member) throws I
 		}
 	}
 
+	@Override
+	public void removeMembers(PerunSession sess, Group group, List<Member> members) throws InternalErrorException, NotGroupMemberException, GroupNotExistsException {
+		// Check if the group is NOT members or administrators group
+		if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
```
stavamichal: It is not important here, but for the future work it is safer to use equals on non-null subject first like:
`VosManager.MEMBERS_GROUP.equals(group.getName())`
or use Object.equals() method to prevent NullPointerException. But in this specific case, group should be always filled with not-null name.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -989,6 +989,42 @@ public void removeMember(PerunSession sess, Group group, Member member) throws I
...
+				}
+			}
+		}
+	}
+
+
+	@Override
+	public void removeMember(PerunSession sess, List<Group> groups, Member member) throws InternalErrorException, NotGroupMemberException, GroupNotExistsException {
+		Collections.sort(groups);
+		Collections.reverse(groups);
```
stavamichal: I think this code is more effective. 
`Collections.sort(groups, Collections.reverseOrder());`

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -314,21 +314,87 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+				List<Group> groups = new ArrayList<>();
+				for (Integer groupInt : groupsInts) {
+					groups.add(ac.getGroupById(groupInt));
+				}
+				ac.getGroupsManager().removeMember(ac.getSession(),
+					ac.getMemberById(parms.readInt("member")),
+					groups);
+			} else {
+				throw new RpcException(RpcException.Type.MISSING_VALUE, "parameter 'groups' not sent");
```
stavamichal: One of parameters 'group' or 'groups'.

stavamichal:
Please create a new pull request from the second commit. You should cherry pick it to the new branch and remove it from this one. Thank you.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 */
 	removeMember {
-
 		@Override
 		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 			ac.stateChangingCheck();
-
-			ac.getGroupsManager().removeMember(ac.getSession(),
+			Member member = ac.getMemberById(parms.readInt("member"));
+			if (!MembershipType.DIRECT.equals(member.getMembershipType())) {
```
zlamalp: You don't want to check ``MembershipType.DIRECT`` on result of ``getMemberById()`` since its retrieved in the context of VO (not group) and this property is set to ``NOT_DEFINED``. Like this you would silently skip removal of single member from single group. We don't want to modify behavior of existing method.

Simply keep previous implementation:
```java
ac.getGroupsManager().removeMember(ac.getSession(), ac.getGroupById(parms.readInt("group")), ac.getMemberById(parms.readInt("member")));
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			if (parms.contains("group")) {
+				ac.getGroupsManager().removeMember(ac.getSession(),
 					ac.getGroupById(parms.readInt("group")),
-					ac.getMemberById(parms.readInt("member")));
+					member);
+			} else if (parms.contains("groups")) {
+				List<Integer> groupsInts = parms.readList("groups", Integer.class);
+				List<Group> groups = new ArrayList<>();
+				for (Integer groupInt : groupsInts) {
+					groups.add(ac.getGroupById(groupInt));
```
zlamalp: Here you should check, if member is not ``DIRECT`` in each group. If yes, then remove such group from the list. You can use ``ac.getGroupsManager().getGroupMemberById()`` as you use in ``removeMembers()``.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @throw GroupNotExistsException When the group doesn't exist
+	 * @throw WrongAttributeAssignmentException When assigning atribute to wrong entity
+	 * @throw AttributeNotExistsException When attribute doesn't exist
+	 * @throw ExternallyManagedException When the group is externally managed
+	 *
 	 * @param group int Group <code>id</code>
 	 * @param member int Member <code>id</code>
 	 */
+	/*#
+	 * Removes a member from groups.
```
zlamalp: You should note specific behavior of this method -> if member is not in the group or is indirect, it is skipped, but the rest of groups are processed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @param group int Group <code>id</code>
+	 * @param members List<Integer> Members - can be empty <code>id</code>
+	 */
+	removeMembers {
+		@Override
+		public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
+			ac.stateChangingCheck();
+			List<Integer> memberInts = parms.readList("members", Integer.class);
+			if (memberInts == null) {
+				throw new RpcException(RpcException.Type.MISSING_VALUE, "parameter 'members' not sent");
```
zlamalp: As @stavamichal mentioned, result can be null, when caller sends param, but with the null value, instead of empty list. In such case ``"parameter 'members' not sent"`` is not exactly right. You should state, that we expect non-empty list. In case of list is empty, no member is removed from the group. This can be stated in the javadoc too.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
-					ac.getGroupById(parms.readInt("group")),
-					ac.getMemberById(parms.readInt("member")));
+				ac.getGroupById(parms.readInt("group")),
+				ac.getMemberById(parms.readInt("member")));
 			return null;
 		}
 	},
 
+	/*#
+	 * Removes members from a group.
```
zlamalp: You should note specific behavior of this method.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @throw WrongAttributeAssignmentException When assigning atribute to wrong entity
+	 * @throw AttributeNotExistsException When attribute doesn't exist
+	 * @throw ExternallyManagedException When the group is externally managed
+	 *
+	 * @param group int Group <code>id</code>
+	 * @param members List<Integer> Members - can be empty <code>id</code>
+	 */
+	removeMembers {
+		@Override
+		public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
```
zlamalp: Return type should by ``Void``, since you always return null.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -308,27 +310,100 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			ac.stateChangingCheck();
+			List<Integer> memberInts = parms.readList("members", Integer.class);
+			if (memberInts == null) {
+				throw new RpcException(RpcException.Type.MISSING_VALUE, "parameter 'members' not sent");
+			}
+			Group group = ac.getGroupById(parms.readInt("group"));
+			List<Member> members = new ArrayList<>();
+			for (Integer memberInt : memberInts) {
+				members.add(ac.getGroupsManager().getGroupMemberById(ac.getSession(), group, memberInt));
```
zlamalp: You should try/catch this call and silently ignore ``MemberNotExistsException`` and ``NotGroupMemberException``.

stavamichal:
To javadoc, you should also add information, that method skips members if they are not found in the group without any warning.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -314,21 +316,108 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			Group group = ac.getGroupById(parms.readInt("group"));
+			List<Member> members = new ArrayList<>();
+			for (Integer memberInt : memberInts) {
+				try {
+					Member member = ac.getGroupsManager().getGroupMemberById(ac.getSession(), group, memberInt);
+					if (MembershipType.DIRECT.equals(member.getMembershipType())) {
+						members.add(member);
+					}
+				} catch(NotGroupMemberException e) {
+					//just skipped
```
stavamichal: Please, write also the reason for skipping there "skipped because user is not member of this group so we don't need to remove him".

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -314,21 +316,108 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+				List<Group> groups = new ArrayList<>();
+				for (Integer groupInt : groupsInts) {
+					try {
+						Group group = ac.getGroupById(groupInt);
+						Member member = ac.getGroupsManager().getGroupMemberById(ac.getSession(), group, parms.readInt("member"));
+						if (MembershipType.DIRECT.equals(member.getMembershipType())) {
+							groups.add(group);
+						}
+					} catch(NotGroupMemberException e) {
+						//just skipped
```
stavamichal: Please, write also the reason for skipping there "skipped because user is not member of this group so we don't need to remove him".

oernst9: done

PR #2310, opened 2019-07-16
Adding thrown exceptions to javadoc

oernst9:
* Added aformentioned information to the javadoc in GroupsManagerMethod


stavamichal:
What exactly was the change mentioned as "Changed message thrown by RpcException in GroupsManagerMethod" in commit message?

zlamalp:
> What exactly was the change mentioned as "Changed message thrown by RpcException in GroupsManagerMethod" in commit message?

You are right. After cherry picking that part of change was discarded and message can be removed from the commit. We must check, if correct code (including this change) is part of original pull-request.

PR #2315, opened 2019-07-17
Completing documentation for create* methods

oernst9:
* added parameters, which are required, ignored or optional

* Category and PerunNotifReceiver can be now created using just attributes

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/CabinetManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/CabinetManagerMethod.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Set;
 
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
+import cz.metacentrum.perun.core.api.exceptions.RpcException;
 import cz.metacentrum.perun.rpc.ApiCaller;
 import cz.metacentrum.perun.rpc.ManagerMethod;
 import cz.metacentrum.perun.rpc.deserializer.Deserializer;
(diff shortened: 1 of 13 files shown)


zlamalp:
Please rebase it to current master, then I will review it. Thank you.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -40,6 +42,8 @@
 	 * Creates a new group in the specific VO defined by object vo in parameter.
```
zlamalp: Please update docs for this. We should state, that this combination creates only top-level group in VO.
Also in order to work ``parentGroupId`` param of ``group`` must not be present or must be set to ``null``.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -24,6 +24,8 @@
 	/*#
 	 * Creates a subgroup of a group.
 	 *
+	 * Group object must contain name. Description and parentGroupId are optional. Other parameters are ignored.
```
zlamalp: We should state, that ``parentGroupId`` param of ``group`` object is ignored in this case, since we use ``parentGroup`` param of the method itself to resolve actual parent group.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/NotificationManagerMethod.java
```diff
@@ -49,15 +50,37 @@ public PerunNotifReceiver call(ApiCaller ac, Deserializer parms) throws PerunExc
...
 	 *
+	 * PerunNotifReceiver must contain: target, locale, templateId, key. Parameters target, locale, key are optional.
+	 *
 	 * @param receiver PerunNotifReceiver PerunNotifReceiver object without <code>id</code>
 	 * @return PerunNotifReceiver PerunNotifReceiver with new <code>id</code> set
 	 */
+	/*#
+	 * Saves PerunNotifReceiver to db and creates <code>id</code>.
+	 *
+	 * @param target String Defines target of receiver, usually contains function to get email or jabber number...
```
zlamalp: Jaber is no longer supported in perun. Please remove it from the javadoc.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/NotificationManagerMethod.java
```diff
@@ -49,15 +50,37 @@ public PerunNotifReceiver call(ApiCaller ac, Deserializer parms) throws PerunExc
...
+	 * @param locale String Locale of receiver - determines a language, which is used for sending messages
+	 * @param templateId int Template Id to which receiver is connected
+	 * @param key String Type of the receiver - 'EMAIL_USER' or 'EMAIL_GROUP'
```
zlamalp: It is weird to use word ``key`` as param name for type of receiver. Please change it to ``type`` or something like this.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/NotificationManagerMethod.java
```diff
@@ -49,15 +50,37 @@ public PerunNotifReceiver call(ApiCaller ac, Deserializer parms) throws PerunExc
 	/*#
 	 * Saves PerunNotifReceiver to db and creates <code>id</code>.
 	 *
+	 * PerunNotifReceiver must contain: target, locale, templateId, key. Parameters target, locale, key are optional.
```
zlamalp: Same here, replace ``key`` with ``type``, since it is a type of receiver.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/CabinetManagerMethod.java
```diff
@@ -86,13 +87,33 @@ public PublicationSystem call(ApiCaller ac, Deserializer parms) throws PerunExce
 	/*#
 	 * Creates new Category for Publications with specified name and rank.
 	 *
-	 * @param category Category new Category object
+	 * Category object's parameters name and rank are optional. Other parameters are ignored.
```
zlamalp: While it is not required on server side API, we always submit Category object with non empty name (limited to 128 chars) and non-empty rank value (double with single digit after decimal point). Please update the docs in this manner, we will later update server side checks.

zlamalp:
Also please rebase this pull-request, so it can be later merged. Thank you.

PR #2319, opened 2019-07-18
Adding methods to add member(s) to the group(s)

oernst9:
* Adding new methods to make adding of more members to one group and one member to more groups easier

* Added:
addMember(PerunSession sess, List<Group> groups, Member member);
addMembers(PerunSession sess, Group group, List<Member> members);

* Added tests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
@@ -292,6 +292,79 @@ public class GroupsManagerEntry implements GroupsManager {
 	}
 
 	@Override
+	public void addMembers(PerunSession sess, Group group, List<Member> members) throws InternalErrorException, MemberNotExistsException, PrivilegeException, AlreadyMemberException, GroupNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternallyManagedException {
+		Utils.checkPerunSession(sess);
+		getGroupsManagerBl().checkGroupExists(sess, group);
+
+		for (Member member : members) {
+			getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
+			// Check if the member and group are from the same VO
+			if (member.getVoId() != (group.getVoId())) {
+				throw new MembershipMismatchException("Member and group are form the different VO");
+			}
+		}
+
+		// Authorization
+		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, group)
+			&& !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
+			throw new PrivilegeException(sess, "addMembers");
+		}
+
+
+
+		// Check if the group is externally synchronized
+		Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+		if ("true".equals(attrSynchronizeEnabled.getValue()) || getGroupsManagerBl().isGroupInStructureSynchronizationTree(sess, group)) {
+			throw new ExternallyManagedException("Adding of member is not allowed. Group is externally managed.");
+		}
+
+		getGroupsManagerBl().addMembers(sess, group, members);
+	}
+
+	@Override
+	public void addMember(PerunSession sess, List<Group> groups, Member member) throws InternalErrorException, MemberNotExistsException, PrivilegeException, AlreadyMemberException, GroupNotExistsException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternallyManagedException {
+		Utils.checkPerunSession(sess);
+		getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
+		for (Group group : groups) {
+			getGroupsManagerBl().checkGroupExists(sess, group);
+			// Authorization
+			if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, group)
+				&& !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
+				throw new PrivilegeException(sess, "addMember");
+			}
+			// Check if the member and group are from the same VO
+			if (member.getVoId() != (group.getVoId())) {
+				throw new MembershipMismatchException("Member and group are form the different VO");
+			}
+			// Check if the group is externally synchronized
+			Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+			if ("true".equals(attrSynchronizeEnabled.getValue()) || getGroupsManagerBl().isGroupInStructureSynchronizationTree(sess, group)) {
+				throw new ExternallyManagedException("Adding of member is not allowed. Group is externally managed.");
+			}
+		}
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -291,6 +291,69 @@ public Group getGroupByName(PerunSession sess, Vo vo, String name) throws GroupN
...
+			// Check if the member and group are from the same VO
+			if (member.getVoId() != (group.getVoId())) {
+				throw new MembershipMismatchException("Member and group are form the different VO");
+			}
+		}
+
+		// Authorization
+		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, group)
+			&& !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, group)) {
+			throw new PrivilegeException(sess, "addMember");
```
zlamalp: Should be "addMembers" in passed param to reflect method name.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -291,6 +291,69 @@ public Group getGroupByName(PerunSession sess, Vo vo, String name) throws GroupN
...
+				throw new MembershipMismatchException("Member and group are form the different VO");
+			}
+			// Check if the group is externally synchronized
+			Attribute attrSynchronizeEnabled = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GROUPSYNCHROENABLED_ATTRNAME);
+			if ("true".equals(attrSynchronizeEnabled.getValue()) || getGroupsManagerBl().isGroupInStructureSynchronizationTree(sess, group)) {
+				throw new ExternallyManagedException("Adding of member is not allowed. Group is externally managed.");
+			}
+		}
+
+
```
zlamalp: I think one empty line as divider between for cycle and next method call is sufficient.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -3483,6 +3483,51 @@ public void addAndRemoveMemberInNonSynchronizedGroup() throws Exception {
...
+
+		vo = setUpVo();
+
+		groupsManagerBl.createGroup(sess, vo, group);
+		groupsManagerBl.createGroup(sess, vo, group2);
+		groupsManagerBl.createGroup(sess, vo, group3);
+
+		List<Group> groups = Arrays.asList(group, group2, group3);
+
+		Attribute synchroAttr = new Attribute(perun.getAttributesManager().getAttributeDefinition(sess, GroupsManager.GROUPSYNCHROENABLED_ATTRNAME));
```
zlamalp: I don't think you need to test "false" case - when synchronization is disabled. We do tests this logic in other tests.

Basically you should have tests for each case - group/member not from same vo, synchronized group, not synchronized group and lastly added logic about group structure synchronization.

But I think it is basically covered in other tests. Please check if there are tests for group structure synchronization case. This could be added for standard add/removeMember() tests.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -3483,6 +3483,51 @@ public void addAndRemoveMemberInNonSynchronizedGroup() throws Exception {
...
+
+
+	@Test
+	public void addMembersToGroup() throws Exception {
+		System.out.println(CLASS_NAME + "addMembersToGroup");
+
+		vo = setUpVo();
+		setUpGroup(vo);
+
+		Attribute synchroAttr = new Attribute(perun.getAttributesManager().getAttributeDefinition(sess, GroupsManager.GROUPSYNCHROENABLED_ATTRNAME));
```
zlamalp: Same as in previous test. Method logic (and test) should work even if attribute is not set at all.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -295,21 +297,94 @@ public Group call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+
+	/*#
+	 * Adds members to a group. If already a member of the group, the member will be skipped.
+	 * Non-empty list of members expected, if empty, no member will be added.
+	 *
+	 * @param groups List<Group> List of groups that the member should be added to <code>id</code>
```
zlamalp: This docs seem to be a copy of ``addMember()`` including params, which are wrong for ``addMembers()``. Param ``groups`` should be a ``group`` and ``member`` should be ``members`` in this case

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -295,21 +297,94 @@ public Group call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+
+			List<Integer> memberInts = parms.readList("members", Integer.class);
+			if (memberInts == null) {
+				throw new RpcException(RpcException.Type.MISSING_VALUE, "Non-empty list of members not sent.");
+			}
+			Group group = ac.getGroupById(parms.readInt("group"));
+			List<Member> members = new ArrayList<>();
+			for (Integer memberInt : memberInts) {
+				Member member = ac.getMemberById(memberInt);
+				if (!ac.getGroupsManager().isGroupMember(ac.getSession(), group, member)) {
```
zlamalp: Maybe you could use ``isDirectGroupMember()`` which is currently present only in impl/dao layer (``GroupsManagerImplApi``). Just add such method to the BL and API (entry) layers and then use it here.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -295,21 +297,94 @@ public Group call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+
+
 	/*#
 	 * Adds a member to a group.
 	 *
 	 * @param group int Group <code>id</code>
 	 * @param member int Member <code>id</code>
 	 */
+	/*#
+	 * Adds a member to a group. If already a member of a group, the group will be skipped.
```
zlamalp: Adds a member to a group. -> Adds a member to the groups.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -295,21 +297,94 @@ public Group call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			} else if (parms.contains("groups")) {
+				List<Integer> groupsInts = parms.readList("groups", Integer.class);
+				if (groupsInts == null) {
+					throw new RpcException(RpcException.Type.MISSING_VALUE, "Non-empty list of groups not sent.");
+				}
+				List<Group> groups = new ArrayList<>();
+				Member member = ac.getMemberById(parms.readInt("member"));
+				for (Integer groupInt : groupsInts) {
+					Group group = ac.getGroupById(groupInt);
+					if (!ac.getGroupsManager().isGroupMember(ac.getSession(), group, member)) {
```
zlamalp: As in previous method, we could save up select by using ``isDirectGroupMember``.

PR #2321, opened 2019-07-19
CORE: Optimized IDs retrieval from auditer log sequence on batchUpdate()

zlamalp:
- We previously retrieved ID for each auditer message by query to the
  DB while constructing batch insert sql.
- Now we pass this nextval function call as a part of batch insert sql
  itself, resulting in a quick single batch insert query.
- Added method Compatiblity.getSequenceNextval(seqName) to return proper
  SQL for each supported DB engine.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
@@ -315,32 +315,23 @@ public class Auditer {
 
 			//Write all messages to the database
 			try {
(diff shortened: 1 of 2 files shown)


zlamalp:
Tested on Postgres and HSQLDB (by tests). We should test on Oracle too. Performance was not measured, we should test this on MU with real big data.

zlamalp:
Tested manually on postgresql. It cuts the time of batch insert of 60.000+ messages from 1 minute to 12 seconds. We need to test if Oracle syntax is ok too, then it can be merged.

zlamalp:
SQL manually tested on Oracle / idm-test.

PR #2322, opened 2019-07-19
CORE: Do not lock resolving audit messages by attribute modules

zlamalp:
- We previously "locked" all transactions before the first one resolved
  any additional audit messages (by attribute module).
  If there was a lot of messages, it could take a lot of time (resolving
  many virtual attributes value changes).
- By moving the lock after additional message resolving we allow other
  transactions to resolve messages on their own and only batch insert
  of messages is now locked to make sure our messages makes sense
  to the reader (like LDAPc).

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
@@ -286,20 +286,20 @@ public class Auditer {
 			return;
 		}
 
+		// Resolve all additional message from attribute modules and add them to the bulk
+		try {
+
+			//Get perun session from the first message (all sessions should be same from the same principal)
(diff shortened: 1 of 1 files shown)


zlamalp:
This change could release the lockup of Perun caused by big transactions, where most of the time is spent on resolving virtual attribute changes. This change is only possible, if we are sure, that call to ``resolveVirtualAttributeValueChange()`` doesn't have race condition in modules themselves and between transactions. Which I'm not currently sure - for example: two transactions adding same member to the different groups and resolving all group names for such member. They could now see only own changes (or already commited during the process) and commit such message with missing value from other transaction. I'm not sure, if this is not currently true too.

zlamalp:
Since we decided to prefer integrity over speed, this pull-request doesn't conform this and can be closed.

PR #2324, opened 2019-07-23
Optimalization of Attribute module user:virt:groupNames

oernst9:
* reducing number of queries for getting names of VOs and groups to one

* mocked tests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
@@ -13,19 +13,28 @@ import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.MemberGroupStatus;
+import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.VosManager;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleAbstract;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleImplApi;
+import org.springframework.dao.EmptyResultDataAccessException;
+import org.springframework.jdbc.core.RowMapper;
 
+import java.sql.ResultSet;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.TreeSet;
 
 /**
  * Contains all group names of the user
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -37,25 +40,32 @@
...
+			" groups_members.source_group_status = ?" +
+			" AND members.user_id = ?", MemberGroupStatus.VALID.getCode(), user.getId());
 
-		List<Member> members = sess.getPerunBl().getMembersManagerBl().getMembersByUser(sess, user);
 		Set<String> voNames = new HashSet<>();
-		for (Member member : members) {
-			Vo vo = sess.getPerunBl().getMembersManagerBl().getMemberVo(sess, member);
-			voNames.add(vo.getShortName());
+		for (Map<String, Object> one : names) {
+			if (!VosManager.MEMBERS_GROUP.equals(one.get("group_name"))) {
```
zlamalp: Move this IF condition after resolving and storing VO name. Like this you are skipping names for vos, where user is only in "members" group. In previous implementation they were included in the resulting list. Just not like ``voShortName:members``, but ``voShortName``.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -37,25 +40,32 @@
...
 		Attribute attribute = new Attribute(attributeDefinition);
 		List<String> groupNames = new ArrayList<>();
+		List<Map<String, Object>> names = sess.getPerunBl().getDatabaseManagerBl().getJdbcPerunTemplate().queryForList("SELECT" +
+			" DISTINCT vos.short_name AS vo_short_name, groups.name AS group_name" +
+			" FROM" +
+			" members" +
+			" JOIN vos ON vos.id = members.vo_id" +
+			" JOIN groups_members ON groups_members.member_id = members.id" +
+			" JOIN groups ON groups_members.group_id = groups.id" +
+			" WHERE" +
```
zlamalp: ``WHERE`` clause is resolved after all joins on tables are performed. This can cause performance issue, since DB will first join whole (possibly large) tables and then perform limitation, discarding most data. While today DBs are optimized for this, it's still better to include limitation as a part of ``JOIN`` caluse.

You can put ``members.user_id = ?`` condition as a part of join with ``vos`` table using ``AND`` operator. Like this: ``JOIN vos ON vos.id = members.vo_id AND members.user_id = ?``.

Then do the same with ``groups_members.source_group_status = ?`` and join on ``groups_members`` table.

After this, fix order of params passed to queryForList(), since they will be switched by above change.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -37,25 +40,32 @@
...
+			" AND members.user_id = ?", MemberGroupStatus.VALID.getCode(), user.getId());
 
-		List<Member> members = sess.getPerunBl().getMembersManagerBl().getMembersByUser(sess, user);
 		Set<String> voNames = new HashSet<>();
-		for (Member member : members) {
-			Vo vo = sess.getPerunBl().getMembersManagerBl().getMemberVo(sess, member);
-			voNames.add(vo.getShortName());
+		for (Map<String, Object> one : names) {
+			if (!VosManager.MEMBERS_GROUP.equals(one.get("group_name"))) {
+				String voShortName = (String) one.get("vo_short_name");
```
zlamalp: You can add retrieved vo names already into ``groupsNames`` variable, since its still a list of strings and no specific order is required.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -37,25 +40,32 @@
...
+		List<Map<String, Object>> names = sess.getPerunBl().getDatabaseManagerBl().getJdbcPerunTemplate().queryForList("SELECT" +
+			" DISTINCT vos.short_name AS vo_short_name, groups.name AS group_name" +
+			" FROM" +
+			" members" +
+			" JOIN vos ON vos.id = members.vo_id" +
+			" JOIN groups_members ON groups_members.member_id = members.id" +
+			" JOIN groups ON groups_members.group_id = groups.id" +
+			" WHERE" +
+			" groups_members.source_group_status = ?" +
+			" AND members.user_id = ?", MemberGroupStatus.VALID.getCode(), user.getId());
```
zlamalp: Please add sorting `` order by vo_short_name, group_name;``. It is not required by any logic, but sometimes the value is displayed to users and alphabetical sort is better. Otherwise same item of the list can be found anywhere, when we ask for value twice (while the list in fact hasn't changed).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -37,25 +40,28 @@
...
-		List<Member> members = sess.getPerunBl().getMembersManagerBl().getMembersByUser(sess, user);
-		Set<String> voNames = new HashSet<>();
-		for (Member member : members) {
-			Vo vo = sess.getPerunBl().getMembersManagerBl().getMemberVo(sess, member);
-			voNames.add(vo.getShortName());
-
-			List<Group> groups = sess.getPerunBl().getGroupsManagerBl().getGroupsWhereMemberIsActive(sess, member);
-			for (Group group : groups) {
-				groupNames.add(vo.getShortName() +":"+ group.getName());
+		List<Map<String, Object>> names = sess.getPerunBl().getDatabaseManagerBl().getJdbcPerunTemplate().queryForList(
```
zlamalp: Sorry I didn't realize before. You can use our api object ``Pair<A,B>`` with both types specified as ``String`` and perform standard ``query()`` with passed ``RowMapper``. See ``RegistrarManagerImpl`` class in ``perun-registrar-lib`` module, around line 1302.

And we should wrap the whole query part in try/catch to handle ``EmptyResultDataAccessException`` - means no row selected, we can return empty list as attribute value in this case. Other exceptions should be catched, logged as error and wrapped in ``InternalErrorException``

Thank you.

oernst9: I've done what you requested and used `Pair<>` and ` query()` but because of that change, the test started to fail. It seems like it's ignoring the result of `when(...).thenReturn(...)` even though the call inside ` when()` is the same as the one that is used. I can amend the code and then fix it if you know what the issue is.

zlamalp: I'll look into it and let you know. If it won't work, we can use your original solution, since it basically works.

zlamalp: It works for me in the test. Catch is, that you need to have "same" parameters when you mock the call as in module. In this case you need static instance of RowMapper in the module and use is also in the test.

In module:
```java
protected static final RowMapper<Pair<String, String>> ROW_MAPPER = new RowMapper<Pair<String, String>>() {
		@Override
		public Pair<String, String> mapRow(ResultSet rs, int rowNum) throws SQLException {
			String voShortName = rs.getString("vo_short_name");
			String groupName = rs.getString("group_name");
			return new Pair<>(voShortName, groupName);
		}
	};
```

In test:
```java
		when(session.getPerunBl().getDatabaseManagerBl().getJdbcPerunTemplate().query("SELECT" +
			" DISTINCT vos.short_name AS vo_short_name, groups.name AS group_name" +
			" FROM" +
			" members" +
			" JOIN vos ON vos.id = members.vo_id AND members.user_id = ?" +
			" JOIN groups_members ON groups_members.member_id = members.id AND groups_members.source_group_status = ?" +
			" JOIN groups ON groups_members.group_id = groups.id" +
			" ORDER BY vo_short_name, group_name", urn_perun_user_attribute_def_virt_groupNames.ROW_MAPPER,
				user.getId(), MemberGroupStatus.VALID.getCode())).thenReturn(dbResult);
```

oernst9: Thanks, done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_groupNames.java
```diff
@@ -37,25 +45,46 @@
...
-		for (Member member : members) {
-			Vo vo = sess.getPerunBl().getMembersManagerBl().getMemberVo(sess, member);
-			voNames.add(vo.getShortName());
-
-			List<Group> groups = sess.getPerunBl().getGroupsManagerBl().getGroupsWhereMemberIsActive(sess, member);
-			for (Group group : groups) {
-				groupNames.add(vo.getShortName() +":"+ group.getName());
+		for (Pair<String, String> one : names) {
+			String voShortName = one.getLeft();
+			groupNames.add(voShortName);
```
zlamalp: This will cause duplicate entries for VO for each non-members group in it. We can use TreeSet to keep unique and sorted values:  ``Set<String> groupNames = new TreeSet<>();``. At the end we will push it to attribute value as list like this: ``attribute.setValue(new ArrayList(groupNames));``

oernst9: done

PR #2329, opened 2019-07-24
Removing attributes minGID and maxGID

oernst9:
* Removed files:
- entityless_def_namespace_maxGID
- entityless_def_namespace_minGID
- facility_virt_maxGID
- facility_virt_minGID

* Removed dependencies of minGID and maxGID (not entirely sure whether correctly)

* Removed from database in tests

* grep didn't find any occurence of 'minGID' or 'maxGID' in https://github.com/CESNET/perun-services.git gen/ folder


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_unixGID_namespace.java
```diff
@@ -233,8 +233,6 @@ public void changedAttributeHook(PerunSessionImpl session, Resource resource, At
 	@Override
 	public List<String> getDependencies() {
 		List<String> dependencies = new ArrayList<>();
-		dependencies.add(AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":namespace-minGID");
-		dependencies.add(AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":namespace-maxGID");
 		//Disallowed because of crosschecks between modules and performance reason
```
zlamalp: I just wonder, why ``group:def:unixGID-namespace:*`` have 2 facility attributes as dependencies, but same attribute for resource ``resource:def:unixGID-namespace:*`` doesn't. @licehammer, @stavamichal ? Commented code for performance reasons doesn't have it either (but maybe orginaly checked them transitively through dependency on group version of attribute).

stavamichal: I think this is just a second level of protection. I would preserve it without any change.

PR #2334, opened 2019-07-30
LDAPc Bug fix: when processing audit message to group creation, VO was also created

xpavlic:
When processing audit message to group creation, LDAPc is also trying to create VO in which group was created.

I added RE that checks to see if it is a given audit message. Without it, the VO bean was also added to the list of beans sent for creation.


perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
```diff
@@ -308,9 +308,17 @@ protected MessageBeans resolveMessage(String msg, Integer idOfMessage) throws In
 
 		//Debug information to check parsing of message.
 		MessageBeans beans = new MessageBeansImpl();
+
+		boolean shouldntAddVo = false;
+		if (msg.matches("Group:[\\d\\D]*created in Vo:[\\d\\D]*"))
+			shouldntAddVo = true;
```
licehammer: Maybe I'm old-school, but I would write it in a following way:
```
boolean suppressAddingVo = msg.matches("Group:[\\d\\D]*created in Vo:[\\d\\D]*");
```

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
```diff
@@ -308,9 +308,17 @@ protected MessageBeans resolveMessage(String msg, Integer idOfMessage) throws In
...
+
 		if(!listOfBeans.isEmpty()){
 			int i=0;
 			for(PerunBean p: listOfBeans) {
+				if (shouldntAddVo && p.getBeanName().equals("Vo"))
+					continue;
```
licehammer: I recommend using curly brackets { } or keep it on one line. It's easier to read.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
```diff
@@ -308,14 +308,21 @@ protected MessageBeans resolveMessage(String msg, Integer idOfMessage) throws In
 
 		//Debug information to check parsing of message.
 		MessageBeans beans = new MessageBeansImpl();
+
+		boolean shouldntAddVo = msg.matches("[\\d\\D]*Group:[\\d\\D]*created in Vo:[\\d\\D]*");
```
zlamalp: I don't think we need ``[\\d\\D]*`` at the start, since relevant messages about groups always start with ``Group ...``.

Generally, I would prefer, if we keep consistent regex usage. If you look at the code, we match strings using ``Pattern`` and ``Matcher`` with modifier ``Pattern.DOTALL``, so you can use ``(.*)`` to match all content of the group/vo object including newlines (which can be in group description).

PR #2335, opened 2019-07-30
Searching in MU ExtSource using non-numeric string issue resolved

oernst9:
* added condition in two methods 'findCandidate'

* if searched by non-numeric string in MU then IllegalArgumentException is thrown


stavamichal:
We don't need this change any more. Was resolved by change in SQL query. I am closing this PR.

PR #2336, opened 2019-07-31
Utils: Added timeout for kerberos password manager

zlamalp:
- We will wait 15 seconds for the KDC response and kill the process
  after another 15 seconds. If timeout occur, we exist script with
  error code 124.
- Use $(...) notation instead of legacy backticks. This prevents possible
  issues with escaping and nesting of commands. Full rationale at:
  https://github.com/koalaman/shellcheck/wiki/SC2006

### DIFF
--- a/perun-utils/password-manager/perun.passwordManager
+++ b/perun-utils/password-manager/perun.passwordManager
@@ -57,7 +57,7 @@ function change() {
 		RET=$(timeout -k $TIMEOUT_KILL $TIMEOUT /usr/bin/kadmin.heimdal --principal=$ADMINPRINCIPAL --keytab=$KEYTAB --realm=$REALM passwd --password="${PASSWORD}" ${USERLOGIN}@${REALM} 2>&1)
 		RET_CODE=$?
 		if [ $RET_CODE -eq 124 ]; then
-			logger "Perun-PasswordChange: setting new password timed out for $LOGINNAMESPACE:$USERLOGIN in $REALM (Warning: this error can mask original error 124 from KDC!)"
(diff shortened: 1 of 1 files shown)


zlamalp:
I'm unsure about ``timeout`` command placement in more complex commands. Should it be at start (as it is now) or after first command which pipes output to the ``kadmin`` command (like ``prinf "whatever" | timeout 123 kadmin params``) ?

zlamalp:
kadmin commands were updated in a following manner: ``prinf "whatever" | timeout 123 kadmin params``, so the timeout always apply to the kadmin command and return code is correct (command after pipe - timeout).

perun-utils/password-manager/perun.passwordManager
```diff
@@ -45,15 +54,25 @@ function change() {
 	case $TYPE in
 	KERBEROS_HEIMDAL)
 		# Change password
-		RET=`/usr/bin/kadmin.heimdal --principal=$ADMINPRINCIPAL --keytab=$KEYTAB --realm=$REALM passwd --password="${PASSWORD}" ${USERLOGIN}@${REALM} 2>&1`
-		if [ $? -ne 0 ]; then
+		RET=$(timeout -k $TIMEOUT_KILL $TIMEOUT /usr/bin/kadmin.heimdal --principal=$ADMINPRINCIPAL --keytab=$KEYTAB --realm=$REALM passwd --password="${PASSWORD}" ${USERLOGIN}@${REALM} 2>&1)
+		RET_CODE=$?
+		if [ $RET_CODE -eq 124 ]; then
+			logger "Perun-PasswordChange: setting new password timed out for $LOGINNAMESPACE:$USERLOGIN in $REALM (Warning: this error can mask original error 124 from KDC!)"
```
licehammer: It would be nice to log content of $RET too. Just in case, kadmin command ends with return code 124 (without timeout).

zlamalp: Fixed with next commit.

PR #2337, opened 2019-07-31
LDAPc Bug fix: when processing audit message to group creation, VO was also created, validating member in VO instead of in Group

xpavlic:
When processing group creation audit message, LDAPc also attempts to create the VO in which the group was created.
I added RE that recognizes this message, and disable adding VO bean to list of beans sent to create.

### DIFF
--- a/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
+++ b/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
@@ -308,6 +308,7 @@ public class EventDispatcherImpl implements EventDispatcher, Runnable {
 
 		//Debug information to check parsing of message.
 		MessageBeans beans = new MessageBeansImpl();
+
 		if(!listOfBeans.isEmpty()){
 			int i=0;
 			for(PerunBean p: listOfBeans) {
--- a/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/SimpleDispatchEventCondition.java
+++ b/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/SimpleDispatchEventCondition.java
@@ -53,8 +53,8 @@ public class SimpleDispatchEventCondition implements DispatchEventCondition {
 	@Override
 	public boolean isApplicable(MessageBeans beans, String msg) {
 		int presentMask = beans.getPresentBeansMask();
-		
-		return (requiredBeans & presentMask) == requiredBeans; 
+
+		return requiredBeans == presentMask;
 	}
 
 	private void addFlagForBeanName(String name) throws InternalErrorException {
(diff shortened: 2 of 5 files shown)


perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
```diff
@@ -308,9 +308,15 @@ protected MessageBeans resolveMessage(String msg, Integer idOfMessage) throws In
 
 		//Debug information to check parsing of message.
 		MessageBeans beans = new MessageBeansImpl();
+
+		boolean shouldntAddVo = msg.matches("Group:(.*)created in Vo:(.*)");
```
zlamalp: I probably wasn't exact in my comment in previous pull-request. I'm not sure this will work here. Since as we do it, you need to construct (ideally static or one instance) ``Pattern`` with specified regex and ``Pattern.DOT_ALL``. Then on runtime get Matcher out of Pattern like this: ``pattern.matcher(inputString)`` and check it by calling ``matcher.find()``.

I know its more code, but it seems more readable to me. We will also hopefully remove regex logic in near future and will map events by their Java type and not string message.

mvocu: I am not sure that I like the way this patch is going. The original problem, I think, is that audit messages with more perun beans trigger handler for both of those beans (see `SimpleDispatchEventCondition.isApplicable()` ). The condition there could be rewritten to return true iff the presentMask contains only the required beans. 

Given that, the Vo bean should be added to condition on "created in Vo:..." and the VO_F case in CreationEventProcessor should be modified to check for the presence of Group bean and skip Vo creation in that case (this change alone would be sufficient to resolve the issue).

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/SimpleDispatchEventCondition.java
```diff
@@ -53,8 +53,8 @@ public String getHandlerMethodName() {
 	@Override
 	public boolean isApplicable(MessageBeans beans, String msg) {
 		int presentMask = beans.getPresentBeansMask();
-		
-		return (requiredBeans & presentMask) == requiredBeans; 
+
+		return ((requiredBeans & presentMask) == requiredBeans && (requiredBeans | presentMask) == requiredBeans);
```
mvocu: Není to nakonec totéž co `requiredBeans == presentMask` ?

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/EventDispatcherImpl.java
```diff
@@ -1,38 +1,20 @@
...
-import cz.metacentrum.perun.core.api.Member;
-import cz.metacentrum.perun.core.api.Pair;
-import cz.metacentrum.perun.core.api.Perun;
-import cz.metacentrum.perun.core.api.PerunBean;
-import cz.metacentrum.perun.core.api.PerunSession;
-import cz.metacentrum.perun.core.api.Resource;
-import cz.metacentrum.perun.core.api.User;
-import cz.metacentrum.perun.core.api.UserExtSource;
-import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.*;
```
zlamalp: This class now has no changes other than new line and grouped imports. In such case please discard all changes in it and set bigger threshold for imports grouping in your IDE (we tend to keep exact import in most cases). Thank you.

PR #2338, opened 2019-07-31
Added checkAttributeSyntax to all layers of AttributesManager

metodej:
- the method was added to interfaces as well as implementations
- usage of this method in setAttribute and possibly in other methods will be added later

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AttributesManagerImplApi.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AttributesManagerImplApi.java
@@ -1746,6 +1746,167 @@ public interface AttributesManagerImplApi {
 	void checkAttributeValue(PerunSession sess, UserExtSource ues, Attribute attribute) throws InternalErrorException;
 
 	/**
+	 * Check if value of this facility attribute has valid syntax.
+	 *
+	 * @param sess perun session
+	 * @param facility facility for which you want to check validity of attribute
+	 * @param attribute attribute to check
+	 *
+	 * @throws InternalErrorException if an exception raise in concrete implementation, the exception is wrapped in InternalErrorException
+	 * @throws WrongAttributeValueException if the attribute value has wrong/illegal syntax
+	 */
+	void checkAttributeSyntax(PerunSession sess, Facility facility, Attribute attribute) throws InternalErrorException, WrongAttributeValueException;
+
+	/**
+	 * Check if value of this vo attribute has valid syntax.
+	 *
+	 * @param sess perun session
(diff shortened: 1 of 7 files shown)


stavamichal:
Please add simple test of call on every such created method in entry layer. You can use new attribute without module or mock the syntax check with some very basic rules.

metodej:
Thanks for review. The simple tests should be there now.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTestAbstract.java
```diff
@@ -7282,6 +7282,245 @@ public void getRequiredHostAttributesFromOneServiceWhenHostNotExists() throws Ex
 
 // already tested in lower layer - package: cz.metacentrum.perun.core.impl.modules.attributes
 
+	@Test
+	public void testCheckAttributeSyntaxVo() throws Exception {
+		attributesManagerBl = getTargetObject(perun.getAttributesManagerBl());
+		Attribute attribute = setAttributeInNamespace(AttributesManager.NS_VO_ATTR_DEF);
+
+		attributesManagerBl.checkAttributeSyntax(sess, vo1, attribute);
```
balcirakpeter: Please call checkAttributeSyntax from attributesManager instead of attributesManagerBl if it is possible (in all tests). We need to check if the code in the Entry layer is working correctly too.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AttributesManagerImplApi.java
```diff
@@ -1745,6 +1745,167 @@
...
+
+	/**
+	 * Check if value of this group attribute has valid syntax.
+	 *
+	 * @param sess perun session
+	 * @param group group for which you want to check validity of attribute
+	 * @param attribute attribute to check
+	 *
+	 * @throws InternalErrorException if an exception raise in concrete implementation, the exception is wrapped in InternalErrorException
+	 * @throws WrongAttributeValueException if the attribute value is wrong/illegal
```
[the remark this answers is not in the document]
metodej: Oh, thanks, must have missed that one.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTestAbstract.java
metodej: Okay, it should use the correct manager now, thank you.

PR #2343, opened 2019-08-06
Removed virtual attribute value calculation from audit messages

metodej:
- the calculation of virtual attribute values in audit messages was removed to speed up storing of audit messages
- added new audit event AttributeChangedForUser to substitute for AttributeSet/RemovedForUser in places where we used to decide which event to use based on virtual attribute value
- the audit messages now contains virtual attribute with null value

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
@@ -1,7 +1,6 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
-import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeRemovedForResourceAndMember;
-import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeSetForResourceAndMember;
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeChangedForResourceAndMember;
 import cz.metacentrum.perun.audit.events.AuditEvent;
 import cz.metacentrum.perun.audit.events.FacilityManagerEvents.BanRemovedForFacility;
 import cz.metacentrum.perun.audit.events.FacilityManagerEvents.BanSetForFacility;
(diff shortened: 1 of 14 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
```diff
@@ -109,9 +108,9 @@ public Attribute getAttributeValue(PerunSessionImpl sess, Member member, Resourc
 		try {
 			Member member = perunSession.getPerunBl().getMembersManagerBl().getMemberById(perunSession, banSetForResource.getMemberId());
 			Resource resource = perunSession.getPerunBl().getResourcesManagerBl().getResourceById(perunSession, banSetForResource.getResourceId());
-			Attribute attribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, member, resource, A_MR_V_isBanned);
-			resolvingMessages.add(new AttributeSetForResourceAndMember(attribute, resource, member));
-		} catch (MemberNotExistsException | ResourceNotExistsException | MemberResourceMismatchException e) {
+			AttributeDefinition attributeDefinition = perunSession.getPerunBl().getAttributesManagerBl().getAttributeDefinition(perunSession, A_MR_V_isBanned);
+			resolvingMessages.add(new AttributeSetForResourceAndMember(new Attribute(attributeDefinition), resource, member));
```
zlamalp: I'm not sure, we should keep using ``AttributeSetForResourceAndMember`` message, while we otherwise use ``AttributeChangedFor*`` class. Same goes for "remove" case.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertDNs.java
```diff
@@ -68,38 +67,17 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
-		if (userExtSourceRemovedFromUser.getUserExtSource().getExtSource() instanceof ExtSourceX509) {
-			User user = userExtSourceRemovedFromUser.getUser();
-			Attribute attrVirtUserCertDNs = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, A_U_V_USER_CERT_DNS);
-			//FIXME: if the attribute value is null or empty, this method should return AttributeRemovedForUser instead
-			resolvingMessages.add(new AttributeSetForUser(attrVirtUserCertDNs, user));
-		}
-
-		return resolvingMessages;
+	private AuditEvent resolveEvent(PerunSessionImpl perunSession, User user) throws AttributeNotExistsException, InternalErrorException {
+		AttributeDefinition attrVirtUserCertDNsDefinition = perunSession.getPerunBl().getAttributesManagerBl().getAttributeDefinition(perunSession, A_U_V_USER_CERT_DNS);
```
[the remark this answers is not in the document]
metodej: Sorry about that, it should be fixed now.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_virt_isBanned.java
metodej: Okay, the messages were changed to use the new type of audit event.

zlamalp:
We will wait until changes in LDAPc are also prepared in a pull-request and then merge it together.

metodej:
PR rebased and the logic was applied also to voShortNames.

PR #2345, opened 2019-08-07
Convert candidate user-attributes to JSON

balcirakpeter:
- Added method which converts candidate's user-attributes and
  user-core-attributes to JSON object. It will be used in User
  synchronizations.
- Added test to check if the method works correctly.
- Added json dependency to perun-base pom.xml

### DIFF
--- a/perun-base/pom.xml
+++ b/perun-base/pom.xml
@@ -136,6 +136,12 @@
 			<artifactId>javax.mail</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
+			<version>${json.version}</version>
+		</dependency>
+
 	</dependencies>
 
 </project>
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/Candidate.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/Candidate.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Set;
 
 import cz.metacentrum.perun.core.api.BeansUtils;
+import org.json.JSONObject;
 
 /**
  * Candidate member of a Virtual Organization.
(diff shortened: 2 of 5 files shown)


perun-base/pom.xml
```diff
@@ -136,6 +136,13 @@
 			<artifactId>javax.mail</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
+			<version>20140107</version>
+			<scope>compile</scope>
```
zlamalp: ``compile`` is default dependency type, you don't have to specify it.

perun-base/pom.xml
```diff
@@ -136,6 +136,13 @@
 			<artifactId>javax.mail</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
+			<version>20140107</version>
```
zlamalp: Can you please use the current version ? I found latest to be ``20190722``. Maybe check, if our parent pom ``spring-boot-starter-parent`` doesn't specify it already.

perun-base/src/main/java/cz/metacentrum/perun/core/api/Candidate.java
```diff
@@ -19,6 +20,9 @@
  */
 public class Candidate extends User {
 
+	private String NS_USER_ATTR_CORE = "urn:perun:user:attribute-def:core:";
+	private String NS_USER_ATTR = "urn:perun:user:attribute-def:";
```
zlamalp: Should be also ``static`` and ``final``.

perun-base/pom.xml
balcirakpeter: Removed

perun-base/pom.xml
balcirakpeter: Changed to 20190722.

perun-base/src/main/java/cz/metacentrum/perun/core/api/Candidate.java
balcirakpeter: Fixed.

perun-base/pom.xml
```diff
@@ -136,6 +136,12 @@
 			<artifactId>javax.mail</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.json</groupId>
+			<artifactId>json</artifactId>
+			<version>20190722</version>
```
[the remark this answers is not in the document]
balcirakpeter: Done.

PR #2346, opened 2019-08-07
User sync failed events

balcirakpeter:
- Created two events for user synchronization: UserSyncFailed and
  UserSyncFinishedWithErrors.


perun-base/src/main/java/cz/metacentrum/perun/audit/events/UserManagerEvents/UserSyncFinishedWithErrors.java
```diff
@@ -0,0 +1,34 @@
...
+	private User user;
+	private String message;
+
+	@SuppressWarnings("unused") // used by jackson mapper
+	public UserSyncFinishedWithErrors() {
+	}
+
+	public UserSyncFinishedWithErrors(User user) {
+		this.user = user;
+		this.message = formatMessage( "%s synchronization failed.", user);
```
stavamichal: Is there right message? Because this one is same like for UserSyncFailed.java event.

balcirakpeter: Copy paste bug, fixed

PR #2348, opened 2019-08-07
Fix of attribute module dependencies removal

Vojtech-Sassmann:
* Problem: during the deletion of an attribute, the old implementation
would always delete the module dependencies from relevant maps
regardless of possible failure of the DB removal. If the DB operation would
fail, the dependencies would be removed but the attribute would not
still be deleted. Next attempt of the attribute removal would fail due
to NullPointerException because the dependencies would no longer contain
the attribute's data.
* Fix: move the dependencies removal after the DB operation. That way,
the dependencies are removed only if the DB operation succeeds.
* Implemented tests that verify the correct handling of attribute
module dependencies during attribute deletion.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -2483,8 +2483,13 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		//Remove services' required attributes
 		//TODO
 
-		AttributeDefinition attributeDef = new AttributeDefinition(attribute);
+		//Remove attribute and all it's values
+		this.deleteAllAttributeAuthz(sess, attribute);
+		getAttributesManagerImpl().deleteAttribute(sess, attribute);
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -2483,8 +2483,13 @@ public void deleteAttribute(PerunSession sess, AttributeDefinition attribute) th
 		//Remove services' required attributes
 		//TODO
 
-		AttributeDefinition attributeDef = new AttributeDefinition(attribute);
+		//Remove attribute and all it's values
+		getPerunBl().getAuditer().log(sess,new AttributeDeleted(attribute));
```
zlamalp: We generally log auditer message after the call to impl layer, please move it after ``getAttributesManagerImpl().deleteAttribute(sess, attribute);``

Vojtech-Sassmann: Fixed.

PR #2350, opened 2019-08-07
Support for multiple identifiers from an extSource

Gaeldrin:
@stavamichal @zlamalp navrh k debate (pondeli?)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
@@ -7,6 +7,7 @@ import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Host;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.PerunPrincipal;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.RichResource;
@@ -484,6 +485,20 @@ public interface UsersManagerBl {
 	 * @throws AttributeNotExistsException if attribute can't be found by it's name
 	 */
 	UserExtSource getUserExtSourceByUniqueAttributeValue(PerunSession sess, String attrName, String uniqueValue) throws InternalErrorException, AttributeNotExistsException, UserExtSourceNotExistsException;
+
+	/**
+	 * Iteratively searches through additional identifiers trying to find userExtSource with the same identifier.
+	 * Returns first found userExtSource or null when no matching userExtSource found.
+	 *
+	 * @param sess PerunSession to retrieve UserExtSource
+	 * @param principal PerunPrincipal which contains additionalIdentifiers
+	 *
+	 * @return UserExtSource found using additionalIdentifiers or null
+	 *
+	 * @throws InternalErrorException When attribute is not defined in Perun
+	 * @throws UserExtSourceNotExistsException When the multiple identifiers received are actually empty
+	 */
+	UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess, PerunPrincipal principal) throws InternalErrorException, UserExtSourceNotExistsException;
 
 	/**
 	 * Get all users userExtSources with last_access not older than (now - m),
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -64,6 +64,8 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 	private static final String UNSET_ROLE = "UNSET";
 	private static final String SET_ROLE = "SET";
 
+	private final static Set<String> extSourcesWithMultipleIdentifiers = BeansUtils.getCoreConfig().getExtSourcesMultipleIdentifiers();
+
 	/**
 	 * Checks if the principal is authorized.
 	 *
@@ -1608,8 +1610,16 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 		Utils.checkPerunSession(sess);
 		log.trace("Refreshing session data for session {}.", sess);
 
+		PerunPrincipal principal = sess.getPerunPrincipal();
+
 		try {
-			User user = perunBl.getUsersManagerBl().getUserByExtSourceNameAndExtLogin(sess, sess.getPerunPrincipal().getExtSourceName(), sess.getPerunPrincipal().getActor());
+			User user;
+				if(extSourcesWithMultipleIdentifiers.contains(principal.getExtSourceName())) {
+					UserExtSource ues = perunBl.getUsersManagerBl().getUserExtSourceFromMultipleIdentifiers(sess, principal);
+					user = perunBl.getUsersManagerBl().getUserByUserExtSource(sess, ues);
+				} else {
+					user = perunBl.getUsersManagerBl().getUserByExtSourceNameAndExtLogin(sess, principal.getExtSourceName(), principal.getActor());
+				}
 			sess.getPerunPrincipal().setUser(user);
 		} catch (Exception ex) {
 			// we don't care that user was not found - clear it from session
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
@@ -62,6 +62,8 @@ import org.slf4j.LoggerFactory;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.ArrayList;
+import java.util.Arrays;
 
 /**
  * Implementation of Perun.
@@ -111,6 +113,7 @@ public class PerunBlImpl implements PerunBl {
 	final static Logger log = LoggerFactory.getLogger(PerunBlImpl.class);
 
 	private final static Set<String> dontLookupUsersForLogins = BeansUtils.getCoreConfig().getDontLookupUsers();
+	private final static Set<String> extSourcesWithMultipleIdentifiers = BeansUtils.getCoreConfig().getExtSourcesMultipleIdentifiers();
 
 	public PerunBlImpl() {
 
@@ -128,14 +131,25 @@ public class PerunBlImpl implements PerunBl {
 			// Get the user if we are completely initialized
 			try {
 				PerunSession internalSession = getPerunSession();
-				User user = usersManagerBl.getUserByExtSourceNameAndExtLogin(internalSession, principal.getExtSourceName(), principal.getActor());
+				User user;
+				if(extSourcesWithMultipleIdentifiers.contains(principal.getExtSourceName())) {
+					UserExtSource ues = usersManagerBl.getUserExtSourceFromMultipleIdentifiers(internalSession, principal);
+					user = usersManagerBl.getUserByUserExtSource(internalSession, ues);
+				} else {
+					user = usersManagerBl.getUserByExtSourceNameAndExtLogin(internalSession, principal.getExtSourceName(), principal.getActor());
+				}
 				principal.setUser(user);
 
 				if (client.getType() != PerunClient.Type.OAUTH) {
 					// Try to update LoA for userExtSource
-					ExtSource es = extSourcesManagerBl.getExtSourceByName(internalSession, principal.getExtSourceName());
-					UserExtSource ues = usersManagerBl.getUserExtSourceByExtLogin(internalSession, es, principal.getActor());
-					if (ues.getLoa() != principal.getExtSourceLoa()) {
+					UserExtSource ues;
+						if(extSourcesWithMultipleIdentifiers.contains(principal.getExtSourceName())) {
+							ues = usersManagerBl.getUserExtSourceFromMultipleIdentifiers(internalSession, principal);
+						} else {
+							ExtSource es = extSourcesManagerBl.getExtSourceByName(internalSession, principal.getExtSourceName());
+							ues = usersManagerBl.getUserExtSourceByExtLogin(internalSession, es, principal.getActor());
+						}
+					if (ues != null && ues.getLoa() != principal.getExtSourceLoa()) {
 						ues.setLoa(principal.getExtSourceLoa());
 						usersManagerBl.updateUserExtSource(internalSession, ues);
 					}
@@ -168,6 +182,7 @@ public class PerunBlImpl implements PerunBl {
 		// update selected attributes for given extsourcetype
 		List<AttributeDefinition> attrs = coreConfig.getAttributesForUpdate().get(ues.getExtSource().getType());
 		if (attrs != null) {
+
 			for (AttributeDefinition attr : attrs) {
 				//get value from authentication
 				String attrValue = additionalAttributes.get(attr.getFriendlyName());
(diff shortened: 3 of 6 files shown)


stavamichal:
@Gaeldrin  For planning a meeting please use Slack. Always use english on github and please write commit message, it can help us to understand what was the purpose of these changes, even if this is just testing one or prototype. Thank you :)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -153,6 +171,34 @@ public PerunSession getPerunSession(PerunPrincipal principal, PerunClient client
...
+			} catch(UserExtSourceNotExistsException ex) {
+				//try to find user ext source using different identifier in the next iteration of for cycle
+			} catch(AttributeNotExistsException ex) {
+				String errorMessage = "Mandatory attribute is not defined: ".concat(additionalIdentifiersPerunAttributeName);
+				log.error(errorMessage);
+				throw new InternalErrorException(errorMessage, ex);
+			}
+		}
+		return ues;
+	}
```
stavamichal: It is really necessary to have this method twice in code?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -111,6 +113,11 @@
 	final static Logger log = LoggerFactory.getLogger(PerunBlImpl.class);
 
 	private final static Set<String> dontLookupUsersForLogins = BeansUtils.getCoreConfig().getDontLookupUsers();
+	private final static Set<String> extSourcesWithMultipleIdentifiers = BeansUtils.getCoreConfig().getExtSourcesMultipleIdentifiers();
+
+	private final static String multivalueAttributeSeparatorRegExp = ";";
+	private final static String additionalIdentifiersAttributeName = "additionalIdentifiers";
+	private final static String additionalIdentifiersPerunAttributeName = AttributesManager.NS_UES_ATTR_DEF + ":" + additionalIdentifiersAttributeName;
```
stavamichal: These constants can be achieved from 1 shared place (like protected or public constants).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -153,6 +171,34 @@ public PerunSession getPerunSession(PerunPrincipal principal, PerunClient client
...
+	/**
+	 * Iteratively searches through additional identifiers trying to find UES with the same identifier.
+	 * Returns first found UES or null when no matching UES found.
+	 *
+	 * @param principal PerunPrincipal which contains additionalIdentifiers
+	 * @param internalSession PerunSession to retrieve UserExtSource
+	 * @return UserExtSource found using additionalIdentifiers or null
+	 * @throws InternalErrorException When attribute is not defined in Perun
+	 */
+	public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunPrincipal principal, PerunSession internalSession) throws InternalErrorException {
```
stavamichal: It is standard for us to have session always as the first parameter if possible. And the name is normally just "session" or "sess".

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -153,6 +171,34 @@ public PerunSession getPerunSession(PerunPrincipal principal, PerunClient client
...
+		String additionalIdentifiers = principal.getAdditionalInformations().get(additionalIdentifiersAttributeName);
+		UserExtSource ues = null;
+		for(String identifier : additionalIdentifiers.split(multivalueAttributeSeparatorRegExp)) {
```
stavamichal: You can get NullPointerException if additionalIdentifiers will be null.

stavamichal:
Please, give me echo as soon as you will fix these mentioned problems. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1,39 +1,8 @@
...
-import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.ResourceNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.SecurityTeamNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.ServiceNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.UserNotAdminException;
-import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.*;
+import cz.metacentrum.perun.core.api.exceptions.*;
```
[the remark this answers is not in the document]
Gaeldrin: Automatic IDE conversion, I will revert that, apologies.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
Gaeldrin: Absolutely not, this was result of some work on the method extraction discussed earlier and I made them exactly the same on purpose to emphasise that only one can be used for both situations. The same goes for those constants from next comment and my question is: where should they be located at and which way should I use to "import" them?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
Gaeldrin: Understood and fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
Gaeldrin: Can that actually happen with the check on lines 139 or 150? Not sure but I expected this check will prevent the attribute to be null.

stavamichal: This is separate public method. If anybody calls it, it's just about returned value of the call "principal.getAdditionalInformations().get(additionalIdentifiersAttributeName)". So if you are sure this call will never return null, than you don't need to check it. Otherwise you should.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
stavamichal: There are two possible location (in my opinion). Because you have session, you can get perunBl from it, so one place where you can add it is to Utils.java. But much better place probably will be UsersManagerBlImpl.java, cause all methods returning object UserExtSource are there.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
Gaeldrin: Makes total sense. How to handle it? Log and throw AttributeNotFoundException?

stavamichal: The question is what does it mean:
- it is unexpected event? Than InternalErrorException is good choice
- It is because there is something missing which should never be missing? Then any error with more specific info is good like - 'CoreAttributeNotFoundException' or something like that (be careful about simple word "Attribute" because it is overloaded in Perun)
- it is normal situation? then you can log it (or not) and return empty ues (null in this case)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -184,7 +201,17 @@ public void setUserExtSourceAttributes(PerunSession session, UserExtSource ues,
 							attributeWithValue = attributesManagerBl.getAttribute(session, ues, attr.getName());
 						}
 					}
-					attributeWithValue.setValue(attrValue);
+
+					//for Array list attributes try to parse string value into individual fields
+					if(attributeWithValue.getType().equals(ArrayList.class.getName())) {
+
+						log.info("TEST: storing list-type attribute (before split) " + attributeWithValue);
+						attributeWithValue.setValue(new ArrayList<String>(Arrays.asList(attrValue.split(multivalueAttributeSeparatorRegExp))));
```
Gaeldrin: I don't like this construction, is there a way to do this better?

stavamichal: Not better, but maybe little easier for reading:
`List<String> value = Arrays.asList(attrValue.split(multivalueAttributeSeparatorRegExp));
attributeWithValue.setValue(value);`

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -128,14 +133,25 @@ public PerunSession getPerunSession(PerunPrincipal principal, PerunClient client
...
 				principal.setUser(user);
 
 				if (client.getType() != PerunClient.Type.OAUTH) {
 					// Try to update LoA for userExtSource
-					ExtSource es = extSourcesManagerBl.getExtSourceByName(internalSession, principal.getExtSourceName());
-					UserExtSource ues = usersManagerBl.getUserExtSourceByExtLogin(internalSession, es, principal.getActor());
-					if (ues.getLoa() != principal.getExtSourceLoa()) {
+					UserExtSource ues;
+						if(extSourcesWithMultipleIdentifiers.contains(principal.getExtSourceName())) {
+							ues = usersManagerBl.getUserExtSourceFromMultipleIdentifiers(internalSession, principal);
```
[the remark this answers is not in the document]
stavamichal: Why it shouldn't be? It looks ok.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1602,8 +1604,16 @@ public static synchronized void refreshSession(PerunSession sess) throws Interna
 		Utils.checkPerunSession(sess);
 		log.trace("Refreshing session data for session {}.", sess);
 
+		PerunPrincipal principal = sess.getPerunPrincipal();
+
 		try {
-			User user = perunBl.getUsersManagerBl().getUserByExtSourceNameAndExtLogin(sess, sess.getPerunPrincipal().getExtSourceName(), sess.getPerunPrincipal().getActor());
+			User user;
+				if(extSourcesWithMultipleIdentifiers.contains(principal.getExtSourceName())) {
+					UserExtSource ues = perunBl.getUsersManagerBl().getUserExtSourceFromMultipleIdentifiers(sess, principal);
```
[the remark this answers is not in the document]
stavamichal: I don't see any problem here.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
Gaeldrin: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
Gaeldrin: Used UserExtSourceNotExistsException as found in UsersManagerImpl.java on line 573.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
```diff
@@ -485,6 +486,17 @@
...
+	/**
+	 * Iteratively searches through additional identifiers trying to find userExtSource with the same identifier.
+	 * Returns first found userExtSource or null when no matching userExtSource found.
+	 *
+	 * @param sess PerunSession to retrieve UserExtSource
+	 * @param principal PerunPrincipal which contains additionalIdentifiers
+	 * @return UserExtSource found using additionalIdentifiers or null
+	 * @throws InternalErrorException When attribute is not defined in Perun
+	 */
+	UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess, PerunPrincipal principal) throws InternalErrorException, UserExtSourceNotExistsException;
```
stavamichal: Missing exception in javadoc.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -682,6 +687,29 @@ private UserExtSource getUserExtSourceByUniqueAttributeValue(PerunSession sess,
...
+		for(String identifier : additionalIdentifiers.split(multivalueAttributeSeparatorRegExp)) {
+			try {
+				ues = perunBl.getUsersManagerBl().getUserExtSourceByUniqueAttributeValue(sess, additionalIdentifiersPerunAttributeName, identifier);
+				log.debug("UserExtSource found using additional identifiers: " + ues);
+				break;
+			} catch (UserExtSourceNotExistsException ex) {
+				//try to find user ext source using different identifier in the next iteration of for cycle
+			} catch (InternalErrorException | AttributeNotExistsException ex) {
+				String errorMessage = "Mandatory attribute is not defined: ".concat(additionalIdentifiersPerunAttributeName);
+				log.error(errorMessage);
```
stavamichal: There is really no need for creating a new string here. You can use the code immediately in the log.error() call.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
Gaeldrin: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
Gaeldrin: In that case, how will I throw the exception with the message on the next line? There was a requirement to log the error and then throw it if I remember well..

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -184,7 +199,15 @@ public void setUserExtSourceAttributes(PerunSession session, UserExtSource ues,
 							attributeWithValue = attributesManagerBl.getAttribute(session, ues, attr.getName());
 						}
 					}
-					attributeWithValue.setValue(attrValue);
+
+					//for Array list attributes try to parse string value into individual fields
+					if(attributeWithValue.getType().equals(ArrayList.class.getName())) {
```
zlamalp: We have also LargeArrayList type of attributes, which are technically lists, but can be larger than 4000bytes. You can use constant ``largeArrayListClassName`` from ``BeansUtils`` to check the type against it.

Gaeldrin: Fixed.

zlamalp: I probably wasn't exact. You need to check both ``ArrayList`` and ``LargeArrayList`` types and handle them as list.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -184,7 +199,15 @@ public void setUserExtSourceAttributes(PerunSession session, UserExtSource ues,
 							attributeWithValue = attributesManagerBl.getAttribute(session, ues, attr.getName());
 						}
 					}
-					attributeWithValue.setValue(attrValue);
+
+					//for Array list attributes try to parse string value into individual fields
+					if(attributeWithValue.getType().equals(BeansUtils.largeArrayListClassName)) {
```
zlamalp: Please check both types, ArrayList and LargeArrayList.

Gaeldrin: Heh, I didn't understand that, hopefully fixed now. :)

PR #2364, opened 2019-08-13
Added use of checkAttributeSyntax

metodej:
- setAttribute now checks syntax before checking semantics
- implementation of checkAttributeSyntax to attribute modules should be added later

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -725,9 +725,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		}
 		//if checkAttributesSemantics fails it causes rollback so no attribute will be stored
 		checkAttributesSemantics(sess, facility, attributesToSet);
-		log.debug("IMPORTANT: ENTERING CHECK ATTRIBUTES DEPENDENCIES");
 		checkAttributesDependencies(sess, facility, null, attributesToSet);
-		log.debug("IMPORTANT: EXITING CHECK ATTRIBUTES DEPENDENCIES");
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -721,7 +721,8 @@ public void setAttributes(PerunSession sess, Facility facility, List<Attribute>
 				}
 			}
 		}
-		//if checkAttributesSemantics fails it causes rollback so no attribute will be stored
+		//if checkAttributesSyntax or checkAttributesSemantics fails it causes rollback so no attribute will be stored
```
balcirakpeter: Please keep comments separated. CheckAttributesSemantics will be moved to a different part of the set method later, so it is better to keep the comments separated.

metodej: Okay, the comments are now separated.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -1961,12 +1996,14 @@ public void setAttribute(PerunSession sess, Resource resource, Member member, At
 		if (!workWithUserAttributes) {
 			if (setAttributeWithoutCheck(sess, member, resource, attribute, false)) {
 				this.checkAttributeDependencies(sess, new RichAttribute<>(resource, member, attribute));
```
zlamalp: This method has different order of methods inside than the others. It should be checks for syntax, semantics, dependencies.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -2229,6 +2270,7 @@ public void setAttribute(PerunSession sess, Resource resource, Group group, Attr
 		}
 		if (setAttributeWithoutCheck(sess, resource, group, attribute)) {
 			this.checkAttributeDependencies(sess, new RichAttribute<>(resource, group, attribute));
```
zlamalp: Again different order of checks in method implementation.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -721,6 +721,8 @@ public void setAttributes(PerunSession sess, Facility facility, List<Attribute>
 				}
 			}
 		}
+		//if checkAttributesSyntax fails it causes rollback so no attribute will be stored
+		checkAttributesSyntax(sess, facility, attributesToSet);
```
zlamalp: As @balcirakpeter noted, you can perform check on syntax before actually storing attributes to the DB (before call ``setAttributeWithoutCheck()``).  I believe this applies to all changed methods.

metodej:
Thank you for review, the order of methods should be now hopefully correct.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -928,6 +942,8 @@ public void setAttributes(PerunSession sess, Member member, Resource resource, L
 			}
 		}
 		removeAttributes(sess, resource, member, attributesToRemove, workWithUserAttributes);
+		//if checkAttributesSyntax fails it causes rollback so no attribute will be stored
+		checkAttributesSyntax(sess, member, resource, attributesToSet, workWithUserAttributes);
```
zlamalp: Not related to this pull-request, but can you please inspect, what is this usage of weird ``Utils.startTimer()`` later in this method ? Seems like just some forgotten debug. It is only present in AttributesManagerBlImpl and just prints time, how long it took to perform specific actions (probably related to adding member). If you don't find any reasonable usage, please remove all such debug messages and then their unused variables filled by timer.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -1959,14 +2008,16 @@ public void setAttribute(PerunSession sess, Resource resource, Member member, At
 			return;
 		}
 		if (!workWithUserAttributes) {
+			checkAttributeSyntax(sess, member, resource, attribute);
```
zlamalp: This method is unused, please remove it. We have standard method to set member-resource attribute and there is no need to support other attribute types like we do for setAttributes(). Otherwise we would have to fill method in API and other layers of AttributesManager, but there is no demand for it.

metodej: Ok, the method should be deleted now.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
metodej: Okay, since I haven't found any reason for those debug logs either, I have deleted them.

PR #2366, opened 2019-08-13
LDAPc Bug:removing data in sync after exception

xpavlic:
When LDAPc sync is getting data and exception is thrown, resource list is empty or not complete and items which are supposed to remain are deleted. (same problem is with attributes)
I added some control variables and edited several try-catch blocks to fix this bug.


mvocu:
The question is, should the synchronization (and ldapc altogether) be allowed to go on in case there is an error reading data from database? Would not it be better to just end the process (and allow for some sort of monitoring to step in)?

balcirakpeter:
Please resolve conflicts in LdapcManagerImpl.

PR #2370, opened 2019-08-15
CORE: Removed usage of getComplementaryObjectsForRole() for filtering

zlamalp:
- In get(Rich)Facilities() we were getting each facility by its ID for
  non perun-admins. Now we use direct select for all such facilities.
- getVos() was also optimized. We can get all vos where user is
  VOADMIN at once and also all groups where is GROUPADMIN to fetch VOs
  from them. Also we don't have to repeat VO retrieval for already
  retrieved VOs.


zlamalp:
Replaced by #2370

PR #2372, opened 2019-08-16
CORE: Removed usage of getComplementaryObjectsForRole() for filtering

zlamalp:
- In get(Rich)Facilities() we were getting each facility by its ID for
  non perun-admins. Now we filter results of select for all.
- getVos() was also optimized. We can get all vos at once and filter
  out those, where user is not VOADMIN or GROUPADMIN, then we add
  vos where user is VOOBSERVER.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
@@ -131,12 +131,19 @@ public class FacilitiesManagerEntry implements FacilitiesManager {
 		if (AuthzResolver.isAuthorized(sess, Role.PERUNADMIN)) {
 			return getFacilitiesManagerBl().getRichFacilities(sess);
 		} else if (AuthzResolver.isAuthorized(sess, Role.FACILITYADMIN)) {
-			// Cast complementary object to Facility
-			List<Facility> facilities = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


zlamalp:
This implementation replaces #2370.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -131,12 +131,15 @@ public Facility getFacilityByName(PerunSession sess, String name) throws Interna
...
-			//Now I create list of richFacilities from facilities
+			List<Facility> facilities = getFacilitiesManagerBl().getFacilities(sess);
+			facilities.removeIf(facility -> {
+				try {
+					return !AuthzResolver.isAuthorized(sess, Role.FACILITYADMIN, facility);
+				} catch (InternalErrorException e) {
+					// if we can't determine authorization prevent returning it
+					return true;
+				}
+			});
```
stavamichal: I am afraid this is not effective for situations where there are lot of facilities and I am admin of just small amount of them. In situation like 100K facilities exists and I am admin only for 2 of them this is very ineffective call. (even if the situation is very unlikely)

zlamalp: For now we have max. hundreds of facilities, so it should be quicker to get data at once than loading 5-10 facilities by id. But I added comment/todo explaining the situation. We should load proper facilities directly, but right now that would be inconsistent with contract of our authz. We keep authorization in session, which is not reflecting authz changes before its refresh. As we agreed, I'm merging this.

PR #2376, opened 2019-08-19
Skipping of value check during dependencies check

Vojtech-Sassmann:
* Problem: during dependencies check, the current implementation is
trying to check the value of all attributes that are dependant on given
attributes. Even though the attribute doesn't need to implement methods
for value checks (semantic, syntax). Because of that, it needs to
calculate its value. This can be slow and it is useless.
* Solution: there is a new annotation for modules of virtual attributes
named SkipValueCheckDuringDependencyCheck. This annotation is used for all
virtual attribute modules that don't check their values. During the
dependencies check, the method now skips all attributes that have a
module which is implementing this new interface.
* NOTE - during a review, I suggest viewing changes commit by commit.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -96,6 +96,7 @@ import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_facility_attr
 import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_member_attribute_def_def_suspensionInfo;
 import cz.metacentrum.perun.core.implApi.AttributesManagerImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.AttributesModuleImplApi;
+import cz.metacentrum.perun.core.implApi.modules.attributes.SkipValueCheckDuringDependencyCheck;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleImplApi;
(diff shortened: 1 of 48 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/AttributeDefinition.java
```diff
@@ -134,6 +134,14 @@ public void setUnique(boolean unique) {
...
 	}
 
+	/**
+	 * Returns true if the attribute definition is for a virtual attribute.
+	 *
+	 * @return true if the attribute definition is for a virtual attribute, false otherwise.
+	 */
+	public boolean isVirtual() {
+		return this.namespace.endsWith("virt");
+	}
```
stavamichal: Did you add this to serialization and de-serialization? We probably can skip it.

Vojtech-Sassmann: I have edited the serializers to skip the value.

zlamalp: While this is technically/java correct solution, I would rather avoid adding such specific method in an object, exactly because it has to be handled in (de)serializer. Method returning type of attribute definition would be more general solution (like we do in ``getEntity()``).
After this change, it has only single usage. We have check methods in ``AttributesManagerImpl`` or some utils class. They are not generic, so we can create a new one or simply check "virtual" state in that single usage place.

PS: If you still want to keep your solution, then please update also (de)serializer in rpclib package. Otherwise we won't be able to read back the AttributeDefinition object from auditer log.

Vojtech-Sassmann: I have changed the check that it is using the method from impl layer.

PR #2382, opened 2019-08-27
Fix replacing of duplicity in shibbolethExtSources module

stavamichal:
- in the situation where duplicity in keys exists, the last login has
 been used and others were replaced. This commit fix this situation by
 using prefix for every key. Prefix is always unique number + ':'


stavamichal:
This PR can't be merge before merging counterpart in perun-services which resolve problem for service du_users_export.

Michal Stava

stavamichal:
Counterpart PR in perun-services is https://github.com/CESNET/perun-services/pull/422

PR #2384, opened 2019-08-28
CORE, GUI, CLI: Support new windows destinations

zlamalp:
- Added constants and checks for new windows type destinations.
- They are expected to contains hostname or user@hostname values,
  hence we treat them as standard host like destinations when they
  are added.
- Added support for their setting from GUI.
- Added support in CLI (documentation, whitespace/typo fixes).

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
@@ -456,7 +456,9 @@ public class ServicesManagerEntry implements ServicesManager {
 		List<Facility> facilitiesByDestination = new ArrayList<>();
 		if(destination.getType().equals(Destination.DESTINATIONHOSTTYPE) ||
 				destination.getType().equals(Destination.DESTINATIONUSERHOSTTYPE) ||
-				destination.getType().equals(Destination.DESTINATIONUSERHOSTPORTTYPE)) {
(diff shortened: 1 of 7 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
```diff
@@ -509,7 +511,9 @@ public Destination addDestination(PerunSession sess, Service service, Facility f
 		List<Facility> facilitiesByDestination = new ArrayList<>();
 		if(destination.getType().equals(Destination.DESTINATIONHOSTTYPE) ||
 				destination.getType().equals(Destination.DESTINATIONUSERHOSTTYPE) ||
-				destination.getType().equals(Destination.DESTINATIONUSERHOSTPORTTYPE)) {
+				destination.getType().equals(Destination.DESTINATIONUSERHOSTPORTTYPE)||
+				destination.getType().equals(Destination.DESTINATIONWINDOWS)||
```
stavamichal: I would prefer same format for all "II" parts (with space before).

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
```diff
@@ -456,7 +456,9 @@ public Destination addDestination(PerunSession perunSession, List<Service> servi
...
```
stavamichal: I would prefer same format for all "II" parts (with space before).

PR #2385, opened 2019-08-28
ExtSource implements ExtSourceSimpleApi

oernst9:
* making ExtSource implement ExtSourceSimpleApi so that classes inheriting from ExtSource don't have to implement it themselves

* this required moving ExtSourceSimpleApi to perun-base, I have also moved ExtSourceApi (so that they wouldn't be separated but ExtSourceApi doesn't have to be in perun-base so it's up to you, ExtSource might implement that as well.


stavamichal:
@oernst9 There shouldn't be a question in commit message :) It should say what you did and why (if needed).

stavamichal:
@oernst9 Thank to this PR we discussed situation about ExtSources more deeply and we decided to do this different way. I will close this one and write conclusion to the redmine task.

PR #2386, opened 2019-08-29
Separation of checkAttributeSyntax from semantics in Group modules

metodej:
- former checkAttributeValue is now separated into checkAttributeSyntax and checkAttributeSemantics in group attribute modules
- also added test for checks

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_collectionID.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_collectionID.java
@@ -5,6 +5,7 @@ import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.GroupAttributesModuleAbstract;
 import cz.metacentrum.perun.core.implApi.modules.attributes.GroupAttributesModuleImplApi;
@@ -17,18 +18,9 @@ import cz.metacentrum.perun.core.implApi.modules.attributes.GroupAttributesModul
 public class urn_perun_group_attribute_def_def_collectionID extends GroupAttributesModuleAbstract implements GroupAttributesModuleImplApi {
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongAttributeValueException {
+	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongReferenceAttributeValueException {
 		// null attribute
-		if (attribute.getValue() == null) throw new WrongAttributeValueException(attribute, "Attribute collectionID cannot be null.");
-
-		// wrong type of the attribute
-		if (!(attribute.getValue() instanceof String)) throw new WrongAttributeValueException(attribute, "Wrong type of the attribute. Expected: String");
-
-		String collectionID = attribute.valueAsString();
-
-		if (collectionID.isEmpty()) {
-			throw new WrongAttributeValueException(attribute, "Attribute collectionID cannot be empty.");
-		}
+		if (attribute.getValue() == null) throw new WrongReferenceAttributeValueException(attribute, "Attribute collectionID cannot be null.");
 	}
 
 	@Override
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_groupStructureSynchronizationTimes.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_groupStructureSynchronizationTimes.java
@@ -19,18 +19,9 @@ public class urn_perun_group_attribute_def_def_groupStructureSynchronizationTime
 	private static final Pattern pattern = Pattern.compile("^(([0-1][0-9])|(2[0-3])):[0-5][0,5]$");
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+	public void checkAttributeSyntax(PerunSessionImpl sess, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException {
 		//Null value is ok, means no settings for group
 		if(attribute.getValue() == null) return;
-
-		try {
-			Attribute foundAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME);
-			if(foundAttribute.getValue() != null) {
-				throw new WrongAttributeAssignmentException("Attribute " + attribute.getName() + " cannot be set because attribute " + GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME + " is already set.");
-			}
-		} catch (AttributeNotExistsException exc) {
-			throw new ConsistencyErrorException("Attribute " + GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME + " is supposed to exist", exc);
-		}
 
 		ArrayList<String> attrValues = attribute.valueAsList();
 
@@ -42,8 +33,18 @@ public class urn_perun_group_attribute_def_def_groupStructureSynchronizationTime
 		}
 	}
 
-
-
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		if (attribute.getValue() == null) return;
+		try {
+			Attribute foundAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME);
+			if(foundAttribute.getValue() != null) {
+				throw new WrongReferenceAttributeValueException(attribute, foundAttribute, group, null, group, null, "Attribute " + attribute.getName() + " cannot be set because attribute " + GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME + " is already set.");
+			}
+		} catch (AttributeNotExistsException exc) {
+			throw new ConsistencyErrorException("Attribute " + GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME + " is supposed to exist", exc);
+		}
+	}
 
 	@Override
 	public List<String> getDependencies() {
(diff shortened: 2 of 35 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_groupStructureSynchronizationTimes.java
```diff
@@ -42,8 +33,17 @@ public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribut
...
 	}
 
-
-
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		try {
+			Attribute foundAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME);
+			if(foundAttribute.getValue() != null) {
+				throw new WrongReferenceAttributeValueException(attribute, foundAttribute, group, null, group, null, "Attribute " + attribute.getName() + " cannot be set because attribute " + GroupsManager.GROUP_STRUCTURE_SYNCHRO_INTERVAL_ATTRNAME + " is already set.");
```
stavamichal: You will throw this exception even if this is the process of removing and the value is  "null". Which is not correct.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_groupSynchronizationTimes.java
```diff
@@ -43,7 +34,17 @@ public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribut
 		}
 	}
 
-
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongAttributeAssignmentException, InternalErrorException, WrongReferenceAttributeValueException {
+		try {
```
stavamichal: Also here must be skipped null value.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_synchronizationEnabled.java
```diff
@@ -27,30 +27,34 @@
...
+	}
 
-					requiredAttribute = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUPEXTSOURCE_ATTRNAME);
-					if (requiredAttribute.getValue() == null) {
-						throw new WrongReferenceAttributeValueException(attribute, requiredAttribute, requiredAttribute.toString() + " must be set in order to enable synchronization.");
-					}
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongAttributeAssignmentException, InternalErrorException, WrongReferenceAttributeValueException {
+		try {
+			if (attribute.valueAsString().equals("true")) {
```
stavamichal: If value will be null, this will throw NullPointerException here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
```diff
@@ -118,7 +118,17 @@ public Attribute fillAttribute(PerunSessionImpl sess, Group group, AttributeDefi
 	}
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException{
+	public void checkAttributeSyntax(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongAttributeValueException, InternalErrorException {
+		//Check if GID is within allowed range
+		try {
+			sess.getPerunBl().getModulesUtilsBl().checkIfGIDIsWithinRange(sess, attribute);
+		} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
```
stavamichal: I am afraid that null is (in specific situation) allowed. So by syntax check it should be allowed always.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
```diff
@@ -133,34 +143,25 @@ public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribut
...
 			}
 
-			//Special behaviour if gid is null
-			Integer attrValue;
-			if(attribute.getValue() == null) {
-				throw new WrongAttributeValueException(attribute, group, "Unix GID must be set");
-			} else {
-				attrValue = (Integer) attribute.getValue();
-			}
+			Integer attrValue = attribute.valueAsInteger();
```
stavamichal: This will not throw an exception if value will be null there so it will continue even if it shouldn't.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
```diff
@@ -133,34 +143,25 @@ public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribut
...
 			//check if gid is not already depleted
 			Attribute usedGids = sess.getPerunBl().getAttributesManagerBl().getAttribute(sess, gidNamespace, A_E_usedGids);
 			//null in value means there is no depleted or used gids
 			if(usedGids.getValue() != null) {
-				Map<String, String> usedGidsValue = (Map<String,String>) usedGids.getValue();
+				Map<String, String> usedGidsValue = usedGids.valueAsMap();
 				//Dx, where x is GID means depleted value for GID x
 				if(usedGidsValue.containsKey("D" + attrValue.toString())) {
-					throw new WrongReferenceAttributeValueException(attribute, usedGids, group, null, gidNamespace, null, "This GID is already depleted.");
+					throw new WrongReferenceAttributeValueException(attribute, usedGids, group, null, null, null, "This GID is already depleted.");
```
stavamichal: The "string" is entity for entityless attributes so it should stay there.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
stavamichal: To be within range is semantic problem here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
```diff
@@ -133,34 +143,25 @@ public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribut
...
 				}
 			}
 
-			//Check if gid GID is within allowed range
-			sess.getPerunBl().getModulesUtilsBl().checkIfGIDIsWithinRange(sess, attribute);
```
stavamichal: This should stay here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
```diff
@@ -283,12 +284,12 @@ public void changedAttributeHook(PerunSessionImpl session, Group group, Attribut
...
-	/*public AttributeDefinition getAttributeDefinition() {
+	public AttributeDefinition getAttributeDefinition() {
 		AttributeDefinition attr = new AttributeDefinition();
 		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
 		attr.setFriendlyName("unixGID-namespace");
 		attr.setType(Integer.class.getName());
 		attr.setDescription("Unix GID namespace.");
 		return attr;
-	}*/
+	}
```
stavamichal: Why did you remove comments here? It is there because of namespace in the module.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGroupName_namespace.java
```diff
@@ -169,12 +174,12 @@ public void changedAttributeHook(PerunSessionImpl session, Group group, Attribut
...
-	/*public AttributeDefinition getAttributeDefinition() {
+	public AttributeDefinition getAttributeDefinition() {
 		AttributeDefinition attr = new AttributeDefinition();
 		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
 		attr.setFriendlyName("unixGroupName-namespace");
 		attr.setType(String.class.getName());
 		attr.setDescription("Unix Group Name namespace.");
 		return attr;
-	}*/
+	}
```
stavamichal: Any reason for removing these comments?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_unixGID_namespace.java
metodej: But I thought that if the value was null, the method would end due to previous condition, would it not?

stavamichal: Yeah, you are right. If value will be null, it will end with an exception or by return. So this is ok.

metodej:
Thank you for review, everything mentioned above should be implemented or fixed now.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_collectionID.java
```diff
@@ -17,18 +18,17 @@
...
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongAttributeValueException {
-		// null attribute
-		if (attribute.getValue() == null) throw new WrongAttributeValueException(attribute, "Attribute collectionID cannot be null.");
+	public void checkAttributeSyntax(PerunSessionImpl sess, Group group, Attribute attribute) throws WrongAttributeValueException {
+		if (attribute.getValue() == null) return;
 
 		// wrong type of the attribute
 		if (!(attribute.getValue() instanceof String)) throw new WrongAttributeValueException(attribute, "Wrong type of the attribute. Expected: String");
```
stavamichal: This whole test can be removed. No need to test instance of the attribute value.

stavamichal:
> There is a potential problem with group modules which check attributes needed for synchronizations. Many necessary checks will not be executed which can cause errors or unexpected behavior. For example, you will be able to enable structure synchronization inside another structure synchronization.

We need to discuss this more. It is not problem of this PR itself, it is problem of the whole concept which need to be resolve before other moves.

balcirakpeter:
> > There is a potential problem with group modules which check attributes needed for synchronizations. Many necessary checks will not be executed which can cause errors or unexpected behavior. For example, you will be able to enable structure synchronization inside another structure synchronization.
> 
> We need to discuss this more. It is not problem of this PR itself, it is problem of the whole concept which need to be resolve before other moves.

We had a discussion and a result is that the concept of splitting checkSyntax/checkSemantics will remain unchanged, the only difference will be in memberResource status extension which will be implemented later. So we can continue with creating/merging pull requests, which are about splitting syntax/smantics, as before.

metodej:
> > > There is a potential problem with group modules which check attributes needed for synchronizations. Many necessary checks will not be executed which can cause errors or unexpected behavior. For example, you will be able to enable structure synchronization inside another structure synchronization.
> > 
> > 
> > We need to discuss this more. It is not problem of this PR itself, it is problem of the whole concept which need to be resolve before other moves.
> 
> We had a discussion and a result is that the concept of splitting checkSyntax/checkSemantics will remain unchanged, the only difference will be in memberResource status extension which will be implemented later. So we can continue with creating/merging pull requests, which are about splitting syntax/smantics, as before.

Okay, I just removed the last unnecessary check about type in collectionID.

PR #2392, opened 2019-09-03
CORE: Use LinkedHashSet for storing resolved auditer messages

zlamalp:
- We recently introduced proper check on duplicity when we stored
  auditer messages resolved from registered attribute modules.
  It used list for messages to preserve their order.
  Now we use LinkedHashSet which improves performance by being a Set
  to prevent duplicate, but also preserver order since its backed
  with linked array.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Auditer.java
```diff
@@ -432,7 +433,7 @@ public void initialize() throws InternalErrorException {
 			//Update list of already resolved messages
 			alreadyResolvedMessages.addAll(addedResolvedMessages);
 			//Continue of processing newly generated messages
-			addedResolvedMessages.addAll(checkRegisteredAttributesModules(session, addedResolvedMessages, alreadyResolvedMessages));
+			addedResolvedMessages.addAll(checkRegisteredAttributesModules(session, new ArrayList<>(addedResolvedMessages), alreadyResolvedMessages));
```
Vojtech-Sassmann: Just a thought. Could we change the type of the argument for the method `checkRegisteredAttributesModules` from `List<AuditerMessage> messages` to `Collection<AuditerMessage> messages` ? With that, it would not be necessary to convert the LinkedHashSet to ArrayList.

zlamalp: Fixed.

PR #2393, opened 2019-09-03
Fixing bug with changing ExtSources

oernst9:
* Changing of ExtSources will no longer remove members who are present in the new ExtSource

* categorizeMembersForLightweightSynchronization() will try to get the user using additional ExtSources, not only the memberSource

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -41,7 +41,6 @@ import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.MemberGroupStatus;
 import cz.metacentrum.perun.core.api.MembershipType;
 import cz.metacentrum.perun.core.api.Pair;
-import cz.metacentrum.perun.core.api.PerunBeanProcessingPool;
 import cz.metacentrum.perun.core.api.PerunClient;
 import cz.metacentrum.perun.core.api.PerunPrincipal;
 import cz.metacentrum.perun.core.api.PerunSession;
@@ -2691,9 +2690,12 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		for(RichMember richMember: groupMembers) {
 			idsOfUsersInGroup.put(richMember.getUserId(), richMember);
 		}
-
 		//try to find users by login and loginSource
 		for(Map<String, String> subjectFromLoginSource : subjects) {
+			if (subjectFromLoginSource == null) {
+				log.error("Null value in the subjects list. Skipping.");
+				continue;
+			}
 			String login = subjectFromLoginSource.get("login");
 			// Skip subjects, which doesn't have login
 			if (login == null || login.isEmpty()) {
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -137,6 +144,69 @@ public static String join(Iterable<?> collection, String separator) {
 		return oBuilder.toString();
 	}
 
+	/**
+	 * Returns additionalUserExtSource according to the attrName
```
stavamichal: This method need to be documented little more. There need to be said that it is used for purpose of synchronization form different ExtSources and that subjectFromLoginSource came from the ExtSource itself.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -137,6 +144,69 @@ public static String join(Iterable<?> collection, String separator) {
...
+	 *
+	 * @param attrName name of the attribute
```
stavamichal: Of which attribute?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -137,6 +144,69 @@ public static String join(Iterable<?> collection, String separator) {
...
+	 * @param subjectFromLoginSource map with the subject
+	 * @param sess perun session
+	 * @param perunBl PerunBl to get ExtSourceManagerBl
+	 * @return UserExtSource if exists, null if parameter attrName or subjectFromLoginSource is null or attrName doesn't start with ExtSourcesManagerImpl.USEREXTSOURCEMAPPING
+	 * @throws InternalErrorException
+	 */
+	public static UserExtSource getAdditionalUserExtSource(String attrName, Map<String, String> subjectFromLoginSource, PerunSession sess, PerunBl perunBl) throws InternalErrorException {
```
stavamichal: Session is always first. You don't need to send perunBl to the method, you can find it in the perun session.

oernst9: I've managed to get it somehow but if there is a better way, please let me know.

stavamichal: Resolved in personal discussion.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -137,6 +145,68 @@ public static String join(Iterable<?> collection, String separator) {
...
+	/**
+	 * Returns additionalUserExtSource according to the attrName. It's used for synchronization from different ExtSources. subjectFromLoginSource was obtained from the login ExtSource.
+	 *
+	 * @param sess perun session
+	 * @param attrName the key for UserExtSource in subjectFromLoginSource map
+	 * @param subjectFromLoginSource map with the subject
+	 * @return UserExtSource if exists, null if parameter attrName or subjectFromLoginSource is null or attrName doesn't start with ExtSourcesManagerImpl.USEREXTSOURCEMAPPING
+	 * @throws InternalErrorException
+	 */
+	public static UserExtSource getAdditionalUserExtSource(PerunSession sess, String attrName, Map<String, String> subjectFromLoginSource) throws InternalErrorException {
```
balcirakpeter: Please change the method so it extracts and returns all additional user extSources from the subjectFromLoginSource. It will be maybe less effective but there will be no need to traverse through the Map each time we want to use this method, so the code will be cleaner.
Change also the name of the method to something like extractAdditionalUserExtSourcesFromSubject, because get method should just return something from an object but this method creates objects and returns them.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -137,6 +144,70 @@ public static String join(Iterable<?> collection, String separator) {
 		return oBuilder.toString();
 	}
 
+	/**
+	 * Returns additionalUserExtSources from the subject. It's used for synchronization from different ExtSources. subjectFromLoginSource was obtained from the login ExtSource.
```
stavamichal: Are you sure it must be subject from login source? I think it can be subject from any source.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -137,6 +144,70 @@ public static String join(Iterable<?> collection, String separator) {
...
+	 *
+	 * @param sess perun session
+	 * @param subjectFromLoginSource map with the subject
+	 * @return List<UserExtSource> all additional ExtSources from the subject, returned list will never contain null value
+	 * @throws InternalErrorException
+	 */
+	public static List<UserExtSource> extractAdditionalUserExtSources(PerunSession sess, Map<String, String> subjectFromLoginSource) throws InternalErrorException {
```
stavamichal: I would just change "subjectFromLoginSource" to "subjectFromExtSource"

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2691,9 +2690,9 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
 		for(RichMember richMember: groupMembers) {
 			idsOfUsersInGroup.put(richMember.getUserId(), richMember);
 		}
-
 		//try to find users by login and loginSource
 		for(Map<String, String> subjectFromLoginSource : subjects) {
+			if (subjectFromLoginSource == null) continue;
```
stavamichal: This can mask problem. At least log it to error or debug log please.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2714,6 +2716,23 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
 					candidate.setAttributes(new HashMap<>());
 				}
 			} catch (UserExtSourceNotExistsException | UserNotExistsException ex) {
+				List<UserExtSource> additionalUserExtSources = Utils.extractAdditionalUserExtSources(sess, subjectFromLoginSource);
```
balcirakpeter: There are some duplicities in these lines of code. Try to rewrite it without the duplicities.

PR #2394, opened 2019-09-05
LDAPC: Fixed Resource handling

zlamalp:
- Fixed Resource creation, we now include all expected facility attributes with it.
- Fixed Resource deletion, which didn't reacted on the auditer message (more beans were in the
  message than in the condition).
- Fixed schema to include entityID attribute with Facility too.

### DIFF
--- a/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
+++ b/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
@@ -5,6 +5,7 @@ import static org.springframework.ldap.query.LdapQueryBuilder.query;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 import javax.naming.Name;
 
@@ -76,13 +77,15 @@ public class PerunResourceImpl extends AbstractPerunEntry<Resource> implements P
 				);
 	}
 
-	public void addResource(Resource resource, String entityID) throws InternalErrorException {
+	public void addResource(Resource resource, Map<String,String> facilityAttributes) throws InternalErrorException {
 		addEntry(resource);
 		// get info about entityID attribute if exists
-		if(entityID != null) {
+		if (!facilityAttributes.isEmpty()) {
 			try {
 				DirContextOperations context = findByDN(buildDN(resource));
-				context.setAttributeValue(PerunAttribute.PerunAttributeNames.ldapAttrEntityID, entityID);
+				for (String ldapAttrName : facilityAttributes.keySet()) {
+					context.setAttributeValue(ldapAttrName, facilityAttributes.get(ldapAttrName));
+				}
 				ldapTemplate.modifyAttributes(context);
 			} catch (Exception e) {
 				throw new InternalErrorException(e);
--- a/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
+++ b/perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
@@ -47,50 +47,38 @@ public class PerunResourceImpl extends AbstractPerunEntry<Resource> implements P
 						PerunAttribute.PerunAttributeNames.ldapAttrCommonName,
 						PerunAttribute.REQUIRED,
 						(PerunAttribute.SingleValueExtractor<Resource>)(resource, attrs) -> resource.getName()
-						),
+				),
 				new PerunAttributeDesc<>(
 						PerunAttribute.PerunAttributeNames.ldapAttrPerunResourceId,
 						PerunAttribute.REQUIRED,
 						(PerunAttribute.SingleValueExtractor<Resource>)(resource, attrs) -> String.valueOf(resource.getId())
-						),
+				),
 				new PerunAttributeDesc<>(
 						PerunAttribute.PerunAttributeNames.ldapAttrPerunFacilityId,
 						PerunAttribute.REQUIRED,
 						(PerunAttribute.SingleValueExtractor<Resource>)(resource, attrs) -> String.valueOf(resource.getFacilityId())
-						),
+				),
 				new PerunAttributeDesc<>(
 						PerunAttribute.PerunAttributeNames.ldapAttrPerunVoId,
 						PerunAttribute.REQUIRED,
 						(PerunAttribute.SingleValueExtractor<Resource>)(resource, attrs) -> String.valueOf(resource.getVoId())
-						),
+				),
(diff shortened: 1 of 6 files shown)


mvocu:
I would like to see this handled in more general way, specifically:
  - add overloaded method `AbstractPerunEntry::addEntry(T bean, Iterable<PerunAttribute<T>> attrs)` which would then use `mapContext()` call to actually set the values (see similar method `modifyEntry()`)
  - in `addResource()` simply call the overloaded `addEntry()`
 
Also, please do not set the ldap attribute names directly from code and directly from their Perun names (they might differ) - the ldap attribute name should be obtained from the XML configuration of PerunAttributeDescription bean (and there would be no need for map as argument type).

More in the inline comments in code.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
```diff
@@ -76,13 +77,15 @@
 				);
 	}
 
-	public void addResource(Resource resource, String entityID) throws InternalErrorException {
+	public void addResource(Resource resource, Map<String,String> facilityAttributes) throws InternalErrorException {
```
mvocu: I suggest naming the second parameter only attributes - it is more generic. Also use list of PerunAttribute objects, not string values.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
```diff
@@ -76,13 +77,15 @@
...
 		addEntry(resource);
 		// get info about entityID attribute if exists
-		if(entityID != null) {
+		if (!facilityAttributes.isEmpty()) {
 			try {
 				DirContextOperations context = findByDN(buildDN(resource));
-				context.setAttributeValue(PerunAttribute.PerunAttributeNames.ldapAttrEntityID, entityID);
+				for (String ldapAttrName : facilityAttributes.keySet()) {
+					context.setAttributeValue(ldapAttrName, facilityAttributes.get(ldapAttrName));
+				}
```
mvocu: I think this is wrong. You should use ldap attribute names as configured in XML context and not suppose they are the same as their Perun counterparts.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/CreationEventProcessor.java
```diff
@@ -69,42 +79,39 @@ public void processEvent(String msg, MessageBeans beans) {
...
+	 * Get facility attributes as map of LDAP attr names to their values. Key in a map is omitted if value is empty
 	 *
 	 * @param facilityId the facilityId
-	 * @return value of entityID or null, if value is null or user not exists yet
+	 * @return value of facility attributes stored in resource or null, if value is null or facility not exists yet
 	 * @throws InternalErrorException if some exception is thrown from RPC
 	 */
-	private String getFacilityEntityIdValue(int facilityId) throws InternalErrorException {
-		Perun perun = ldapcManager.getPerunBl();
+	private Map<String,String> getFacilityAttributes(int facilityId) throws InternalErrorException {
```
mvocu: Please make this method return the list of PerunAttribute objects. This way you can use existing mechanism of mapContext/extractors to obtain the ldap attribute names and values.

zlamalp:
> I would like to see this handled in more general way, specifically:
> 
>     * add overloaded method `AbstractPerunEntry::addEntry(T bean, Iterable<PerunAttribute<T>> attrs)` which would then use `mapContext()` call to actually set the values (see similar method `modifyEntry()`)
> 
>     * in `addResource()` simply call the overloaded `addEntry()`
> 

I don't think, that we will need this for any other object than resource. Since it is the only case, when we want to push attributes of a different object with it. Also, this logic will be removed, once facility attributes will be switched to Facility object only on the side of proxy. I believe it will be sooner than later.

> 
> Also, please do not set the ldap attribute names directly from code and directly from their Perun names (they might differ) - the ldap attribute name should be obtained from the XML configuration of PerunAttributeDescription bean (and there would be no need for map as argument type).

In order to do it, I would need to make some methods of ``PerunResourceImpl`` and ``AbstractPerunEntry`` public, so I could use them outside of the entry object in ``CreationEventProcessor`` where I actually load facility attributes. Since I can't load data from Perun inside the ``PerunResourceImpl`` bean.

Technically I like your suggestion/solution, but due to its removal in a near future, I don't want re-done whole logic.

When I look more closely, I think I could actually use ``beginSynchronizeEntry()`` and ``commitSyncOperation()``. It technically synces passed list of Attributes, which I have directly loaded in ``CreationEventProcessor`` and it can be publicly used. So I don't have to add any new method to the ``PerunResourceImpl`` itself or ``AbstractPerunEntry``.

I will test it and push it, if it works. I need to deploy this fix on monday, so this is also a reason, why I don't want to change the "concept", but rather "fix the bug".

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
zlamalp: I removed the second param all together.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/impl/PerunResourceImpl.java
zlamalp: I no longer have specific addResource() method, I use sync operation instead.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/CreationEventProcessor.java
zlamalp: It now returns ``List<Attribute>`` which is passed to the sync operation, which was already implemented.

perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/processor/impl/CreationEventProcessor.java
```diff
@@ -34,32 +36,35 @@ public void processEvent(String msg, MessageBeans beans) {
 					log.debug("Adding new group: {}", beans.getGroup());
 					perunGroup.addGroup(beans.getGroup());
 					break;
-					
+
 				case MessageBeans.RESOURCE_F:
 					log.debug("Adding new resource: {}", beans.getResource());
-					perunResource.addResource(beans.getResource(), getFacilityEntityIdValue(beans.getResource().getFacilityId()));
+					perunResource.addResource(beans.getResource());
```
mvocu: Now this is probably not necessary - sync will create the entry anyway.

PR #2397, opened 2019-09-09
Make the external configuration file optional, add extension points to configuration

mvocu:
Instead of rewriting beans (`perunGroup`, `perunUser` etc.) to extend them it is now possible to
define just the additional attributes using the `perunGroupConfigurer` etc. beans:
- adds configuration beans infrastructure
- adds wildcard to spring context import (making the imported file optional)
- get extended attribute configuration from configuration beans

### DIFF
--- a/perun-ldapc/src/main/resources/perun-ldapc.xml
+++ b/perun-ldapc/src/main/resources/perun-ldapc.xml
@@ -613,6 +613,7 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 				</bean>
 			</list>
 		</property>
+		<property name="attributeDescriptionsExt" ref="perunUserAttributesExt" />
 	</bean>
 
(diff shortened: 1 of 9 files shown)


perun-ldapc/src/main/resources/perun-ldapc.xml
```diff
@@ -706,6 +726,6 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 
 	<ldap:ldap-template />
 
-	<import resource="file:${perun.conf.custom}/perun-ldapc-attributes.xml" />
+	<import resource="file:${perun.conf.custom}/perun-ldapc-*.xml" />
```
zlamalp: Are there any other expected file names ? Because we have empty ``perun-ldapc-attributes.xml`` present in our deployment template. I believe we can still use it to re-define all attribute configurers.

PR #2408, opened 2019-09-23
Various changes in SCIM module

oernst9:
* Replaced inheritance with composition in GroupSCIM which extended Resource object

* Getters which return collections return unmodifiable lists in UserSCIM now

* Removed wild card in imports in AbstractSCIMTest.java

* method getAllResourceTypes() in ResourceTypesEndpointController.java changed to static

* closing PrintWriter in Api.java moved to finally block

* edited catch block in UserResourceEndpointController.java method getEmail() so that the error log contains different message for different exception. Specifically WrongAttributeAssignmentException and InternalErrorException

### DIFF
--- a/perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/GroupSCIM.java
+++ b/perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/GroupSCIM.java
@@ -1,5 +1,6 @@
 package cz.metacentrum.perun.scim.api.entities;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 
@@ -13,7 +14,10 @@ import java.util.List;
  */
 
 @JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
-public class GroupSCIM extends Resource {
+public class GroupSCIM {
+
+	@JsonIgnore
+	private Resource resource;
 
 	@JsonProperty
 	private List<String> schemas;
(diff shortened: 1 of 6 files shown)


perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/GroupSCIM.java
```diff
@@ -13,7 +13,10 @@
  */
 
 @JsonSerialize(include = JsonSerialize.Inclusion.NON_NULL)
-public class GroupSCIM extends Resource {
+public class GroupSCIM {
```
zlamalp: When I look at it now, it seems, that both ``GroupSCIM`` and ``UserSCIM`` are DTO objects. So regarding our original request "we should use composition and not extension", we must make sure:
- we align also ``UserSCIM`` to use composition and not extension.
- we must prevent ``getResource()`` and ``setResource(Resource)`` to be (de)serialized to JSON, since former implementation has no property "resource" but instead its internal properties (id,externalId,meta) as a part of to level object (group/user).

perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/GroupSCIM.java
```diff
@@ -62,18 +75,59 @@ public void setMembers(List<MemberSCIM> members) {
 		this.members = members;
 	}
 
+	@JsonSetter("id")
```
balcirakpeter: Please remove JsonSetter and getter annotations from both GroupScim and UserScim classes. They are unnecessary because Json serializer can determine these fields automatically. Someone could be later confused why are these annotations used.

oernst9: Ok, I wasn't sure whether it is necessary or not

perun-scim/src/main/java/cz/metacentrum/perun/scim/api/entities/UserSCIM.java
```diff
@@ -41,14 +53,19 @@ public UserSCIM(Long id, Long externalId, Meta meta, List<String> schemas, Strin
...
+		resource.setId(id);
+		resource.setExternalId(externalId);
+		resource.setMeta(meta);
 	}
 
 	public UserSCIM() {
+		resource = new Resource();
 	}
 
+	@JsonIgnore
```
balcirakpeter: We want to ignore just resource and its getters and setters, so please remove JsonIgnore annotation for getSchemas() and also for getEmails().

PR #2417, opened 2019-09-30
Addition of RichUserExtSource

Vojtech-Sassmann:
* We need an easy way of accesing the userExtSources with theirs
attributes. This now can be done using the new RichUserExtSource object.

* Added new method for accessing all users RichUserExtSources.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
@@ -394,6 +394,32 @@ public interface UsersManager {
 	List<UserExtSource> getUserExtSources(PerunSession perunSession, User user) throws InternalErrorException, UserNotExistsException, PrivilegeException;
 
 	/**
+	 * Gets list of all user's external sources with attributes.
+	 *
+	 * @param perunSession session
+	 * @param user user for which should be returned rich ext sources
+	 * @return list of user's external sources with attributes
+	 * @throws InternalErrorException internal error
+	 * @throws UserNotExistsException if given user doesn't exist
+	 * @throws PrivilegeException insufficient permissions
+	 */
+	List<RichUserExtSource> getRichUserExtSources(PerunSession perunSession, User user) throws InternalErrorException, UserNotExistsException, PrivilegeException;
+
+	/**
+	 * Gets list of all user's external sources with attributes. If any of the attribute names is incorrect
+	 * then the value is silently skipped. If the attrsNames is null, then this method returns all ues attributes.
+	 *
+	 * @param perunSession session
+	 * @param user user for which should be returned rich ext sources
+	 * @param attrsNames list of attribute names that should be found
+	 * @return list of user's external sources with attributes
+	 * @throws InternalErrorException internal error
+	 * @throws UserNotExistsException if given user doesn't exist
+	 * @throws PrivilegeException insufficient permissions
+	 */
+	List<RichUserExtSource> getRichUserExtSources(PerunSession perunSession, User user, List<String> attrsNames) throws InternalErrorException, UserNotExistsException, PrivilegeException;
+
+	/**
 	 * Get the user ext source by its id.
 	 *
 	 * @param sess
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -31,6 +31,7 @@ import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.RichResource;
 import cz.metacentrum.perun.core.api.RichUser;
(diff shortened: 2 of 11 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -508,6 +509,26 @@ public UserExtSource updateUserExtSource(PerunSession sess, UserExtSource userEx
...
+			!AuthzResolver.isAuthorized(sess, Role.ENGINE) &&
+			!AuthzResolver.isAuthorized(sess, Role.RPC) &&
+			!AuthzResolver.isAuthorized(sess, Role.SELF, user) &&
+			!AuthzResolver.isAuthorized(sess, Role.PERUNOBSERVER)) {
+			throw new PrivilegeException(sess, "getRichUserExtSources");
+		}
+
+		getUsersManagerBl().checkUserExists(sess, user);
+
+		return getUsersManagerBl().getRichUserExtSources(sess, user);
```
zlamalp: We should implement ``filterOnlyAllowedAttributes()`` method to perform authorization check on list of ``RichUserExtSource`` as we have for groups and members. It will be part of Bl layers and called from here.

Also, I don't know if we have authorization check on ues attributes. I believe, that right now they are accessible only to perun admin, and will need to configure it on each instance on deployment.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -393,6 +393,18 @@
...
+	 * Gets list of all user's external sources with attributes.
+	 *
+	 * @param perunSession session
+	 * @param user user for which should be returned rich ext sources
+	 * @return list of user's external sources with attributes
+	 * @throws InternalErrorException internal error
+	 * @throws UserNotExistsException if given user doesn't exist
+	 * @throws PrivilegeException insufficient permissions
+	 */
+	List<RichUserExtSource> getRichUserExtSources(PerunSession perunSession, User user) throws InternalErrorException, UserNotExistsException, PrivilegeException;
```
zlamalp: As we talked, please add also version, where you can specify list of attribute names to retrieve with each RichUserExtSource.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -496,6 +496,35 @@ public UserExtSource call(ApiCaller ac, Deserializer parms) throws PerunExceptio
 		}
 	},
 
+	/*#
+	 * Gets list of all user's external sources with attributes. If attrsNames is empty
```
zlamalp: If this version of the method doesn't specify ``attrsNames`` parameter, you shouldn't mention it in the general text.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -496,6 +496,35 @@ public UserExtSource call(ApiCaller ac, Deserializer parms) throws PerunExceptio
...
+	 * return all attributes. If attrsNames is null, this methods returns all attributes.
+	 *
+	 * @param user int User <code>id</code>
+	 * @return List<UserExtSource> list of user's external sources with attributes
+	 */
+	/*#
+	 * Gets list of all user's external sources with specified attributes.
```
zlamalp: You should mention the possibility about null value for ``attrsNames`` here. It will return all attributes with non-null value. Empty list will return no attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -639,6 +641,22 @@ public void updateUserExtSourceLastAccess(PerunSession sess, UserExtSource userE
...
+			.map(ues -> new RichUserExtSource(ues,
+				attrsNames == null ? getPerunBl().getAttributesManagerBl().getAttributes(sess, ues)
+					: getPerunBl().getAttributesManagerBl().getAttributes(sess, ues, attrsNames)))
+			.collect(Collectors.toList());
+	}
+
+	@Override
+	public List<RichUserExtSource> filterOnlyAllowedAttributesForRichUserExtSources(PerunSession sess, List<RichUserExtSource> richUserExtSources) {
+		richUserExtSources.forEach(rues -> rues.setAttributes(
+			AuthzResolverBlImpl.filterNotAllowedAttributes(sess, rues.asUserExtSource(), rues.getAttributes())));
```
zlamalp: It doesn't have to be part of this pull-request, but asking for authorization on all attributes can be really time consuming operation. We could "cache" results of authorization for one caller and all UserExtSources of single user. Like shared context, as we do in filtering attributes for RichMember and Groups. Because if you are allowed to read "cn" attribute of UserExtSource for one user, you have same authorization for the rest of his UserExtSources.

Vojtech-Sassmann:
Hi @zlamalp, I have fixed the documentation. Regarding the privileges validation optimization, I think it should be done in a separate PR as you mentioned.

PR #2419, opened 2019-09-30
new facility attributes

mvocu:
remove facility attributes from resource entries and set them for facility entries
add new facility attributes

### DIFF
--- a/perun-utils/ldapc-scripts/schemas/perun-schema.ldif
+++ b/perun-utils/ldapc-scripts/schemas/perun-schema.ldif
@@ -31,12 +31,19 @@ olcAttributeTypes: {26}( 1.3.6.1.4.1.8057.2.80.30 NAME 'loa' DESC 'Level of auth
 olcAttributeTypes: {27}( 1.3.6.1.4.1.8057.2.80.31 NAME 'internalUserIdentifiers' DESC 'All identifiers collected from UserExtSource attribute additionalIdentifiers' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 olcAttributeTypes: {28}( 1.3.6.1.4.1.8057.2.80.32 NAME 'perunFacilityDn' DESC 'DN of Facility the resource belongs to' EQUALITY distinguishedNameMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 SINGLE-VALUE )
 olcAttributeTypes: {29}( 1.3.6.1.4.1.8057.2.80.35 NAME 'eduPersonOrcid' DESC 'All OrcID identifiers of user' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {30}( 1.3.6.1.4.1.8057.2.80.36 NAME 'checkGroupMembership' DESC 'OIDC check group membership' EQUALITY caseExactMatch SYNTAX  1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {31}( 1.3.6.1.4.1.8057.2.80.37 NAME 'voShortNames' DESC 'Short names of assigned VOs' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {32}( 1.3.6.1.4.1.8057.2.80.38 NAME 'wayfFilter' DESC 'WAYF filter for the facility' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {33}( 1.3.6.1.4.1.8057.2.80.39 NAME 'wayfEFilter' DESC 'WAYF filter for the facility' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {34}( 1.3.6.1.4.1.8057.2.80.40 NAME 'isTestSp' DESC 'is test SP?' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {35}( 1.3.6.1.4.1.8057.2.80.41 NAME 'proxyIdentifiers' DESC 'Proxy identifiers' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {36}( 1.3.6.1.4.1.8057.2.80.42 NAME 'showOnServiceList' DESC 'show on service list?' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 -
 replace: olcObjectClasses
 olcObjectClasses: {0}( 1.3.6.1.4.1.8057.2.80.4 NAME 'perunUser' DESC 'User managed by Perun' SUP inetOrgPerson STRUCTURAL MUST ( perunUserId $ isServiceUser  $ isSponsoredUser ) MAY ( preferredMail $ userCertificateSubject $ uidNumber $ login $ eduPersonPrincipalNames $ userPassword $ memberOfPerunVo $ libraryIDs $ schacHomeOrganizations $ eduPersonScopedAffiliations $ bonaFideStatus $ groupNames $ institutionsCountries $ isCesnetEligible $ loa $ internalUserIdentifiers $ eduPersonOrcid ) )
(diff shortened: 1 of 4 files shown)


perun-utils/ldapc-scripts/schemas/perun-schema.ldif
```diff
@@ -31,12 +31,19 @@ olcAttributeTypes: {26}( 1.3.6.1.4.1.8057.2.80.30 NAME 'loa' DESC 'Level of auth
...
+olcObjectClasses: {2}( 1.3.6.1.4.1.8057.2.80.15 NAME 'perunResource' DESC 'Resource managed by Perun' SUP top STRUCTURAL MUST ( cn $  perunResourceId $ perunVoId $ perunFacilityId ) MAY (uniqueMember $ businessCategory $ seeAlso $ owner $ ou $ o $ description $ assignedGroupId $ perunFacilityDn ))
 olcObjectClasses: {3}( 1.3.6.1.4.1.8057.2.80.6 NAME 'perunVO' DESC 'VO managed by Perun' SUP organization STRUCTURAL MUST perunVoId MAY uniqueMember )
-olcObjectClasses: {4}( 1.3.6.1.4.1.8057.2.80.34 NAME 'perunFacility' DESC 'Facility managed by Perun' SUP top STRUCTURAL MUST ( cn $ perunFacilityId ) MAY ( description $ entityID $ OIDCClientID ) )
+olcObjectClasses: {4}( 1.3.6.1.4.1.8057.2.80.34 NAME 'perunFacility' DESC 'Facility managed by Perun' SUP top STRUCTURAL MUST ( cn $ perunFacilityId ) MAY ( description $ entityID $ OIDCClientID $ checkGroupMembership $ voShortNames $ wayfFilter $ wayfEFilter $ $ isTestSp $ proxyIdentifiers $ showOnServiceList ) )
```
zlamalp: Doubled ``$`` in ``wayfEFilter $ $ isTestSp``.

perun-utils/ldapc-scripts/schemas/perun-schema.ldif
```diff
@@ -31,12 +31,19 @@ olcAttributeTypes: {26}( 1.3.6.1.4.1.8057.2.80.30 NAME 'loa' DESC 'Level of auth
...
 olcAttributeTypes: {29}( 1.3.6.1.4.1.8057.2.80.35 NAME 'eduPersonOrcid' DESC 'All OrcID identifiers of user' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {30}( 1.3.6.1.4.1.8057.2.80.36 NAME 'checkGroupMembership' DESC 'OIDC check group membership' EQUALITY booleanMatch SYNTAX  1.3.6.1.4.1.1466.115.121.1.7 )
+olcAttributeTypes: {31}( 1.3.6.1.4.1.8057.2.80.37 NAME 'voShortNames' DESC 'Short names of assigned VOs' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {32}( 1.3.6.1.4.1.8057.2.80.38 NAME 'wayfFilter' DESC 'WAYF filter for the facility' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {33}( 1.3.6.1.4.1.8057.2.80.39 NAME 'wayfEFilter' DESC 'WAYF filter for the facility' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {34}( 1.3.6.1.4.1.8057.2.80.40 NAME 'isTestSp' DESC 'is test SP?' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
```
zlamalp: This attribute is actually also boolean type in perun (it was wrongly defined in the issue). We can then map it to the boolean syntax in LDAP and not string.

perun-utils/ldapc-scripts/schemas/perun-schema.ldif
```diff
@@ -31,12 +31,19 @@ olcAttributeTypes: {26}( 1.3.6.1.4.1.8057.2.80.30 NAME 'loa' DESC 'Level of auth
 olcAttributeTypes: {27}( 1.3.6.1.4.1.8057.2.80.31 NAME 'internalUserIdentifiers' DESC 'All identifiers collected from UserExtSource attribute additionalIdentifiers' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 olcAttributeTypes: {28}( 1.3.6.1.4.1.8057.2.80.32 NAME 'perunFacilityDn' DESC 'DN of Facility the resource belongs to' EQUALITY distinguishedNameMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.12 SINGLE-VALUE )
 olcAttributeTypes: {29}( 1.3.6.1.4.1.8057.2.80.35 NAME 'eduPersonOrcid' DESC 'All OrcID identifiers of user' EQUALITY caseExactMatch SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+olcAttributeTypes: {30}( 1.3.6.1.4.1.8057.2.80.36 NAME 'checkGroupMembership' DESC 'OIDC check group membership' EQUALITY booleanMatch SYNTAX  1.3.6.1.4.1.1466.115.121.1.7 )
```
zlamalp: Boolean syntax for attribute seems to be problematic, since from perun/java will provide "true", "false", but ldap syntax expect "TRUE" and "FALSE". We will need to provide own ``ValueTransformer`` for such attributes (java+context usage) or handle it more generally in ``SingleAttributeValueExtractor`` for all boolean atributes.
Also using plain string values in LDAP is possible, but it might be more desirable to use boolean. I will discuss it internally and let you know.

perun-utils/ldapc-scripts/schemas/perun-schema.ldif
zlamalp: Not relevant anymore.

perun-utils/ldapc-scripts/schemas/perun-schema.ldif
zlamalp: Ok, I realized, that in Perun, boolean attributes are either "true" or null, since we do not store false values. False can be probably returned only from virtual attributes, but in audit messages no event "set to false" is logged. It is converted to null and stored as "removed".

So please change all boolean attributes to string syntax in ldap with single-value. Proxy will check such attributes on presence. Thank you.

perun-utils/ldapc-scripts/schemas/perun-schema.ldif
mvocu: Fixed.

perun-utils/ldapc-scripts/schemas/perun-schema.ldif
mvocu: Changed syntax and matching rules to string for relevant attributes.

mvocu:
Není to nějaký renonc z lokálního perun-ldapc-attributes.xml? V současném kódu nikde u Resource nastavení entityID nevidím (v ResourceSynchronizer ani žádné atributy vyjmenované nejsou).

zlamalp:
> Isn't it some remnant from local perun-ldapc-attributes.xml? There is no mention of entityID attribute with Resource in current code (and ResourceSynchronizer does not contain list of attributes itself).

Well, if I pull your branch on current master, I can see in the content of ``ResourceSynchronizer`` code, where you retrieve Facility object and then its attributes and pass it to ``synchronizeResource()``.
We now expect to store only resource attributes within Resource object in LDAP. So this implementation needs to be adjusted to work with resource attributes from perun.

Above probably isn't related to reported issue (SchemaViolationException) but it wouldn't synchronize any resource attributes (mismatch in passed names and used method params).

I do tests against dev instance and there is no custom xml. You can build and deploy ldapc from your branch on devel and restart the service to test it.

When I update schema and deploy new version and sync it, I expect it to correct LDAP content. If current implementation of sync keeps existing entry attributes, which are no longer defined in context or allowed by schema, we should somehow removed them from the java object (entry) before we call update to LDAP. The fact, that it probably doesn't work that way now, caused mentioned SchemaViolationException.

mvocu:
Fixed ResourceSynchronizer (though the code was incorrect, it could not have influenced the result - mapping of attributes is always done according to the list of attributes defined in the respective ldap model bean).

As for the schema changes, I do not agree it is the responsibility of LDAPc to remove attributes that it does not know about; that could lead to serious and undesirable side effects in potential installations. 

If the schema is updated, it should be responsibility of updater (process or person) to remove non-conforming data from the tree.

mvocu:
Rebased.

PR #2420, opened 2019-09-30
Audit log for sending invitations to VO/group

oernst9:
* created new class InvitationSentEvent extending AuditEvent

* logged InvitationSentEvent right after the mail has been successfully sent

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/audit/events/MailManagerEvents/InvitationSentEvent.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/audit/events/MailManagerEvents/InvitationSentEvent.java
@@ -1,0 +1,62 @@
+package cz.metacentrum.perun.audit.events.MailManagerEvents;
+
+import cz.metacentrum.perun.audit.events.AuditEvent;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.Vo;
+
+public class InvitationSentEvent extends AuditEvent {
+
+	private User sender;
+	private String email;
+	private String language;
+	private String message;
+	private Vo vo;
+	private Group group;
+
+	@SuppressWarnings("unused") // used by jackson mapper
+	public InvitationSentEvent() {
+	}
+
+	public InvitationSentEvent(User user, String mail, String language, Group group, Vo vo) {
+		this.sender = user;
+		this.email = mail;
+		this.language = language;
+		this.group = group;
+		this.vo = vo;
+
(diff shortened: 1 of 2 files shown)


perun-base/src/main/java/cz/metacentrum/perun/audit/events/MailManagerEvents/InvitationSentEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+	private User sender;
+	private String email;
+	private String language;
+	private String message;
+
+	public InvitationSentEvent() {
+	}
+
+	public InvitationSentEvent(User user, String mail, String language) {
```
zlamalp: Can you please add also VO and Group object to the event ? We need to know, into which entity was somebody invited.

perun-base/src/main/java/cz/metacentrum/perun/audit/events/MailManagerEvents/InvitationSentEvent.java
```diff
@@ -0,0 +1,45 @@
...
+import cz.metacentrum.perun.core.api.User;
+
+public class InvitationSentEvent extends AuditEvent {
+
+	private User sender;
+	private String email;
+	private String language;
+	private String message;
+
+	public InvitationSentEvent() {
```
zlamalp: Other audit event classes have following annotation above empty constructor:
```java
@SuppressWarnings("unused") // used by jackson mapper
```

perun-base/src/main/java/cz/metacentrum/perun/audit/events/MailManagerEvents/InvitationSentEvent.java
```diff
@@ -0,0 +1,62 @@
...
+
+	public InvitationSentEvent(User user, String mail, String language, Group group, Vo vo) {
+		this.sender = user;
+		this.email = mail;
+		this.language = language;
+		this.group = group;
+		this.vo = vo;
+
+		this.message = formatMessage("Invitation to VO: %s, group: %s, sent by user: %s, to email: %s, in language: %s",
+			vo, group == null ? "None" : group, sender, email, language);
```
zlamalp: You don't have to pass string "None", group will resolve to null in resulting message. Since we still sometimes parse/check messages, you should probably omit ``VO:`` and ``group:`` before objects, since they will be resolved to something like ``VO: Vo:[....]``.
I suggest to modify message based on passed VO/group: 
```java
formatMessage("Invitation to %s, sent by %s, to email: %s, in language: %s.", (group != null) ? group : vo, sender, email, language);
```

PR #2424, opened 2019-10-02
added rpc method: deleteAttributes

DanoFecko:
* before there was only method `deleteAttribute` and GUI needed to do multiple calls to rpc
* `deleteAttributes` takes list of attribute definitions to delete

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
@@ -1370,10 +1370,8 @@ public enum AttributesManagerMethod implements ManagerMethod {
 	},
 
 	/*#
-	 * Deletes attribute definition from Perun.
-	 *
-	 * Deletion fails if any entity in Perun has
-	 * any value for this attribute set.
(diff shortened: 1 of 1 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -1389,6 +1389,27 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Deletes multiple attribute definitions from Perun.
+	 *
+	 * Deletion fails if any entity in Perun has
+	 * any value for any of these attributes set.
+	 *
+	 * @param attributes List<int> AttributeDefinition <code>id</code>
```
zlamalp: This method actually deletes AttributeDefinition including all values set for any entity. Can you please update javadoc accordingly and fix the same docs for deleteAttribute() ? Also in GUI please include big warning about it. Perun admin must know what he does. Thank you.

DanoFecko: @zlamalp Done

PR #2425, opened 2019-10-02
updated Spring Boot Starter Parent to 2.1.9 / Spring 5.1.10

martin-kuba:


martin-kuba:
Travis cache of Maven repo does not contain a needed file. This is a local Travis problem, but we cannot solve it. Closing PR.

PR #2429, opened 2019-10-07
Javadoc for exceptions

oernst9:
* added javadoc for exceptions explaining when they are thrown

* added javadoc for constructors and getters of the exceptions so that it is easy to find out how to use them


oernst9:
resolving the problem

PR #2434, opened 2019-10-10
Testing functionality of cache

metodej:
- Cache is now enabled.
- All attributes are now retrieved from cache as well as from db and the results are compared. If results do not match, both versions of attributes and some additional information is logged. The methods always return values from db.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
@@ -852,13 +852,8 @@ public class AttributesManagerImpl implements AttributesManagerImplApi {
 							"where namespace=? or (namespace in (?,?) and (attr_value is not null or attr_value_text is not null))",
 					new SingleBeanAttributeRowMapper<>(sess, this, facility), facility.getId(), AttributesManager.NS_FACILITY_ATTR_CORE, AttributesManager.NS_FACILITY_ATTR_DEF, AttributesManager.NS_FACILITY_ATTR_OPT);
 
-			if (attributesCache != null) {
-				Collections.sort(attributesDB);
-				Collections.sort(attributesCache);
-				if (!attributesDB.equals(attributesCache)) {
-					log.error("The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility);
-				}
-			}
+			String message = "The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility;
+			compareAttributesFromCacheAndDB(attributesCache, attributesDB, message);
 
 			return attributesDB;
 		} catch (EmptyResultDataAccessException ex) {
@@ -900,13 +895,8 @@ public class AttributesManagerImpl implements AttributesManagerImplApi {
 							"where namespace in ( :nSC,:nSO,:nSD,:nSV ) and attr_names.attr_name in ( :attrNames )",
 					parameters, new SingleBeanAttributeRowMapper<>(sess, this, facility));
 
-			if (attributesCache != null) {
-				Collections.sort(attributesDB);
-				Collections.sort(attributesCache);
-				if (!attributesDB.equals(attributesCache)) {
-					log.error("The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility + " and their names are: " + attrNames);
-				}
-			}
+			String message = "The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility + " and their names are: " + attrNames;
+			compareAttributesFromCacheAndDB(attributesCache, attributesDB, message);
 
 			return attributesDB;
 		} catch (EmptyResultDataAccessException ex) {
@@ -937,13 +927,8 @@ public class AttributesManagerImpl implements AttributesManagerImplApi {
 		}
 		List<Attribute> attributesDB = getVirtualAttributes(new SingleBeanAttributeRowMapper<>(sess, this, facility), AttributesManager.NS_FACILITY_ATTR_VIRT);
 
-		if (attributesCache != null) {
-			Collections.sort(attributesDB);
-			Collections.sort(attributesCache);
-			if (!attributesDB.equals(attributesCache)) {
-				log.error("The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility);
-			}
-		}
+		String message = "The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility;
+		compareAttributesFromCacheAndDB(attributesCache, attributesDB, message);
 
 		return attributesDB;
 	}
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -1645,22 +2092,33 @@ public Object mapRow(ResultSet rs, int i) throws SQLException {
 
 	@Override
 	public String getEntitylessAttrValueForUpdate(PerunSession sess, int attrId, String key) throws InternalErrorException, AttributeNotExistsException {
```
stavamichal: I am little afraid of this one. This specific one using "SELECT FOR UPDATE" in SQL, not sure how it should work, if both cache and DB are working together. We should think about this one.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -839,17 +839,28 @@ public Object mapRow(ResultSet rs, int i) throws SQLException {
...
 		}
 
 		try {
-			return jdbc.query("select " + getAttributeMappingSelectQuery("fac") + " from attr_names " +
+			List<Attribute> attributesDB = jdbc.query("select " + getAttributeMappingSelectQuery("fac") + " from attr_names " +
 							"left join facility_attr_values fac    on id=fac.attr_id and fac.facility_id=? " +
 							"where namespace=? or (namespace in (?,?) and (attr_value is not null or attr_value_text is not null))",
 					new SingleBeanAttributeRowMapper<>(sess, this, facility), facility.getId(), AttributesManager.NS_FACILITY_ATTR_CORE, AttributesManager.NS_FACILITY_ATTR_DEF, AttributesManager.NS_FACILITY_ATTR_OPT);
 
+			if (attributesCache != null) {
```
zlamalp: I would like to see this as some private method. It is repeated too many times. You can always initialize ``attributesCache`` to empty list, since as far as I looked, cache methods return empty list if nothing is found and not null. You can also generally use ``Objects.equals(a,b)`` and don't care about null in any of the variables.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -839,17 +839,28 @@ public Object mapRow(ResultSet rs, int i) throws SQLException {
...
+				Collections.sort(attributesDB);
+				Collections.sort(attributesCache);
+				if (!attributesDB.equals(attributesCache)) {
+					log.error("The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility);
```
zlamalp: You should use:
```java
log.error("text {} text {}", var1, var2);
```
instead of 
```java
log.error("text " + var1 + "text " + var2);
```
otherwise objects are always printed, even if we are not logging this class or not logging in ERROR level.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -839,17 +839,23 @@ public Object mapRow(ResultSet rs, int i) throws SQLException {
...
 		}
 
 		try {
-			return jdbc.query("select " + getAttributeMappingSelectQuery("fac") + " from attr_names " +
+			List<Attribute> attributesDB = jdbc.query("select " + getAttributeMappingSelectQuery("fac") + " from attr_names " +
 							"left join facility_attr_values fac    on id=fac.attr_id and fac.facility_id=? " +
 							"where namespace=? or (namespace in (?,?) and (attr_value is not null or attr_value_text is not null))",
 					new SingleBeanAttributeRowMapper<>(sess, this, facility), facility.getId(), AttributesManager.NS_FACILITY_ATTR_CORE, AttributesManager.NS_FACILITY_ATTR_DEF, AttributesManager.NS_FACILITY_ATTR_OPT);
 
+			String message = "The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + facility;
```
zlamalp: This is even less effective than previously. Right now you serialize all objects all the time, not just when attributes doesn't match. You can pass string with placeholders and array of objects like This:

```java
// changed private method
private void compareAttributesFromCacheAndDB(List<Attribute> attributesCache, List<Attribute> attributesDB, String message, Object... messageParams)

// generic strings with placeholders
String message = "The attributes do not match, attributes from cache are: {} while attributes from db are: {} the attributes belong to {}";

// call of private method
compareAttributesFromCacheAndDB(attributesCache, attributesDB, message, facility);

// implementation in private method
log.error(message, attributesCache, attributesDb, messageParams);
```

You can probably even simplify it by having message fixed inside private methods and then care only about passed messageParams (which will be perun beans). It doesn't really matter if the text will end like: "_attributes belong to [[user], [facility]]_".
You could also append placeholders dynamically to the string base on the length of "messageParams" (as I mention in another comment). Then it wouldn't be placed as an array into the message. It might be necessary, if we want to pass exception as the last param to get stack trace from it.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -2159,13 +2631,20 @@ public Attribute getAttributeById(PerunSession sess, Group group, int id) throws
...
 			Attribute attr = perun.getCacheManager().getAttributeById(id, new Holder(host.getId(), Holder.HolderType.HOST), null);
-			return setValueOfAttribute(sess, attr, host, null);
+			attributesCache = setValueOfAttribute(sess, attr, host, null);
 		}
 
 		try {
-			return jdbc.queryForObject("select " + getAttributeMappingSelectQuery("host_attr_values") + " from attr_names left join host_attr_values on id=attr_id and host_id=? where id=?", new SingleBeanAttributeRowMapper<>(sess, this, host), host.getId(), id);
+			Attribute attributesDB = jdbc.queryForObject("select " + getAttributeMappingSelectQuery("host_attr_values") + " from attr_names left join host_attr_values on id=attr_id and host_id=? where id=?", new SingleBeanAttributeRowMapper<>(sess, this, host), host.getId(), id);
+
+			String message = "The attributes do not match, attribute from cache is: " + attributesCache + " while attribute from db is: " + attributesDB + " the attribute belongs to " + host + " and its id is " + id;
```
zlamalp: You probably doesn't have to specify ID of attribute in the log message, because that is the identifier, by which we retrieve it from the cache/DB. Then if results are not the same, at lease one must not be empty and hence logged attributes will contain attribute id.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -2971,19 +3544,26 @@ public void deleteAllAttributeAuthz(PerunSession sess, AttributeDefinition attri
...
 		try {
-			return jdbc.query("select " + getAttributeMappingSelectQuery("host") + " from attr_names " +
+			List<Attribute> attributesDB = jdbc.query("select " + getAttributeMappingSelectQuery("host") + " from attr_names " +
 							"left join      host_attr_values   host    on      attr_names.id=host.attr_id    and   host_id=? " +
 							"where namespace in (?,?,?) " +
 							"and attr_names.id in (select distinct service_required_attrs.attr_id from service_required_attrs " +
 							"join resource_services on service_required_attrs.service_id=resource_services.service_id and resource_services.resource_id=?)",
 					new SingleBeanAttributeRowMapper<>(sess, this, host), host.getId(), AttributesManager.NS_HOST_ATTR_CORE, AttributesManager.NS_HOST_ATTR_DEF, AttributesManager.NS_HOST_ATTR_OPT, resourceToGetServicesFrom.getId());
+
+			String message = "The attributes do not match, attributes from cache are: " + attributesCache + " while attributes from db are: " + attributesDB + " the attributes belong to " + host + " from service with resource: " + resourceToGetServicesFrom;
```
zlamalp: The resource only specify, which services required attributes will be retrieved for the host. I'm not sure we need to log that. It somehow complicate my idea about generic message in private method with passed params, since you need to know their number. You can probably check its length and create placeholders in the message accordingly.

When I think about it, it would be more interesting to print stack trace, to identify which method failed on this check. I'm not sure, if there is any other way, than throwing and catching exception inside the private method to get stack trace.

metodej:
Thank you for review, I hope the log message works better now.

zlamalp:
As we decided to remove the cache in  #2786, this pull-request is now obsolete. Closing.

PR #2438, opened 2019-10-15
Initialization of roles from configuration file

balcirakpeter:
- roles will be loaded into database from configuration file
  perun-roles.yml instead of the enum class role.
- PerunRolesLoader component was created for this purpose which will be
  called from AuthzResolverImpl initialize() method. PerunRolesLoader
  will be later used also for loading policies.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
@@ -1,0 +1,67 @@
+package cz.metacentrum.perun.core.impl;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.core.JdbcPerunTemplate;
+import org.springframework.core.io.Resource;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
```diff
@@ -0,0 +1,61 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.core.JdbcPerunTemplate;
+import org.springframework.core.io.Resource;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+public class PerunRolesLoader {
```
zlamalp: Maybe a simple class javadoc about purpose of this class and notes about where it get configuration from.

perun-core/src/main/resources/perun-core.xml
```diff
@@ -495,6 +496,11 @@ http://www.springframework.org/schema/context http://www.springframework.org/sch
 			<property name="auditer" ref="auditer"/>
 			<property name="cacheManager" ref="cacheManager"/>
 		</bean>
+
+		<!-- PerunRolesLoader implementation -->
+		<bean id="perunRolesLoader" class="cz.metacentrum.perun.core.impl.PerunRolesLoader">
+			<property name="configurationPath" value="file:@perun.conf@perun-roles.yml"/>
```
zlamalp: Probably not necessary now, but it would be nice, if it could fallback also to ``file:${perun.conf.custom}perun-roles.yml``, since we fill default path on build and later is for runtime change by settings ENV variable ``perun.conf.custom`` for tomcat.

Also, we can probably use this principle to switch fixed path of ``/etc/perun/perun-extSources.xml`` to build/runtime relative in ExtSourcesManager.

zlamalp:
I suggest to merge this together with #2436

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
balcirakpeter: Added

PR #2439, opened 2019-10-15
CORE: Allow cookies when synchronizing with other Perun instances

zlamalp:
- To prevent unnecessary session initialization on remote side we
  should keep session by using cookies.
  We perform many subsequent calls to remote Perun API and it generates
  a lot of auditer messages there.
- BasicCookieStore is a thread safe and shared between all instances of
  ExtSourcePerun, since for each group synchronization, new instance
  is created and performs own callback.
  Only if we call same perun instance with different authorizations,
  then it performs session init on remote side, since the "actor"
  has changed (and cookie is overwritten back and forth).


zlamalp:
Manually tested local against devel. With cookie, creates single session on all subsequent synces. Without cookie session is recreated a new on each sync.

PR #2440, opened 2019-10-15
Javadoc for exceptions

oernst9:
* added javadoc for exceptions explaining when they are thrown

* added javadoc for constructors and getters of the exceptions so that it is easy to find out how to use them

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/AlreadyAdminException.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/AlreadyAdminException.java
@@ -3,7 +3,7 @@ package cz.metacentrum.perun.core.api.exceptions;
 import cz.metacentrum.perun.core.api.*;
 
 /**
- * Should be named AlreadyInRoleException.
+ * Thrown when trying to assign role to a user/group while the user/group already has the role.
  *
  */
 public class AlreadyAdminException extends PerunException {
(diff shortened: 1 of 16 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/AlreadyAdminException.java
```diff
@@ -3,7 +3,7 @@
 import cz.metacentrum.perun.core.api.*;
 
 /**
- * Should be named AlreadyInRoleException.
+ * Thrown when trying to assign role to a user in which the user already is
```
stavamichal: I would prefer something like this:
`Thrown when trying to assign role to a user while the user already has the role.`

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ExtendMembershipException.java
```diff
@@ -4,7 +4,7 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * Checked version of ExtendMembershipException.
+ * Thrown when the membership could not be extended for various reasons which is sent to the constructor of the exception
```
stavamichal: If we really want to inform about reason in the constructor, I would do that in separate sentence.
`The reason is sent as a parameter of the exception.`

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/AlreadyAdminException.java
Vojtech-Sassmann: Also, the role can be assigned to a group instead of a user.

PR #2446, opened 2019-10-17
Changing adding of admins to the entities

oernst9:
* Audit events are now logged in setRole in AuthzResolverBlImpl for every entity

* started using setRole/unsetRole in AuthzResolverBlImpl directly from the entry layer for setting admin instead of addAdmin/removeAdmin in Bl layer of the managers

* Removed addAdmin and removeAdmin methods from Bl layer of the managers


oernst9:
I didn't use some code from addAdmin/removeAdmin methods in Bl layer that checked whether the user/group already is admin because it threw the very same "AlreadyAdminException" that the actual attempt of the database to add the current admin as admin again. I might be wrong, if so, I will add it.

PR #2451, opened 2019-10-21
REGISTRAR: Separate mail verification from auto approval

zlamalp:
- Exception thrown from auto-approval process formerly failed mail
  validation with not-so-much explanatory exception.
  Now we explicitly perform verification only and based on result we
  continue with auto-approval. If it fails, than mail verification
  ignores it, while createApplication() throws former exception.
- This bug prevented mail verification and VO admin approval on
  applications with LoA=1 where module prevented auto-approval for
  such application, but allowed LoA=2.


zlamalp:
Tested manually. Merging.

PR #2456, opened 2019-10-29
Creating AuditEvents about removing admins of a deleted entity

oernst9:
* added code that removes admin users and admin groups using unsetRole where AuditEvents are logged by default

* entitites where change has been made: Group, Facility, Resource, SecurityTeam, Vo

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
@@ -340,6 +340,28 @@ public class FacilitiesManagerBlImpl implements FacilitiesManagerBl {
 			}
 		}
 
+		//remove admins of this facility
+		List<Group> adminGroups = getFacilitiesManagerImpl().getAdminGroups(sess, facility);
+		for (Group adminGroup : adminGroups) {
+			try {
+				AuthzResolverBlImpl.unsetRole(sess, adminGroup, facility, Role.FACILITYADMIN);
+			} catch (GroupNotAdminException e) {
+				log.warn("When trying to unsetRole FacilityAdmin for group {} in the facility {} the exception was thrown {}", adminGroup, facility, e);
+				//skip and log as warning
+			}
+		}
+
+		List<User> adminUsers = getFacilitiesManagerImpl().getAdmins(sess, facility);
+
+		for (User adminUser : adminUsers) {
+			try {
+				AuthzResolverBlImpl.unsetRole(sess, adminUser, facility, Role.FACILITYADMIN);
+			} catch (UserNotAdminException e) {
+				log.warn("When trying to unsetRole FacilityAdmin for user {} in the facility {} the exception was thrown {}", adminUser, facility, e);
+				//skip and log as warning
+			}
+		}
+
 		//remove hosts
 		List<Host> hosts = this.getHosts(sess, facility);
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -340,6 +340,28 @@ public void deleteFacility(PerunSession sess, Facility facility, Boolean force)
 			}
 		}
 
+		//remove admins of this facility
+		List<Group> adminGroups = getFacilitiesManagerImpl().getAdminGroups(sess, facility);
+		for (Group adminGroup : adminGroups) {
+			try {
+				AuthzResolverBlImpl.unsetRole(sess, adminGroup, facility, Role.FACILITYADMIN);
+			} catch (GroupNotAdminException e) {
+				log.warn("When trying to unsetRole FacilityAdmin for group with id: " + adminGroup.getId() + ", in the facility with id: " + facility.getId() + " , this exception was thrown: " + e.toString());
```
stavamichal: You should use format like this:
`log.warn("When trying to unsetRole FacilityAdmin for group {} in the facility {} the exception was thrown {}", adminGroup, facility, e.ToString());`
for any logging in code.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -273,6 +273,29 @@ public void deleteMembersGroup(PerunSession sess, Vo vo) throws InternalErrorExc
...
+		List<User> adminUsers = getGroupsManagerImpl().getAdmins(sess, group);
+
+		for (User adminUser : adminUsers) {
+			try {
+				AuthzResolverBlImpl.unsetRole(sess, adminUser, group, Role.GROUPADMIN);
+			} catch (UserNotAdminException e) {
+				log.warn("When trying to unsetRole GroupAdmin for user with id: " + adminUser.getId() + ", in the group with id: " + group.getId() + " , this exception was thrown: " + e.toString());
+				//skip and log as warning
+			}
+		}
```
stavamichal: You should probably do this for all subgroups of this group.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
zlamalp: Just a note: exception object can be passed as it is, there is no need to convert it to string using ``e.toString()``, logging framework will do it by itself.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
zlamalp: I just wonder, why we don't simply call ``deleteAnyGroup()`` on each subgroup, after we retrieve them and sort them in reverse order. By the quick look,  only difference is, that we don't care about their relations and delete them anyway (like we would call this main method with ``forceDelete=TRUE``).

zlamalp: Also we might want to do all of this after we check members and subgroups, since if deletion is not forced, it will fail more quickly and has to rollback fewer changes.

PR #2458, opened 2019-10-29
Load and get perun policies

balcirakpeter:
- PerunRolesLoader was extended to also support loading perun policies
  from the configuration file.
- New class PerunPoliciesContainer was created which will store the
  policies.
- AuthzResolverImpl will load the policies to the PerunPoliciesContainer
  during the initialize() method. It has now also method
  getPolicy(String policyName) which will fetch the given policy from the
  container.

### DIFF
--- a/perun-base/src/test/resources/perun-roles.yml
+++ b/perun-base/src/test/resources/perun-roles.yml
@@ -20,4 +20,12 @@ perun_roles:
   - SECURITYADMIN
   - CABINETADMIN
   - UNKNOWN
+
+#A list of Perun policies that are loaded to the PerunPoliciesContainer.
+perun_policies:
+
+  default_policy:
+    policy_roles:
+      - PERUNADMIN:
+    include_policies: []
 ...
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
@@ -14,7 +14,10 @@ import org.springframework.core.io.Resource;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.HashMap;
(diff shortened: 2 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
```diff
@@ -31,17 +34,10 @@
...
 
-		try (InputStream is = configurationPath.getInputStream()) {
-			JsonNode rolesNode = mapper.readTree(is).get("perun_roles");
-			roles = new ObjectMapper().convertValue(rolesNode, new TypeReference<List<String>>() {});
-		} catch (FileNotFoundException e) {
-			throw new InternalErrorException("Configuration file not found for perun roles. It should be in: " + configurationPath, e);
-		} catch (IOException e) {
-			throw new InternalErrorException("IO exception was thrown during the processing of the file: " + configurationPath, e);
-		}
+		JsonNode rolesNode =rootNode.get("perun_roles");
```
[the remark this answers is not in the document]
balcirakpeter: Done

balcirakpeter:
Changes in the configuration file were also reflected in the perun_roles.yml in perun-dev.cesnet.cz, so the PR can be safely merged.

perun-base/src/test/resources/perun-roles.yml
```diff
@@ -20,4 +20,12 @@ perun_roles:
   - SECURITYADMIN
   - CABINETADMIN
   - UNKNOWN
+
+#A list of Perun policies that are loaded to the perunPolicies map in AuthzResolverImpl
+perun_policies:
```
zlamalp: If this stub is required on all deployed instances, then please add it to the perun-ansible repository too.

balcirakpeter: PR was created CESNET/perun-ansible#122

PR #2461, opened 2019-10-31
createSponsoredMember will accept separated name with titles in a map

oernst9:
* original functionality has been preserved, accepting separated name is a new way of creating user to avoid issues with parsing the name as the whole

* changes have been made in RPC as well as in Core

* tests added

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
@@ -1154,7 +1154,7 @@ public interface MembersManager {
 	 * @param session actor
 	 * @param vo virtual organization  for the member
 	 * @param namespace namespace for selecting password module
-	 * @param guestName a designation, usually a full name
+	 * @param name a map containing the full name or its parts (mandatory: firstName, lastName; optionally: titleBefore, titleAfter)
 	 * @param password  password
 	 * @param sponsor sponsoring user or null for the caller
 	 * @return new Member in the Vo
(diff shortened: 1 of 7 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -1453,7 +1453,7 @@
 	 * @param session perun session
 	 * @param vo virtual organization
 	 * @param namespace used for selecting external system in which guest user account will be created
-	 * @param guestName full name or other designation
+	 * @param name a map containing the full name or parts of it
```
stavamichal: You should tell what parts you expect there.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1154,7 +1154,7 @@
 	 * @param session actor
 	 * @param vo virtual organization  for the member
 	 * @param namespace namespace for selecting password module
-	 * @param guestName a designation, usually a full name
+	 * @param name a map containing the full name or parts of it
```
stavamichal: You should tell what parts you expect there.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -145,15 +146,23 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
 		@Override
 		public RichMember call(ApiCaller ac, Deserializer params) throws PerunException {
```
stavamichal: You should change javadoc to add information about possible options here.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1154,7 +1154,7 @@
 	 * @param session actor
 	 * @param vo virtual organization  for the member
 	 * @param namespace namespace for selecting password module
-	 * @param guestName a designation, usually a full name
+	 * @param name a map containing the full name or its parts (required: firstName, lastName; voluntary: titleBefore, titleAfter)
```
stavamichal: More often we are using "mandatory" and "optionally".

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -141,19 +142,43 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 			String password = params.readString("password");
 			Vo vo =  ac.getVoById(params.readInt("vo"));
 			String namespace = params.readString("namespace");
 			User sponsor = null;
 			if(params.contains("sponsor")) {
 				sponsor = ac.getUserById(params.readInt("sponsor"));
 			}
-			return ac.getMembersManager().createSponsoredMember(ac.getSession(), vo, namespace, guestName, password, sponsor);
+			Map<String, String> name = new HashMap<>();
+			if (params.contains("firstName")) {
```
zlamalp: I would prefer to differentiate behavior based on presence of ``guestName`` rather than ``firstName``. It is used like this later in BL logic and conforms original implementation.
Like this you could pass ``guestName`` and ``lastName`` and it would work and just ignored ``lastName`` value without letting you know. You can probably create third branch where you throw exception if wrong combination of params is passed. Like:
```java
if (params.contains("guestName")) {
...
} else if (params.contains("firstName") && params.contains("lastName")) {
...
} else {
throw new Exception()
}

PR #2465, opened 2019-11-03
loadAuthorizationComponents() method

balcirakpeter:
- Created method loadAuthorizationComponents() to be able to load perun
  policies and roles without the tomcat restart. method was also added
  to RPC module.
- Moved log information about initializing AuthzResolverImpl in read
  only version from PerunRolesLoader to AuthzResolverImpl.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
@@ -611,6 +611,17 @@ public enum AuthzResolverMethod implements ManagerMethod {
 		public String call(ApiCaller ac, Deserializer parms) throws PerunException {
 			return "OK";
 		}
-	};
+	},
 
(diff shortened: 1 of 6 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -611,6 +611,17 @@ public PerunPrincipal call(ApiCaller ac, Deserializer parms) throws PerunExcepti
 		public String call(ApiCaller ac, Deserializer parms) throws PerunException {
 			return "OK";
 		}
-	};
+	},
 
+	/*#
+	 * Load perun roles an policies from the configuration file perun-roles.yml.
```
zlamalp: an -> and

balcirakpeter: Fixed

PR #2469, opened 2019-11-05
CORE: Fixed null pointer when storing UserExtSource attributes

zlamalp:
- When splitting list of incoming attribute values from federation
  we must check on null.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
@@ -202,7 +202,10 @@ public class PerunBlImpl implements PerunBl {
 
 					//for Array list attributes try to parse string value into individual fields
 					if(attributeWithValue.getType().equals(ArrayList.class.getName()) || attributeWithValue.getType().equals(BeansUtils.largeArrayListClassName)) {
-						List<String> value = new ArrayList<>(Arrays.asList(attrValue.split(UsersManagerBlImpl.multivalueAttributeSeparatorRegExp)));
+						List<String> value = new ArrayList<>();
+						if (attrValue != null) {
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -202,7 +202,10 @@ public void setUserExtSourceAttributes(PerunSession session, UserExtSource ues,
...
 					//for Array list attributes try to parse string value into individual fields
 					if(attributeWithValue.getType().equals(ArrayList.class.getName()) || attributeWithValue.getType().equals(BeansUtils.largeArrayListClassName)) {
-						List<String> value = new ArrayList<>(Arrays.asList(attrValue.split(UsersManagerBlImpl.multivalueAttributeSeparatorRegExp)));
+						List<String> value = new ArrayList<>();
+						if (attrValue != null) {
+							value = new ArrayList<>(Arrays.asList(attrValue.split(UsersManagerBlImpl.multivalueAttributeSeparatorRegExp)));
+						}
 						attributeWithValue.setValue(value);
 					} else {
 						attributeWithValue.setValue(attrValue);
```
stavamichal: What about to rewrite this code as:
```
if(attrValue != null && (attributeWithValue.getType().equals(ArrayList.class.getName()) || attributeWithValue.getType().equals(BeansUtils.largeArrayListClassName))) {
  attrValue = new ArrayList<>(Arrays.asList(attrValue.split(UsersManagerBlImpl.multivalueAttributeSeparatorRegExp)));
}
attributeWithValue.setValue(attrValue);
```
Cause we are always setting attrValue no matter if it is null or splitted.

zlamalp: I will keep it, since like this there would be conflicting type in attrValue variable.

PR #2473, opened 2019-11-07
ENGINE: Store TaskResults for failed GEN Tasks

zlamalp:
- We now store TaskResults for all destinations present in a Task
  after data generation fails. When it succeed, we continue with
  sending and handle TaskResults as usual.


zlamalp:
Manually tested, works as expected - writes TaskResult when generation fails.

PR #2474, opened 2019-11-07
upgraded Spring Boot Starter Parent to 2.2.1

martin-kuba:


martin-kuba:
A temporary problem at Travis has occurred, we have to wait a day:
`Failure to find org.springframework.boot:spring-boot-maven-plugin:jar:2.2.1.RELEASE in https://maven-central.storage-download.googleapis.com/repos/central/data/ was cached in the local repository, resolution will not be reattempted until the update interval of google-maven-central has elapsed or updates are forced`

PR #2489, opened 2019-11-19
Another fix of VosManagerEntry authorization

balcirakpeter:
- added VOOBSERVER priviledge to getCompleteCandidates() methods
- getVos() method was filtering vo if a principal was not VOADMIN
  and not GROUPADMIN for the vo. Then it was adding back these vos
  for which the principal was VOOBSERVER.
  Now it filters a vo if a principal is not VOADMIN, not GROUPADMIN
  and not VOOBSERVER for the vo, at one place.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
@@ -76,20 +76,15 @@ public class VosManagerEntry implements VosManager {
 				tempVos.removeIf(vo -> {
 					try {
 						return !AuthzResolver.isAuthorized(sess, Role.VOADMIN, vo) &&
-								!AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, vo);
+								!AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, vo) &&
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
```diff
@@ -278,6 +273,7 @@ public Vo getVoById(PerunSession sess, int id) throws VoNotExistsException, Inte
 
 		// Authorization
 		if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, vo) &&
+				!AuthzResolver.isAuthorized(sess, Role.VOOBSERVER, vo) &&
```
zlamalp: Now when I think about it. VOOBSERVER can't actually add candidate as a member of VO. So there is no real reason, why he should be able to read them. Also it retrieves data from ext sources - people who are not yet in the VO (possibly perun), but the role VOOBSERVER should grant privilege to see data in Perun only.

stavamichal: I think VOOBSERVER can see anything what VOADMIN can see, so there is no problem since there is no modification.

zlamalp: OK.

PR #2493, opened 2019-11-21
Replaced old authorization by the new one in VosManagerEntry

balcirakpeter:
- In Vo manager was completely replaced the old authorization, except the deprecated methods.
  policies used in the VosManagerEntry.

### DIFF
--- a/perun-base/src/test/resources/perun-roles.yml
+++ b/perun-base/src/test/resources/perun-roles.yml
@@ -2154,4 +2154,198 @@ perun_policies:
       - PERUNOBSERVER:
     include_policies:
       - default_policy
+
+  #VosManagerEntry
+  getVos_policy:
+    policy_roles:
+      - GROUPADMIN:
+      - VOOBSERVER:
+      - VOADMIN:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  filter-getVos_policy:
+    policy_roles:
+      - GROUPADMIN: Vo
+      - VOOBSERVER: Vo
+      - VOADMIN: Vo
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  getAllVos_policy:
+    policy_roles:
+      - FACILITYADMIN:
+      - GROUPADMIN:
+      - VOOBSERVER:
+      - VOADMIN:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  deleteVo_Vo_boolean_policy:
+    policy_roles: []
+    include_policies:
+      - default_policy
+
+  deleteVo_Vo_policy:
+    policy_roles: []
+    include_policies:
+      - default_policy
+
+  createVo_Vo_policy:
+    policy_roles:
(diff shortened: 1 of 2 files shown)


stavamichal:
We should prepare some tests of this functionality first.

balcirakpeter:
The configuration file in server has to be updated before merge

balcirakpeter:
> We should prepare some tests of this functionality first.

The tests were created.

perun-core/src/main/java/cz/metacentrum/perun/core/api/VosManager.java
```diff
@@ -46,7 +46,7 @@
 	 * @throws RelationExistsException
 	 * @throws InternalErrorException
 	 */
-	List<Vo> getVos(PerunSession perunSession) throws InternalErrorException, PrivilegeException;
+	List<Vo> getVos(PerunSession perunSession) throws InternalErrorException, PrivilegeException, PolicyNotExistsException;
```
stavamichal: Is this exception still needed in Entry/Api?

balcirakpeter:
> You said, that tests were created. Are they somewhere else? I can't see them in this PR.

They were added in the pull request #2501

perun-core/src/main/java/cz/metacentrum/perun/core/api/VosManager.java
balcirakpeter: No, it is not. Fixed

perun-core/src/test/java/cz/metacentrum/perun/core/entry/SecurityTeamsManagerEntryIntegrationTest.java
```diff
@@ -961,7 +962,7 @@ private void setUpBlacklists() throws PrivilegeException, InternalErrorException
 	}
 
 
-	private Group setUpGroup(User u0, User u1) throws PrivilegeException, InternalErrorException, UserNotExistsException, VoExistsException, GroupExistsException, VoNotExistsException, GroupNotExistsException, AlreadyMemberException, MemberNotExistsException, WrongReferenceAttributeValueException, WrongAttributeValueException, ExtendMembershipException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternallyManagedException {
+	private Group setUpGroup(User u0, User u1) throws PrivilegeException, InternalErrorException, UserNotExistsException, VoExistsException, GroupExistsException, VoNotExistsException, GroupNotExistsException, AlreadyMemberException, MemberNotExistsException, WrongReferenceAttributeValueException, WrongAttributeValueException, ExtendMembershipException, WrongAttributeAssignmentException, AttributeNotExistsException, ExternallyManagedException, PolicyNotExistsException {
```
Vojtech-Sassmann: Please, remove this useless throw declaration. Otherwise, it is working, but it could be better.

balcirakpeter: Fixed

PR #2495, opened 2019-11-22
Separation of checkAttributeSyntax from semantics in Member-Resource …

metodej:
…modules

- former checkAttributeValue is now separated into checkAttributeSyntax and checkAttributeSemantics in member-resource attribute modules
- also added test for checks

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataLimit.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataLimit.java
@@ -5,6 +5,7 @@ import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
@@ -13,6 +14,7 @@ import cz.metacentrum.perun.core.api.exceptions.MemberResourceMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.blImpl.ModulesUtilsBlImpl;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.MemberResourceAttributesModuleAbstract;
 import cz.metacentrum.perun.core.implApi.modules.attributes.MemberResourceAttributesModuleImplApi;
@@ -33,8 +35,8 @@ public class urn_perun_member_resource_attribute_def_def_dataLimit extends Membe
 	private static final String A_R_defaultDataQuota = AttributesManager.NS_RESOURCE_ATTR_DEF + ":defaultDataQuota";
 	private static final String A_MR_dataQuota = AttributesManager.NS_MEMBER_RESOURCE_ATTR_DEF + ":dataQuota";
 	private static final String A_F_readyForNewQuotas = AttributesManager.NS_FACILITY_ATTR_DEF + ":readyForNewQuotas";
-	private static final Pattern numberPattern = Pattern.compile("[0-9]+([.,])?[0-9]*");
-	private static final Pattern letterPattern = Pattern.compile("[A-Z]");
+	private static final Pattern testingPattern = Pattern.compile("^[0-9]+([.][0-9]+)?[KMGTPE]$");
+
 	final long K = 1024;
 	final long M = K * 1024;
 	final long G = M * 1024;
@@ -43,15 +45,18 @@ public class urn_perun_member_resource_attribute_def_def_dataLimit extends Membe
 	final long E = P * 1024;
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
-		Attribute attrDataQuota;
-		String dataQuota;
-		String dataLimit;
+	public void checkAttributeSyntax(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException {
+		if (attribute.getValue() == null) return;
 
-		String dataQuotaNumber = null;
-		String dataQuotaLetter = null;
-		String dataLimitNumber = null;
-		String dataLimitLetter = null;
+		Matcher testMatcher = testingPattern.matcher(attribute.valueAsString());
+		if (!testMatcher.find())
+			throw new WrongAttributeValueException(attribute, member, resource, "Format of quota must be something like ex.: 1.30M or 2500K, but it is " + attribute.getValue());
+	}
+
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+		Attribute attrDataQuota;
+		String dataQuota = null;
 
 		//Get attrDataQuota attribute
 		try {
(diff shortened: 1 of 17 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataQuota.java
```diff
@@ -70,29 +61,13 @@ public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member
...
-		else quotaNumber = new BigDecimal("0");
+
+		Pair<BigDecimal, String> quotaNumberAndLetter = ModulesUtilsBlImpl.getNumberAndUnitFromString(attribute.valueAsString());
+		BigDecimal quotaNumber = quotaNumberAndLetter.getLeft();
+		String dataQuotaLetter = quotaNumberAndLetter.getRight();
 
 		if (quotaNumber.compareTo(BigDecimal.valueOf(0)) < 0) {
-			throw new WrongAttributeValueException(attribute, resource, member, attribute + " cant be less than 0.");
+			throw new WrongReferenceAttributeValueException(attribute, null, resource, member, attribute + " cant be less than 0.");
 		}
```
stavamichal: isn't this part more syntax than semantic check?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataQuotas.java
```diff
@@ -33,15 +33,20 @@
 	public static final String A_R_maxUserDataQuotas = AttributesManager.NS_RESOURCE_ATTR_DEF + ":maxUserDataQuotas";
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+	public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
```
stavamichal: There is no syntax check, but method checkAndTransferQuotas do syntax check. So for this reason I would use it in syntax check too.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_fileQuotas.java
```diff
@@ -33,15 +33,20 @@
 	public static final String A_R_maxUserFileQuotas = AttributesManager.NS_RESOURCE_ATTR_DEF + ":maxUserFileQuotas";
 
 	@Override
-	public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongAttributeValueException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
+	public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member, Resource resource, Attribute attribute) throws InternalErrorException, WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
```
stavamichal: There is no syntax check, but method checkAndTransferQuotas do syntax check. So for this reason I would use it in syntax check too.

metodej:
Thanks for review, there is now syntax check for the four mentioned modules.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataLimit.java
```diff
@@ -33,8 +35,8 @@
 	private static final String A_R_defaultDataQuota = AttributesManager.NS_RESOURCE_ATTR_DEF + ":defaultDataQuota";
 	private static final String A_MR_dataQuota = AttributesManager.NS_MEMBER_RESOURCE_ATTR_DEF + ":dataQuota";
 	private static final String A_F_readyForNewQuotas = AttributesManager.NS_FACILITY_ATTR_DEF + ":readyForNewQuotas";
-	private static final Pattern numberPattern = Pattern.compile("[0-9]+([.,])?[0-9]*");
-	private static final Pattern letterPattern = Pattern.compile("[A-Z]");
+	private static final Pattern testingPattern = Pattern.compile("^[0-9]+([.][0-9]+)?[KMGTPE]$");
```
balcirakpeter: Nit: testingPattern name sounds like it is a pattern used in tests.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_filesLimit.java
```diff
@@ -48,30 +54,30 @@ public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member
...
 		} else {
 			try {
 				attribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, resource, A_R_defaultFilesLimit);
 			} catch (AttributeNotExistsException ex) {
 				throw new ConsistencyErrorException("Attribute with defaultFilesLimit from resource " + resource.getId() + " could not obtained.", ex);
 			}
 			if(attribute != null && attribute.getValue() != null) {
-				filesLimit = (Integer) attribute.getValue();
+				filesLimit = attribute.valueAsInteger();
+				if (filesLimit < 0) throw new WrongReferenceAttributeValueException(attribute, null, resource, member, attribute + " cannot be less than 0.");
```
balcirakpeter: This condition is already in the syntax check. Thus, it is not necessary for the semantics check.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_filesQuota.java
```diff
@@ -48,30 +54,30 @@ public void checkAttributeSemantics(PerunSessionImpl perunSession, Member member
...
 		} else {
 			try {
 				attribute = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, resource, A_R_defaultFilesQuota);
 			} catch (AttributeNotExistsException ex) {
 				throw new ConsistencyErrorException("Attribute with defaultFilesQuota from resource " + resource.getId() + " could not obtained.", ex);
 			}
 			if(attribute != null && attribute.getValue() != null) {
-				filesQuota = (Integer) attribute.getValue();
+				filesQuota = attribute.valueAsInteger();
+				if(filesQuota < 0) throw new WrongReferenceAttributeValueException(attribute, null, resource, member, attribute + " cannot be less than 0.");
```
balcirakpeter: Same here.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_filesLimitTest.java
```diff
@@ -0,0 +1,90 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.*;
```
balcirakpeter: We would like to avoid wildcard imports in the code.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataLimit.java
metodej: I borrowed the name and pattern from other modules with similar check (like urn_perun_group_resource_attribute_def_def_projectDataLimit)...

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_filesQuota.java
metodej: The reason for this check is that _attribute_ does not necessarily hold input attribute but it can hold _defaultFilesLimit_ attribute unlike in syntax check where it's definitely attribute given by the method. Should I change the check to check only if it's _defaultFilesLimit_?

balcirakpeter: Sorry I didn't see that. You don't have to change it

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_filesLimit.java
balcirakpeter: Same here. Just ignore it... :D

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_member_resource_attribute_def_def_dataLimit.java
balcirakpeter: It is not wrong and you don't have to change it. Just my first thought, when I had read it for the first time, was that you forgot to erase a pattern for some tests. :)

PR #2498, opened 2019-11-23
Separation of checkAttributeSyntax from semantics in Resource modules

metodej:
- former checkAttributeValue is now separated into checkAttributeSyntax and checkAttributeSemantics in resource attribute modules
- also added test for checks

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataLimit.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataLimit.java
@@ -4,6 +4,7 @@ import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.Facility;
+import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
@@ -11,6 +12,7 @@ import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.blImpl.ModulesUtilsBlImpl;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.attributes.ResourceAttributesModuleAbstract;
 import cz.metacentrum.perun.core.implApi.modules.attributes.ResourceAttributesModuleImplApi;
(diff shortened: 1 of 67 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataQuota.java
```diff
@@ -119,7 +91,7 @@ public void checkAttributeSemantics(PerunSessionImpl perunSession, Resource reso
 		//Compare DefaultDataQuota with DefaultDataLimit
 		if (quotaNumber.compareTo(BigDecimal.valueOf(0)) == 0) {
 			if (limitNumber.compareTo(BigDecimal.valueOf(0)) != 0) {
-				throw new WrongReferenceAttributeValueException(attribute, attrDefaultDataLimit, resource, null, resource, null, "Try to set unlimited quota, but limit is still " + defaultDataLimitNumber + defaultDataLimitLetter);
+				throw new WrongReferenceAttributeValueException(attribute, attrDefaultDataLimit, resource, null, resource, null, "Try to set unlimited quota, but limit is still " + limitNumber + defaultDataLimitLetter);
 			}
 		} else if (limitNumber.compareTo(BigDecimal.valueOf(0)) != 0 && defaultDataLimitLetter != null && defaultDataQuotaLetter != null) {
```
balcirakpeter: I think you can erase the check if defaultDatalimitLetter is null as you did in the defaultDataLimit module. Moreover, there is the same question on defaultDataQuotaLetter as in the previous module.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_defaultDataLimit.java
```diff
@@ -121,7 +97,7 @@ public void checkAttributeSemantics(PerunSessionImpl perunSession, Resource reso
 			if (limitNumber.compareTo(BigDecimal.valueOf(0)) != 0) {
 				throw new WrongReferenceAttributeValueException(attribute, attrDefaultDataQuota, resource, null, resource, null, "Try to set limited limit, but there is still set unlimited Quota.");
 			}
-		} else if ((quotaNumber.compareTo(BigDecimal.valueOf(0)) != 0) && (limitNumber.compareTo(BigDecimal.valueOf(0)) != 0) && defaultDataLimitLetter != null && defaultDataQuotaLetter != null) {
+		} else if ((quotaNumber.compareTo(BigDecimal.valueOf(0)) != 0) && (limitNumber.compareTo(BigDecimal.valueOf(0)) != 0) && defaultDataQuotaLetter != null) {
```
balcirakpeter: Why do you need to check if defaultDataQuotaLetter is null when it came from the same method as defaultDataLimitLetter?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_def_unixGID_namespace.java
```diff
@@ -94,26 +94,30 @@ public Attribute fillAttribute(PerunSessionImpl sess, Resource resource, Attribu
...
+				throw new WrongReferenceAttributeValueException(attribute, null, resource, null, "Unix GID must be set");
 			} else {
-				attrValue = (Integer) attribute.getValue();
+				attrValue = attribute.valueAsInteger();
 			}
 
 			//Check if GID is within allowed range
-			sess.getPerunBl().getModulesUtilsBl().checkIfGIDIsWithinRange(sess, attribute);
+			try {
+				sess.getPerunBl().getModulesUtilsBl().checkIfGIDIsWithinRange(sess, attribute);
```
balcirakpeter: I think the method checkIfGIDIsWithinRange should be reworked, so it does not throw WrongAttributeValueException. This Exception is thrown on two places: first if the attribute is null where we usually do not throw Value exception. Second, from method checkAndConvertGIDRanges which check reference attribute so we should catch Value exception and throw the Reference exception.

metodej: We already discussed this method earlier with group module and decided to this solution.

metodej:
Thanks for review, the missed unnecessary checks were removed.

PR #2501, opened 2019-11-25
Tests for AuthzResolver.authorized() method

oernst9:
- added 4 tests for this method

### DIFF
--- a/perun-base/src/test/resources/perun-roles.yml
+++ b/perun-base/src/test/resources/perun-roles.yml
@@ -28,4 +28,100 @@ perun_policies:
     policy_roles:
       - PERUNADMIN:
     include_policies: []
+
+  test_resourceselfservice:
+    policy_roles:
+       - RESOURCESELFSERVICE: Resource
+         GROUPADMIN: Group
+    include_policies: []
+
+  test_sponsor:
+    policy_roles:
+       - SPONSOR: Vo
+    include_policies: []
+
+  test_cabinet:
+    policy_roles:
+       - CABINETADMIN:
+    include_policies: []
+
+  test_self:
+    policy_roles:
+       - SELF: User
+    include_policies: []
+
+  test_voobserver_and_topgroupcreator:
+    policy_roles:
+      - VOOBSERVER: Vo
+        TOPGROUPCREATOR:
+    include_policies: []
+
+  test_security_admin:
+    policy_roles:
+      - SECURITYADMIN: SecurityTeam
+    include_policies: []
+
+  test_group_or_vo:
+    policy_roles:
+      - GROUPADMIN: Group
+      - VOADMIN: Vo
+    include_policies: []
+
+  test_resource_admin:
+    policy_roles:
+      - RESOURCEADMIN: Resource
+    include_policies: []
+
+  test_resource_and_facility_admin:
+    policy_roles:
+      - RESOURCEADMIN: Resource
+        FACILITYADMIN: Facility
+    include_policies: []
+
+  test_authorized_vo_admin:
+    policy_roles:
+      - VOADMIN: Vo
+    include_policies: []
+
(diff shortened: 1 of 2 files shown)


balcirakpeter:
And another thing. I think that if you need just one group to be created in the Vo, you can use setUpGroup() method.

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -64,6 +64,378 @@ public void authorizedPerunAdmin() throws Exception {
...
+
+		perun.getResourcesManager().addResourceSelfServiceGroup(sess, resource, createdGroup);
+
+		AuthzResolver.refreshAuthz(session);
+		assertTrue(AuthzResolver.authorized(session, "test_resourceselfservice", Arrays.asList(resource)));
+	}
+
+	/*
+	@Test
+	public void authorizedServiceUser() throws Exception {
```
balcirakpeter: You can erase this test for now. I think SERVICEUSER does not have any rights now.

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -64,6 +64,378 @@ public void authorizedPerunAdmin() throws Exception {
...
+
+		AuthzResolver.setRole(sess, createdUser, createdVo, Role.VOADMIN);
+
+		PerunSession session = getHisSession(createdMember);
+		AuthzResolver.refreshAuthz(session);
+		assertTrue(AuthzResolver.authorized(session, "test_authorized_vo_admin",  Arrays.asList(createdVo)));
+	}
+
+	@Test
+	public void authorizedVoAdminCycleAdmin() throws Exception {
```
balcirakpeter: Please change the names of the tests which have aasertFalse() so they start with unauthorized prefix. It will be easier to understand what is the test checking.

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -64,6 +64,378 @@ public void authorizedPerunAdmin() throws Exception {
 		assertTrue(AuthzResolver.authorized(session, "default_policy", Collections.emptyList()));
 	}
 
+	@Test
+	public void authorizedVoAdmin() throws Exception {
```
balcirakpeter: Please add the System.out.println(CLASS_NAME + "The_actual_test_name"); to each test you created. I know there are many other tests which do not have that, we will fix that later.

perun-base/src/test/resources/perun-roles.yml
```diff
@@ -24,8 +24,108 @@ perun_roles:
...
+    include_policies:
+      - test_cycle_voadmin
+
+  test_authorized_group_admin:
+    policy_roles:
+      - GROUPADMIN: Group
+    include_policies:
+      - test_authorized_vo_admin
+
   default_policy:
```
balcirakpeter: Please move this policy to the first place.

perun-base/src/test/resources/perun-roles.yml
```diff
@@ -24,8 +24,108 @@ perun_roles:
 #A list of Perun policies that are loaded to the PerunPoliciesContainer.
 perun_policies:
 
+  test_resourceselfservice:
+    policy_roles:
+       - RESOURCESELFSERVICE: Resource
```
balcirakpeter: RESOURCESELFSERVICE role will be always in the rule together with the GROUPADMIN role (AND connection between them). Please add the "GROUPADMIN: Group" to the rule and correct the tests if needed.

PR #2503, opened 2019-11-26
implemented Perun RPC client in Python

martin-kuba:


stavamichal:
Because library can be generated, I would remove all such files from this PR.

martin-kuba:
generated file removed from PR

PR #2509, opened 2019-11-27
Lists Facilities where host is assigned.

zoraseb:
Tool for listing of facilities where hostname is used. Tests if hostname is not
only alias. In that case are facilities for canonical hostname listed too.

### DIFF
--- a/perun-cli/getFacilitiesByHostName
+++ b/perun-cli/getFacilitiesByHostName
@@ -1,0 +1,49 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Getopt::Long qw(:config no_ignore_case);
+use Text::ASCIITable;
+use Perun::Agent;
+use Perun::Common qw(printMessage getSortingFunction printTable);
+#use Data::Dumper;
+
+sub help {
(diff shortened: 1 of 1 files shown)


perun-cli/getFacilitiesByHostName
```diff
@@ -0,0 +1,59 @@
...
+unless (@facilities) {
+	printMessage "No facilities found for $hostname", $batch;
+} else {
+	#output
+	printTable($sortingFunction, @facilities);
+}	
+
+#zjisteni aliasu
+my $row;
+open (ALIAS,"host $hostname |") or die "system command host is not available\n";
```
zlamalp: I don't think we need any of this. We want to call existing API method, which will find host by its hostname (i believe it is unique in perun) and it will return list of facilities (simply because we allow adding same host to multiple facilities). So in most cases it will simply return table with single facility.
There shouldn't be necessary any logic to guess "facility name" based on user provided / arbitrary "hostname". Admin should know, for which host he wants to get facility, since next step is to delete it (but it requires knowledge of facility name/id at the time).

PR #2513, opened 2019-12-02
Added new exceptions and their usage to registrar

metodej:
- MailManager should now throw newly created exceptions instead of DuplicateKey and some InternalError
- the exceptions also added to JsonErrorHandler

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
@@ -28,6 +28,9 @@ import cz.metacentrum.perun.core.bl.GroupsManagerBl;
 import cz.metacentrum.perun.core.bl.MembersManagerBl;
 import cz.metacentrum.perun.core.bl.UsersManagerBl;
 import cz.metacentrum.perun.core.impl.Compatibility;
+import cz.metacentrum.perun.registrar.exceptions.ApplicationMailAlreadyRemovedException;
+import cz.metacentrum.perun.registrar.exceptions.ApplicationMailExistsException;
+import cz.metacentrum.perun.registrar.exceptions.ApplicationMailNotExistsException;
 import cz.metacentrum.perun.registrar.exceptions.FormNotExistsException;
 import cz.metacentrum.perun.registrar.exceptions.RegistrarException;
 import cz.metacentrum.perun.audit.events.MailManagerEvents.InvitationSentEvent;
@@ -120,7 +123,7 @@ public class MailManagerImpl implements MailManager {
 
 	@Override
 	@Transactional(rollbackFor = Exception.class)
-	public Integer addMail(PerunSession sess, ApplicationForm form, ApplicationMail mail) throws PerunException, DuplicateKeyException {
+	public Integer addMail(PerunSession sess, ApplicationForm form, ApplicationMail mail) throws ApplicationMailExistsException, PrivilegeException {
 
 		if (form.getGroup() != null) {
 			if (!AuthzResolver.isAuthorized(sess, Role.VOADMIN, form.getVo()) && !AuthzResolver.isAuthorized(sess, Role.GROUPADMIN, form.getGroup())) {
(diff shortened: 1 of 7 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -120,7 +123,7 @@ protected void initialize() throws PerunException {
 
 	@Override
 	@Transactional(rollbackFor = Exception.class)
-	public Integer addMail(PerunSession sess, ApplicationForm form, ApplicationMail mail) throws PerunException, DuplicateKeyException {
+	public Integer addMail(PerunSession sess, ApplicationForm form, ApplicationMail mail) throws PerunException {
```
zlamalp: Can you please convert this to throw specific exceptions instead of PerunException ? Thank you.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -171,7 +182,7 @@ public void deleteMailById(PerunSession sess, ApplicationForm form, Integer id)
 		ApplicationMail mail = getMailById(sess, id);
 
 		int result = jdbc.update("delete from application_mails where id=?", id);
-		if (result == 0) throw new InternalErrorException("Mail notification with id="+id+" doesn't exists!");
+		if (result == 0) throw new ApplicationMailAlreadyRemovedException("Mail notification with id="+id+" doesn't exists!");
```
zlamalp: Can you please use specific exceptions in the method definition instead of PerunException ? Thank you.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -198,7 +209,7 @@ public ApplicationMail getMailById(PerunSession sess, Integer id) throws Interna
 			// set
 			if (mails.size() != 1) {
 				log.error("[MAIL MANAGER] Wrong number of mail definitions returned by unique params, expected 1 but was: {}.", mails.size());
-				throw new InternalErrorException("Wrong number of mail definitions returned by unique params, expected 1 but was: "+mails.size());
+				throw new ApplicationMailNotExistsException("Wrong number of mail definitions returned by unique params, expected 1 but was: "+mails.size());
```
zlamalp: This is not the case mail not exists. Not Exists case is catched below in handling ``EmptyResultDataAccessException``. This covers cases, when there are more than 1 mail definitions, so it should be consistency error (probably). You can fix above condition to be more specific ``>1``.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -256,9 +271,11 @@ public void setSendingEnabled(PerunSession sess, List<ApplicationMail> mails, bo
 		for (ApplicationMail mail : mails) {
 			// update sending (enabled / disabled)
 			if (Compatibility.isPostgreSql()) {
-				jdbc.update("update application_mails set send=? where id=?", (enabled) ? "1" : "0", mail.getId());
+				int existence = jdbc.update("update application_mails set send=? where id=?", (enabled) ? "1" : "0", mail.getId());
+				if (existence < 1) throw new ApplicationMailNotExistsException("Application mail does not exist.", mail);
```
zlamalp: You should try/catch the ``jdbc.update()`` call. Handle ``EmptyResultDataAccessException`` as ``ApplicationMailNotExistsException`` and wrap ``RuntimeException`` to ``InternalErrorException``.

metodej:
Thank you for review, hopefully everything mentioned above should be fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -256,9 +271,16 @@ public void setSendingEnabled(PerunSession sess, List<ApplicationMail> mails, bo
 		for (ApplicationMail mail : mails) {
 			// update sending (enabled / disabled)
 			if (Compatibility.isPostgreSql()) {
-				jdbc.update("update application_mails set send=? where id=?", (enabled) ? "1" : "0", mail.getId());
+				try {
+					jdbc.update("update application_mails set send=? where id=?", (enabled) ? "1" : "0", mail.getId());
```
zlamalp: ``jdbc.update()`` doesn't throws ``EmptyResultDataAccessException`` since it doesn't select the result, but it only returns number of affected rows. You should do the same, as in Oracle DB case below.
Also you should wrap calls ``jdbc.update()`` in try/catch to cache ``RuntimeException`` and wrap it in our ``InternalErrorException``. It should be done everywhere, but for the purpose of this pull-request it will be sufficient to fix it in modified methods.

metodej: My bad, it should be fixed now.

PR #2515, opened 2019-12-03
Mail manager refactor

dBucik:
- removed code duplicates
- improved code readability
- extracted constants
- added documentation to some methods
- implemented sending name of sender in mail headers

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
@@ -388,7 +388,7 @@ public class MailManagerImpl implements MailManager {
 			// different behavior based on mail type
 			switch (mail.getMailType()) {
 				case APP_CREATED_USER:
-					appCreatedUser(app, mail, data, reason, exceptions);
+					sendUserMessage(app, mail, data, reason, exceptions, MailType.APP_CREATED_USER);
 					break;
 				case APP_CREATED_VO_ADMIN:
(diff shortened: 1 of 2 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -63,28 +66,39 @@
 	private static final String MAILS_SELECT_BY_PARAMS = "select id,app_type,form_id,mail_type,send from application_mails where form_id=? and app_type=? and mail_type=?";
 	private static final String MAIL_TEXTS_SELECT_BY_MAIL_ID= "select locale,subject,text from application_mail_texts where mail_id=?";
 
+	private static final String URN_VO_FROM_NAME_EMAIL = "urn:perun:vo:attribute-def:def:fromNameEmail";
```
zlamalp: You might want to add new attributes to the ``RegistrarManagerImpl.initialize()``

dBucik: Thank you for pointing it out, I have added the new attributes there. Just a question, shouldn't we make the attributes (at least it's full names) public and use the constants that are used in initialize in the classes as well? I.e. I can make use of the full urn name from RegistrarManagerImpl in the MailManagerImpl. When attribute will change in one class, we do not have to look for every place it has been mentioned and change it accordingly.

zlamalp: Yes, you can make necessary constants from RegistrarManager public and reuse them in MailManager.

PR #2523, opened 2019-12-12
Logging info about missing roles in config

balcirakpeter:
- When a role exists in the database, but it is missing in the
  configuration file then it is logged because we do not delete roles from
  the database.


zlamalp:
Can you please rebase it? Also can you please add ``log.error()`` before throwing any exception? I realized, that we will shut down ultimate logger and we would have hard times to find cause for perun not starting. Thank you.

PR #2524, opened 2019-12-13
New authorization in AuditMessagesManagerEntry

balcirakpeter:
- In AuditMessagesManagerEntry was completely replaced the old authorization.
  policies used in the AuditMessagesManagerEntry.


balcirakpeter:
The config on devel needs to be updated before merge.

PR #2525, opened 2019-12-14
New authorization in DatabaseManagerEntry

balcirakpeter:
- In DatabaseManagerEntry was completely replaced the old authorization.
  policies used in the DatabaseManagerEntry.


balcirakpeter:

PR #2526, opened 2019-12-14
New authorization in OwnersManagerEntry

balcirakpeter:
- In OwnersManagerEntry was completely replaced the old authorization.
  policies used in the OwnersManagerEntry.


balcirakpeter:

PR #2529, opened 2019-12-20
New authorization in ExtSourceManagerEntry

balcirakpeter:
- In ExtSourceManagerEntry was completely replaced the old authorization.
  policies used in the ExtSourceManagerEntry.
  in the future.


balcirakpeter:

PR #2533, opened 2020-01-02
testing GitHub CI with PR

martin-kuba:
just test, please ignore


martin-kuba:
unsuccessful, the slack-notify action needs a secret, but secrets are not passed for actions triggered by forked repository

PR #2535, opened 2020-01-06
Forbid indirect relationship for group unions

stavamichal:
- indirect relationship can exist between groups in the same structure,
 for example if we have 3 groups "A", "A:B", "A:B:C", there is direct
 relationship between "A" and "A:B" and also direct relationship between
 "A:B" and "A:B:C". Because all members of A:B:C are included by this
 relationship to "A:B" and all such members are automatically included
 into "A", there is indirect relationship between groups "A" and "A:B:C"
 not saved in the database, but processed the same way
 - we want to prevent creating of unions between groups with such
 indirect relationship, so if someone want to create union from "A:B:C"
 to "A", it should be prohibited, because there is no benefit in such
 opration (all members of "A:B:C" are already indirect members of "A"
 through "A:B")
 - this situation need to be checked when new union has been created or
 existing group has been moved and there is a possibility that wrong
 union already exists
 - test for both of these use-case were added

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -671,11 +671,41 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 				throw new GroupMoveNotAllowedException("There is already group union between moving group: " + movingGroup + " and destination group: " + destinationGroup + ".", movingGroup, destinationGroup);
 			}
 
+			//prevent existence forbidden unions between groups after moving
+			//example: there are groups "A", "A:B" and "C", where "C" is included into "A" (there is a union), if we move "C" under "A:B", existing union will change to forbidden one
+			//Get all unions for moving group and all of its subgroups
+			Set<Group> movingGroupStructureUnions = new HashSet<>();
+			movingGroupStructureUnions.addAll(getGroupUnions(sess, movingGroup, true));
+			for(Group subGroup: getAllSubGroups(sess, movingGroup)) {
+				movingGroupStructureUnions.addAll(getGroupUnions(sess, subGroup, true));
+			}
+			//remove direct relationship from moving group to it's parent group (if exists)
+			//prevent wrong exception in situation like "A", "A:B", "A:C" and moving "A:C" under "A:B", which is correct
+			if(movingGroup.getParentGroupId() != null) {
+				try {
+					Group parentGroupOfMovingGroup = getParentGroup(sess, movingGroup);
+					movingGroupStructureUnions.remove(parentGroupOfMovingGroup);
+				} catch (ParentGroupNotExistsException ex) {
+					throw new InternalErrorException("Can't find parentGroup for " + movingGroup);
+				}
+			}
+
+			//Get all group stucture of destination group (destination group and all its parent groups)
+			Set<Group> destinationGroupStructure = new HashSet<>();
+			destinationGroupStructure.add(destinationGroup);
+			destinationGroupStructure.addAll(getParentGroups(sess, destinationGroup));
+
+			movingGroupStructureUnions.retainAll(destinationGroupStructure);
+			if(!movingGroupStructureUnions.isEmpty()) {
+				throw new GroupMoveNotAllowedException("After moving of moving group: " + movingGroup + " under destination group: " + destinationGroup + " there will exist forbidden indirect relationship.", movingGroup, destinationGroup);
+			}
+
 			processRelationsWhileMovingGroup(sess, destinationGroup, movingGroup);
 
 			// We have to set group attributes so we can update it in database
 			movingGroup.setParentGroupId(destinationGroup.getId());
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4109,7 +4109,14 @@ public Group createGroupUnion(PerunSession sess, Group resultGroup, Group operan
 		// check if there is already a record of these two groups
 		if (this.groupsManagerImpl.isRelationBetweenGroups(resultGroup, operandGroup)) {
 			throw new GroupRelationAlreadyExists("There is already a relation defined between result group " + resultGroup +
-					" and operand group " + operandGroup + " or they are in hierarchical structure.");
+					" and operand group " + operandGroup + " or they are in direct hierarchical structure.");
+		}
+
+		// check indirect relationships (for example "A" with "A:B:C" are in indirect relationship using "A:B")
+		List<Group> allParentGroups = this.getParentGroups(sess, operandGroup);
+		if(allParentGroups.contains(operandGroup)) {
```
Vojtech-Sassmann: Shouldn't this verify the presence of the resultGroup instead?
```suggestion
		if(allParentGroups.contains(resultGroup)) {
```

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -245,6 +245,18 @@ public void expireMemberInSubGroupAndInUpperGroup() throws Exception {
...
+	@Test (expected=GroupRelationNotAllowed.class)
+	public void createGroupUnionForIndirectRelationship() throws Exception {
+		System.out.println(CLASS_NAME + "createGroupUnionForIndirectRelationship");
+
+		Vo vo = setUpVo();
+		groupsManagerBl.createGroup(sess, vo, group2);
+		groupsManagerBl.createGroup(sess, group2, group3);
+		groupsManagerBl.createGroup(sess, group3, group4);
+
+		groupsManagerBl.createGroupUnion(sess, group4, group2, false);
```
Vojtech-Sassmann: I think there is an invalid order of the groups. The way it is implemented now, this test creates this relation:
2
|
3
|
4
And calling `createGroupUnion(sess, group4, group2, false)` tries to import members from group2 to group4. This causes a cycle, and I am not sure if we can detect this in the current implementation. Also, the test passes because of the error which I have pointed out above.
```suggestion
		groupsManagerBl.createGroupUnion(sess, group2, group4, false);
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: You are right, this would be always true. I checked wrong group. Fixed.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
stavamichal: Right, groups are switched here.

stavamichal:
> I also have one more question. Shouldn't we check this even in the `moveGroup` method? In case there is a union between some groups, shouldn't we forbid the move?




> I also have one more question. Shouldn't we check this even in the `moveGroup` method? In case there is a union between some groups, shouldn't we forbid the move?

You are right, there is such possibility, I prepared a code change and also a test for moveGroup logic too.

stavamichal:
> The changes seem to be OK. I have just one more question. How about a situation when you are moving group C under group A, but there is also a subgroup of C - C:B, and a union from B to A?

Yes, you are right again. For this reason I changed the checking code in moveGroup method to check all possible forbidden indirect relationship between all moving group subgroups and all destination group parent groups cause it could be anywhere in the structure. Test for mentioned situation was added.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -671,11 +671,42 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
...
+				try {
+					Group parentGroupOfMovingGroup = getParentGroup(sess, movingGroup);
+					movingGroupStructureUnions.remove(parentGroupOfMovingGroup);
+				} catch (ParentGroupNotExistsException ex) {
+					throw new InternalErrorException("Can't find parentGroup for " + movingGroup);
+				}
+			}
+
+			//Get all group stucture of destination group (destination group all all its parent groups)
+			List<Group> destinationGroupStructure = new ArrayList<>();
```
Vojtech-Sassmann: I think the use of Sets, with their intersection operation, would improve the performance on larger group structures.

stavamichal: Changed to using intersection between sets.

PR #2545, opened 2020-01-17
initial perun-synchronizer

mvocu:
- move scheduling and top level synchronizer class away from Perun core (and RPC) into new module
perun-synchronizer,
- add starter class to start-up independent process

This is not yet complete, the infrastructure for actually starting the system process (OS startup file & conf) is still missing.


zlamalp:
We decided not to exclude synchronization logic from the perun-core, but rather make it "switchable" by some flag. So that one instance can be primary and others can be replicas with this logic disabled. We will keep this pull-request for reference at graveyard.

PR #2549, opened 2020-01-28
OpenApi - attribute entity

Vojtech-Sassmann:
* Added property 'entity' to attribute.

* This values is not present in the Attribute class itself, but it is
encoded from the getter method.

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -184,6 +184,9 @@ components:
             displayName: { type: string }
             writable: { type: boolean }
             unique: { type: boolean }
+            entity:
+              type: string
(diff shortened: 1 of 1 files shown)


perun-openapi/openapi.yml
```diff
@@ -184,6 +184,7 @@ components:
             displayName: { type: string }
             writable: { type: boolean }
             unique: { type: boolean }
+            entity: { type: string }
```
martin-kuba:
```suggestion
            entity: { type: string, readOnly: true }
```

Vojtech-Sassmann: Fixed.

PR #2551, opened 2020-01-30
Add new method for testing database performance

stavamichal:
- new method getTimeOfQueryPerformance in DatabaseManager on all layers
 (RPC, api, bl, blImpl), can be called by anyone with role SELF and it
 will count processing time (in nanoseconds) of update operation (set
 timestamp to database table configurations to predefined property)
 - for this purpose, other methods was created to support this
 funcionality like createProperty (to create a new property in
 configurations table), propertyExists (to check if specific property
 already exists or need to be created)
 - creating of this new property was added to initialization process of
 DatabaseManager (if property doesn't exist, it will be created on start
 of Perun system)
 - database schemas were modified to remove existing constraint on
 configurations table to allow any property (not just database version)
 - basic tests of new methods were added
 - database version was increased

 IMPORTANT: database need to be changed manually for this commit,
 constraint need to be removed before using properly

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/DatabaseManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/DatabaseManagerMethod.java
@@ -1,6 +1,8 @@
 package cz.metacentrum.perun.rpc.methods;
 
+import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.rpc.ApiCaller;
 import cz.metacentrum.perun.rpc.ManagerMethod;
 import cz.metacentrum.perun.rpc.deserializer.Deserializer;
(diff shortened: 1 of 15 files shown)


stavamichal:
Method was added also to openapi.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/DatabaseManagerMethod.java
```diff
@@ -50,5 +52,21 @@ public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 *	Get time in ns "nanoseconds" of calling 1 simple update query to DB.
+	 *  This query will update property for this purpose in configurations table.
+	 *
+	 * @exampleResponse "196920"
+	 * @return time of processing query in nanoseconds
+	 */
+	getTimeOfQueryPerformance {
+
+		@Override
+		public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
```
martin-kuba: Please change the return type of the method to Long:
```suggestion
		public Long call(ApiCaller ac, Deserializer parms) throws PerunException {
```

stavamichal: Done.

PR #2560, opened 2020-02-07
OpenApi - attributesManager methods

Vojtech-Sassmann:
* Added new descriptions of methods to the openApi.
* Created new schemas for optional id parameters.

* getAttributeDefinitionWithRights
* removeAttribute
* updateAttributeDefinition
* getAttributeRights
* setAttributeRights

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -202,6 +202,31 @@ components:
       discriminator:
         propertyName: beanName
 
+    ActionType:
+      type: string
+      description: "Action types for attributes."
+      enum:
+        - "WRITE"
+        - "WRITE_VO"
+        - "WRITE_PUBLIC"
+        - "READ"
+        - "READ_VO"
+        - "READ_PUBLIC"
+
+    AttributeRights:
+      type: object
+      required:
+        - attributeId
+        - rights
+        - role
+      properties:
+        attributeId: { type: integer }
+        role: { type: string }
+        rights:
+          type: array
+          items:
+            $ref: '#/components/schemas/ActionType'
+
     Facility:
       allOf:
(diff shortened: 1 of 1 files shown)


perun-openapi/openapi.yml
```diff
@@ -202,6 +202,27 @@ components:
...
+      enum:
+        - "WRITE"
+        - "WRITE_VO"
+        - "WRITE_PUBLIC"
+        - "READ"
+        - "READ_VO"
+        - "READ_PUBLIC"
+
+    AttributeRights:
+      allOf:
```
martin-kuba: allOf is not needed when the type does not inherit from another type

perun-openapi/openapi.yml
```diff
@@ -2892,7 +2994,28 @@ paths:
...
 
-#TODO getAttributesDefinitionWithRights
+  /urlinjsonout/attributesManager/getAttributesDefinitionWithRights:
+    get:
+      tags:
+        - AttributesManager
+      operationId: getAttributesDefinitionWithRights
+      summary: Returns all AttributeDefinitions for every entity and possible
+        combination of entities with rights. Only attribute definition of 
+        attributes user can read (or write) you will get.
```
martin-kuba: The javadoc also contains the following, please add it too:
```
Combination of entities is based on provided parameters, which are optional (at least one must be present).
```

perun-openapi/openapi.yml
Vojtech-Sassmann: Fixed.

perun-openapi/openapi.yml
Vojtech-Sassmann: Summary updated.

PR #2561, opened 2020-02-07
Fixed problem with leap year

metodej:
- the exception is no longer thrown if expiration date is set to 29. 2. of a leap year
- if extension of membership is set to 29. 2. of a non-leap year, it's set to 28. 2. instead


metodej:
Thank you for review, the chosen leap year is now 2000 and the condition is now hopefully more readable.

PR #2565, opened 2020-02-11
Fixed bug in extendDateByStaticDate

metodej:
- the method now works only with today's date

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
@@ -1518,33 +1518,32 @@ public class Utils {
 	}
 
 	/**
-	 * Extends given date by values from given matcher.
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1539,7 +1539,7 @@ public static LocalDate extendDateByStaticDate(LocalDate localDate, Matcher matc
 		extensionInNextYear = extensionDate.isBefore(LocalDate.now());
 
 		// Set the date to which the membership should be extended, can be changed if there was grace period, see next part of the code
-		localDate = LocalDate.of(year, month, day);
+		LocalDate localDate = LocalDate.of(year, month, day);
```
stavamichal: At this moment, you don't need to use localDate, you can use extensionDate and return it.

PR #2570, opened 2020-02-19
Openapi facility

xkureck:
* added new descriptions of methods to the openApi
* also added new schemas, responses and parameters where needed
* make different operationId of every operation


perun-openapi/openapi.yml
```diff
@@ -5032,60 +5160,457 @@ paths:
...
+      parameters:
+        - { name: hostname, schema: { type: string }, in: query, required: true, description: "specific hostname"}
+      responses:
+        '200':
+          $ref: '#/components/responses/ListOfFacilitiesResponse'
+        default:
+          $ref: '#/components/responses/ExceptionResponse'
 
-  #TODO getRichAdminsWithSpecificAttributes
+  /json/facilitiesManager/getAllowedUsers:
```
Vojtech-Sassmann:
```suggestion
  /json/facilitiesManager/getAllowedUsers/v-f-s:
```
There are also other versions of this method and might be in the future added to this file.

Vojtech-Sassmann: You should also add a TODO mentioning that there are missing some versions of this method.

PR #2581, opened 2020-02-24
bumped version in openapi.yml to 3.10.0

martin-kuba:
to keep in sync with Perun API


zlamalp:
Thank you, I will keep in mind to bump openapi.yml version too next time.

PR #2591, opened 2020-03-02
New authorization in TasksManagerEntry

balcirakpeter:
- In TasksManagerEntry was completely replaced the old authorization.
  policies used in the TasksManagerEntry.
  in the future.

### DIFF
--- a/perun-base/src/test/resources/perun-roles.yml
+++ b/perun-base/src/test/resources/perun-roles.yml
@@ -466,4 +466,103 @@ perun_policies:
       - PERUNOBSERVER:
     include_policies:
       - default_policy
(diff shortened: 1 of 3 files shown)


perun-base/src/test/resources/perun-roles.yml
```diff
@@ -124,4 +124,122 @@ perun_policies:
...
+
+
+
+
+
+
+
+
+
+
```
Vojtech-Sassmann: Those empty lines are there on purpose?

balcirakpeter: fixed

PR #2594, opened 2020-03-02
OpenAPI - service manager implemented

xkureck:
* service manager methods completed
* added new schemas, responses, parameters


perun-openapi/openapi.yml
```diff
@@ -8290,296 +8354,1032 @@ paths:
...
+
+  /urlinjsonout/servicesManager/addRequiredAttributes:
+    post:
+      tags:
+        - ServicesManager
+      operationId: addRequiredAttributes
+      summary: Batch version of addRequiredAttribute.
+      parameters:
+        - $ref: '#/components/parameters/serviceId'
+        - name: attributes
```
Vojtech-Sassmann: There is an already created schema named `attributes`. You can reuse it here. Moreover, when passing a list of values in a URL parameter, the name of the parameter should end with the `[]` brackets.

perun-openapi/openapi.yml
```diff
@@ -8290,296 +8354,1032 @@ paths:
...
+
+  /urlinjsonout/servicesManager/removeRequiredAttributes:
+    post:
+      tags:
+        - ServicesManager
+      operationId: removeRequiredAttributes
+      summary: Remove required attributes from service.
+      parameters:
+        - $ref: '#/components/parameters/serviceId'
+        - name: attributes
```
Vojtech-Sassmann: Same here.

perun-openapi/openapi.yml
```diff
@@ -8290,296 +8354,1032 @@ paths:
...
+          $ref: '#/components/responses/ExceptionResponse'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              title: InputAddDestinationToMultipleServices
+              description: "input to addDestinationToMultipleServices"
+              type: object
+              required:
```
Vojtech-Sassmann: If all of the possible input fields are required, it is impossible to call this method with only some of them, which is actually allowed by the implementation. You should probably split this method into multiple versions.

Vojtech-Sassmann: @xkureck this is not fixed

Vojtech-Sassmann:
@xkureck, please, rebase this PR.

Vojtech-Sassmann:
Hi @xkureck, once you rebase this PR, we can probably merge it.

PR #2603, opened 2020-03-03
implemented PHP CLI example

martin-kuba:


perun-cli-php/README.md
```diff
@@ -0,0 +1,13 @@
...
+This module shows how to call Perun RPC from PHP using classes generated from its
+OpenAPI description.
+
+Usage:
+
+```$bash
+sudo apt install php-cli composer phpunit php-xml php-mbstring php-curl
+./generate.sh
+
+./get_user.py --PERUN_USER="perun/test" --user 12345
```
zlamalp: Probably should be ``.php``

martin-kuba: fixed

PR #2606, opened 2020-03-05
CORE: Store mail used in pwd reset notifications

zlamalp:
- We allow sending password reset to non-standard mail addresses by
  choosing source mail attribute. When password is actually re-set,
  we should notify user on the same address as was used first time.
- This change contains DB version update.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -2234,9 +2234,14 @@ public class UsersManagerBlImpl implements UsersManagerBl {
 	public void changeNonAuthzPassword(PerunSession sess, User user, String m, String password, String lang) throws InternalErrorException, LoginNotExistsException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException {
 
 		String requestId = Utils.cipherInput(m, true);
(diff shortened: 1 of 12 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -2234,9 +2234,14 @@ public User convertUserEmptyStringsInObjectAttributesIntoNull(User user) {
...
-		String namespace = getUsersManagerImpl().loadPasswordResetRequest(user, Integer.parseInt(requestId));
+		Pair<String,String> resetRequest = getUsersManagerImpl().loadPasswordResetRequest(user, Integer.parseInt(requestId));
 
-		if (namespace.isEmpty()) throw new InternalErrorException("Password reset request is not valid anymore or doesn't existed at all for User: "+user);
+		if (resetRequest == null) {
+			throw new InternalErrorException("Password reset request is not valid anymore or doesn't existed at all for User: "+user);
+		}
+
+		String namespace = resetRequest.getLeft();
+		String mail = resetRequest.getRight();
```
stavamichal: You still can get null in left or right part. Shouldn't we check it here too?

zlamalp: Namespace is non-null in DB. Mail can be null and I added check before adding its value between known mails at line 2268.

PR #2610, opened 2020-03-08
Overload method authorized

balcirakpeter:
- Both methods authorizedInternal and authorizedExternal have overloaded
  version which does not take list of PerunBeans.
- It was created for scenarios, where there is no need to pass beans
  to autrhorization. It erase necessity to call the authorization with
  Collections.emptyList() argument.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
@@ -20,6 +20,7 @@ import cz.metacentrum.perun.core.impl.AuthzRoles;
 import cz.metacentrum.perun.core.impl.Privileges;
 import cz.metacentrum.perun.core.impl.Utils;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -39,6 +40,20 @@ public static boolean authorizedExternal(PerunSession sess, String policyDefinit
...
+	 * Checks if the principal is authorized.
+	 * Used when there are no PerunBeans needed for authorization.
+	 * This method should be accessed through external components.
+	 *
+	 * @param sess PerunSession which contains the principal.
+	 * @param policyDefinition of policy which contains authorization rules.
+	 * @return true if the principal has particular rights, false otherwise.
+	 * @throws PolicyNotExistsException when the given policyDefinition does not exist in the PerunPoliciesContainer.
+	 */
+	public static boolean authorizedExternal(PerunSession sess, String policyDefinition) throws PolicyNotExistsException {
```
Vojtech-Sassmann: I think it is not necessary to provide this version of the method in RPC.

balcirakpeter: Erased

PR #2611, opened 2020-03-08
New authorization in ServicesManagerEntry

balcirakpeter:
- In ServicesManagerEntry was completely replaced the old authorization.
  policies used in the ServicesManagerEntry.
  in the future.


perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
```diff
@@ -38,10 +39,13 @@
 import cz.metacentrum.perun.core.bl.ServicesManagerBl;
 import cz.metacentrum.perun.core.blImpl.PerunBlImpl;
 import cz.metacentrum.perun.core.impl.Utils;
+import org.checkerframework.checker.units.qual.C;
```
Vojtech-Sassmann: Please, remove this unused import. Otherwise, I went through all of the records, and it seems it should work the same way as with the old authorization.

balcirakpeter: Fixed

PR #2614, opened 2020-03-09
marked most of Maven submodules as skipped during deploy

martin-kuba:
most of Maven submodules are separate just for organization purposes, only perun-rpc and perun-openapi may need to be deployed

### DIFF
--- a/perun-engine/pom.xml
+++ b/perun-engine/pom.xml
@@ -21,6 +21,8 @@
 		<start-class>cz.metacentrum.perun.engine.main.EngineStarter</start-class>
 		<!-- module properties used by all profiles -->
 		<perun.test.groups>unit-tests</perun.test.groups>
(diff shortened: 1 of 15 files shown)


perun-engine/pom.xml
```diff
@@ -21,79 +21,21 @@
...
-
-			<!-- Prepare META-INF/MANIFEST.MF entries for JAR file -->
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-jar-plugin</artifactId>
-				<configuration>
-					<archive>
-						<index>true</index>
-						<manifestEntries>
-							<SplashScreen-Image>perun-engine-splash.png</SplashScreen-Image>
```
zlamalp: Probably you can remove image from the resources too.

PR #2620, opened 2020-03-13
New login attribute for SURF RAM

licehammer:


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_surf_ram_persistent_shadow.java
```diff
@@ -0,0 +1,92 @@
...
+		} catch (ExtSourceNotExistsException ex) {
+			throw new InternalErrorException("IdP external source for SURF RAM doesn't exist.", ex);
+		}
+	}
+
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_USER_ATTR_DEF);
+		attr.setFriendlyName("login-namespace:surf-ram-persistent-shadow");
+		attr.setDisplayName("FENIX login");
```
zlamalp: Probably should be ``SURF RAM login``, same for description.

licehammer: Cup&paste error. Thanks for spotting it.

PR #2625, opened 2020-03-17
isAuthorized method Deprecated

balcirakpeter:
- Method isAuthorized() was deprecated because of the new authorization logic,
  which is evaluated through the method authorizedInternal().

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
@@ -83,6 +83,7 @@ public class AuthzResolver {
 	 * @return true if the principal authorized, false otherwise
 	 * @throws InternalErrorException if something goes wrong
 	 */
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -83,6 +83,7 @@ public static boolean authorizedInternal(PerunSession sess, String policyDefinit
 	 * @return true if the principal authorized, false otherwise
 	 * @throws InternalErrorException if something goes wrong
 	 */
+	@Deprecated
```
zlamalp: What about ``isAuthorized(PerunSession sess, String role)`` at line 350/351. Shouldn't it be deprecated too?

balcirakpeter: Fixed

PR #2628, opened 2020-03-18
getUserExtSourceFromMultipleIdentifiers null

balcirakpeter:
- Previous implementation of the method returned null
  when the UserExtSource was not found. That caused NullPointerException
  later in the method getUserByUserExtSource which was called by the method
  getPerunSession.
- Therefore, the method getUserExtSourceFromMultipleIdentifiers was changed
  to throw the UserExtSourceNotExistException when the UserExtSource was not found.
- In this method was also check for null additionalIdentifiers, which was throwing
  the UserExtSourceNotExistsException. This exception was replaced by the
  WrongReferenceAttributeValueException, which is usually used when an attribute
  is null and it cannot be null. The exception is catched in getPerunSession same as
  the USerExtSourceNotExistsException so the behaviour will not change.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -706,10 +706,10 @@ public class UsersManagerBlImpl implements UsersManagerBl {
 	}
 
 	@Override
-	public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess, PerunPrincipal principal) throws InternalErrorException, UserExtSourceNotExistsException {
+	public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess, PerunPrincipal principal) throws UserExtSourceNotExistsException {
 		String additionalIdentifiers = principal.getAdditionalInformations().get(additionalIdentifiersAttributeName);
 		if (additionalIdentifiers == null) {
-			throw new UserExtSourceNotExistsException("Mandatory attribute is not defined: ".concat(additionalIdentifiersAttributeName));
+			throw new InternalErrorException("Entry " + additionalIdentifiersAttributeName + " is not defined in the principal's additional information. Either it was not provided by external source used for sign-in or the mapping configuration is wrong.");
 		}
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -706,10 +706,10 @@ private UserExtSource getUserExtSourceByUniqueAttributeValue(PerunSession sess,
 	}
 
 	@Override
-	public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess, PerunPrincipal principal) throws InternalErrorException, UserExtSourceNotExistsException {
+	public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess, PerunPrincipal principal) throws InternalErrorException, UserExtSourceNotExistsException, WrongReferenceAttributeValueException {
 		String additionalIdentifiers = principal.getAdditionalInformations().get(additionalIdentifiersAttributeName);
 		if (additionalIdentifiers == null) {
-			throw new UserExtSourceNotExistsException("Mandatory attribute is not defined: ".concat(additionalIdentifiersAttributeName));
+			throw new WrongReferenceAttributeValueException("Principal's mandatory attribute is not defined: ".concat(additionalIdentifiersAttributeName));
```
licehammer: Are you sure this is not rather a configuration error? That would mean using InternalErrorException.

stavamichal: InternalErrorException would be definitely better here. This is probably not attribute as we are using it in Perun. It is more like missing information.

balcirakpeter: Changed to InternalErrorexception.

licehammer: What about reworking the exception message as well? I will probablly have hard time to figure out that I have to change the config file after I will find this message in logs.

balcirakpeter: Done

PR #2632, opened 2020-03-21
eduTEAMS fenix extSourceName and domainName

balcirakpeter:
- Added support to configure extSourceName and domainName properties,
  which are used in the module urn_perun_user_attribute_def_def_login_namespace_fenix_persistent_shadow
  because they are different for fenix and fenix-pilot instances.


licehammer:
I can imagine having a configuration folder for configuration file for modules. Something like /etc/perun/modules/urn_perun_user_attribute_def_def_login_namespace_fenix_persistent_shadow

If it's not possible I can live with suggestion from @zlamalp. Maybe one day we will deploy different modules for different instances, which will solve it as well :-)

PR #2634, opened 2020-03-23
Fix problem with creating a group union

stavamichal:
- there was a problem when two groups (for example 'A' and 'AB' was incorrectly checked as predecessors. Only prefix of groups should be test there
- for this reason, we need to set result group as the full prefix using ':' character on it's name end

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -4161,7 +4161,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 		// check indirect relationships (for example "A" with "A:B:C" are in indirect relationship using "A:B")
 		// looking for situation where result group is predecessor of operand group (by name) but not a parent of it (which is ok)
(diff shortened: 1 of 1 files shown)


stavamichal:
I know about test failure, will be checking it later.

perun-base/src/main/java/cz/metacentrum/perun/core/api/Group.java
```diff
@@ -116,6 +116,19 @@ public void setShortName(String shortName) {
...
 	}
 
+	/**
+	 * Get whole parent group prefix.
+	 * For example for group A:B:C return "A:B"
+	 * For group A return empty String (there is no prefix in such case)
+	 *
+	 * @return whole prefix (without the last ':' character)
+	 */
+	public String getPrefix() {
```
zlamalp: This will affect all (de)serialization of the Group object. It you make it utils method, we can have this change as a quick fix in the nearest release. Otherwise it must me updated on other places too and we must notify other parties before deployment.

stavamichal: Logic was changed.

stavamichal:
The whole logic of fix was changed, it is now much simple.

PR #2641, opened 2020-03-26
Few OpenAPI methods added

xkostka2:
* getUserRoleNames method was added into authzResolver manager
* createAlternativePassword method was added into user manager
* deleteAlternativePassword method was added into user manager

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -2278,6 +2278,20 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/authzResolver/getUserRoleNames:
+    get:
+      tags:
+        - AuthzResolver
+      operationId: getUserRoleNames
+      summary: Returns list of user's role names. Perun system uses role names in the upper case format. However, for now, they are converted to the lower case format because of the compatibility with external systems.
+      parameters:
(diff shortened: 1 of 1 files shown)


perun-openapi/openapi.yml
```diff
@@ -5454,6 +5468,64 @@ paths:
...
+            type: string
+          }
+          in: query
+          required: true
+        - name: password
+          schema: {
+            type: string
+          }
+          in: query
+          required: true
```
martin-kuba:
```suggestion
        - name: description, schema: { type: string }, in: query, required: true
        - name: loginNamespace, schema: { type: string }, in: query, required: true
        - name: password, schema: { type: string }, in: query, required: true
```
Please use the shorter YAML syntax for parameters, it is easier to read.

perun-openapi/openapi.yml
```diff
@@ -5454,6 +5468,64 @@ paths:
...
+        - name: passwordId
+          schema: {
+            type: string
+          }
+          in: query
+          required: true
```
martin-kuba:
```suggestion
        - name: loginNamespace, schema: { type: string }, in: query, required: true
        - name: passwordId, schema: { type: string }, in: query, required: true
```

PR #2643, opened 2020-03-28
New authorization in ResourcesManagerEntry

balcirakpeter:
- In ResourcesManagerEntry was completely replaced the old authorization.
  policies used in the ResourcesManagerEntry.
  in the future.
- If there was a situation, where tvwo objects could not be in the same
  list, one of them was put to the separation check with its own policy.

### DIFF
--- a/perun-base/src/test/resources/perun-roles.yml
+++ b/perun-base/src/test/resources/perun-roles.yml
@@ -1554,6 +1554,648 @@ perun_policies:
     include_policies:
       - default_policy
 
+  #ResourcesManagerEntry
+  getResourceById_int_policy:
+    policy_roles:
+      - RPC:
+      - VOADMIN: Vo
+      - VOOBSERVER: Vo
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  getRichResourceById_int_policy:
+    policy_roles:
+      - FACILITYADMIN: Facility
+      - VOADMIN: Vo
+      - VOOBSERVER: Vo
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  getResourceByName_Vo_Facility_String_policy:
+    policy_roles:
+      - RPC:
+      - VOADMIN: Vo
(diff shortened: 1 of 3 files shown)


perun-base/src/test/resources/perun-roles.yml
```diff
@@ -167,4 +167,616 @@ perun_policies:
...
+    include_policies:
+      - default_policy
+
+  filter-getResourcesWhereUserIsAdmin_Facility_Vo_User_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+      - RESOURCEADMIN: Resource
+      - VOADMIN: Vo
+    include_policies:
+      - default_policy
```
stavamichal, line 623: There is missing rules for method getResourcesWhereUserIsAdmin(PerunSession sess, Vo vo, User authorizedUser) You are using the same rules for two methods.

balcirakpeter: Fixed

balcirakpeter:
Conflicts will be resolved before merge.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -1045,19 +906,15 @@ public void copyGroups(PerunSession sess, Resource sourceResource, Resource dest
...
-		}
-
-		//Resource manager can see only resources where he has role resource manager (filter them)
-		if(AuthzResolver.isAuthorized(sess, Role.RESOURCEADMIN)) {
-			return filterNotAuthorizedResource(sess, resources);
-		} else {
+		//Authorization
+		List<PerunBean> beans = new ArrayList<>(resources);
+		beans.addAll(Arrays.asList( vo, authorizedUser));
+		if(!AuthzResolver.authorizedInternal(sess, "getResourcesWhereUserIsAdmi_Vo_User_policy", beans)){
```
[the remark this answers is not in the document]
balcirakpeter: fixed

PR #2648, opened 2020-03-30
Fix wrong exception when updating group with existing name

stavamichal:
- when we are trying to update a group's name with name which is
 already occupied in the same VO, we need to throw GroupExistsException
 instead of SQL constraint exception

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
@@ -200,7 +200,7 @@ public class GroupsManagerEntry implements GroupsManager {
 	}
 
 	@Override
-	public Group updateGroup(PerunSession sess, Group group) throws GroupNotExistsException, InternalErrorException, PrivilegeException {
+	public Group updateGroup(PerunSession sess, Group group) throws GroupNotExistsException, GroupExistsException, InternalErrorException, PrivilegeException {
 		Utils.checkPerunSession(sess);
 		getGroupsManagerBl().checkGroupExists(sess, group);
 		Utils.notNull(group, "group");
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
@@ -29,6 +29,7 @@ import cz.metacentrum.perun.core.blImpl.AuthzResolverBlImpl;
 import cz.metacentrum.perun.core.implApi.GroupsManagerImplApi;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.JdbcPerunTemplate;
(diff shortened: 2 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -200,7 +200,7 @@ public void deleteGroups(PerunSession perunSession, List<Group> groups, boolean
 	}
 
 	@Override
-	public Group updateGroup(PerunSession sess, Group group) throws GroupNotExistsException, InternalErrorException, PrivilegeException {
+	public Group updateGroup(PerunSession sess, Group group) throws GroupNotExistsException, GroupExistsException, InternalErrorException, PrivilegeException {
```
zlamalp: We can check this using SQL constraint itself in impl/dao layer, since we update name and description separately. We can catch ``DataIntegrityViolationException`` when updating name and then throw ``GroupExistsException`` rather than current ``InternalErrorException``. See example in ``FacilitiesManagerImpl`` method ``updateFacility()``.

stavamichal: Yes, you are right. I changed that logic to be same as for Facility.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -212,6 +214,8 @@ public Group updateGroup(PerunSession sess, Group group) throws InternalErrorExc
 			try {
 				jdbc.update("update groups set name=?,modified_by=?, modified_by_uid=?, modified_at=" + Compatibility.getSysdate() + " where id=?", dbGroup.getName(),
 						sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getUserId(), dbGroup.getId());
+			} catch (DataIntegrityViolationException e) {
+				throw new GroupExistsException("The name must be unique and it's already occupied.", e);
```
zlamalp: Maybe we should include used name in the exception text.

stavamichal: I don't really think we need to know that. And it is exactly the same copy of the logic like in facilityManager.

stavamichal: Do you insist? Of course I can add it, but I don't think it is important except the situation when database constraint works improperly.

zlamalp: Not really, but since its in DAO layer, it might fail also some inner logic (move group?, group structure sync?). I'm not sure. Right now it won't be possible to know, which group caused the problem, unless it is logged anywhere else, since even passed SQL exception doesn't contain input data.

stavamichal: Ok - added.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -18,6 +18,7 @@
 import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyMemberException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
+import cz.metacentrum.perun.core.api.exceptions.FacilityExistsException;
```
balcirakpeter: Unused import.

PR #2651, opened 2020-04-03
Module for attribute virt:resource:assignedGroups

stavamichal:
- module will compute list of group names assigned to the resource
 (optionally also with all their subgroups)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_virt_assignedGroups.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_virt_assignedGroups.java
@@ -1,0 +1,100 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.Resource;
+import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_virt_assignedGroups.java
```diff
@@ -0,0 +1,104 @@
...
+	 * @return computed group name
+	 */
+	private String computeGroupName(Vo vo, Group group) {
+		return vo.getShortName() + ":" + group.getName();
+	}
+
+	@Override
+	public List<String> getStrongDependencies() {
+		List<String> strongDependencies = new ArrayList<>();
+		strongDependencies.add(AttributesManager.NS_VO_ATTR_CORE + ":shortName");
```
zlamalp, line 90: Why ? We retrieve value from the object ``Vo`` and do not call ``getAttribute()`` on such core attribute.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_virt_assignedGroups.java
```diff
@@ -0,0 +1,104 @@
...
+		for(Group assignedGroup: assignedGroups) {
+			allUniqueGroupNames.add(computeGroupName(vo, assignedGroup));
+			if(isAssignedWithSubgroups) {
+				for(Group subgroup: sess.getPerunBl().getGroupsManagerBl().getAllSubGroups(sess, assignedGroup)) {
+					allUniqueGroupNames.add(computeGroupName(vo, subgroup));
+				}
+			}
+		}
+
+		attribute.setValue(new ArrayList(allUniqueGroupNames));
```
zlamalp, line 71: Maybe we should sort list of group names to ensure consistent value is returned on each request. I think we can use ``SortedSet`` to do it on the fly.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_virt_assignedGroups.java
stavamichal: Copy&Paste mistake. Fixed. :)

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_resource_attribute_def_virt_assignedGroups.java
stavamichal: Added.

PR #2656, opened 2020-04-03
Added method to reactivate member in group

metodej:
- the method validates member and then sets membershipExpirationAttribute based on membershipExpirationRules as if he was added to group today
- also added tests for this method

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -4948,4 +4948,31 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 	public List<Facility> getFacilitiesWhereGroupIsAdmin(PerunSession perunSession, Group group) throws InternalErrorException {
 		return this.getGroupsManagerImpl().getFacilitiesWhereGroupIsAdmin(perunSession, group);
 	}
+
+	@Override
+	public void reactivateMember(PerunSession sess, Member member, Group group) throws InternalErrorException, MemberNotExistsException {
+		if (!isGroupMember(sess, group, member)) throw new MemberNotExistsException("Member does not belong to this group");
+
+		validateMemberInGroup(sess, member, group);
+
+		// Get current membershipExpiration date attribute
+		Attribute membershipExpirationAttribute = getMemberExpiration(sess, member, group);
+		// Set new value of the membershipExpiration for the member
+		membershipExpirationAttribute.setValue(null);
+
+		try {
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4948,4 +4948,61 @@ private boolean isMemberInGracePeriod(Map<String, String> membershipExpirationRu
...
 		return this.getGroupsManagerImpl().getFacilitiesWhereGroupIsAdmin(perunSession, group);
 	}
+
+	@Override
+	public void reactivateMember(PerunSession sess, Member member, Group group) throws InternalErrorException, MemberNotExistsException {
+		if (!isGroupMember(sess, group, member)) throw new MemberNotExistsException("Member does not belong to this group");
+
+		validateMemberInGroup(sess, member, group);
+
+		Attribute membershipExpirationRulesAttribute = getMembershipExpirationRulesAttribute(sess, group);
```
zlamalp, line 4958: I just wonder, if setting group membership expiration to null and calling extendMembershipInGroup() wouldn't do the same job, so we could prevent code duplicity?

stavamichal: We should definitely prevent code duplicity if possible.

metodej: Yes, it would, thank you, it should be fixed now.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4948,4 +4948,30 @@ private boolean isMemberInGracePeriod(Map<String, String> membershipExpirationRu
...
+		} catch (WrongAttributeValueException e) {
+			throw new InternalErrorException("Wrong value: " + membershipExpirationAttribute.getValue(),e);
+		} catch (WrongReferenceAttributeValueException | WrongAttributeAssignmentException e) {
+			throw new InternalErrorException(e);
+		}
+
+		try {
+			extendMembershipInGroup(sess, member, group);
+		} catch (ExtendMembershipException ex) {
+			// This exception should not be thrown for null membershipExpiration attribute
```
stavamichal, line 4974: If exception shouldn't be thrown (you are not expecting it), you should throw an InternalErrorException otherwise.

metodej: Okay, it should be fixed now.

PR #2660, opened 2020-04-07
Force multiple identifiers

balcirakpeter:
Added force option for multiple identifiers which permits to use
multiple identifiers even when theprincipal's extSource is not in the
multipleIdentifiers extSources.


balcirakpeter:
This fix works but I probably figured out better and much simpler solution for this problem.

balcirakpeter:
The problem is solved in PR #2661

PR #2661, opened 2020-04-08
Multiple Identifiers with shibIdentityProvider

balcirakpeter:
Problem: Multiple identifiers did not work alongside with original IdPs functionality,
                because configuration for multiple identifiers had set proxy id
		but principals extSource contained original Idp id. Therefore,
		the principal's extSource was never found in multiple identifiers list,
		thus the additional identifiers were not used.
Solution:  Checking, if the principal's extSource is in the list of multiple identifiers configuration,
                was changed. It does not check principal's extSource but principal's originIdentityProvider,
	        which is set to proxy id when a user logs in through some federation.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -1742,10 +1742,11 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 		log.trace("Refreshing session data for session {}.", sess);
 
 		PerunPrincipal principal = sess.getPerunPrincipal();
+		String shibIdentityProvider = principal.getAdditionalInformations().get(PerunBl.ORIGIN_IDENTITY_PROVIDER_KEY);
 
 		try {
 			User user;
-				if(extSourcesWithMultipleIdentifiers.contains(principal.getExtSourceName())) {
+				if(shibIdentityProvider != null && extSourcesWithMultipleIdentifiers.contains(shibIdentityProvider)) {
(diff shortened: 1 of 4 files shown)


balcirakpeter:
Functionality has been tested on acceptance instance.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -118,6 +118,7 @@
 
 	private final static Set<String> dontLookupUsersForLogins = BeansUtils.getCoreConfig().getDontLookupUsers();
 	private final static Set<String> extSourcesWithMultipleIdentifiers = BeansUtils.getCoreConfig().getExtSourcesMultipleIdentifiers();
+	private final static String shibIdentityProvider_key = "originIdentityProvider";
```
stavamichal, line 121: You can use one public static String identifier, instead create a duplicate of it.

balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -97,6 +97,7 @@
 	private static final String SET_ROLE = "SET";
 
 	private final static Set<String> extSourcesWithMultipleIdentifiers = BeansUtils.getCoreConfig().getExtSourcesMultipleIdentifiers();
+	private final static String shibIdentityProvider_key = "originIdentityProvider";
```
zlamalp, line 100: I believe this can be removed now.

balcirakpeter: My bad, fixed.

PR #2662, opened 2020-04-08
Default loa for IdP

balcirakpeter:
Problem: When proxy did not send loa attribute, perun set it to 2. In eduTeams we need to configure this default value.
Solution: New configuration property was created for this purpose. when it is not set, the default value remains 2.


balcirakpeter:
Changes have been tested on the acceptance instance.

PR #2666, opened 2020-04-09
REGISTRAR: More robust mail validation

zlamalp:
- User triggered mail validation is now more robust. It fakes
  OK response on already approved or rejected applications
  to prevent their processing.
- We try to verify only new applications.
- We try to auto-approve only verified applications.
- Make sure we do not send mail validation notification for
  mail addresses, which have assurance level >= 1, since it
  means address is verified. It was previously possible to
  notify also loa=2 addresses.


zlamalp:
Tested on devel, it prevents multiple attempts on auto-approval, which spams admin about errors during auto-approval process.

PR #2668, opened 2020-04-14
Modules configuration

balcirakpeter:
- Created functionality for modules' configuration. All modules can have their separate config while, for
  instance specific properties which will be accesible through the singleton class ModulesConfigContainer.
- new model is  already applied in urn_perun_user_attribute_def_def_login_namespace_fenix_persistent_shadow
  module.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/ModulesConfigContainer.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/ModulesConfigContainer.java
@@ -1,0 +1,110 @@
+package cz.metacentrum.perun.core.api;
+
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Singleton class ModulesConfigContainer used to store and fetch modules configurations.
+ * It is loaded by the Spring and it is accessible through the static getInstance() method.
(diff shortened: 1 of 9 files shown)


balcirakpeter:
The functionality has been tested on eduTeams acceptance instance.
IMPORTANT -> It is necessary to create a new directory called "modules" in /etc/perun/ before deployment.

perun-base/src/main/java/cz/metacentrum/perun/core/api/ModulesConfigContainer.java
```diff
@@ -0,0 +1,110 @@
...
+	 * @param moduleName name of the module for which the property is stored
+	 * @param propertyName name of the property
+	 * @return property value
+	 */
+	private String fetchPropertyValue(String moduleName, String propertyName) {
+		ModuleConfiguration moduleConfig = modulesProperties.get(moduleName);
+		if (moduleConfig == null)
+			throw new InternalErrorException("Configuration for module " + moduleName + " does not exists.");
+		String propertyValue = moduleConfig.getModuleProperties().get(propertyName);
+		if (propertyValue == null || propertyValue.isEmpty())
```
zlamalp, line 106: I just wonder, if we couldn't use empty value as a correct input for some logic. Like value is either set or not, then some logic performs specific or generic behavior. You could fall-back to default values by type like empty list / null / false etc.

zlamalp:
Implemented in #2705, closing this one.

PR #2671, opened 2020-04-17
CORE: Fixed audit message resolving for eduPersonScopedAffiliations

zlamalp:
- Shorter code when resolving affiliations from users groups.
- Added getMembersByUserWithStatus() into MembersManagerBl, it
  returns only those members of users, which have specified status,
  eg. VALID.
- Module for user:virt:eduPersonScopedAffiliations now reacts also
  on necessary group membership and member status events.
  Implemented new message resolving within module for case of source
  group attribute changes, since we must iterate over current group
  users.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
@@ -151,7 +151,7 @@ public class MembersManagerImpl implements MembersManagerImplApi {
 	public List<Member> getMembersByUserWithStatus(PerunSession sess, User user, Status status) throws InternalErrorException {
 		try {
 			return jdbc.query("SELECT " + memberMappingSelectQuery + " FROM" +
-							" members WHERE members.user_id=? and members.status "+Compatibility.castToInteger()+"=?",
+							" members WHERE members.user_id=? and members.status"+Compatibility.castToInteger()+"=?",
 					MEMBER_MAPPER, user.getId(), status.getCode());
 		} catch (EmptyResultDataAccessException ex) {
 			throw new InternalErrorException(new MemberNotExistsException("user=" + user));
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
@@ -148,6 +148,19 @@ public class MembersManagerImpl implements MembersManagerImplApi {
 	}
 
 	@Override
+	public List<Member> getMembersByUserWithStatus(PerunSession sess, User user, Status status) throws InternalErrorException {
+		try {
+			return jdbc.query("SELECT " + memberMappingSelectQuery + " FROM" +
+							" members WHERE members.user_id=? and members.status "+Compatibility.castToInteger()+"=?",
+					MEMBER_MAPPER, user.getId(), status.getCode());
+		} catch (EmptyResultDataAccessException ex) {
+			throw new InternalErrorException(new MemberNotExistsException("user=" + user));
+		} catch (RuntimeException err) {
+			throw new InternalErrorException(err);
+		}
+	}
+
+	@Override
 	public Member getMemberById(PerunSession sess, int id) throws InternalErrorException, MemberNotExistsException {
 		try {
 			return jdbc.queryForObject("SELECT " + memberMappingSelectQuery + " FROM members "
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
@@ -1,20 +1,38 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AllAttributesRemovedForGroup;
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AllAttributesRemovedForUser;
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeChangedForUser;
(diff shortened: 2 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -147,6 +147,19 @@ public Member getMemberByUserId(PerunSession sess, Vo vo, int userId) throws Int
 		}
 	}
 
+	@Override
+	public List<Member> getMembersByUserWithStatus(PerunSession sess, User user, Status status) throws InternalErrorException {
+		try {
+			return jdbc.query("SELECT " + memberMappingSelectQuery + " FROM" +
+							" members WHERE members.user_id=? and members.status "+Compatibility.castToInteger()+"=?",
```
stavamichal, line 154: Are you sure this is necessary?  I can't find casting to integer in any other method (using status) in this manager.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -210,28 +207,133 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+
+			AttributeDefinition attributeDefinition = perunSession.getPerunBl().getAttributesManagerBl().getAttributeDefinition(perunSession, getDestinationAttributeName());
+			// TODO - get only active group users, since expired are not affected by current group affiliations
+			List<User> users = perunSession.getPerunBl().getGroupsManagerBl().getGroupUsers(perunSession, ((AttributeRemovedForGroup) message).getGroup());
+			for (User user : users) {
+				resolvingMessages.add(new AttributeChangedForUser(new Attribute(attributeDefinition), user));
+			}
+
+		} else if (message instanceof AllAttributesRemovedForGroup &&
+				!VosManager.MEMBERS_GROUP.equals(((AllAttributesRemovedForGroup) message).getGroup().getName())) {
```
stavamichal, line 310: Wrong indentation.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
```diff
@@ -210,28 +207,133 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
...
+
 		});
-		handleIdentifiers.add(auditEvent -> {
-			if (auditEvent instanceof AttributeSetForUser && ((AttributeSetForUser) auditEvent).getAttribute().getFriendlyName().equals(getSecondarySourceAttributeFriendlyName())) {
-				return ((AttributeSetForUser) auditEvent).getUser().getId();
-			} else {
-				return null;
+
+		return handleIdentifiers;
+	}
```
stavamichal, line 244: I am not sure if this will work when you are overriding method resolveVirtualAttributeValueChange.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
zlamalp: I didn't want to end up with doubling the select based on DB type, like in ```getMembersCount()```. It is used like that in ```getGroupMembers()``` in ```GroupsManagerImpl```. The problem is, that jdbc postgres driver fails to convert passed number for the ```char(1)``` column, while oracle driver has no problem. But we could probably clean up all usages to always convert passed ```Status``` object to the string.

Anyway, I'll test it on devel, since I noticed whitespace before that cast and I'm not sure it won't fail the sql.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
zlamalp: I believe, calling ```super.resolveVirtualAttributeValueChange()``` at the start of own implementation of ``resolveVirtualAttributeValueChange()`` should still in its logic call ```getHandleIdentifiers()``` on actual module/class, since its overridden too. Inside of it it calls super to add generic message handlers, then adds our member messages handlers and everything is processed by super logic of resolveVirtualAttributeValueChange. Then it process group messages and returns resolved messages.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
zlamalp: File is auto aligned consistently with two tabs, if its break of expression or method call chain. It seems to be like that in other files too. One tab is reserved for default indent, like class and its content, or method and its content or generally any code block and its content.

zlamalp:
I manually tested it on devel. Member and group-member status changes create necessary auditer message (usually together with groupNames). Also value is correctly calculated for those changes. Group attribute value change is also reflected.
Used code on postgres perform mentioned conversion to integer without error. We can generally unify its usage for member like methods in own pull-request.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
stavamichal: My bad, I missed that line.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliations.java
stavamichal: I see it now. Missed that exclamation mark (now i see it is part of if clause).

stavamichal:
Ok, if it's working, it can be merged.

PR #2675, opened 2020-04-24
CORE: Fix allowed resource recognition, assignGroup optimization

zlamalp:
- Set group and group_resource required attributes together.
- We now set required attributes only to the valid, expired
  and suspended group members (same as assign service does).
- Optimized isTrulyRequired(user,facility), now we resolve
  allowed resources and check required attributes on them.
  For true it previously double checked allowed facilities,
  assigned and allowed resources.
- Fixed getAllowedResources(facility,user) which previously returned
  also resources, where member was disabled.
- Fixed getAllowedResources(member) which previously returned
  also resources, where member was disabled.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
@@ -5097,7 +5097,8 @@ public class AttributesManagerImpl implements AttributesManagerImplApi {
 	 */
 	GroupResourceVirtualAttributesModuleImplApi getResourceGroupVirtualAttributeModule(PerunSession sess, AttributeDefinition attribute) throws InternalErrorException {
 		Object attributeModule = getAttributesModule(sess, attribute);
-		if (attributeModule == null) return null;
+		if (attributeModule == null)
+			throw new ModuleNotExistsException("Virtual attribute module for " + attribute + " doesn't exists.");
 
 		if (attributeModule instanceof GroupResourceVirtualAttributesModuleImplApi) {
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -5097,7 +5097,8 @@ private UserExtSourceAttributesModuleImplApi getUserExtSourceAttributeModule(Per
 	 */
 	GroupResourceVirtualAttributesModuleImplApi getResourceGroupVirtualAttributeModule(PerunSession sess, AttributeDefinition attribute) throws InternalErrorException {
 		Object attributeModule = getAttributesModule(sess, attribute);
-		if (attributeModule == null) return null;
+		if (attributeModule == null)
+			throw new ModuleNotExistsException("Virtual attribute module for " + attribute + " doesn't exists.");
```
Vojtech-Sassmann, line 5101: I just would like to ask, why is this change made only to these three methods and not to all of them? This way, the module accessing methods behave differently for different combinations of entities.

zlamalp: Module is "required" only for virtual attributes, since it implements ``getAttributeValue()`` to actually compute the value. Other getters for virtual attribute modules were already correct, so this commit fixes implementation to be consistent for all virtual attributes.

Vojtech-Sassmann: I see, thanks for the explanation.

PR #2678, opened 2020-04-27
ExtSourceSql with pooling connections

martin-kuba:
Changed implementation of **ExtSourceSql** and **ExtSourcesManagerImpl** to use database connections obtained from pooling DataSources instead of keeping a special connection open for each one.
- changed parsing of XML file perun-extSources.xml in ExtSourcesManagerImpl 
  - replaced parsing library DOM with DOM4J, it is much easier to use 
  - replaced project dependency on JDOM, which was not used, with dependency on DOM4J
  - parsing of perun-extSources.xml detects new tags **&lt;dbpool>** and creates a read-only Hikari pool for each one (special mark **&lt;main/>** is available for referencing the main pool instead of creating a new one)
  - extSources of type ExtSourceSql  now support a new attribute **dbpool** referencing a db pool
  - for backward compatibility, extSources of type ExtSourceSql with legacy attributes **url**, **driver**, **user** and **password** are detected and dbpools are created for them
- added method **destroy()** to PerunBlImpl, ExtSourcesManagerBl, ExtSourcesManagerBlImpl, ExtSourcesManagerImplApi and ExtSourcesManagerImpl which closes created DataSources on shutdown
- changed implementation of methods **querySource()** and **groupQuery()** to be faster by pre-processing result set metadata outside of the reading loop and reading only available columns, thus avoiding throwing-and-catching exceptions just to detect missing columns on each row
- rewrote  **ExtSourcesManagerBlImpl** to cache instances of ExtSources in memory to avoid a lot of database queries
- replaced injection of PerunBl instance into static variable in ExtSource* classes from perun-core.xml with all ExtSource descendants implementing a new class EtxSourceImpl that has a setter for PerunBl, and this is called right after creation of a new instance in EXTSOURCE_MAPPER
- replaced usages of EXT_SOURCES_EXTRACTOR with EXTSOURCE_MAPPER in ExtSourcesManagerImpl and changed SQL queries to spare useless extracting and immediate throwing away of attribute values when reading just ExtSources
- in AuthzResolverIntegrationTest added refreshing ExtSources before every test

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
@@ -27,7 +27,6 @@ import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationExc
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.SubjectNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
-
 import cz.metacentrum.perun.core.bl.ExtSourcesManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.impl.Utils;
@@ -36,6 +35,7 @@ import cz.metacentrum.perun.core.implApi.ExtSourcesManagerImplApi;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import javax.sql.DataSource;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -43,6 +43,7 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 public class ExtSourcesManagerBlImpl implements ExtSourcesManagerBl {
 	final static Logger log = LoggerFactory.getLogger(ExtSourcesManagerBlImpl.class);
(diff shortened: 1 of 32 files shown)


martin-kuba:
deleted branch by mistake

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
```diff
@@ -119,299 +125,153 @@ public ExtSourceSql() {
...
-						map.put("login", rs.getString("login"));
-					} catch (SQLException e) {
-						// If the column doesn't exists, ignore it
-						map.put("login", null);
+	protected List<Map<String,String>> querySource(String query, String searchString, int maxResults) throws InternalErrorException {
+		log.debug("Searching for '{}' in external source '{}'", searchString, getName());
+		try (Connection con = dataSource.getConnection()) {
+			try (PreparedStatement st = con.prepareStatement(query)) {
+				// Substitute the ? in the query by the searchString
+				if (searchString != null && !searchString.isEmpty()) {
```
zlamalp, line 152: We can use ``StringUtils.isNotEmpty(searchString)`` or even ``StringUtils.isNotBlank(searchString)``. Its not necessary, I just like those methods, since they can make code more readable if we check multiple conditions.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
```diff
@@ -119,299 +125,153 @@ public ExtSourceSql() {
...
 								} catch (IOException ex) {
-									log.error("Unable to read BLOB for column {}", columnName);
-									throw new InternalErrorException("Unable to read BLOB data for column: " + columnName, ex);
+									throw new InternalErrorException("Unable to read BLOB data for column: " + columnMapping.columnName, ex);
 								}
 							} else {
-								// let driver to convert type to string
-								attributeValue = rs.getString(i);
-							}
-							if (rs.wasNull()) {
```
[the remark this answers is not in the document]
martin-kuba: The javadoc for getString() states that it returns Java null for NULL.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
```diff
@@ -119,299 +125,153 @@ public ExtSourceSql() {
...
+							columnMappings.add(new ColumnMapping(columnName, i, columnName.toLowerCase(), false));
+						}
+					}
+					// process each row
+					List<Map<String, String>> subjects = new ArrayList<>();
+					while (rs.next()) {
+						Map<String, String> map = new HashMap<>();
+						for (ColumnMapping columnMapping : columnMappings) {
+							if (columnMapping.blob) {
+								try (InputStream in = rs.getBinaryStream(columnMapping.columnIndex)) {
```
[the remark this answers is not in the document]
martin-kuba: Base64 uses only safe ASCII characters

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
martin-kuba: OK, this code was there before my change, replaced.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -62,67 +74,111 @@ public ExtSourcesManagerBlImpl(ExtSourcesManagerImplApi extSourcesManagerImpl) {
...
+		ExtSourceUnity.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceCSV.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceISMU.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceLdap.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceSql.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourcePerun.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceREMS.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceTCS.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceSqlComplex.setPerunBlImpl((PerunBlImpl) perunBl);
+		ExtSourceXML.setPerunBlImpl((PerunBlImpl) perunBl);
```
stavamichal, line 88: Are you sure it can't be done in more general way? To prevent anybody from forgetting of adding new ExtSource to method for initialization?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -62,67 +74,111 @@ public ExtSourcesManagerBlImpl(ExtSourcesManagerImplApi extSourcesManagerImpl) {
...
+		this.extSourcesManagerImpl.destroy();
 	}
 
 	@Override
 	public ExtSource createExtSource(PerunSession sess, ExtSource extSource, Map<String, String> attributes) throws InternalErrorException, ExtSourceExistsException {
 		getPerunBl().getAuditer().log(sess, new ExtSourceCreated(extSource));
-		return getExtSourcesManagerImpl().createExtSource(sess, extSource, attributes);
+		ExtSource extSource1 = getExtSourcesManagerImpl().createExtSource(sess, extSource, attributes);
+		extSourcesByIdMap.put(extSource1.getId(), extSource);
+		extSourcesByNameMap.put(extSource1.getName(), extSource);
```
stavamichal, line 120: Is there any reason to use old extSource instead of returned one? If yes, than at least change name of the variable to 'returnedExtSource' or 'createdExtSource'.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -62,67 +74,111 @@ public ExtSourcesManagerBlImpl(ExtSourcesManagerImplApi extSourcesManagerImpl) {
...
+		Map<String,ExtSource> extSourcesByNameMap = new HashMap<>();
+		// read all ExtSources from database
+		List<ExtSource> extSources = getExtSourcesManagerImpl().getExtSources(sess);
+		// cache them in hashmaps
+		for(ExtSource extSource : extSources) {
+			extSourcesByIdMap.put(extSource.getId(), extSource);
+			extSourcesByNameMap.put(extSource.getName(), extSource);
+		}
+		this.extSourcesByIdMap = extSourcesByIdMap;
+		this.extSourcesByNameMap = extSourcesByNameMap;
```
stavamichal, line 107: We should consider of locking these maps while changing to prevent race conditions. What do you think? At least there is a small window when extSourceByIdMap could be different from extSourceByNameMap.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -259,23 +314,15 @@ public Candidate getCandidate(PerunSession perunSession, Map<String,String> subj
...
 		if(subjectData.get("isSponsoredUser") == null) {
 			candidate.setSponsoredUser(false);
 		} else {
 			String isSponsoredUser = subjectData.get("isSponsoredUser");
-			if(isSponsoredUser.equals("true")) {
-				candidate.setSponsoredUser(true);
-			} else {
-				candidate.setSponsoredUser(false);
-			}
+			candidate.setSponsoredUser(isSponsoredUser.equals("true"));
```
stavamichal, line 325: These two changes are not part of the logic of this PR. If you insist to change them here, you need to add them to the commit message.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -570,6 +570,11 @@ public void initialize() throws InternalErrorException {
 		}
 	}
 
+	public void destroy() {
+		log.debug("destroying");
+		this.extSourcesManagerBl.destroy();
+	}
```
stavamichal, line 576: Please add at least small javadoc here with information, that this method is usually called when Perun is going down to clean all opened Resources. I know it is standard behavior, but it is little confusing to see it destroying only extSources manager.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
martin-kuba: It can, I have made all the classes to extend a new class ExtSourceImpl which has the method setPerunBl() that is called when a new instance is created.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
martin-kuba: There returned class is a subclass of ExtSource based on the specified type.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
martin-kuba: I think there cannot be a race condition. Assigning is atomic, so any code sees either the state before reloading, or after reloading.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
martin-kuba: done

stavamichal:
Please rebase this PR. Thank you.

martin-kuba:
done as #3895

PR #2679, opened 2020-04-28
CORE,RPC: Unified password manager modules logic

zlamalp:
- Added GenericPasswordManagerModule as default implementation
  for password manager logic replacing former implementation
  from inside of managePassword(), which was removed.
- Each outer API method to manage password now calls proper
  module / implementation.
- Method UsersManagerBl.getPasswordManagerModule() is now public
  and always returns module for namespace or generic implementation.
  If instantiation fails, then exception is thrown.
- Removed generateAccount() from UsersManagerImpl, it is now
  only in Bl, reusing bl method to get module.
- Removed long time deprecated createPassword() methods from all APIs.
- Removed unnecessary and incomplete javadoc from bl methods.

- Added createAlternativePassword() and deleteAlternativePassword()
  to PasswordManagerModule.
- Modified logic in UsersManagerBlImpl to use modules instead of
  shared manageAlternativePassword().
- Added generic implementation for alternative passwords inside
  GenericPasswordManagerModule.

- New Einfra/DusambaPasswordManagerModule and will pass necessary data as ENV
  variables for the new pwd manager script, avoiding STDIN.
- Both modules append their namespace after pwd manager script path, allowing custom
  implementation (unless its /bin/true). On deployment symlink or new script must be present.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -83,6 +83,7 @@ import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.bl.UsersManagerBl;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.impl.Utils;
+import cz.metacentrum.perun.core.impl.modules.pwdmgr.GenericPasswordManagerModule;
 import cz.metacentrum.perun.core.implApi.UsersManagerImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleImplApi;
 import cz.metacentrum.perun.core.implApi.modules.pwdmgr.PasswordManagerModule;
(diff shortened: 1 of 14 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -2345,8 +2036,22 @@ public int getUsersCount(PerunSession sess) throws InternalErrorException {
...
 	}
 
-	private PasswordManagerModule getPasswordManagerModule(PerunSession session, String namespace) throws InternalErrorException {
-		return getUsersManagerImpl().getPasswordManagerModule(session, namespace);
+	@Override
+	public PasswordManagerModule getPasswordManagerModule(PerunSession session, String namespace) throws InternalErrorException {
+		PasswordManagerModule module = getUsersManagerImpl().getPasswordManagerModule(session, namespace);
+		if (module == null) {
+			log.info("Password manager module for '{}' not found. Loading 'generic' password manager module instead.", namespace);
+			module = getUsersManagerImpl().getPasswordManagerModule(session, "generic");
```
stavamichal, line 2044: I would use this String as constant.

PR #2685, opened 2020-05-04
removed limit on column ext_sources_attributes.attr_value

martin-kuba:
The value of ext_sources_attributes.attr_value may hold very large SQL texts, the original limit of 4000 bytes was too small. The changes:
- in PostgreSQL used VARCHAR without limit (the documentation says the limit is about 1GB)
- in Oracle used CLOB (the documentation says limit is 4GB)
- in HSQLDB used LONGVARCHAR (the documentation says limit is 16MB)

### DIFF
--- a/perun-core/src/main/resources/oracleChangelog.txt
+++ b/perun-core/src/main/resources/oracleChangelog.txt
@@ -7,7 +7,7 @@
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
 3.1.59
-ALTER TABLE EXT_SOURCES_ATTRIBUTES ADD (tmpcolumn  CLOB);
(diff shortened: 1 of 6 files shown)


perun-core/src/main/resources/oracleChangelog.txt
```diff
@@ -6,6 +6,13 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.59
+ALTER TABLE EXT_SOURCES_ATTRIBUTES ADD (tmpcolumn  CLOB);
```
zlamalp, line 10: Two spaces between ```tmpcolumn``` and  ```CLOB```.

PR #2689, opened 2020-05-08
CORE: Fixed getAssignedServices(resource)

zlamalp:
- It previously retrieved service IDs and then asked
  for each service by ID in the BL. Now we retrieve
  list of services directly in DAO layer.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
@@ -560,9 +560,11 @@ public class ResourcesManagerImpl implements ResourcesManagerImplApi {
 	}
 
 	@Override
-	public List<Integer> getAssignedServices(PerunSession sess, Resource resource) throws InternalErrorException {
+	public List<Service> getAssignedServices(PerunSession sess, Resource resource) throws InternalErrorException {
 		try {
-			return jdbc.query("select service_id as id from resource_services where resource_id=?", Utils.ID_MAPPER, resource.getId());
+			return jdbc.query("select "+ ServicesManagerImpl.serviceMappingSelectQuery +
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
```diff
@@ -560,9 +560,11 @@ public boolean isGroupAssigned(PerunSession sess, Group group, Resource resource
 	}
 
 	@Override
-	public List<Integer> getAssignedServices(PerunSession sess, Resource resource) throws InternalErrorException {
+	public List<Service> getAssignedServices(PerunSession sess, Resource resource) throws InternalErrorException {
 		try {
-			return jdbc.query("select service_id as id from resource_services where resource_id=?", Utils.ID_MAPPER, resource.getId());
+			return jdbc.query("select "+ ServicesManagerImpl.serviceMappingSelectQuery +
+					" from services join resource_services on services.id=resource_services.service_id and resource_services.resource_id=?",
```
Vojtech-Sassmann, line 566:
```suggestion
					" from services join resource_services on services.id=resource_services.service_id where resource_services.resource_id=?",
```
I just wanted to ask, if there is a reason to use the condition in the join like you do, instead of writing WHERE?

zlamalp: Joined table can be constrained before actual joining. When you use condition in where, then whole tables are joined first and then invalid data is discarded. For our numbers of records within tables it doesn't really matter + clever sql planners can optimize it even if it`s in a where clause.

Vojtech-Sassmann: Ok, I thought it might be something like that. Thanks.

PR #2690, opened 2020-05-08
CORE: Fixed getting group members by expiration on Oracle DB

zlamalp:
- Oracle SQL query planner determine order of join/where
  conditions by itself. For some reason it prefers to
  convert all attr values to_date() instead of taking
  only values of specified attribute (by index on attr_id).
  This results in exception, since only one attribute has
  values, that can be converted to date.
- We now use Oracle specific hint (incorporated as SQL
  comment) to force its planner to use index search first,
  so that query succeeds. Note - we must reference renamed
  table (alias).
- Added debug and error log messages to expiration switching
  logic.


zlamalp:
Tested and deployed on production so tonight it should switch expired group members.

PR #2691, opened 2020-05-11
Removed lazy loading of attributes modules

Vojtech-Sassmann:
* All attribute's modules should be loaded during the initialization
phase. If the method for accessing modules failed to find some module,
it tried to load it again but that should never be possible.


stavamichal:
This would work only if you'll have chance to add new module while perun is running. I don't think we support such thing.

PR #2698, opened 2020-05-13
added command GetOwnerFacilities to java CLI

martin-kuba:
- class GetOwnerFacilities implementing command was added to perun-cli-java module
- the main pom.xml and the pom.xml for perun-cli-java were modified not to tinker with maven-shade-plugin configuration inherited from spring-boot-starter-parent, because it was causing problems when building the single module


stavamichal:
There are two changes in pom files not mentioned in commit message. Can you remove it or add a message about these changes too please?

martin-kuba:
added message about pom.xml files

PR #2702, opened 2020-05-15
CORE: Fixed condition in getAllowedFacilities(user)

zlamalp:
- This method previously returned also facilities, where user
  was not allowed (was in disabled state) since it checked
  only invalid status of processed users members.
  It also iterated over those members.
- New implementation performs direct DB query to get all
  facilities, where user is assigned with allowed members.

- Added getAllowedFacilities(member), which can be later used
  (eg. when getting all dependant user-facility attrs for
  member attribute on input).
  We generally perform too many selects on resolving allowed
  state from entities and perform their intersection later.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
@@ -466,12 +466,12 @@ public class FacilitiesManagerBlImpl implements FacilitiesManagerBl {
 	}
 
 	@Override
-	public List<Facility> getAllowedFacilities(PerunSession sess, User user) throws InternalErrorException {
+	public List<Facility> getAllowedFacilities(PerunSession sess, User user) {
 		return getFacilitiesManagerImpl().getAllowedFacilities(sess, user);
 	}
 
 	@Override
-	public List<Facility> getAllowedFacilities(PerunSession sess, Member member) throws InternalErrorException {
+	public List<Facility> getAllowedFacilities(PerunSession sess, Member member) {
 		return getFacilitiesManagerImpl().getAllowedFacilities(sess, member);
 	}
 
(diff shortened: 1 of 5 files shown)


zlamalp:
Ok, I will add tests, since I still want to debug and check correctness of those other existing SQLs for allowed state. I think they can be in some cases optimized to be quicker.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -467,16 +467,12 @@ public PerunBl getPerunBl() {
...
-			if(!getPerunBl().getMembersManagerBl().haveStatus(sess, member, Status.INVALID)) {
-				assignedFacilities.addAll(this.getAssignedFacilities(sess, member));
-			}
-		}
+		return getFacilitiesManagerImpl().getAllowedFacilities(sess, user);
+	}
 
-		return new ArrayList<>(assignedFacilities);
+	@Override
+	public List<Facility> getAllowedFacilities(PerunSession sess, Member member) throws InternalErrorException {
```
balcirakpeter, line 474: You do not have to throw InternalErrorException.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/FacilitiesManagerImpl.java
```diff
@@ -449,6 +449,44 @@ public void removeOwner(PerunSession sess, Facility facility, Owner owner) throw
...
+			return jdbc.query("select distinct " + facilityMappingSelectQuery + " from groups_resources" +
+							" join groups on groups_resources.group_id=groups.id" +
+							" join groups_members on groups.id=groups_members.group_id" +
+							" join members on groups_members.member_id=members.id " +
+							" join resources on groups_resources.resource_id=resources.id " +
+							" join facilities on resources.facility_id=facilities.id " +
+							" where members.user_id=? and members.status!=? and members.status!=?",
+					FACILITY_MAPPER, user.getId(),
+					String.valueOf(Status.INVALID.getCode()), String.valueOf(Status.DISABLED.getCode()));
+		} catch (EmptyResultDataAccessException e) {
```
balcirakpeter, line 464: This exception is probably not thrown for method jdbc.query.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
zlamalp: You are right, fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/FacilitiesManagerImpl.java
zlamalp: Again you are right, we should clean up the code everywhere. Here fixed.

PR #2705, opened 2020-05-15
Attribute modules configuration

Vojtech-Sassmann:
* Some attribute modules need to have a separate configuration file.
These configurations should be located inside the /etc/perun/modules/
folder.

* Created ModulesConfigLoader component that is used to load a module's
configurations.

* In the module of u:def:login_namespace_fenix_persistent_shadow
attribute added loading of extSourceNameFenix and domainNameFenix from
its configuration file. If the module is not used, the configuration
file does not need to exist.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesConfigLoader.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesConfigLoader.java
@@ -1,0 +1,108 @@
+package cz.metacentrum.perun.core.impl.modules;
+
+import cz.metacentrum.perun.core.api.exceptions.rt.ModulePropertyNotFoundException;
+
+import java.util.List;
+
+/**
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+public interface ModulesConfigLoader {
+
+	/**
+	 * For module with the given name, find a String configuration property with given name.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
+	 * @return found String property for given module
+	 * @throws ModulePropertyNotFoundException when the specified property value is null or is not found
+	 *                                         or the specified module configuration is not found
+	 */
+	String loadString(String moduleName, String property) throws ModulePropertyNotFoundException;
+
+	/**
+	 * For module with the given name, find a String configuration property with given name.
+	 * If the module configuration does not contain the specified property, return the default value.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
+	 * @param defaultValue default value
+	 * @return found String property for given module, or default value if not found
+	 * @throws ModulePropertyNotFoundException when the specified module configuration is not found
+	 */
+	String loadStringOrDefault(String moduleName, String property, String defaultValue)
+		throws ModulePropertyNotFoundException;
+
+	/**
+	 * For module with the given name, find an Integer configuration property with given name.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
+	 * @return found Integer property for given module
+	 * @throws ModulePropertyNotFoundException when the specified property value is null or is not found
+	 *                                         or the specified module configuration is not found
+	 */
+	Integer loadInteger(String moduleName, String property) throws ModulePropertyNotFoundException;
+
+	/**
+	 * For module with the given name, find an Integer configuration property with given name.
+	 * If the module configuration does not contain the specified property, return the default value.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
+	 * @param defaultValue default value
+	 * @return found Integer property for given module, or default value if not found
+	 * @throws ModulePropertyNotFoundException when the specified module configuration is not found
+	 */
+	Integer loadIntegerOrDefault(String moduleName, String property, Integer defaultValue)
+		throws ModulePropertyNotFoundException;
+
+	/**
+	 * For module with the given name, find a list of Strings configuration property with given name.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
+	 * @return found list of Strings property for given module
+	 * @throws ModulePropertyNotFoundException when the specified property value is null or is not found
+	 *                                         or the specified module configuration is not found
+	 */
+	List<String> loadStringList(String moduleName, String property) throws ModulePropertyNotFoundException;
+
+	/**
+	 * For module with the given name, find a list of Strings configuration property with given name.
+	 * If the module configuration does not contain the specified property, return the default value.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
+	 * @param defaultValue default value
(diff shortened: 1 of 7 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesConfigLoader.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+public interface ModulesConfigLoader {
+
+	/**
+	 * For module with the given name, find a String configuration property with given name.
+	 *
+	 * @param moduleName name of a module
+	 * @param property name of a property
```
balcirakpeter, line 14: You should mention that the property name can be chained by dots.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesYamlConfigLoader.java
```diff
@@ -0,0 +1,94 @@
...
+	private JsonNode loadPropertyNode(String moduleName, String propertyName) {
+		notNull(moduleName, "configFile");
+		notNull(propertyName, "propertyName");
+
+		String path = modulesDirPath + moduleName + ".yaml";
+
+		JsonNode root = loadModulesYamlFile(path);
+
+		JsonNode propertyNode = parsePropertyNode(root, propertyName);
+		if (propertyNode == null) {
```
balcirakpeter, line 70: I am citing Pavel Z. who commented on this behavior in the previous PR:
"I just wonder, if we couldn't use empty value as a correct input for some logic. Like value is either set or not, then some logic performs specific or generic behavior. You could fall-back to default values by type like empty list / null / false etc."

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesYamlConfigLoader.java
```diff
@@ -0,0 +1,94 @@
...
+			String[] split = propertyName.split("\\.", 2);
+			currentNode = currentNode.get(split[0]);
+			propertyName = split[1];
+		}
+		return currentNode.get(propertyName);
+	}
+
+	private static JsonNode loadModulesYamlFile(String path) {
+		try (InputStream is = new FileInputStream(path)) {
+			return yamlObjectMapper.readTree(is);
```
martin-kuba, line 88: Here I would prefer this way of leading a YAML file:
```java
YAMLMapper mapper = new YAMLMapper();
return mapper.readValue(new File(path), JsonNode.class);
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesYamlConfigLoader.java
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+
+public class ModulesYamlConfigLoader implements ModulesConfigLoader {
+
+	private final static Logger log = LoggerFactory.getLogger(ModulesYamlConfigLoader.class);
+
+	private static final ObjectMapper yamlObjectMapper = new ObjectMapper(new YAMLFactory());
+	private String modulesDirPath = "/etc/perun/modules/";
```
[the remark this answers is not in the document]
Vojtech-Sassmann: It seems that this variable can only be used inside pom file. I would have to create a separate properties file and load it from there. In other places in code, the path is also hardcoded. So I would leave it that way.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesYamlConfigLoader.java
```diff
@@ -0,0 +1,94 @@
...
+		JsonNode root = loadModulesYamlFile(path);
+
+		JsonNode propertyNode = parsePropertyNode(root, propertyName);
+		if (propertyNode == null) {
+			throw new InternalErrorException("Failed to parse property: '" + propertyName + "' from file: " + path);
+		}
+		return propertyNode;
+	}
+
+	private static JsonNode parsePropertyNode(JsonNode root, String propertyName) {
```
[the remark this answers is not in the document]
Vojtech-Sassmann: I have added javadoc to all private methods.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesConfigLoader.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesYamlConfigLoader.java
Vojtech-Sassmann: I have added new versions of the methods that allow you to specify a default value that is returned, if the found value is null or not found. I don't think it is a good idea to generally return some fallback default values. It might now be clear that it does return these default values and what they actually are.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/ModulesYamlConfigLoader.java
Vojtech-Sassmann: I have updated the implementation corresponding to your suggestion.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_fenix_persistent_shadow.java
```diff
@@ -41,7 +54,7 @@ public Attribute fillAttribute(PerunSessionImpl perunSession, User user, Attribu
 		Attribute filledAttribute = new Attribute(attribute);
 
 		if (attribute.getFriendlyName().equals(attrNameFenix)) {
-			filledAttribute.setValue(sha1HashCount(user, domainNameFenix).toString() + domainNameFenix);
+			filledAttribute.setValue(sha1HashCount(user, getDomainNameFenix()).toString() + getDomainNameFenix());
```
balcirakpeter, line 57: I've just realized that the domain name property will be stored without the "@" character. However, we need to include the "@" while setting the attribute value.
So, please could you change the code to something like this?

```
if (attribute.getFriendlyName().equals(attrNameFenix)) {
    String domain = "@" + getDomainNameFenix();
    filledAttribute.setValue(sha1HashCount(user, domain).toString() + domain);
```

Vojtech-Sassmann: I have applied your suggested changes.

PR #2707, opened 2020-05-19
CORE: Check that Member and Group are from the same VO

zlamalp:
- We could and probably should check, that Member and Group on the input
  are from the same VO. We already check group-resource and member-resource
  versions and this check was missing.
- Added check to the expected methods working with member-group attributes.
  It throws MemberGroupMismatchException. This exception should be catched
  by inner method logic, if there is no member/group in the method input and converted
  to either InternalErrorException or ConsistencyErrorException.
- This check might be missing from methods, which already check member-resource
  and group-resource, since they transitively must come from the same VO. 
- Some tests were fixed, since we used member from different VO than group.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -1631,12 +1631,13 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 	}
 
 	@Override
-	public Attribute getAttributeById(PerunSession sess, Member member, Group group, int id) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException {
+	public Attribute getAttributeById(PerunSession sess, Member member, Group group, int id) throws InternalErrorException, AttributeNotExistsException, WrongAttributeAssignmentException, MemberGroupMismatchException {
 		AttributeDefinition attributeDefinition = getAttributeDefinitionById(sess, id);
 
 		if (getAttributesManagerImpl().isFromNamespace(attributeDefinition, AttributesManager.NS_MEMBER_GROUP_ATTR)) {
 			Attribute attribute = getAttributesManagerImpl().getAttributeById(sess, member, group, id);
 			getAttributesManagerImpl().checkNamespace(sess, attribute, NS_MEMBER_GROUP_ATTR);
+			checkMemberIsFromTheSameVoLikeGroup(sess, member, group);
 			return attribute;
 		} else if (getAttributesManagerImpl().isFromNamespace(attributeDefinition, AttributesManager.NS_MEMBER_ATTR)) {
 			return getAttributesManagerImpl().getAttributeById(sess, member, id);
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -364,6 +364,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 	@Override
 	public List<Attribute> getAttributes(PerunSession sess, Group group, Member member, Resource resource, List<String> attrNames, boolean workWithUserAttributes) throws InternalErrorException, MemberResourceMismatchException, GroupResourceMismatchException, MemberGroupMismatchException {
 		checkGroupIsFromTheSameVoLikeResource(sess, group, resource);
+		checkMemberIsFromTheSameVoLikeGroup(sess, member, group);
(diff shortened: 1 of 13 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTestAbstract.java
```diff
@@ -5607,7 +5607,8 @@ public void testUniqAttributes() throws Exception {
 							break;
 						case "member_group":
 							attributesManager.setAttribute(sess, member1OfUser1, group1InVo1, a);
-							attributesManager.setAttribute(sess, member2OfUser1, group1InVo1, b);
+							attributesManager.setAttribute(sess, member2OfUser2, group2InVo1, b);
+							//attributesManager.setAttribute(sess, member2OfUser1, group1InVo1, b);
```
stavamichal, line 5611: You should remove this comment.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -356,7 +362,7 @@ public AttributesManagerBlImpl(AttributesManagerImplApi attributesManagerImpl) {
 	}
 
 	@Override
-	public List<Attribute> getAttributes(PerunSession sess, Group group, Member member, Resource resource, List<String> attrNames, boolean workWithUserAttributes) throws InternalErrorException, MemberResourceMismatchException, GroupResourceMismatchException {
+	public List<Attribute> getAttributes(PerunSession sess, Group group, Member member, Resource resource, List<String> attrNames, boolean workWithUserAttributes) throws InternalErrorException, MemberResourceMismatchException, GroupResourceMismatchException, MemberGroupMismatchException {
```
stavamichal, line 365: Why are you adding exception MemberGroupMismatchException to the methods, where is no control of it?

zlamalp: Its enforced by the calling ```getAttributes(sess, member, group)``` within the method, but you are right that we should check it, since for the case of non-empty ```attrNames``` it wouldn't be checked at all.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTestAbstract.java
zlamalp: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
zlamalp: Fixed.

zlamalp:
I'll add exceptions to the RPC javadoc soon. It seems, that we are not mentioning them much for the existing methods either.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
stavamichal: It is not just in this method. So it was more a rhetorical question. If they need to have this check just because of under-layer method, then it is ok.

zlamalp: Ok. I'll check other methods, to be sure check is not skipped like it was there.

zlamalp:
RPC javadoc added for changed methods. I added check also for ``getAttributeById()``. I checked that all methods do check member-group relation either by check method call or its checked transitively.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -3499,28 +3515,31 @@ public void checkAttributesSemantics(PerunSession sess, Member member, Resource
...
 		if (!workWithUserAttributes) {
 			getAttributesManagerImpl().checkNamespace(sess, attributes, NS_MEMBER_GROUP_ATTR);
+			this.checkMemberIsFromTheSameVoLikeGroup(sess, member, group);
 
 			for (Attribute attribute : attributes) {
 				if (attribute.getValue() == null && !isTrulyRequiredAttribute(sess, member, group, attribute)) continue;
 				getAttributesManagerImpl().checkAttributeSemantics(sess, member, group, attribute);
 			}
 		} else {
+			this.checkMemberIsFromTheSameVoLikeGroup(sess, member, group);
```
Vojtech-Sassmann, line 3542: I think you can move the check before the if.

zlamalp: Thanks for noticing, I fixed it and added lazy loading of user object for later checks.

PR #2709, opened 2020-05-20
Tool for reservation of logins for users comming from IT4I

zoraseb:
Required data are: first_name,last_name,e-mail,organization and login


zoraseb:
Patri to semjen z casti.

PR #2714, opened 2020-05-22
CORE: Optimize getting value of user:virt:tcsMails:mu

zlamalp:
- Previously we calculated value of user:virt:tcsMails:mu by
  selecting each single source attribute at the time.
- Now we retrieve values of all source attributes at once and
  then iterate over them in expected order and create resulting
  value as previously.
  This lower number of DB queries from 5 to 1.

- Fixed tests to align with new implementation.
- Fixed attribute variable/name, fixed expected order.
- Fixed idea warnings.


zlamalp:
I'll fix the tests over the weekend.

PR #2716, opened 2020-05-22
New virt attribute module for user certificates

balcirakpeter:
- Attribute module urn_perun_user_attribute_def_virt_userCertificatesLimited was created
 for MUNI purposes. They need to get only the 10 newest user certificates. So the module
 takes all user certificates and return only the 10 newest according to
 certificate expirations.
- For the purpose of this attribute and the attribute
 user:virt:userCertExpirations was created method
 retrieveCertsExpiration, which extracts expiration for all user
 certificates and return them as a map.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
@@ -27,12 +27,19 @@ import cz.metacentrum.perun.core.bl.ModulesUtilsBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.impl.Utils;
+import org.apache.commons.codec.binary.Base64;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.math.BigDecimal;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.security.cert.CertificateException;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -1164,11 +1171,48 @@ public class ModulesUtilsBlImpl implements ModulesUtilsBl {
 		return new Pair<>(number, unit);
 	}
 
+	/**
+	 * Extracts expiration of the given certificates.
+	 *
+	 * @param certificates as a map where the key is a DN and the value is a certificate
+	 * @return map where the key is certificate DN and the value is a certificate expiration
+	 */
+	public static Map<String, String> retrieveCertificatesExpiration(Map<String, String> certificates) {
+		Utils.notNull(certificates, "certificates");
+		Map<String, String> resultMap = new LinkedHashMap<>();
+		certificates.forEach((key, value) -> resultMap.put(key, getCertificateExpiration(value)));
+		return resultMap;
+	}
+
 	public PerunBl getPerunBl() {
 		return this.perunBl;
 	}
 
 	public void setPerunBl(PerunBl perunBl) {
 		this.perunBl = perunBl;
+	}
+
+	/**
+	 * Retrieve expiration of the given certificate
+	 *
+	 * @param certificate in PEM format from which the expiration will be retrieved
+	 * @return Certificate expiration as String
+	 */
+	private static String getCertificateExpiration(String certificate) {
+		Utils.notNull(certificate, "certificate");
+
+		String certWithoutBegin = certificate.replaceFirst("-----BEGIN CERTIFICATE-----", "");
+		String rawCert = certWithoutBegin.replaceFirst("-----END CERTIFICATE-----", "");
+
+		try (ByteArrayInputStream bis = new ByteArrayInputStream(Base64.decodeBase64(rawCert.getBytes()))) {
+			CertificateFactory certFact = CertificateFactory.getInstance("X.509");
+			X509Certificate cert = (X509Certificate)certFact.generateCertificate(bis);
+			DateFormat dateFormat = DateFormat.getDateInstance();
+			return dateFormat.format(cert.getNotAfter());
+		} catch (IllegalArgumentException e) {
+			throw new ConsistencyErrorException("Certificate is not in base64 format!", e);
+		} catch (IOException | CertificateException e) {
+			throw new InternalError(e);
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This attribute module is used to filter user certificates only to ten newest according to certificates expiration
+ */
```
zlamalp, line 27: You should add ``@SkipValueCheckDuringDependencyCheck``, otherwise attribute value will be retrieved during dependency checks and then no check will be preformed, since module doesn't have one.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
```diff
@@ -0,0 +1,103 @@
...
+public class urn_perun_user_attribute_def_virt_userCertificates extends UserVirtualAttributesModuleAbstract implements UserVirtualAttributesModuleImplApi {
+
+
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDefinition attributeDefinition) {
+		Attribute resultAttribute = new Attribute(attributeDefinition);
+
+		Attribute userCertsAttribute = getUserCertsAttribute(sess, user);
```
zlamalp, line 35: I know that this is clear and correct implementation, but we should minimize number of selects. Like retrieve both attributes at once. Also dependency on another virtual attributes is problematic. As a result, you twice retrieve value of ``user:def:userCertificates`` and then do some work with it. Maybe you should extract logic from ``getValue()`` of ``user:def:userCertificates`` and make it ``ModulesUtilsBlImpl`` method. Then you could use it directly in both modules to select user attributes only once and construct expiration mapping locally in this module (without dependency on virtual attribute).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
balcirakpeter: Added

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
balcirakpeter: Reworked

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
```diff
@@ -1164,11 +1171,46 @@ public User getUserFromMessage(PerunSessionImpl sess, String message) throws Int
...
+
+	/**
+	 * Retrieve expiration of the given certificate
+	 *
+	 * @param certificate in PEM format from which the expiration will be retrieved
+	 * @return Certificate expiration as String
+	 */
+	private static String getCertificateExpiration(String certificate) {
+		String certWithoutBegin = certificate.replaceFirst("-----BEGIN CERTIFICATE-----", "");
+		String rawCert = certWithoutBegin.replaceFirst("-----END CERTIFICATE-----", "");
```
stavamichal, line 1203: You should test not-null on certificate variable.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
```diff
@@ -0,0 +1,97 @@
...
+
+			List<String> tenEarliestDNs = new ArrayList<>();
+
+			certsExpirationInMilliSeconds.entrySet().stream()
+				.sorted(Collections.reverseOrder(Map.Entry.comparingByValue()))
+				.limit(10)
+				.forEachOrdered(entry -> tenEarliestDNs.add(entry.getKey()));
+
+			certs.entrySet().removeIf(entry -> !tenEarliestDNs.contains(entry.getKey()));
+			result = new ArrayList<>(certs.values());
```
stavamichal, line 63: You can call this instead:
```
.forEachOrdered(entry -> result.add(certs.get(entry.getKey())));
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This attribute module is used to filter user certificates only to ten newest according to certificates expiration
+ */
+@SkipValueCheckDuringDependencyCheck
+public class urn_perun_user_attribute_def_virt_userCertificates extends UserVirtualAttributesModuleAbstract implements UserVirtualAttributesModuleImplApi {
```
stavamichal, line 31: We should think about the name. This is too simple and it will be hard to understand that this specific certificate attribute is limited by 10 last certificates. I would prefer something like userCertificatesLimited or something similar to that. In the future, the limit itself could be attribute of facility and '10' could be the default value.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
balcirakpeter: Added

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
balcirakpeter: Reworked

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificates.java
balcirakpeter: Renamed

zlamalp: Please fix attribute name also in commit message. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userCertificatesLimited.java
```diff
@@ -0,0 +1,93 @@
...
+	public List<String> getStrongDependencies() {
+		return Collections.singletonList(AttributesManager.NS_USER_ATTR_DEF + ":userCertificates");
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_USER_ATTR_VIRT);
+		attr.setFriendlyName("userCertificates");
+		attr.setDisplayName("User certificates");
```
stavamichal, line 88: You need to change content of this method too (while changing the name).

PR #2719, opened 2020-05-26
New module for generating unique IDs for Groups

stavamichal:
- new module with implemented fillValue method to generate
 unperdictable unique ID for Group using Perun group's id
 - add test for this module

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_uniqueID.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_uniqueID.java
@@ -1,0 +1,79 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.BeansUtils;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_uniqueID.java
```diff
@@ -0,0 +1,73 @@
...
+	 * @return Builder to get string ID
+	 * @throws InternalErrorException When generation fails
+	 */
+	protected StringBuilder sha1HashCount(Group group) throws InternalErrorException {
+		try {
+			String salt = BeansUtils.getCoreConfig().getInstanceId();
+			MessageDigest mDigest = MessageDigest.getInstance("SHA1");
+			// counts sha1hash and converts output to hex
+			int length = 4+salt.getBytes(StandardCharsets.UTF_8).length;
+			byte[] result = mDigest.digest(ByteBuffer.allocate(length).putInt(group.getId()).put(salt.getBytes(StandardCharsets.UTF_8)).array());
```
Vojtech-Sassmann, line 50: NIT: This is really difficult to read. Something like this would be much easier:
```
byte[] result = mDigest.digest(ByteBuffer
				.allocate(length)
				.putInt(group.getId())
				.put(salt.getBytes(StandardCharsets.UTF_8))
				.array());
```

stavamichal: Changed.

PR #2720, opened 2020-05-27
CORE: 'tomcat9' shouldn't be allowed as login

zlamalp:
- By default we shouldn't allow 'tomcat9' as unix login for users.


zlamalp:
Ok, you are right, ``tomcat9`` is a service name.

zlamalp:
I'll handle this in a new pull-request related to einfra login/password strength checking. Closing this one.

PR #2725, opened 2020-06-02
Refactorization of persistent shadow modules

Vojtech-Sassmann:
* Persistent shadow modules contained duplicit code that could be
easily extracted and resued.
* This makes it easy to create new persistent shadow modules. Also, this
ensures that all of them works exactly the same way.
* Created abstract classes for modules that encapsulates common
behaviour.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_bbmri_test_persistent_shadow.java
```diff
@@ -23,75 +12,41 @@
  */
 
 
-public class urn_perun_user_attribute_def_def_login_namespace_bbmri_test_persistent_shadow extends urn_perun_user_attribute_def_def_login_namespace {
+public class urn_perun_user_attribute_def_def_login_namespace_bbmri_test_persistent_shadow
```
zlamalp, line 15: I think, that ``bbmri-test`` namespace is no longer required and module can be removed. I'll ask and let you know.

zlamalp: I have it confirmed, module for ``bbmri-test`` can be removed from the code.

Vojtech-Sassmann: I have removed the module.

PR #2726, opened 2020-06-03
Added new object EnrichedHost and new method getEnrichedHosts

metodej:
- added new object EnrichedHost which contains host and its attributes
- added new method to FacilitiesManager getEnrichedHosts which returns all enrichedHosts from given facility (attributes in object are filtered by rights)
- added method to get attributes selected by list of attribute names for host
- also added tests for these methods

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/EnrichedHost.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/EnrichedHost.java
@@ -1,0 +1,70 @@
+package cz.metacentrum.perun.core.api;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Host with list of all its attributes
+ * @author Metodej Klang <metodej.klang@gmail.com>
+ */
+public class EnrichedHost {
+
+	private Host host;
+	private List<Attribute> hostAttributes;
+
+	public EnrichedHost(){
+	}
+
+	public EnrichedHost(Host host, List<Attribute> hostAttributes) {
+		this.host = host;
+		this.hostAttributes = hostAttributes;
+	}
+
+	public Host getHost() {
+		return host;
+	}
+
+	public List<Attribute> getHostAttributes() {
+		return hostAttributes;
+	}
+
+	public void setHost(Host host) {
+		this.host = host;
+	}
+
+	public void setHostAttributes(List<Attribute> hostAttributes) {
+		this.hostAttributes = hostAttributes;
+	}
+
+	@Override
+	public String toString() {
+		StringBuilder str = new StringBuilder();
+
+		return str.append("EnrichedHost:[host='").append(host.toString()).append("', hostAttributes='").append(hostAttributes).append("']").toString();
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj)
+			return true;
+		if (obj == null)
+			return false;
+		if (!(obj instanceof EnrichedHost))
+			return false;
+		EnrichedHost other = (EnrichedHost) obj;
+		if (host == null) {
+			if (other.host != null)
+				return false;
+		} else if (!host.equals(other.host))
+			return false;
+		if (hostAttributes == null && other.getHostAttributes() == null)
+			return true;
+		if (hostAttributes == null
+			|| other.getHostAttributes() == null
+			|| hostAttributes.size() != other.getHostAttributes().size())
+			return false;
+		Collections.sort(hostAttributes);
+		Collections.sort(other.getHostAttributes());
+		return hostAttributes.equals(other.getHostAttributes());
+	}
+}
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/EnrichedHost.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/EnrichedHost.java
@@ -1,7 +1,7 @@
 package cz.metacentrum.perun.core.api;
 
-import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Host with list of all its attributes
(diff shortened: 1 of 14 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -491,6 +491,18 @@
 	 */
 	List<Attribute> getAttributes(PerunSession sess, Host host) throws InternalErrorException;
 
+	/**
+	 * Get all attributes associated with the host which have name in list attrNames (empty and virtual too).
+	 *
```
stavamichal, line 496: Add information to javadoc, that empty list will return no attributes. In some specific cases we have exactly different behavior (where empty means all). Same for other layers (impl, api, rpc).

perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -491,6 +491,18 @@
...
+	 * @param sess perun session
+	 * @param host host to get attributes from
```
stavamichal, line 498: 'to get attributes for' Same for other layers (impl, api, rpc).

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -660,6 +662,30 @@ public void setPerunBl(PerunBl perunBl) {
 		return getFacilitiesManagerBl().getHosts(sess, facility);
 	}
 
+	@Override
+	public List<EnrichedHost> getEnrichedHosts(PerunSession sess, Facility facility, List<String> attrNames) throws AttributeNotExistsException, FacilityNotExistsException {
+		Utils.checkPerunSession(sess);
+		getFacilitiesManagerBl().checkFacilityExists(sess, facility);
+
+		List<Host> hosts = getFacilitiesManagerBl().getHosts(sess, facility);
```
stavamichal, line 670: You are not checking rights to get hosts for facility. You need to check it there or call method from api to correctly check rights.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -660,6 +662,30 @@ public void setPerunBl(PerunBl perunBl) {
...
+		Host host1 = hosts.get(0);
```
stavamichal, line 671: This could lead to null pointer exception if no hosts is on facility.

perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
```diff
@@ -276,6 +276,19 @@
...
+	 *
+	 * @param sess perun session
+	 * @param facility facility
+	 * @param attrNames attribute names
+	 * @return list of enriched hosts
+	 *
+	 * @throws AttributeNotExistsException if some attribute does not exist
+	 * @throws FacilityNotExistsException if facility does not exist
+	 */
+	List<EnrichedHost> getEnrichedHosts(PerunSession sess, Facility facility, List<String> attrNames) throws AttributeNotExistsException, FacilityNotExistsException;
```
stavamichal, line 290: Probably missing privileged exception and testing of rights for a user.

perun-base/src/main/java/cz/metacentrum/perun/core/api/EnrichedHost.java
```diff
@@ -0,0 +1,70 @@
...
+		} else if (!host.equals(other.host))
+			return false;
+		if (hostAttributes == null && other.getHostAttributes() == null)
+			return true;
+		if (hostAttributes == null
+			|| other.getHostAttributes() == null
+			|| hostAttributes.size() != other.getHostAttributes().size())
+			return false;
+		Collections.sort(hostAttributes);
+		Collections.sort(other.getHostAttributes());
```
Vojtech-Sassmann, line 67: We have discussed this with @stavamichal, and we don't want to compare enriched hosts based on their attributes. We want to say that two enriched hosts are equal if they represent the same host. There are specific scenarios where also comparing attributes could lead to some troubles.

metodej:
Thanks for review. Everything mentioned above should be fixed now.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntryIntegrationTest.java
```diff
@@ -825,6 +826,35 @@ public void getHostsWhenFacilityNotExists()throws Exception{
...
+		attrNames.add(attr.getName());
+
+		List<Attribute> hostAttributes = new ArrayList<>();
+		hostAttributes.add(attr);
+		EnrichedHost actualEnrichedHost = new EnrichedHost(createdHost, hostAttributes);
+
+		List<EnrichedHost> expectedEnrichedHosts = facilitiesManagerEntry.getEnrichedHosts(sess, facility, attrNames);
+		EnrichedHost expectedHost = expectedEnrichedHosts.get(0);
+
+		assertEquals("Created and returned enrichedHost should be the same", expectedHost, actualEnrichedHost);
```
stavamichal, line 855: Yo changed the logic. So now you need to check if there are expected attributes manually.

metodej: Yes, it should be fixed now as well.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -1064,6 +1064,40 @@ public Object mapRow(ResultSet rs, int i) throws SQLException {
...
+			List<String> controlledAttrNames = new ArrayList<>();
+
+			for(String attributeName: attrNames) {
+				//check namespace
+				if(attributeName.startsWith(AttributesManager.NS_HOST_ATTR)) controlledAttrNames.add(attributeName);
+			}
+
+			List<Attribute> attrs = perun.getCacheManager().getAttributesByNames(controlledAttrNames, new Holder(host.getId(), Holder.HolderType.HOST), null);
+			return this.setValuesOfAttributes(sess, attrs, host, null);
+		}
```
Vojtech-Sassmann, line 1079: We have recently agreed on removing the cache completely. So please, remove this code as well.

metodej: Ok, removed.

PR #2731, opened 2020-06-07
requestPreferredEmailChange method added to OpenAPI

xkostka2:


perun-openapi/openapi.yml
```diff
@@ -3639,6 +3639,21 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/attributesManager/getAttributes/user-facility:
```
Vojtech-Sassmann, line 3642: This method is already in the specification aa `/json/attributesManager/getAttributes/f-u`

PR #2732, opened 2020-06-08
Removed throws declarations of unchecked exceptions

Vojtech-Sassmann:
* Since we have changed the InternalErrorException to unchecked
exception, we can remove all of its throws declarations.
* I run the IDEAs cleanup tool to find all throws declarations of
unchecked exceptions.

### DIFF
--- a/perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
+++ b/perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
@@ -29,7 +29,7 @@ public class AuditParser {
 	static final Pattern pointyAndSquareBracket = Pattern.compile(".\\[|.\\]|[^\\\\](\\\\\\\\)*(<|>)");
 
 
-	public static List<PerunBean> parseLog(String log) throws InternalErrorException {
+	public static List<PerunBean> parseLog(String log) {
 		List<PerunBean> listPerunBeans = new ArrayList<PerunBean>();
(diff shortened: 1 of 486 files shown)


Vojtech-Sassmann:
Hi @zlamalp, I have done this only to clean up the code from the unnecessary declarations. Because right now, when we create new methods (often by copy-paste), these unnecessary declarations are copied as well

>  And by "just removing" it from all the declarations, we will lose necessity to react properly on some application states - aka - we will have to remember on any change in the code to check, if we have to catch and interpret InternalErrorException or not.

I don't entirely agree with this. The purpose of the InternalErrorException (as from my understanding) is to be thrown when some unexpected problem occurred. This kind of error might happen anytime and anywhere (it might not even be the InternallErrorException), and you don't want to handle all possible runtime errors of each method call. 

The only time I see you would like to do that is something like processing a list of members and not wanting to finish if the processing of one member unexpectedly fails. And in that case, you usually do something like `catch (Exception)` anyway.

> Technically, I would rather see cleanup from the impl by each manager and checking, if something better can't be thrown instead of InternalErrorException, since it was sometimes quicker to use generic exception instead of creating specific one.

This makes sense, but I don't think it is relevant to this PR. What you could then actually do is to go through all the places where this Exception is instantiated.

> And it probably breaks all the other 27 pull-requests :-( 

I agree this is not really nice. But the kind of rebasing that would have to be done, I think, is a pretty simple one. We might do this by some parts, but that way, it would be in some inconsistent state.

PR #2737, opened 2020-06-11
CORE,GUI: Login / password checks for einfra

zlamalp:
CORE
- Logic of attribute checks for login syntax was moved into
  tha password manager modules.
- Helping methods in ModulesUtilsBl were updated to work just with
  login and namespace (without Attribute object).
- New configuration option "login-namespace:[namespace]:allowedExceptions"
  in /etc/perun/perun-namespaces.properties can be used to allow
  otherwise invalid logins (syntactically or unpermitted).
  This is necessary to enforce correct checks on new logins, but
  leave out currently existing invalid logins.
- Added checked InvalidLoginException, which is thrown if login
  is invalid by syntax or is not permitted.
- isLoginAvailable() now also checks if login is syntactically
  correct and allowed within namespace before returning info about
  availability. This allows us to have input check in current
  forms.
- Updated generic user:def:login-namespace attribute module to wrap
  new logic and throw WrongAttributeValue as expected.

- Implemented password strength check in password manager modules.
- GenericPasswordManagerModule checks only emptiness, to conform
  previous behavior.
  EinfraPasswordManagerModule checks various password properties
  and its logic will be extended later. It is also applied to
  SambaduPasswordManagerModule.

- Various generic methods working with login/password now throws those
  two new exceptions, and usually also methods using them.
  Core and RPC javadoc for such methods was updated.

GUI
- Check for login availability can handle InvalidLoginException.
- Show einfra specific help and perform checks on password.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
@@ -10,6 +10,7 @@ import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidLoginException;
 import cz.metacentrum.perun.core.api.exceptions.QuotaNotInAllowedLimitException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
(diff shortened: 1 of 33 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
```diff
@@ -236,17 +237,50 @@
...
+	 * Eg. when syntax check normally doesn't allow such value, but its manually allowed here
+	 * in order to support already existing (historic) wrong values.
+	 * It can be defined for each namespace in /etc/perun/perun-namespaces.properties
+	 *
+	 * @param namespace Namespace to perform check in
+	 * @param login Login to check
+	 * @return TRUE if login value is within exceptions / FALSE otherwise
 	 */
-	void checkUnpermittedUserLogins(Attribute loginAttribute) throws WrongAttributeValueException;
+	boolean isLoginException(String namespace, String login);
```
stavamichal, line 283: The name of this method is misleading. It sounds like the login could be a java Exception. I would change that for something like 'isLoginExceptionallyAllowed' or something like that.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
```diff
@@ -539,20 +540,66 @@ public void checkReservedUnixGroupNames(Attribute groupNameAttribute) throws Wro
...
+			// check syntax or if its between exceptions
+			if (!defaultRegex.matcher(login).matches() && !isLoginException(namespace, login)) {
+				log.warn("Login '{}' in {} namespace doesn't match regex: {}", login, namespace, regex);
+				throw new InvalidLoginException("Login doesn't matches expected regex: \"" + defaultRegex +"\"");
+			}
+		}
+	}
+
+	@Override
+	public boolean checkIfUserLoginIsPermitted(String namespace, String login) {
```
stavamichal, line 574: I was thinking about this method and also there would be better name something like "isUserLoginPermitted", because it returns true or false. Check is more like OK or Exception in our code.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1292,8 +1294,8 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
 	public RichMember createSponsoredMember(PerunSession session, Vo vo, String namespace, Map<String, String> name, String password, User sponsor)
 			throws InternalErrorException, PrivilegeException, AlreadyMemberException,
 			LoginNotExistsException, PasswordCreationFailedException,
-		ExtendMembershipException,
-			WrongAttributeValueException, ExtSourceNotExistsException, WrongReferenceAttributeValueException, UserNotInRoleException {
+			ExtendMembershipException,
```
stavamichal, line 1297: Why is this exception alone on the line?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
```diff
@@ -19,10 +20,11 @@
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
- * Password manager for EINFRA login-namespace. It performs custom checks
- * on password strength and login format.
+ * Password manager for EINFRA login-namespace. It provides custom checks on login format
+ * adn password strength. Also implementation for alternative passwords is customized.
```
stavamichal, line 27: Typo 'adn'.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/GenericPasswordManagerModule.java
```diff
@@ -55,64 +64,65 @@ public void setActualLoginNamespace(String actualLoginNamespace) {
...
-	public void reserveRandomPassword(PerunSession session, String userLogin) throws InternalErrorException {
+	public void reserveRandomPassword(PerunSession sess, String userLogin) throws InvalidLoginException {
+		checkLoginFormat(sess, userLogin);
 		Process process = createPwdManagerProcess(PASSWORD_RESERVE_RANDOM, actualLoginNamespace, userLogin);
 		handleExit(process, actualLoginNamespace, userLogin);
 	}
 
 	@Override
 	public void checkPassword(PerunSession sess, String userLogin, String password) {
+		// use custom check instead of strength, since this is about empty input
```
stavamichal, line 90: I am not sure what does this comment mean.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
zlamalp: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
zlamalp: I aligned all exceptions in three rows.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ModulesUtilsBl.java
zlamalp: Renamed to suggested name. I have no better.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
zlamalp: Renamed, makes sense.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/GenericPasswordManagerModule.java
zlamalp: I rephrase that. Basically we must not check password strength here, since we check "old" possibly weaker password. We just must make sure, that input is not empty.

PR #2746, opened 2020-06-14
Updated methods in FacilitiesManagerMethod to work with facilityName

metodej:
- methods in rpc FacilitiesManagerMethod were extended to work with facility name where previously only facility id was allowed
- addOwner, removeOwner were also extended to work with owner name and not just owner id
- removeHostByHostname added to rpc to simulate removeHost with hostname as parameter
- methods also added to openapi
- new method removeHostByHostname removes host by given hostname if there is unique host to be removed after privilege check, otherwise throws exception
- getOwnerByName was added to all layers


perun-core/src/main/java/cz/metacentrum/perun/core/api/OwnersManager.java
```diff
@@ -71,6 +71,20 @@
...
+	 * @param perunSession perun session
+	 * @param name name of the owner
+	 *
+	 * @return Owner with specified name
+	 *
+	 * @throws OwnerNotExistsException if owner with given name does not exist
+	 * @throws InternalErrorException
+	 * @throws PrivilegeException if user does not have sufficient permissions
+	 */
+	Owner getOwnerByName(PerunSession perunSession, String name) throws OwnerNotExistsException, InternalErrorException, PrivilegeException;
```
zlamalp, line 86: We don't want to mention ``InternalErrorException`` in method definition and javadoc anymore, since its runtime.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -565,6 +746,21 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Remove host from the Facility based on hostname. If there is ambiguity, method throws exception and no host is removed.
```
zlamalp, line 750: Maybe you should mention that exception:
```
@throw HostNotExistsException When host doesn't exist or is not unique by name.
```

metodej:
Thank you for review, I've updated javadoc and method definitions as mentioned.

PR #2754, opened 2020-06-17
Implementation of independent propagation task scheduling

mvocu:
Implementation of independent scheduling of propagations:
  * adds "schedule_expr" column to tasks table
  * value of this column can be:
     - "rate=LONG" - propagations will be scheduled every LONG seconds
     - "cron=EXPR" - propagations will be scheduled according to the EXPR cron expression

### DIFF
--- a/perun-base/src/test/resources/test-schema.sql
+++ b/perun-base/src/test/resources/test-schema.sql
@@ -1156,6 +1156,7 @@ create table tasks (
 	start_time timestamp,
 	end_time timestamp,
 	engine_id integer,
+	schedule_expr varchar(256),
 	created_at timestamp default current_date not null,
 	err_message varchar(4000),
 	created_by_uid integer,
--- a/perun-base/src/test/resources/test-schema.sql
+++ b/perun-base/src/test/resources/test-schema.sql
@@ -1753,7 +1753,7 @@ CREATE INDEX ufauv_idx ON user_facility_attr_u_values (user_id, facility_id, att
 CREATE INDEX vauv_idx ON vo_attr_u_values (vo_id, attr_id) ;
 
 -- set initial Perun DB version
-insert into configurations values ('DATABASE VERSION','3.1.59');
+insert into configurations values ('DATABASE VERSION','3.1.60');
 insert into membership_types (id, membership_type, description) values (1, 'DIRECT', 'Member is directly added into group');
 insert into membership_types (id, membership_type, description) values (2, 'INDIRECT', 'Member is added indirectly through UNION relation');
 insert into action_types (id, action_type, description) values (nextval('action_types_seq'), 'read', 'Can read value.');
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/TasksManagerImplApi.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/TasksManagerImplApi.java
@@ -116,4 +116,6 @@ public interface TasksManagerImplApi {
 	 */
 	List<TaskResult> getTaskResultsForDestinations(List<String> destinationsNames) throws InternalErrorException;
 
+	List<Task> listAllSchedulableTasks();
+
 }
--- a/perun-db/postgres.sql
+++ b/perun-db/postgres.sql
@@ -1153,7 +1153,7 @@ create table tasks (
 	start_time timestamp,                    --real start time of task
 	end_time timestamp,                      --real end time of task
 	engine_id integer, --identifier of engine which executing the task (engines.id)
-    schedule_expr varchar(256), -- cron scheduling expr
+	schedule_expr varchar(256), -- cron scheduling expr
 	created_at timestamp default statement_timestamp() not null,
 	err_message varchar(4000),          --return message in case of error
 	created_by_uid integer,
--- a/perun-db/postgres.sql
+++ b/perun-db/postgres.sql
@@ -1,4 +1,4 @@
--- database version 3.1.59 (don't forget to update insert statement at the end of file)
+-- database version 3.1.60 (don't forget to update insert statement at the end of file)
 
 -- VOS - virtual organizations
 create table vos (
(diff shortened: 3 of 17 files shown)


perun-base/src/test/resources/test-schema.sql
```diff
@@ -1156,6 +1156,7 @@ create table tasks (
 	start_time timestamp,
 	end_time timestamp,
 	engine_id integer,
+	schedule_expr varchar(256),
```
zlamalp, line 1159: DB version needs to be updated also at the end of this file.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/TasksManagerImplApi.java
```diff
@@ -116,4 +116,6 @@
 	 */
 	List<TaskResult> getTaskResultsForDestinations(List<String> destinationsNames) throws InternalErrorException;
 
+	List<Task> listAllSchedulableTasks();
```
zlamalp, line 119: There is missing implementation for this method in ``TasksManagerImpl``.

perun-db/postgres.sql
```diff
@@ -1153,6 +1153,7 @@ create table tasks (
 	start_time timestamp,                    --real start time of task
 	end_time timestamp,                      --real end time of task
 	engine_id integer, --identifier of engine which executing the task (engines.id)
+    schedule_expr varchar(256), -- cron scheduling expr
```
zlamalp, line 1156: alignment

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
```diff
@@ -82,8 +91,62 @@ public void setDispatcherProperties(Properties dispatcherProperties) {
...
+				schedulingPool.scheduleTask(task, -1);
+			} else {
+				log.info("[{}] Task in state {}, skipping regular schedule.", task.getStatus());
+			}
+		}
+			
+	}
+	
+	/**
+	 *
```
zlamalp, line 125: Either add javadoc or remove this.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
```diff
@@ -82,8 +91,62 @@ public void setDispatcherProperties(Properties dispatcherProperties) {
...
+		private Task task;
+	
+		public TaskExecution(Task task) {
+			super();
+			this.task = task;
+		}
+		
+		@Override
+		public void run() {
+			if(task.getStatus().equals(TaskStatus.DONE)) {
```
zlamalp, line 114: You should probably schedule all Tasks in other ending states too: ``TaskStatus.ERROR, TaskStatus.GENERROR, TaskStatus.SENDERROR``. Other states are considered "processing".

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
```diff
@@ -107,6 +170,8 @@ public void run() {
 			return;
 		}
 
+		this.createIndependentTaskSchedules();
```
zlamalp, line 173: You should make sure, that repetitively scheduled tasks are not re-sheduled in below methods. I think, that all Tasks are in the shared pool and other threads work with them, so you should eg. skip Tasks with ``scheduleExpr != NULL`` in their processing.

perun-base/src/test/resources/test-schema.sql
mvocu: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/TasksManagerImplApi.java
mvocu: Fixed.

perun-db/postgres.sql
mvocu: fixed

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
mvocu: Added javadoc.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
mvocu: I disagree with that. Tasks that ended in one of the error states should be rescheduled according to the default rules for failed tasks - ie. with task delay and recurrence. Only after succesfull completion they should be subject to the periodic scheduling again.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
mvocu: I disagree with that too. I tend to think that tasks should be scheduled at least once in two days (as a safety in case the periodic scheduling fails for whatever reason), should be run when new data appears, and as above, should be subject to the same reschedule rules in case of failure.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
zlamalp: I'm not saying, that its wrong in the principle, but it means, that if Task fails, It will perform just 2 more tries in cca 10 minutes (our default) and then it will wait 2 days for the next try, since these Tasks shouldn't be triggered by source data changes, but the cron only. 
If there is some kind of temporary fail and cron is set to "every hour", it won't try to start the propagation for the 2 whole days and I would expect that "try" to be every hour, just like the cron is set. I'll think about it more and discuss it within the core team.

mvocu: Then what about scheduling error tasks by cron only after they have exhausted the retries?

zlamalp:
We decided, that this feature is no longer needed, as we can schedule/force service propagation using outside cron and we have single requestor for this feature. Still its interesting feature, so I marked the pull-request with graveyard tag.

PR #2761, opened 2020-06-24
Group structure sync - additional attributes

Vojtech-Sassmann:
* Implemented loading of additional attributes data returned from the
query.
* Now, from the extsource, it is possible to load additional attribute
data, if the columns have a correct name.
* Loaded data will be set as group def attributes, other attributes
cannot be set right now, although it is possible to load them from the
extSourceSql.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
@@ -5,7 +5,6 @@ import cz.metacentrum.perun.core.api.GroupsManager;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.SubjectNotExistsException;
-import cz.metacentrum.perun.core.blImpl.GroupsManagerBlImpl;
 import cz.metacentrum.perun.core.blImpl.PerunBlImpl;
 import cz.metacentrum.perun.core.implApi.ExtSourceSimpleApi;
 import org.apache.commons.codec.binary.Base64;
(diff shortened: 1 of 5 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupStructureSynchronizationIntegrationTest.java
```diff
@@ -447,6 +454,104 @@ public void removeAllGroupsTest() throws Exception {
 		assertTrue("List of sub groups of base group should be empty!", subGroups.isEmpty());
 	}
 
+	@Test
+	public void additionalStringAttributeIsSet() throws Exception {
+        System.out.println(CLASS_NAME + "additionalStringAttributeIsSet");
+
+		AttributeDefinition loginAttrDef = setGroupAttribute(ADDITIONAL_STRING);
```
stavamichal, line 461: You probably want to change name of this variable. It is a little confusing now.

Vojtech-Sassmann: Yes, I do. I have changed it.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ExtSourceSql.java
```diff
@@ -358,6 +323,94 @@ public void close() {
...
+	}
+
+	/**
+	 * Add additional data from the given result set to the given map.
+	 *
+	 * @param map map with attributeName-value entries
+	 * @param rs result set with attribute data
+	 * @throws SQLException SQLException
+	 */
+	private void addAdditionalAttributeData(Map<String, String> map, ResultSet rs) throws SQLException {
```
balcirakpeter, line 354: I would prefer to return the new resulting Map instead of just changing the input one.

Vojtech-Sassmann: I have changed it so it returns the data.

PR #2764, opened 2020-06-25
Set login as argument of base group members query in structure sync

stavamichal:
- replace short name of group by its login in query for sub-groups of
 base group while setting members query attribute
 - the main identification of group is now its login, not name

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -1724,7 +1724,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		updateExistingGroupsWhileSynchronization(sess, baseGroup, groupsToUpdate, loginAttributeDefinition, skippedGroups);
 		removeFormerGroupsWhileSynchronization(sess, baseGroup, groupsToRemove, skippedGroups);
 
-		setUpSynchronizationAttributesForAllSubGroups(sess, baseGroup, source);
+		setUpSynchronizationAttributesForAllSubGroups(sess, baseGroup, source, loginAttributeDefinition);
 
 		log.info("Group structure synchronization {}: ended.", baseGroup);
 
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3627,14 +3627,15 @@ private void saveInformationAboutGroupStructureSynchronization(PerunSession sess
 	 *
 	 * @param sess
 	 * @param baseGroup from which are sub groups taken
-	 * @param source from which members are synchronized
+	 * @param source from which members are synchronized\
```
balcirakpeter, line 3630: Nit: typo

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3662,7 +3663,10 @@ private void setUpSynchronizationAttributesForAllSubGroups(PerunSession sess, Gr
 				log.info("ExtSource already assigned to group: {}", group);
 			}
 
-			membersQueryAttribute.setValue(baseMembersQuery.getValue().toString().replace("?", group.getShortName()));
+			//we want to set login of group
+			Attribute loginAttribute = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, loginAttributeDefinition.getName());
+			if(loginAttribute.getValue() == null) throw new InternalErrorException("For purpose of setting attributes for group " + group + " we need to have not empty group login " + loginAttribute);
```
balcirakpeter, line 3668: Nit: I think you can write the message like this "...for " + group + " ...", because if you leave it like it is now, the final message will be with duplicate "group" word.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: Removed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
stavamichal: Changed.

PR #2765, opened 2020-06-25
removed artificial limits on varchar lengths

martin-kuba:
In PostgreSQL and HSQL, limiting the length of a varchar database column is useless, it does not save any space and adds a small overhead.

- replaced all "varchar(N)" with "varchar" in postgres.sql
- added corresponding ALTER TABLE ALTER COLUMN lines into postgresChangelog.txt
- replaced all "vachar(N)" with "longvarchar" in test-schema.sql (longvarchar is a synonym for varchar but it does not need the limit)
- removed useless lines from hsqldbChangelog.txt and postgresChangelog.txt


perun-db/postgres.sql
```diff
@@ -84,16 +84,16 @@ create table cabinet_publications (
...
+	isbn varchar,
 	categoryId integer not null, --identifier of category (cabinet_categories.id)
-	createdBy varchar(1300) default user not null,
+	createdBy varchar default user not null,
 	createdDate timestamp not null,
 	rank numeric (38,1) default 0 not null,
-	doi varchar(256),
-	locked varchar(1) default 0 not null,
+	doi varchar,
+	locked char(1) default 0 not null,
```
zlamalp, line 96: Change ``varchar(1)`` to ``char(1)`` is missing in postgresql changelog. Also we should probably default to ``'0'`` (character not int) as other char columns do. It can be later updated to boolean, but I would like to have converts from char to boolean as a separate pull-request.

perun-db/postgres.sql
```diff
@@ -563,9 +563,9 @@ create table application_data (
 create table application_mails (
 	id integer not null,
 	form_id integer not null,       --identifier of form (application_form.id)
-	app_type varchar(30) not null,  --application type (initial/extension)
-	mail_type varchar(30) not null, --type of mail (user/administrator)
-	send varchar(1) not null,       --sent (Y/N)
+	app_type varchar not null,  --application type (initial/extension)
+	mail_type varchar not null, --type of mail (user/administrator)
+	send char(1) not null,       --sent (Y/N)
```
zlamalp, line 568: Again this change is missing from postgresql changelog.

perun-base/src/test/resources/test-schema.sql
```diff
@@ -567,9 +567,9 @@ create table application_data (
 create table application_mails (
 	id integer not null,
 	form_id integer not null,       --identifier of form (application_form.id)
-	app_type varchar(30) not null,  --application type (initial/extension)
-	mail_type varchar(30) not null, --type of mail (user/administrator)
-	send varchar(1) not null,       --sent (Y/N)
+	app_type longvarchar not null,  --application type (initial/extension)
+	mail_type longvarchar not null, --type of mail (user/administrator)
+	send longvarchar not null,       --sent (Y/N)
```
zlamalp, line 572: Should be probably char(1) as in postgres.

perun-base/src/test/resources/test-schema.sql
```diff
@@ -88,16 +88,16 @@ create table cabinet_publications (
...
+	isbn longvarchar,
 	categoryId integer not null, --identifier of category (cabinet_categories.id)
-	createdBy varchar(1300) default user not null,
+	createdBy longvarchar default user not null,
 	createdDate timestamp not null,
 	rank numeric (38,1) default 0 not null,
-	doi varchar(256),
-	locked varchar(1) default 0 not null,
+	doi longvarchar,
+	locked longvarchar default 0 not null,
```
zlamalp, line 100: Should be probably char(1) and default to '0' as in postgres.

perun-db/postgres.sql
martin-kuba: I omitted it on purpose, I think that we do not need to migrate this column in the existing databases, it will be changed to boolean anyway.

perun-db/postgres.sql
martin-kuba: The same, no need to migrate it in existing databases.

perun-base/src/test/resources/test-schema.sql
martin-kuba: fixed

perun-base/src/test/resources/test-schema.sql
martin-kuba: fixed

perun-db/postgres.sql
zlamalp: ok

PR #2773, opened 2020-07-01
CORE: Move generating random password into PasswordManagerModule

zlamalp:
- PasswordManagerModule can now generate own random password.
  It does it from defined chars in expected length and performs
  check on strength. Only passwords passing that check are returned.
  Null is returned on fail.
- Added module for 'vsup' namespace, which have specific requirements
  for random password generation. Eg. omit chars that user can mistake
  on keyboard: l vs I, yY<->zZ, special chars.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -88,12 +88,10 @@ import cz.metacentrum.perun.core.impl.modules.pwdmgr.GenericPasswordManagerModul
 import cz.metacentrum.perun.core.implApi.UsersManagerImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleImplApi;
 import cz.metacentrum.perun.core.implApi.modules.pwdmgr.PasswordManagerModule;
-import org.apache.commons.lang3.RandomStringUtils;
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -2064,43 +2061,38 @@ public void removeAllUserExtSources(PerunSession sess, User user) {
...
+		String userLogin;
 		try {
-			// FIXME - we want to re-generate password if it was considered weak
-			changePassword(session, user, loginNamespace, null, newRandomPassword, false);
-		} catch (PasswordDoesntMatchException | PasswordStrengthFailedException e) {
-			// should not happen when we are not using the old password
+			Attribute userLoginAttribute = getPerunBl().getAttributesManagerBl().getAttribute(session, user, AttributesManager.NS_USER_ATTR_DEF + ":login-namespace:" + namespace);
+			userLogin = (String) userLoginAttribute.getValue();
+		} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+			// should not happen since the changePassword method passed
```
Vojtech-Sassmann, line 2072: I can't see the changePassword method call.

zlamalp: It was moved to the line 2087, since we now first check if user has login and then perform other logic.

PR #2774, opened 2020-07-01
Modification of group structure synchronization

stavamichal:
- order of operations was change from (add,update,remove) to
 (remove,add,update) to be able to change login of groups. Without it we
 can't create two groups with same name but different login
 - to be able to correctly work with group structure after changing of
 operations order, filtering of non-existent parent groups must be done
 in updating
 - for better efficiency the removing is starting from leaves if
 possible (groups for removing were sorted by names from longest to
 shortest)
 - login prefix for one group structure tree can be now set on base
 group and all newly synchronized groups will be updated to have this
 prefix. Prefix is optional and extSource don't need to prepare any data
 for using it
 - test of newly added prefixes was created

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -138,6 +138,8 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import static java.util.Collections.reverseOrder;
+import static java.util.Comparator.comparingInt;
 
 import static cz.metacentrum.perun.core.impl.PerunLocksUtils.lockGroupMembership;
 
@@ -1710,6 +1712,8 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 		//get login attribute for structure
 		AttributeDefinition loginAttributeDefinition = getLoginAttributeForGroupStructure(sess, baseGroup);
+		//get login prefix if exists
+		String loginPrefix = getLoginPrefixForGroupStructure(sess, baseGroup);
 
 		List<CandidateGroup> candidateGroupsToAdd = new ArrayList<>();
 		Map<CandidateGroup, Group> groupsToUpdate = new HashMap<>();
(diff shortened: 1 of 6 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupStructureSynchronizationIntegrationTest.java
```diff
@@ -350,6 +347,51 @@ public void removeIntermediaryGroupTest() throws Exception {
 		assertTrue("Leaf group should not have any children!", 0 == groupsManagerBl.getSubGroupsCount(sess, subBaseGroupChild));
 	}
 
+	@Test
+	public void replaceStructureWithPrefix() throws Exception {
+		System.out.println(CLASS_NAME + "removeIntermediaryGroupTestWithPrefix");
```
Vojtech-Sassmann, line 352:
```suggestion
		System.out.println(CLASS_NAME + "replaceStructureWithPrefix");
```

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupStructureSynchronizationIntegrationTest.java
```diff
@@ -350,6 +347,51 @@ public void removeIntermediaryGroupTest() throws Exception {
...
+
+		String loginPrefix = "prefix-";
+		setLoginPrefixForStructure(baseGroup, loginPrefix);
+
+		final Group subBaseGroup = new Group("baseSubGroup", "child of base group");
+		groupsManagerBl.createGroup(sess, baseGroup, subBaseGroup);
+		setLoginToGroup(baseGroup, subBaseGroup, "baseSubGroup");
+
+		final Group interGroup = new Group("interGroup", "child of sub base group");
```
Vojtech-Sassmann, line 361: `interGroup` I think you don't want to burry the group in the ground :wink:

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4217,7 +4251,8 @@ private void addMissingGroupsWhileSynchronization(PerunSession sess, Group baseG
 			}
 		}
 		// update newly added groups cause the hierarchy could be incorrect
-		updateExistingGroupsWhileSynchronization(sess, baseGroup, groupsToUpdate, loginAttributeDefinition, skippedGroups);
+		//no need to send list of removed parent groups here, because it is no need to resolve it for new groups at all
+		updateExistingGroupsWhileSynchronization(sess, baseGroup, groupsToUpdate, new ArrayList<>(), loginAttributeDefinition, skippedGroups);
```
balcirakpeter, line 4255: Nit: you can use Collections.emptyList() instead of new ArrayList().

PR #2775, opened 2020-07-02
New method forceAllSubGroupsSynchronization

stavamichal:
- to be able to force synchronization of all subgroups (whole tree) of
 any base group, we need to add new method for it
 - in forcing group process we need to also check if group is enabled to
 be synchronized, if not, new exception should be thrown. For this
 reason GroupSynchronizationNotEnabledException was created
 - add new method also to rpc and openapi

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -78,6 +78,7 @@ import cz.metacentrum.perun.core.api.exceptions.GroupRelationNotAllowed;
 import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.GroupStructureSynchronizationAlreadyRunningException;
 import cz.metacentrum.perun.core.api.exceptions.GroupSynchronizationAlreadyRunningException;
+import cz.metacentrum.perun.core.api.exceptions.GroupSynchronizationNotEnabledException;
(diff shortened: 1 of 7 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -1739,20 +1740,27 @@ protected boolean hasCandidateExistingMember(Candidate candidate, RichMember ric
...
-	public void forceGroupSynchronization(PerunSession sess, Group group) throws GroupSynchronizationAlreadyRunningException {
-		//Check if the group is not currently in synchronization process
-		if(poolOfSynchronizations.putGroupToPoolOfWaitingGroups(group, true)) {
-			log.debug("Scheduling synchronization for the group {} by force!", group);
+	public void forceGroupSynchronization(PerunSession sess, Group group) throws GroupSynchronizationAlreadyRunningException, GroupSynchronizationNotEnabledException {
+		//Check if the group should be synchronized (attribute synchronizationEnabled is set to 'true')
+		Boolean syncEnabled = false;
+		try {
+			Attribute syncEnabledAttr = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUPSYNCHROENABLED_ATTRNAME);
+			if(syncEnabledAttr.getValue() != null) syncEnabled = syncEnabledAttr.valueAsBoolean();
```
balcirakpeter, line 1749: This attribute is not boolean but String so I do not think you can cast it to boolean like that.

stavamichal: You are right, fixed.

PR #2786, opened 2020-07-14
Attributes cache removed

Vojtech-Sassmann:
* The implementation of the attributes cache was not working very well.
The Perun was actually slower with the cache enabled. We are not able to
improve it so it can be removed.
* Removed all classes related to the attributes cache and all of theirs
usages.


Vojtech-Sassmann:
Hi @zlamalp, @balcirakpeter. I forgot to remove some of the parts of the cache. I have removed them in a new commit, so you can easily see it. Please, also review these changes.

PR #2788, opened 2020-07-14
New method in Core - get policies

Vojtech-Sassmann:
* Created a new method which can be used to retrieve all of the loaded
perun policies in the AuthzResolver. This method will be used in the
NGUI for authorization.
* The new method `getPolicies` has been added to the openApi
specification as well.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
@@ -1,10 +1,12 @@
 package cz.metacentrum.perun.rpc.methods;
 
+import cz.metacentrum.perun.core.api.AuthzResolver;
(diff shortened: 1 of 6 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -613,6 +615,18 @@ public String call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 
+	/*#
+	 * Return all loaded perun policies.
+	 *
+	 * @return List<PerunPolicy> all loaded policies
+	 */
+	getPolicies {
+		@Override
+		public List<PerunPolicy> call(ApiCaller ac, Deserializer parms) throws PerunException {
+			return AuthzResolver.getAllPolicies();
```
balcirakpeter, line 626: Please unify the names of these methods.

PR #2789, opened 2020-07-15
Core - secondary regex for group name

Vojtech-Sassmann:
* On some instances, we need to be able to restrict the group names
even more. For this purpose, I have added an implementation of
validation against a secondary regexes. If the secondary regex is not
specified, it is skipped during the validation.
* Secondary regex for short name can be specified with the
`perun.group.nameSecondaryRegex` core property.
* Secondary regex for full name can be specified with the
`perun.group.fullNameSecondaryRegex` core property.
* One should define both, or none of this properties to ensure a correct
behaviour of the whole system.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
@@ -23,6 +23,7 @@ import cz.metacentrum.perun.core.api.exceptions.GroupExistsException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.GroupRelationDoesNotExist;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidGroupNameException;
 import cz.metacentrum.perun.core.api.exceptions.NotGroupMemberException;
 import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
 import cz.metacentrum.perun.core.blImpl.AuthzResolverBlImpl;
@@ -122,9 +123,10 @@ public class GroupsManagerImpl implements GroupsManagerImplApi {
 			}
 		}
 
-		// Check the group name, it can contain only a-Z0-9_- and space
-		if (!group.getShortName().matches("^[- a-zA-Z.0-9_]+$")) {
-			throw new InternalErrorException(new IllegalArgumentException("Wrong group name, group name can contain only a-Z0-9.-_: and space characters. " + group));
+		try {
+			Utils.validateGroupName(group.getShortName());
+		} catch (InvalidGroupNameException e) {
+			throw new InternalErrorException(e);
 		}
 
 		try {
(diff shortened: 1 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/GroupsManagerBl.java
```diff
@@ -45,6 +45,7 @@
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidGroupNameException;
```
zlamalp, line 48: Not sure why added if not used.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/EduGain.java
```diff
@@ -34,7 +35,12 @@ public Application approveApplication(PerunSession session, Application app) thr
 
 			AuthzResolver.setRole(session, user, vo, Role.TOPGROUPCREATOR);
 
-			Group membersGroup = session.getPerun().getGroupsManager().getGroupByName(session, vo, "members");
+			Group membersGroup = null;
+			try {
+				membersGroup = session.getPerun().getGroupsManager().getGroupByName(session, vo, "members");
+			} catch (InvalidGroupNameException e) {
+				// shouldn't happen, "members" is a valid name
```
zlamalp, line 42: You should probably fail with ``InternalErrorException`` here, since module wants to add you as an admin of the members group, not just give you generic groupadmin role (with null group).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -122,9 +123,10 @@ public Group createGroup(PerunSession sess, Vo vo, Group group) throws GroupExis
 			}
 		}
 
-		// Check the group name, it can contain only a-Z0-9_- and space
-		if (!group.getShortName().matches("^[- a-zA-Z.0-9_]+$")) {
-			throw new InternalErrorException(new IllegalArgumentException("Wrong group name, group name can contain only a-Z0-9.-_: and space characters. " + group));
+		try {
+			Utils.validateGroupName(group.getShortName());
```
zlamalp, line 127: This check is a safe-guard repetition of check from entry layer. Maybe we should add the same check for ``updateGroup()``.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/GroupsManagerBl.java
Vojtech-Sassmann: Import removed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/EduGain.java
Vojtech-Sassmann: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
Vojtech-Sassmann: I have added the check also to the update group method.

PR #2790, opened 2020-07-15
DISPATCHER: Reintroduced configurable re-schedule interval

zlamalp:
- If Tasks are DONE or ERROR for too many hours, they will
  be re-scheduled, even if source data hasn't changed. 
  It was previously configurable, then we
  switched to fixed "two days" period.
  Now, it is configurable again, defaulting to two days
  (setting is in hours in perun-dispatcher.properties).

### DIFF
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
@@ -6,7 +6,6 @@ import cz.metacentrum.perun.core.api.PerunPrincipal;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.bl.PerunBl;
-import cz.metacentrum.perun.core.bl.TasksManagerBl;
 import cz.metacentrum.perun.taskslib.model.Task;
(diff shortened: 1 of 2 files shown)


perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/PropagationMaintainer.java
```diff
@@ -186,11 +191,11 @@ private void rescheduleErrorTasks() {
 				howManyMinutesAgo = task.getDelay() + 1;
 			}
 
-			log.trace("[{}] Task in ERROR state completed {} minutes ago: {}.", new Object[]{task.getId(), howManyMinutesAgo, task});
+			log.trace("[{}] Task in ERROR state completed {} minutes ago: {}.", task.getId(), howManyMinutesAgo, task);
 
 			// If DELAY time has passed, we reschedule...
 			int recurrence = task.getRecurrence() + 1;
-			LocalDateTime twoDaysAgo = LocalDateTime.now().minusDays(2);
+			LocalDateTime toManyHoursAgo = LocalDateTime.now().minusHours(oldRescheduleHours);
```
mvocu, line 198: Probably just a typo here: "toManyHoursAgo" -> "tooManyHoursAgo".

zlamalp: Thanks for noticing, fixed.

PR #2792, opened 2020-07-20
CORE: Handle members.status as integer

zlamalp:
- Converted members.status column to integer and bumped
  DB version, updated changelogs.
- Pass integer value to the SQL working with members.status.
- Fixed tests.


perun-core/src/main/resources/hsqldbChangelog.txt
```diff
@@ -3,6 +3,9 @@
 
 -- this update is not supported on hsql since its used only as in-memory db
 
+3.1.62
```
Vojtech-Sassmann:
```suggestion
3.1.63
```

PR #2795, opened 2020-07-20
GUI - secondary group name regex

Vojtech-Sassmann:
* Gui can now work with a secondary group name regex. This regex has to
be defined as a perun-web-gui property with a name
`groupNameSecondaryRegex`.
* Also added an option to specify a custom error message, when the group
name is invalid. This text can be specified using the
`groupNameErrorMessage` perun-web-gui property.
* If the `groupNameErrorMessage` is not specified, a default message
is shown.
* If the secondary regex is not specified, its validation is skipped.


Vojtech-Sassmann:
> Please do the same for the `EditGroupDetailsTabItem`. Thank you.

Done.

PR #2798, opened 2020-07-21
CORE: Removed all unused "status" columns

zlamalp:
- Changed "status" column to integer for "specific_user_users" table.
- Fixed SQL working with "specific_user_users" table.
- Removed all unused "status" columns of char(1) types from all tables.
- Updated DB schema and test data to 3.1.64.


perun-base/src/test/resources/test-schema.sql
```diff
@@ -423,11 +410,11 @@ create table specific_user_users (
 	modified_by_uid integer,
 	modified_at timestamp default current_date not null,
 	type longvarchar default 'service' not null,
-	status char(1) default '0' not null, --is it service user?
+	status integer default 0 not null, --is it service user?
```
stavamichal, line 413: This comment is misleading. Status in this table means 'enabled or disabled ownership'. Can you please change it here?

zlamalp: Fixed in both DB definition files.

PR #2799, opened 2020-07-21
Core - search users optimizations

Vojtech-Sassmann:
* The most expensive operation when searching users as a Perun Admin was
checking, if the attribute definition passed to the authorization method
is correct. This doesn't need to be performed here, because this is
usually verified in earlier calls or to this method are passed only
definitions retrieved from the database.
* Also removed duplicit conversion of richUsers to
richUsersWithAttributes.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -357,8 +357,7 @@ public class UsersManagerBlImpl implements UsersManagerBl {
 	@Override
 	public List<RichUser> convertRichUsersToRichUsersWithAttributes(PerunSession sess, List<RichUser> richUsers)  throws UserNotExistsException {
 		for (RichUser richUser: richUsers) {
-			User user = getPerunBl().getUsersManagerBl().getUserById(sess, richUser.getId());
-			List<Attribute> userAttributes = getPerunBl().getAttributesManagerBl().getAttributes(sess, user);
+			List<Attribute> userAttributes = getPerunBl().getAttributesManagerBl().getAttributes(sess, richUser);
 
 			richUser.setUserAttributes(userAttributes);
 		}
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -1729,7 +1729,7 @@ public RichUser convertUserToRichUserWithAttributesByNames(PerunSession sess, Us
 	public List<RichUser> findRichUsersWithAttributes(PerunSession sess, String searchString, List<String> attrsName) throws UserNotExistsException {
 
 		if(attrsName == null || attrsName.isEmpty()) {
-			return convertRichUsersToRichUsersWithAttributes(sess, findRichUsers(sess, searchString));
+			return findRichUsers(sess, searchString);
```
zlamalp, line 1732: Same optimization can be done for ``findRichUsersWithAttributesByExactMatch()`` below.

Also I wonder, if we can optimize implementation of ``convertRichUsersToRichUsersWithAttributes()`` itself, which retrieves user by id. We can easily pass ``RichUser`` instead of ``User`` into ``getAttributes()`` method, but I'm not sure that some code doesn't rely on the exact class to be used.

Vojtech-Sassmann: > Same optimization can be done for `findRichUsersWithAttributesByExactMatch()` below.

Done.

> Also I wonder, if we can optimize implementation of `convertRichUsersToRichUsersWithAttributes()` itself, which retrieves user by id. We can easily pass `RichUser` instead of `User` into `getAttributes()` method, but I'm not sure that some code doesn't rely on the exact class to be used.

I have noticed this as well. I took a look at it and I think this should not cause any problems. So I have added also this optimization.

Vojtech-Sassmann: Done.

PR #2800, opened 2020-07-21
Core - get admins for resource manager role

Vojtech-Sassmann:
* The getRichAdmin and getAdminGroups methods were not able to work with
the resourceManager. Now they can, but in the future, these methods has
to be reimplemented, so they can work will all currently defined roles
in the policies configuration.


Vojtech-Sassmann:
> Please update also Javadoc and RPC javadoc for those method. They enumerate supported roles.

Fixed.

PR #2802, opened 2020-07-21
CORE: All attribute tables now have single attr_value column

zlamalp:
- All values of large attributes from "attr_value_text" column
  were moved to the standard "attr_value" column.
  Default column type changed to "text" to contain large values.
- Updated DB schema and changelog 3.1.65.
- Simplified attribute value reading where possible.

- We expect to remove "large" attribute types from the API/logic
  in the future.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
@@ -1,8 +1,6 @@
 package cz.metacentrum.perun.core.impl;
 
 import cz.metacentrum.perun.core.api.Attribute;
-import cz.metacentrum.perun.core.api.AttributeDefinition;
-import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.BeansUtils;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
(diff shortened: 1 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/SearcherImpl.java
```diff
@@ -186,62 +183,37 @@ private void insertWhereClausesAndQueryParametersFromAttributes(StringBuilder qu
...
+				whereClauses.add("val" + counter + ".attr_value IS NULL ");
 			} else {
 				if (key.getType().equals(Integer.class.getName())) {
 					key.setValue(Integer.valueOf(value));
 					whereClauses.add("val" + counter + ".attr_value=:v" + counter + " ");
 					whereClauses.add("nam" + counter + ".type=:n" + counter + " ");
 					parameters.addValue("n" + counter, Integer.class.getName());
 					parameters.addValue("v" + counter, BeansUtils.attributeValueToString(key));
-				} else if (key.getType().equals(String.class.getName())) {
+				} else if (key.getType().equals(String.class.getName()) || key.getType().equals(BeansUtils.largeStringClassName)) {
```
stavamichal, line 194: There is nothing like large objects anymore, so can't we remove these special parts too?

zlamalp: I'd like to, but for now we have existing attributes with these types and I'm not sure, that nobody has that harcoded when calling API. But If we agree and let others know, we can convert all such attributes to their default type using DB changelog and then it can be removed from everywhere (gui/cli/core).

PR #2803, opened 2020-07-22
GUI: Introduce password strength check for presence of login

zlamalp:
- Check that password doesn't contain login (case insensitive
  and unaccented) or backwards login.
  Login is checked only if longer than 2 chars to prevent
  false positive on short strings.

### DIFF
--- a/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/memberstabs/CreateServiceMemberInVoTabItem.java
+++ b/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/memberstabs/CreateServiceMemberInVoTabItem.java
@@ -204,6 +204,18 @@ public class CreateServiceMemberInVoTabItem implements TabItem, TabItemWithUrl {
 						return false;
 					}
 
+					// check on login in password if login is longer than 2 chars
+					// TODO - check for name/surname too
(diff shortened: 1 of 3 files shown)


perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/memberstabs/CreateServiceMemberInVoTabItem.java
```diff
@@ -204,6 +204,20 @@ public boolean validateTextBox() {
 						return false;
 					}
 
+					// check on login in password if login is longer than 2 chars
+					// TODO - check for name/surname too
+					String pass = serviceUserPassword.getTextBox().getValue();
+					String login = serviceUserLogin.getTextBox().getValue();
+					if (login.length() > 2) {
+						if (pass.toLowerCase().contains(login.toLowerCase()) ||
+								pass.toLowerCase().contains(Utils.reverseString(login.toLowerCase())) ||
```
Vojtech-Sassmann, line 213: Do you need to perform these checks? Wouldn't the following conditions handle these situations?

zlamalp: You are right. There is no point in duplicate check.

PR #2811, opened 2020-07-22
Validate sponsored members after creating of sponsorship

stavamichal:
- sponsored members could be expired (after first sponsorship ends), so
 we need to be sure that new sponsorship will validate them again (and
 also remove their expiration date)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2425,6 +2425,19 @@ public Member sponsorMember(PerunSession session, Member sponsoredMember, User s
 		getMembersManagerImpl().addSponsor(session, sponsoredMember, sponsor);
 		getPerunBl().getAuditer().log(session, new SponsorshipEstablished(sponsoredMember, sponsor));
```
Vojtech-Sassmann, line 2426: I would suggest moving this log to the end of the method. If any of the next operations fail, the process of the AuditEvent would be done unnecessarily.

stavamichal: You are right. Done.

PR #2816, opened 2020-07-24
Roles - default politics file

Vojtech-Sassmann:
* With the old implementation, it was difficult to release new versions
of perun with new policits. All configurations would have to be changed
and there was a lot of places to do a manual error. Even right now, we
don't plan to have different politics on different instances.
* Created a default politics cofiguration file
`perun-base/src/main/resources/perun-roles.yml` which is packed into the
war archive and will be used in production and even in tests.
* Created a test configuration file `test-roles.yml` which contains
politics specially created for tests.
* Changed behaviour of loading of the politics. On production, the
default politics configuration is used and also, the configuration file
at `/etc/perun/perun-roles.yml` can be used to override some politics.
This file is ignored during tests.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
```diff
@@ -71,36 +77,22 @@ public void loadPerunRoles(JdbcPerunTemplate jdbc) {
...
-				List<String> includePolicies = new ArrayList<>(objectMapper.convertValue(policyNode.get("include_policies"), new TypeReference<List<String>>() {
-				}));
-
-				policies.add(new PerunPolicy(policyName, perunRoles, includePolicies));
+			JsonNode rootNode;
+			if (secondaryConfigurationPath != null) {
+				rootNode = loadConfigurationFile(secondaryConfigurationPath);
+				policies.addAll(loadPoliciesFromJsonNode(rootNode));
 			}
+			rootNode = loadConfigurationFile(configurationPath);
```
zlamalp, line 89: Shouldn't we first load normal configuration and then overwrite it with secondary config (as we do in ```loadPerunRolesManagement()```)?

zlamalp: Ok, @balcirakpeter corrected me, that its on purpose, because of behavior of adding policies method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
```diff
@@ -116,36 +107,52 @@ public void loadPerunRoles(JdbcPerunTemplate jdbc) {
...
-			throw new InternalErrorException("The configuration file " + configurationPath.getFilename() + " has invalid syntax.", e);
+			throw new InternalErrorException("One of the roles configuration file has invalid syntax. Configuration files: " +
+					configurationPath.getFilename() +
+					(secondaryConfigurationPath == null ? "not defined" : secondaryConfigurationPath.getFilename()), e);
 		}
 
 		return rolesManagementRules;
 	}
 
+	private Set<RoleManagementRules> loadPerunRolesManagementGromJsonNode(JsonNode rootNode) {
```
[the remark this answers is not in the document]
Vojtech-Sassmann: fixed

PR #2817, opened 2020-07-27
OpenAPI getUsersBySpecificUser

bodnara:
* added method which returns user owners in Users Manager

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -7372,6 +7372,20 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/usersManager/getUsersBySpecificUser:
+    get:
+      tags:
+        - UsersManager
+      operationId: getUsersBySpecificUser
+      summary: Return all users who owns the specific user.
+      parameters:
+        - $ref: '#/components/parameters/specificUserId'
+      responses:
+        '200':
+          $ref: '#/components/responses/ListOfUsersResponse'
+        default:
+          $ref: '#/components/responses/ExceptionResponse'
+
 
 
   #################################################


perun-openapi/openapi.yml
```diff
@@ -7372,6 +7372,20 @@ paths:
...
           $ref: '#/components/responses/ExceptionResponse'
 
+  /urlinjsonout/usersManager/getUsersBySpecificUser:
+    post:
+      tags:
+        - UsersManager
+      operationId: getUsersBySpecificUser
+      summary: Return all users who owns the specific user.
+      parameters:
+        - $ref: '#/components/parameters/specificUserId'
```
balcirakpeter, line 7382: I think the correct parameter should be '#/components/parameters/specificUser'
You can find the parameter name in UsersManagerMethod.java in the call of the getUsersBySpecificUser method.

perun-openapi/openapi.yml
```diff
@@ -7372,6 +7372,20 @@ paths:
...
+      responses:
+        '200':
+          $ref: '#/components/responses/VoidResponse'
```
balcirakpeter, line 7385: The correct response type should be '#/components/responses/ListOfUsersResponse'
Again the response type can be found in the UsersManagerMethod.java in the JavaDoc of the getUsersBySpecificUser.

perun-openapi/openapi.yml
```diff
@@ -7372,6 +7372,20 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /urlinjsonout/usersManager/getUsersBySpecificUser:
+    post:
```
balcirakpeter, line 7376: I think the operation is get and not post.

perun-openapi/openapi.yml
```diff
@@ -7372,6 +7372,20 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /urlinjsonout/usersManager/getUsersBySpecificUser:
```
balcirakpeter, line 7375: The URL should be '/json/usersManager/getUsersBySpecificUser' because the method is GET type and the parameter is just a user id.

perun-openapi/openapi.yml
bodnara: This is actually correct, specificUserId is defined with the name specificUser, it is the same as with userId and user.

balcirakpeter: Sorry, you are right. Missed the reference.

PR #2818, opened 2020-07-27
CORE: Entityless attributes must be marked as writable

zlamalp:
- In order to edit attributes in GUI we must mark them as editable.
  It was used in old methods, but was missing in new methods.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
@@ -756,7 +756,13 @@ public class AttributesManagerEntry implements AttributesManager {
 			throw new PrivilegeException(sess, "getEntitylessAttributesWithKeys");
 		}
 
-		return attributesManagerBl.getEntitylessAttributesWithKeys(sess, attrName);
+		Map<String, Attribute> result = attributesManagerBl.getEntitylessAttributesWithKeys(sess, attrName);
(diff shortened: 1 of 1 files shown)


zlamalp:
> Is there any reason why you are not using the method "isAuthorizedForAttribute" to get info about read and write rights for these attributes? It would be a much cleaner way in my opinion.

I basically re-used logic from ``getEntitylessAttributes()``. Its probably quicker this way, since ``isAuthorizedForAttribute()`` perform some duplicite checks. But you are right, we should look more closely into it. There is different authorization used eg. in ``set/removeAttribute()``. First is using new policies, second is using ``isAuthorizedForAttribute()``.

stavamichal:
> > Is there any reason why you are not using the method "isAuthorizedForAttribute" to get info about read and write rights for these attributes? It would be a much cleaner way in my opinion.
> 
> I basically re-used logic from `getEntitylessAttributes()`. Its probably quicker this way, since `isAuthorizedForAttribute()` perform some duplicite checks. But you are right, we should look more closely into it. There is different authorization used eg. in `set/removeAttribute()`. First is using new policies, second is using `isAuthorizedForAttribute()`.

Firstly you are checking if principal has access rights to call the method itself. Then you are checking if he has the right to read or write a value of the attribute.

zlamalp:
Entityless attributes are special case, since former methods don't contain "key", hence can't be checked by current authorization. I added authorization to the new methods, but it will be replaced/unified once we will have new authorization for attributes.

PR #2831, opened 2020-08-04
Edit of findUsers

metodej:
- findUsers now uses only one query with multiple joins and conditions
- the attributes to search users or members by are now stored in perun.properties and loaded by CoreConfig (it supports only full names of attributes)
- these changes were also applied to findMembers
- findUsers and findUsersByExactMatch now use private method with switch which selects whether name should be search only by exact match or not
- also added tests for this method and fixed existing tests that were affected

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
@@ -6,6 +6,7 @@ import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.MemberGroupStatus;
 import cz.metacentrum.perun.core.api.MembershipType;
+import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.PerunPrincipal;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
@@ -509,9 +510,9 @@ public class MembersManagerImpl implements MembersManagerImplApi {
 
 		String userNameQueryString;
 		if (Compatibility.isPostgreSql()) {
-			userNameQueryString= " strpos(lower("+Compatibility.convertToAscii("COALESCE(u.first_name,'') || COALESCE(u.middle_name,'') || COALESCE(u.last_name,'')")+"),?) > 0 ";
+			userNameQueryString= " strpos(lower("+Compatibility.convertToAscii("COALESCE(users.first_name,'') || COALESCE(users.middle_name,'') || COALESCE(users.last_name,'')")+"),:nameString) > 0 ";
 		} else if (Compatibility.isHSQLDB()) {
-			userNameQueryString=" lower("+Compatibility.convertToAscii("COALESCE(u.first_name,'') || COALESCE(u.middle_name,'') || COALESCE(u.last_name,'')")+") like '%' || ? || '%' ";
+			userNameQueryString=" lower("+Compatibility.convertToAscii("COALESCE(users.first_name,'') || COALESCE(users.middle_name,'') || COALESCE(users.last_name,'')")+") like '%' || :nameString || '%' ";
 		} else {
 			throw new InternalErrorException("Unsupported db type");
 		}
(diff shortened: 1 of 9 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -533,18 +548,18 @@ public MemberGroupStatus getUnifiedMemberGroupStatus(PerunSession sess, User use
 		Set<Member> members = new HashSet<>(jdbc.query("select distinct " + memberMappingSelectQuery +
 				" from members " +
 				" left join users u on members.user_id=u.id " +
-				" left join member_attr_values mav1 on members.id=mav1.member_id and mav1.attr_id in (select id from attr_names where attr_name='" + MembersManagerImpl.A_D_MEMBER_MAIl + "') " +
-				" left join user_attr_values uav1 on u.id=uav1.user_id and uav1.attr_id in (select id from attr_names where attr_name='" + MembersManagerImpl.A_D_USER_PREFERRED_MAIL + "') " +
-				" left join user_attr_values uav2 on u.id=uav2.user_id and uav2.attr_id in (select id from attr_names where friendly_name like 'login-namespace:%') " +
+				" left join member_attr_values mav on members.id=mav.member_id and mav.attr_id in (select id from attr_names where attr_name in (" + memberAttributes.toString() + "))" +
```
zlamalp, line 551: There is possible SQL injection, since you put values from the config right into the SQL. Please use ``?`` placeholder instead and pass it as a lists of strings. If necessary, you can use ``namedParameterJdbcTemplate`` instead of ``jdbc``, but I believe ``query()`` should be able to work with the list params.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -794,90 +795,92 @@ public void removeAllUserExtSources(PerunSession sess, User user) {
...
+		// Search by member attributes
+		// Search by user attributes
+		// Search by login in userExtSources
+		// Search by userExtSource attributes
+		// Search by user id
+		// Search by name for user
+		Set<User> users = new HashSet<>(jdbc.query("select distinct " + userMappingSelectQuery +
+			" from users " +
+			" left join members m on users.id=m.user_id " +
+			" left join member_attr_values mav on m.id=mav.member_id and mav.attr_id in (select id from attr_names where attr_name in (" + memberAttributes.toString() + "))" +
```
zlamalp, line 869: Same as in ``MembersManagerImpl``, please use ``?`` placeholders.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
metodej: Oh, thank you, it should be fixed now, although I was unable to make it work without `namedParameterJdbcTemplate `.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -794,90 +795,112 @@ public void removeAllUserExtSources(PerunSession sess, User user) {
...
-			users.addAll(jdbc.query("select " + userMappingSelectQuery + " from users where id=?", USER_MAPPER, userId));
+			int id = Integer.parseInt(searchString);
+			idQueryString = " users.id=" + id + " or ";
 		} catch (NumberFormatException e) {
-			// IGNORE
+			// IGNORE wrong format of ID
+		}
+
+		// Divide attributes received from CoreConfig into member, user and userExtSource attributes
+		List<String> allAttributes = BeansUtils.getCoreConfig().getAttributesToSearchUsersAndMembersBy();
```
balcirakpeter, line 846: The next lines of code seem to be the same as in MembersManagerImpl. Could you extract it to a separate method that can be used in both managers?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -794,90 +795,112 @@ public void removeAllUserExtSources(PerunSession sess, User user) {
...
+	 * Returns list of users who matches the searchString, searching name, id, member attributes, user attributes
+	 * and userExtSource attributes (listed in CoreConfig).
+	 *
+	 * @param searchString string used to search by
+	 * @param exactMatch if true, searches name only by exact match
+	 * @return list of users
+	 */
+	private List<User> findUsers(String searchString, boolean exactMatch) {
+		String userNameQueryString;
+		if (exactMatch) {
```
balcirakpeter, line 816: Both conditional branches can be extracted to separate methods. One could be called something like "prepareUserSearchQueryExactMatch" and the second one "prepareUserSearchQuerySimilarMatch" or something like that. It will be more readable. Moreover, the second method can be used also in MembersManagerImpl so you eliminate duplicities.

metodej: Ok, I've created new methods in Utils.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
metodej: Ok, I've created new methods in Utils, although I'm not sure if it is not too complicated, I couldn't think of anything nicer.

PR #2836, opened 2020-08-05
New modules and tests for attributes

stavamichal:
- attribute g:d:o365EmailAddresses:o365mu is new attribute for
 migration data from gr:d:o365EmailAddresses:mu
   - it has the same value -> array list of email addresses
   - syntaxCheck: duplicity is forbidden, email pattern is checked
   - semanticCheck: if group has attribute g:d:adName:o365mu set
 to not-null, it has to be not-null too (there need to be a value).
 Duplicities between other groups and members with
 m:d:o365EmailAddresses:mu are forbidden.
   - fill: if group has attribute g:d:adName:o365mu set to not-null, it
 will fill list with 1 value generated from adName

 - attribute g:d:adName:o365mu is new attribute for migration data from
 gr:d:adName
   - is has unique String value (defined on attribute itself)
   - syntaxCheck: regex pattern is checked
   - semanticCheck: group can't be assigned to more than 1 resource with
 set attribute r:d:adOuName in the sama facility

 - tests for both new modules were added


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_o365EmailAddresses_o365mu.java
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
+		attr.setFriendlyName("o365EmailAddresses:o365mu");
+		attr.setDisplayName("MU O365 email addresses in o365mu");
+		attr.setType(ArrayList.class.getName());
+		attr.setUnique(true);
```
zlamalp, line 138: Judging by the commit message this attribute is not mentioned as ``unique`` by definition, you already check duplicates in ``checkAttributeSyntax()``. Or you can have it unique and then remove that check, but it will prevent two different groups to have same email (not just unique within one attribute value), which I'm not sure is wanted or not.

Also do we care about case-insensitive or non-accented uniqueness too (remembering the problems with tcsMails:mu attribute)?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_o365EmailAddresses_o365mu.java
```diff
@@ -0,0 +1,142 @@
...
+
+		//check for duplicities among all members and groups
+		AttributesManagerBl attributesManagerBl = sess.getPerunBl().getAttributesManagerBl();
+		Set<Integer> setOfGroupsDefinedByAttribute = attributesManagerBl.getPerunBeanIdsForUniqueAttributeValue(sess, attribute).stream().map(Pair::getLeft).collect(Collectors.toSet());
+		setOfGroupsDefinedByAttribute.remove(group.getId());
+
+		if (!setOfGroupsDefinedByAttribute.isEmpty()) {
+			throw new WrongReferenceAttributeValueException(attribute, attribute, group, null, "some of the email addresses are already assigned to the following group pairs: " + setOfGroupsDefinedByAttribute);
+		}
+		Attribute memberO365EmailAddresses = new Attribute(new urn_perun_member_attribute_def_def_o365EmailAddresses_mu().getAttributeDefinition());
```
zlamalp, line 101: I'm not sure we rely on attribute definition from the module except the tests, since it might differ from the real perun instance state. Also subsequent call of ``getPerunBeanIdsForUniqueAttributeValue()`` requires attribute definition to have real id, which can't be present in this construct as is set to 0.
I suggest to get attribute definition by name (urn).

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_o365EmailAddresses_o365mu.java
```diff
@@ -0,0 +1,142 @@
...
+
+				return;
+			}
+		} catch (AttributeNotExistsException e) {
+			throw new InternalErrorException(e.getMessage(), e);
+		}
+
+		//check for duplicities among all members and groups
+		AttributesManagerBl attributesManagerBl = sess.getPerunBl().getAttributesManagerBl();
+		Set<Integer> setOfGroupsDefinedByAttribute = attributesManagerBl.getPerunBeanIdsForUniqueAttributeValue(sess, attribute).stream().map(Pair::getLeft).collect(Collectors.toSet());
```
zlamalp, line 95: If the attribute will be ``unique``, we don't have to check uniqueness through the same attribute, since its checked on inserting the value.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_o365EmailAddresses_o365mu.java
stavamichal: It is my mistake. The attribute is already defined as unique and it will stay in this format even for the group.
Because we just need to copy the existing version of module to the new one, there is no need to fix all situations like "case sensitivity" etc.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_o365EmailAddresses_o365mu.java
stavamichal: Changed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_o365EmailAddresses_o365mu.java
stavamichal: Yes, this is not needed here. Only members check against this attribute.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_adName_o365mu.java
```diff
@@ -0,0 +1,107 @@
...
+		Map<Integer, Set<Resource>> mapOfResourcesByFacility = new HashMap<>();
+		for (Resource resource: sess.getPerunBl().getResourcesManagerBl().getAssignedResources(sess, group)) {
+			if (!mapOfResourcesByFacility.containsKey(resource.getFacilityId())) {
+				Set<Resource> setOfResourceWithSameFacilityId = new HashSet<>();
+				setOfResourceWithSameFacilityId.add(resource);
+				mapOfResourcesByFacility.put(resource.getFacilityId(), setOfResourceWithSameFacilityId);
+			} else {
+				mapOfResourcesByFacility.get(resource.getFacilityId()).add(resource);
+			}
+		}
```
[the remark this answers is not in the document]
stavamichal: I like it. Changed. Thank you.

PR #2837, opened 2020-08-05
Implemented working with Form Items

dBucik:
- new Bean - FormItem
- added methods to RegistrarAgent for getting and updating FormItems for group

### DIFF
--- a/perun-cli/Perun/beans/ApplicationFormItem.pm
+++ b/perun-cli/Perun/beans/ApplicationFormItem.pm
@@ -1,4 +1,4 @@
-package Perun::beans::FormItem;
+package Perun::beans::ApplicationFormItem;
 
 use strict;
 use warnings;
@@ -75,11 +75,11 @@ sub TO_JSON
         $regex = undef;
     }
 
-    my $applicationTypes;
+    my @applicationTypes;
     if (defined($self->{_applicationTypes})) {
-        $applicationTypes = $self->{_applicationTypes};
+        @applicationTypes = @{$self->{_applicationTypes}};
     } else {
-        $applicationTypes = undef;
+        @applicationTypes = undef;
     }
 
     my $ordnum;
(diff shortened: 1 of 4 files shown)


perun-cli/Perun/beans/FormItem.pm
```diff
@@ -0,0 +1,294 @@
...
+        type                      => $type,
+        federationAttribute       => $federationAttribute,
+        perunSourceAttribute      => $perunSourceAttribute,
+        perunDestinationAttribute => $perunDestinationAttribute,
+        regex                     => $regex,
+        applicationTypes          => $applicationTypes,
+        ordnum                    => $ordnum,
+        forDelete                 => $forDelete,
+        i18n                      => $i18n,
+        beanName                  => 'FormItem'
```
zlamalp, line 118: Judging by the Java implementation of ``getBeanName()`` it should return ``ApplicationFormItem``.

perun-cli/Perun/beans/FormItem.pm
```diff
@@ -0,0 +1,294 @@
...
+    $self->{_regex} = shift;
+
+    return;
+}
+
+sub getApplicationTypes
+{
+    my $self = shift;
+
+    return $self->{_applicationTypes};
```
zlamalp, line 238: I think this is an array with string values of ``INITIAL`` and/or ``EXTENSION`` and should be returned like
```perl
return @{$self->{_applicationTypes}};
```
Maybe also fix ``sub TO_JSON``, see ``AttributeRights.pm``

perun-cli/Perun/beans/FormItem.pm
```diff
@@ -0,0 +1,294 @@
...
+{
+    my $self = shift;
+
+    return $self->{_forDelete};
+}
+
+sub setForDelete
+{
+    my $self = shift;
+    $self->{_forDelete} = shift;
```
zlamalp, line 274: ``forDelete`` is boolean property. See, how we handle getter/setter in ``User.pm`` for ``serviceUser`` property.

perun-cli/Perun/beans/ApplicationFormItem.pm
```diff
@@ -0,0 +1,313 @@
...
+    $self->{_ordnum} = shift;
+
+    return;
+}
+
+sub isForDelete
+{
+    my $self = shift;
+
+    return ($self->{_forDelete}) ? 'true' : 'false';
```
zlamalp, line 277: Depending on your usage you might want to retrieve original value (boolean or 0/1) which you set using ``setForDelete``. This kind of handling is useful for printing the value in tables in a consistent way. You can have ``sub isForDeleteToPrint`` for returning true/false as strings.

PR #2838, opened 2020-08-06
Fix invalid entry calls

Vojtech-Sassmann:
* There were some methods that didn't call directly entry layer but
instead, they called methods from ApiCaller which are for a different
purpose. Because of that in the entry layer, there was an invalid
Principal.
* Added role SELF to policy for the `getGroupById` method because after
this fix, group members would not be able to call this method.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ExtSourcesManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ExtSourcesManagerMethod.java
@@ -62,7 +62,7 @@ public enum ExtSourcesManagerMethod implements ManagerMethod {
 
 		@Override
 		public ExtSource call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getExtSourceById(parms.readInt("id"));
+			return ac.getExtSourcesManager().getExtSourceById(ac.getSession(), parms.readInt("id"));
 		}
 	},
 
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
@@ -501,7 +501,7 @@ public enum MembersManagerMethod implements ManagerMethod {
 	getMemberById {
 		@Override
 		public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getMemberById(parms.readInt("id"));
+			return ac.getMembersManager().getMemberById(ac.getSession(), parms.readInt("id"));
 		}
 	},
 
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
@@ -24,7 +24,7 @@ public enum ResourcesManagerMethod implements ManagerMethod {
 
 		@Override
 		public Resource call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getResourceById(parms.readInt("id"));
+			return ac.getResourcesManager().getResourceById(ac.getSession(), parms.readInt("id"));
 		}
 	},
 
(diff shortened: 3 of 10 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ExtSourcesManagerMethod.java
```diff
@@ -62,7 +62,7 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 
 		@Override
 		public ExtSource call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getExtSourceById(parms.readInt("id"));
+			return ac.getExtSourcesManager().getExtSourceById(ac.getSession(), parms.readInt("id"));
```
zlamalp, line 65: This is called by CLI when using tools to add/remove UES or set/remove their attributes. We should check, if user (SELF) can call these methods and if yes, we should enable SELF role in this methods policy.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -501,7 +501,7 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
 	getMemberById {
 		@Override
 		public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getMemberById(parms.readInt("id"));
+			return ac.getMembersManager().getMemberById(ac.getSession(), parms.readInt("id"));
```
zlamalp, line 504: I believe we should allow also SELF on Member/User in this methods policy.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -24,7 +24,7 @@
 
 		@Override
 		public Resource call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getResourceById(parms.readInt("id"));
+			return ac.getResourcesManager().getResourceById(ac.getSession(), parms.readInt("id"));
```
zlamalp, line 27: A would allow this method for ``FACILITYADMIN`` too. They probably use version with RichResource from GUI, but its not consistent with expected logic that facility admin can retrieve resources of his facility (by any identifier).

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -399,7 +399,7 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 
 		@Override
 		public Service call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getServiceById(parms.readInt("id"));
+			return ac.getServicesManager().getServiceById(ac.getSession(), parms.readInt("id"));
```
zlamalp, line 402: Its strange, that this method is not allowed for facility admins (same for ByName version), but facility admins can retrieve all services together. We should allow it for them in both methods.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -966,7 +966,7 @@ public ServiceAttributes call(ApiCaller ac, Deserializer parms) throws PerunExce
 
 		@Override
 		public ServicesPackage call(ApiCaller ac, Deserializer parms) throws PerunException {
-			return ac.getServicesPackageById(parms.readInt("servicesPackageId"));
+			return ac.getServicesManager().getServicesPackageById(ac.getSession(), parms.readInt("servicesPackageId"));
```
zlamalp, line 969: Same goes for all methods working with service packages, since facility admin is the one to get/assign service packages to the facility and perun admin manages those packages.

zlamalp (changes requested):
Its OK. I just think, that we can prevent most of unexpected breakage for clients by allowing more roles to call the API methods. Changes are suggested in comments.

Vojtech-Sassmann:
@zlamalp I have updated the politics based on your suggestions.

zlamalp (approved):

PR #2842, opened 2020-08-10
Fixh authz for rich objets

balcirakpeter:
- Rich objects were not correctly processed during the authorization. It
  was because they extend base objects but their bean name starts with
  Rich, which is not saved in the database.
- To fix this, method fetchAllRelatedObjects had to be changed. It does
  not simply add source object to the result list but it creates base
  object in case the source object is of type "Rich". This new object is
  then added to the result list.


balcirakpeter:
> I think it might be relevant to also support the `UserExtSource` to fetch the related user object.

Added

PR #2845, opened 2020-08-11
Perun roles documentation

balcirakpeter:
- Added documentation for roles which are used in Perun. The
  documentation is located in perun-roles.yml.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -1,5 +1,84 @@
 ---
 # A list of Perun roles that are loaded to the database
+#
+# PERUNADMIN role can perform all the actions possible in the Perun system.
+# It is the only role that is able to also set this role to other users.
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -1,5 +1,84 @@
 ---
 # A list of Perun roles that are loaded to the database
+#
+# PERUNADMIN role can perform all the actions possible in the Perun system.
+# It is the only role that is able to also set this role to other users.
+#
+# PERUNOBSERVER role has access to all the information, but it is not able to perform any changes
+# except for the delegation of the role to other users or groups. The role can be set only by Perun administrators.
```
Vojtech-Sassmann, line 8: This is not true.

balcirakpeter: Fixed

PR #2847, opened 2020-08-11
New method to convert unique attribute to nonunique

metodej:
- new method convertAttributeToNonunique converts unique attribute to nonunique by setting uniqueness flag in attribute definition to false and removing values of attribute from table _attr_u_values which has unique constraint
- convertAttributeToNonunique was added to all layers of Perun including rpc and openapi, also added test
- updateAttributeDefinition fixed to update uniqueness of attribute definition (and not just set it to true)
- method convertAttributeToUnique also added to openapi
- removed forgotten debugging of convertAttributeToUnique method

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -8116,10 +8116,19 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		log.info("converting attribute {} to unique",attrDef.getName());
 		attrDef.setUnique(true);
 		this.updateAttributeDefinition(session, attrDef);
-		long startTime = System.currentTimeMillis();
+
 		attributesManagerImpl.convertAttributeValuesToUnique(session, attrDef);
-		long endTime = System.currentTimeMillis();
-		log.debug("Attribute {} was converted to unique in {} ms",attrDef.getName(),(endTime-startTime));
+	}
+
+	@Override
+	public void convertAttributeToNonunique(PerunSession session, int attrId) throws AttributeNotExistsException {
+		AttributeDefinition attrDef = getAttributeDefinitionById(session, attrId);
+		if(!attrDef.isUnique()) return;
+		log.info("converting unique attribute {} to nonunique",attrDef.getName());
+		attrDef.setUnique(false);
+		this.updateAttributeDefinition(session, attrDef);
+
+		attributesManagerImpl.convertAttributeValuesToNonunique(session, attrDef);
 	}
 
 	@Override
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -8124,7 +8124,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 	@Override
 	public int convertAttributeToNonunique(PerunSession session, int attrId) throws AttributeNotExistsException, AttributeNotMarkedUniqueException {
 		AttributeDefinition attrDef = getAttributeDefinitionById(session, attrId);
-		if(!attrDef.isUnique()) throw new AttributeNotMarkedUniqueException("Cannot convert attribute because it is already marked as unique", attrDef);
+		if(!attrDef.isUnique()) throw new AttributeNotMarkedUniqueException("Cannot convert attribute because it is already marked as nonunique", attrDef);
 		log.info("converting unique attribute {} to nonunique",attrDef.getName());
 		attrDef.setUnique(false);
 		this.updateAttributeDefinition(session, attrDef);
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -75,6 +75,7 @@ import cz.metacentrum.perun.core.api.exceptions.ActionTypeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeAlreadyMarkedUniqueException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeDefinitionExistsException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotMarkedUniqueException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
 import cz.metacentrum.perun.core.api.exceptions.FacilityNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
@@ -8121,14 +8122,14 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 	}
 
 	@Override
-	public void convertAttributeToNonunique(PerunSession session, int attrId) throws AttributeNotExistsException {
+	public int convertAttributeToNonunique(PerunSession session, int attrId) throws AttributeNotExistsException, AttributeNotMarkedUniqueException {
 		AttributeDefinition attrDef = getAttributeDefinitionById(session, attrId);
-		if(!attrDef.isUnique()) return;
+		if(!attrDef.isUnique()) throw new AttributeNotMarkedUniqueException("Cannot convert attribute because it is already marked as unique", attrDef);
 		log.info("converting unique attribute {} to nonunique",attrDef.getName());
 		attrDef.setUnique(false);
 		this.updateAttributeDefinition(session, attrDef);
 
-		attributesManagerImpl.convertAttributeValuesToNonunique(session, attrDef);
+		return attributesManagerImpl.convertAttributeValuesToNonunique(session, attrDef);
 	}
 
 	@Override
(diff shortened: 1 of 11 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
```diff
@@ -4522,6 +4522,18 @@ public void convertAttributeToUnique(PerunSession session, int attrId) throws Pr
 		getAttributesManagerBl().convertAttributeToUnique(session, attrId);
 	}
 
+	@Override
+	public void convertAttributeToNonunique(PerunSession session, int attrId) throws PrivilegeException, AttributeNotExistsException {
+		Utils.checkPerunSession(session);
+
+		// Authorization
+		if(!AuthzResolver.authorizedInternal(session, "convertAttributeToNonunique_int_policy")) {
+			throw new PrivilegeException("This operation can do only PerunAdmin.");
```
stavamichal, line 4531: For this exception, I would not use this kind of specific message, because you can't be sure what is really set in the policy and if anyone changes the policy, this text would be misleading. Use more generic message like:
```
throw new PrivilegeException(sess, "convertAttributeToNonunique");
```

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AttributesManagerImplApi.java
```diff
@@ -2601,4 +2601,9 @@
 	 * Copies all values of the attribute to table _attr_u_values which has unique constraint.
 	 */
 	void convertAttributeValuesToUnique(PerunSession session, AttributeDefinition attrDef);
+
+	/**
+	 * Deletes all values of the attribute from table _attr_u_values which has unique constraint.
+	 */
+	void convertAttributeValuesToNonunique(PerunSession session, AttributeDefinition attrDef);
```
stavamichal, line 2608: I am thinking about the way how to check that all expected values were removed. What about changing the behavior of both IMPL and BL methods for conversion to return a number of rows added or deleted? Then this could be at least tested in tests. What do you think?

perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
metodej: Ok, fixed (even in convertAttributeToUnique).

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AttributesManagerImplApi.java
metodej: Yes, that sounds nice, I've already tested it manually, because the test already in use basically tests only the change of the attribute to unique/nonunique state... So I've added one more that checks exactly how many rows were changed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4519,6 +4515,16 @@ public void convertAttributeValuesToUnique(PerunSession session, AttributeDefini
 		}
 	}
 
+	@Override
+	public int convertAttributeValuesToNonunique(PerunSession session, AttributeDefinition attrDef) {
+		String tablePrefix = attributeToTablePrefix(attrDef);
+		try {
+			return jdbc.update("DELETE FROM " + tablePrefix + "_attr_u_values WHERE attr_id=?", attrDef.getId());
```
Vojtech-Sassmann, line 4522: @stavamichal Do we really want to delete all values when doing this conversion? Couldn't we just move those values to a different table? I understand that when you do the opposite conversion, you have to remove the old values because there could be duplicates. But in this direction, I think it is not necessary.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTestAbstract.java
```diff
@@ -5499,6 +5499,158 @@ public void testConvertingToUniqAttribute() throws Exception {
 		}
 	}
 
+	@Test
+	public void testConvertingToNonuniqAttribute() throws Exception {
```
Vojtech-Sassmann, line 5503: Hi @metodej, It is a good practice that each test is testing only one thing. For example, you would have one test to test the conversion of a string attribute, one for a list attribute etc. If you test multiple scenarios in one test, one has to debug what has actually gone wrong. If this test would fail, it isn't easy to understand why, because the test itself is long and with two for loops and two switches. 

To write multiple tests, you wouldn't even have to duplicate the code. You could write a private test method that would contain the shared code, and you could pass some arguments to this method, which would differ for different scenarios.

If you would decide to rewrite this and had some trouble, just write to me and we will figure something out.

perun-openapi/openapi.yml
```diff
@@ -5938,7 +5938,35 @@ paths:
...
+
+  /urlinjsonout/attributesManager/convertAttributeToNonunique:
+    post:
+      tags:
+        - AttributesManager
+      operationId: convertAttributeToNonunique
+      summary: "Converts attribute to nonunique - unmarks unique flag from attribute definition, and deletes all values
+      from a special table with unique constraint that ensures that all values remain unique."
+      parameters:
+        - $ref: '#/components/parameters/attributeInteger'
```
Vojtech-Sassmann, line 5964: This parameter has a name `attribute` and in the `AttributesManagerMethod` you are expecting a parameter named `attrDefId`. These two values have to be the same. I suggest changing the name in the OpenApi because the method `convertAttributeToUnique` could be already used by someone and if you would change it, it would stop working for them.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
stavamichal: Unique attributes has values in both tables "_attr_values" and "_attr_u_vaues". So yes, you need to delete them. If you let them here, you can have a big problem when you want to convert attribute to unique again. There will be inconsistency in data.

Vojtech-Sassmann: I see. I thought they are only in this table. Thanks for your explanation.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -8116,10 +8117,19 @@ public void convertAttributeToUnique(PerunSession session, int attrId) throws At
...
+
 		attributesManagerImpl.convertAttributeValuesToUnique(session, attrDef);
-		long endTime = System.currentTimeMillis();
-		log.debug("Attribute {} was converted to unique in {} ms",attrDef.getName(),(endTime-startTime));
+	}
+
+	@Override
+	public int convertAttributeToNonunique(PerunSession session, int attrId) throws AttributeNotExistsException, AttributeNotMarkedUniqueException {
+		AttributeDefinition attrDef = getAttributeDefinitionById(session, attrId);
+		if(!attrDef.isUnique()) throw new AttributeNotMarkedUniqueException("Cannot convert attribute because it is already marked as unique", attrDef);
```
zlamalp, line 8127: Exception message should state ``non unique`` at the end.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -3974,6 +3975,25 @@ public Void call(ApiCaller ac, Deserializer parms) throws AttributeAlreadyMarked
 		}
 	},
 
+	/*#
+	 * Converts attribute to nonunique - unmarks unique flag from attribute definition, and deletes all values
+	 * from a special table with unique constraint that ensures that all values remain unique.
+	 * Already nonunique attribute is skipped.
+	 *
+	 * @param attrDefId int AttributeDefinition <code>id</code>
+	 * @throw PrivilegeException Insufficient permissions.
```
zlamalp, line 3984: We do not mention ``PriviligeException`` in RPC docs, since its automatically added by docs parsing too.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -3974,6 +3975,25 @@ public Void call(ApiCaller ac, Deserializer parms) throws AttributeAlreadyMarked
 		}
 	},
 
+	/*#
+	 * Converts attribute to nonunique - unmarks unique flag from attribute definition, and deletes all values
+	 * from a special table with unique constraint that ensures that all values remain unique.
+	 * Already nonunique attribute is skipped.
```
zlamalp, line 3981: I don't think that ``Already nonunique attribute is skipped.`` sentence is true. I think, that exception is thrown instead.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTestAbstract.java
metodej: Ok, so I broke down the for cycle that iterated by types into several tests each testing one type and also moved duplicate code to private methods used by these parts.

perun-openapi/openapi.yml
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
metodej: Fixed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
metodej: Removed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
metodej: Removed.

zlamalp (approved):

Vojtech-Sassmann (approved):

PR #2850, opened 2020-08-12
BASE: Removed unused getter/setter for cache enabled config option

zlamalp:


perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
```diff
@@ -134,14 +134,6 @@ public void setReadOnlyPerun(boolean readOnlyPerun) {
 		this.readOnlyPerun = readOnlyPerun;
 	}
 
-	public boolean isCacheEnabled() {
-		return cacheEnabled;
```
Vojtech-Sassmann, line 138: Couldn't we also remove the `cacheEnabled` field as well?

zlamalp: You are right, fixed.

PR #2855, opened 2020-08-13
Ldap attribute fix

Vojtech-Sassmann:
* The ldapc was not working properly with attribute
EduPersonPrincipalNames. It contained all IDP login even those in
invalid format.
* The ldapc was fixed so the attribute EduPersonPrincipalNames contains
only valid EEPN and EPUID logins.
* Created new ldap attribute userIdentities that contains all IDP
logins.


zlamalp (changes requested):
Because of the deployment troubles, may I ask you to modify this pull-request? For now, please comment-out the condition which filters out non-EPPN-like identities from ``eduPersonPrincipalNames`` attribute, so that both old and new attribute are filled with the same values (all identity identifiers). Please add FIXME to it and we will remove unwanted values from the eppn attribute later in another pull-request/next release. Thank you.

zlamalp (approved):

PR #2859, opened 2020-08-18
Core - RichDestinations, service denial info

Vojtech-Sassmann:
* Added a new field in RichDestination with information about
destination blocks. If the destination is blocked, the service denial
field contains a non-null ServiceDenial object. This object has
information about who and when created this denial.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichDestination.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichDestination.java
@@ -11,17 +11,23 @@ import cz.metacentrum.perun.core.api.BeansUtils;
 public class RichDestination extends Destination implements Comparable<PerunBean> {
 	private Service service;
 	private Facility facility;
+	private boolean blocked;
 
 	public RichDestination(){
 	}
 
 	public RichDestination(Destination destination, Facility facility, Service service) {
+		this(destination, facility, service, false);
+	}
+
+	public RichDestination(Destination destination, Facility facility, Service service, boolean blocked) {
 		super(destination.getId(), destination.getDestination(), destination.getType(), destination.getCreatedAt(),
 				destination.getCreatedBy(), destination.getModifiedAt(), destination.getModifiedBy(),
 				destination.getCreatedByUid(), destination.getModifiedByUid());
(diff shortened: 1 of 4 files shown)


zlamalp (commented):
Generally, its working OK. I just don't feel need for new ``ServiceDenial`` object, unless somebody is really interested in metadata (when/who blocked the service on destination). Also I don't think its necessary to have it in the auditer log (serializeToString()).
Since ``RichDestination`` already contains combination of service/facility/destination, plus currently it doesn't consider (select) facility wide blocks (which is expected), I think that boolean flag would be sufficient. Maybe discuss it with the rest of the core team.

Vojtech-Sassmann:
> Generally, its working OK. I just don't feel need for new `ServiceDenial` object, unless somebody is really interested in metadata (when/who blocked the service on destination). Also I don't think its necessary to have it in the auditer log (serializeToString()).
> Since `RichDestination` already contains combination of service/facility/destination, plus currently it doesn't consider (select) facility wide blocks (which is expected), I think that boolean flag would be sufficient. Maybe discuss it with the rest of the core team.

At first, I wanted to create just a boolean flag but I thought the metadata might be interesting for someone (it might be useful in the future). But I could remove it from the serializeToString method.

stavamichal:
Is there any final decision about the ServiceDenial object?

stavamichal:
> > Generally, its working OK. I just don't feel need for new `ServiceDenial` object, unless somebody is really interested in metadata (when/who blocked the service on destination). Also I don't think its necessary to have it in the auditer log (serializeToString()).
> > Since `RichDestination` already contains combination of service/facility/destination, plus currently it doesn't consider (select) facility wide blocks (which is expected), I think that boolean flag would be sufficient. Maybe discuss it with the rest of the core team.
> 
> At first, I wanted to create just a boolean flag but I thought the metadata might be interesting for someone (it might be useful in the future). But I could remove it from the serializeToString method.

IMHO if the problem is in the bigger bulk of data we need to send between Engine and Dispatcher, I would change the object to a boolean variable. If the only reason is that no one needs it at this moment, I would preserve code as it is. (if there is no other problem)

zlamalp (approved):

stavamichal (approved):

PR #2860, opened 2020-08-18
Core - dont remove non-required u-f attributes

Vojtech-Sassmann:
* The old implementation of deleteMember was trying to find out, which
attributes for corresponding user will not be required anymore, after the
member deletion. This is not something we want. We want to persist u-f
attributes, even if they are not required, because the user might return
to the facility and we want him to get the same attributes. Also, if the
member was at first manually removed from his groups, this would not
happen and all u-f would remain set.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -131,22 +131,6 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	public void deleteMember(PerunSession sess, Member member) throws MemberAlreadyRemovedException {
 		Vo vo = this.getMemberVo(sess, member);
 
-		User user;
-		try {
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -138,15 +138,6 @@ public void deleteMember(PerunSession sess, Member member) throws MemberAlreadyR
 			throw new ConsistencyErrorException("Removing member who doesn't have corresponding user.", e1);
 		}
 
-		List<Facility> allowedFacilities = getPerunBl().getFacilitiesManagerBl().getAllowedFacilities(sess, user);
```
balcirakpeter, line 141: It seems that the user object is no longer needed, so you can remove also its initialization

Vojtech-Sassmann: You are right, I have removed it.

balcirakpeter (approved):

stavamichal (approved):

zlamalp (approved):

PR #2861, opened 2020-08-19
Extend RESOURCEADMIN privileges

balcirakpeter:
- RESOURCEADMIN privileges were extended to get methods for groups, vos
  and facilities.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -616,6 +616,7 @@ perun_policies:
 
   getAssignedFacilities_Group_policy:
     policy_roles:
+      - RESOURCEADMIN: Vo
       - GROUPADMIN: Group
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -1618,6 +1643,7 @@ perun_policies:
 
   getGroupUnions_Group_boolean_policy:
     policy_roles:
+      - RESOURCEADMIN: Vo
```
Vojtech-Sassmann, line 1646: I think this is not necessary.

Vojtech-Sassmann (changes requested):

perun-base/src/main/resources/perun-roles.yml
balcirakpeter: removed

PR #2864, opened 2020-08-20
Fix too long processing of members in the tree of groups

stavamichal:
- there is a place in the method addIndirectMembers where information
 about further processing of members in parentGroup is counted from
 actual members of the group. This operation is slow and unefficient
 especially if members are served one by one (like in synchronization)
 - now we are testing if member was part of the group before adding and
 if yes, he need to be processed further in parentGroup
 - this operation is very efficient for one member and the only problem
 is that there must be a query for every member in the method so for
 really big groups like hundred of thousands members it can be little
 slower than the old solution


Vojtech-Sassmann (approved):

balcirakpeter (approved):

stavamichal:
This specific PR also prevents bug with membership locks. Because the lock was acquired after reading the old state of members in the group. It should be acquired before all actions in the method!

zlamalp (approved):

PR #2865, opened 2020-08-20
Core - filterMembers method optimization

Vojtech-Sassmann:
* The method filterMembersByMembershipTypeInGroup needed to be
optimized, it was running very slowly. The old implementation was slow
because of use of iterators with removal. Each removal caused the whole
array to be moved in memory and this was really slow.
* From my measurements, the new implementation should be atleast 10x
faster (depending on the ratio of DIRECT and INDIRECT relationships for
given members).

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -93,7 +93,6 @@ import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ParserException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordDeletionFailedException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordOperationTimeoutException;
-import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.RelationExistsException;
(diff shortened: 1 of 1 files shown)


stavamichal (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2401,36 +2402,61 @@ public void checkGroupExists(PerunSession sess, Group group) throws GroupNotExis
...
-		while(membersIterator.hasNext()) {
-			Member m = membersIterator.next();
-			if(m.getMembershipType().equals(MembershipType.DIRECT)) {
-				filteredMembers.add(m);
-				membersIterator.remove();
-			}
-		}
+		List<Member> directMembers = members.stream()
+				.filter(m -> m.getMembershipType().equals(MembershipType.DIRECT))
+				.distinct()
```
balcirakpeter, line 2411: I am just curious. Is this distinct necessary?

balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
Vojtech-Sassmann: The old implementation used Set, for some reason, and I wasn't sure if this could happen for the direct members as well. Probably not. But, this shouldn't cause any unwanted side effects.

zlamalp (approved):

PR #2884, opened 2020-09-03
engine: WARN state

mvocu:
Adds WARNING state both to TaskResult and Task, entered in case the send task completes with 0 return code, but not empty error output. The WARNING state is handled in the same way as DONE, but is marked differently in GUI.
The pull request contains:
  - required database changes (constraints changed)
  - core modifications to handle updated db
  - changes to the dispatcher and engine to handle the new state
  - changes to the (old) GUI to display the state

### DIFF
--- a/perun-engine/src/main/java/cz/metacentrum/perun/engine/runners/SendCollector.java
+++ b/perun-engine/src/main/java/cz/metacentrum/perun/engine/runners/SendCollector.java
@@ -8,7 +8,10 @@ import cz.metacentrum.perun.engine.scheduling.SchedulingPool;
 import cz.metacentrum.perun.engine.scheduling.impl.BlockingSendExecutorCompletionService;
 import cz.metacentrum.perun.taskslib.exceptions.TaskStoreException;
 import cz.metacentrum.perun.taskslib.model.SendTask;
+import cz.metacentrum.perun.taskslib.model.SendTask.SendTaskStatus;
+import cz.metacentrum.perun.taskslib.model.TaskResult.TaskResultStatus;
 import cz.metacentrum.perun.taskslib.model.Task;
+import cz.metacentrum.perun.taskslib.model.Task.TaskStatus;
 import cz.metacentrum.perun.taskslib.runners.impl.AbstractRunner;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -25,10 +28,11 @@ import java.util.Objects;
  * For each SendTask its outcome is reported to Dispatcher as a TaskResult.
  *
  * If any of SendTasks fails its processing (has ERROR status), whole Task is set to SENDERROR.
+ * If any of SendTasks completes its processing with WARNING status, whole Task is set to WARNING, unless there was other failure..
  * Otherwise SENDING or DONE is kept for whole Task.
  * Once all SendTasks are finished Task status is reported to Dispatcher.
  *
- * Expected Task status change is SENDING -> DONE | SENDERROR based on all SendWorkers outcome.
+ * Expected Task status change is SENDING -> DONE | WARNING | SENDERROR based on all SendWorkers outcome.
  *
  * @see BlockingSendExecutorCompletionService
  * @see SchedulingPool#createTaskResult(int, int, String, String, int, Service)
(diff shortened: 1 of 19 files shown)


perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/SendWorkerImpl.java
```diff
@@ -92,14 +94,20 @@ public SendTask call() throws TaskExecutionException {
 						task.getId(), getReturnCode(), getStdout(), getStderr());
 
 				sendTask.setStatus(ERROR);
+				// XXX: why exception? There is nothing exceptional about the situation.
```
zlamalp, line 97: I guess its historically by the way the Collectors are written. By their expectations, if Task succeeds, then Task/SendTask is returned. When it fails Exception is thrown and it contains all relevant error codes and messages, since in case of Task, there is no property for them to be passed. Like this collector has basically one positive and one negative branch, instead of three.

perun-engine/src/main/java/cz/metacentrum/perun/engine/runners/SendCollector.java
```diff
@@ -83,9 +87,15 @@ public void run() {
 				 all Destinations (whole Task). Default rescheduleTime is 3 hours * no.of destinations.
 				 */
 				task.setSendEndTime(LocalDateTime.now());
-				if (!Objects.equals(task.getStatus(), Task.TaskStatus.SENDERROR)) {
-					// keep SENDING status only if if task previously doesn't failed
+				// XXX: why is this necessary? Rewriting status with every completed destination? 
+				if (!Objects.equals(task.getStatus(), Task.TaskStatus.SENDERROR) &&
```
zlamalp, line 91: Not sure, it might be related to the possible race conditions with stuck task processing as with ``sendEndTime`` above. But value assignment for variable shouldn't cost us too much time. We can generally improve this logic in another pull-requests.

perun-engine/src/main/java/cz/metacentrum/perun/engine/runners/SendCollector.java
```diff
@@ -83,9 +87,15 @@ public void run() {
...
+						!Objects.equals(task.getStatus(), Task.TaskStatus.WARNING) &&
+						!Objects.equals(sendTask.getStatus(), SendTaskStatus.WARNING)) {
+					// keep SENDING status only if task previously hasn't failed
 					task.setStatus(Task.TaskStatus.SENDING);
+				} else if(!Objects.equals(task.getStatus(), Task.TaskStatus.SENDERROR) &&
+						sendTask.getStatus() == SendTaskStatus.WARNING) {
```
zlamalp, line 97: I would prefer ``Objects.equals(a,b)`` as in conditions above.

zlamalp (changes requested):

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/json/tasksManager/GetFacilityServicesState.java
```diff
@@ -246,6 +246,9 @@ else if (Arrays.asList("GENERATING","GENERATED","SENDING").contains(row.getStatu
 				else if (row.getStatus().equalsIgnoreCase("DONE")){
 					return "rowgreen";
 				}
+				else if (row.getStatus().equalsIgnoreCase("WARNING")){
+					return "rowgreenyellow";
```
zlamalp, line 250: We discussed color implications and we want to make it more like a semaphore, so that warning would be visible, but not so much outstanding like now. Basically styles should be as follows:

**Task**
```
WAITING = empty (white - table default)
PLANNED = rowlightyellow
GENERATING, GENERATED, SENDING = rowyellow
DONE = rowgreen
WARNING = roworange (this is added)
GENERROR, SENDERROR, ERROR = rowred
```
On some places in GUI, where "rowred" style is used for the Task, it is compared only to the ERROR, but should support other states (GENERROR/SENDERROR). Please find them and update them. Thank you.

**TaskResult**
```
DONE = rowgreen
DENIED = rowyellow (while I don't recall ever seeing TaskResult in a denied state).
WARNING = roworange (this is changed)
ERROR = rowred (this is changed from orange to red)
```
As we never ever used ``FATAL_ERROR`` from ``TaskResultStatus`` we can remove it from the object in core and here in GUI and use its color for the error state as mentioned above.

You can then remove newly added ``rowgreenyellow`` styles from CSS.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/json/tasksManager/GetFacilityState.java
```diff
@@ -187,7 +188,10 @@ else if (row.getState().equalsIgnoreCase("OPEN")){
 				else if (row.getState().equalsIgnoreCase("ERROR")){
 					return "rowred";
 				}
-		return "";
+				else if (row.getState().equalsIgnoreCase("WARNING")){
```
zlamalp, line 191: ``FacilityState`` is object of its own with states ``OK, ERROR, PROCESSING, NOT_DETERMINED`` so please remove ``OPEN`` in above "else if" and use colors green, red, yellow, white accordingly. Also ``WARNING`` shouldn't be here.

zlamalp (changes requested):

zlamalp:
I took the liberty to make the requested changes in #2901, since I want this to be merged and pushed to the production soon. Please review it. You can also comment on this pull-request, if you don't agree with some of the suggested changes.

zlamalp:
It was finished in #2901, closing this pull-request.

PR #2887, opened 2020-09-07
Removed functionality for creating sponsored users

balcirakpeter:
- We want to use new functionality for sponsored members. We don't want
  to Create sponsored users anymore.
- Therefore, all places where it could be called from outside were
  removed or forbiden.
- We do not want to remove all logic around sponsored users (old way) for now,
  because there are still users which are set this way.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
@@ -162,16 +162,6 @@ public class UsersManagerEntryIntegrationTest extends AbstractPerunIntegrationTe
 		setUpUser();
 		User owner = user;
 
-		assertTrue("User should be sponsored", sponsoredUser.isSponsoredUser());
-		usersManager.unsetSpecificUser(sess, sponsoredUser, SpecificUserType.SPONSORED);
-		User user1 = usersManager.getUserById(sess, sponsoredUser.getId());
-		assertTrue("User shouldn't be sponsored", !user1.isSponsoredUser());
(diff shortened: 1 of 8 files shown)


balcirakpeter:
Changes in Perun GUI were tested.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -156,33 +156,6 @@ public void getSpecificUsers() throws Exception {
...
-
-		assertTrue("User should be service user", serviceUser1.isServiceUser());
-		usersManager.unsetSpecificUser(sess, serviceUser1, SpecificUserType.SERVICE);
-		User user2 = usersManager.getUserById(sess, serviceUser1.getId());
-		assertTrue("User shouldn't be service user", !user2.isServiceUser());
-		usersManager.setSpecificUser(sess, user2, SpecificUserType.SERVICE, owner);
-		user2 = usersManager.getUserById(sess, user2.getId());
-		assertTrue("User should be service user again", user2.isServiceUser());
-		owners = usersManager.getUsersBySpecificUser(sess, user2);
-		assertTrue("There should be just our owner", owners.size() == 1 && owners.contains(owner));
```
Vojtech-Sassmann, line 183: I think you don't need to delete this test, just the first part.

Vojtech-Sassmann (changes requested):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
balcirakpeter: Fixed

Vojtech-Sassmann (approved):

PR #2889, opened 2020-09-08
add OpenAPI: methods and added role to getRichAdmins 

DanoFecko:
* methods to membersManager in OpenAPI
* method getRichAdmins supports role SPONSOR

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -2329,6 +2329,87 @@ components:
       in: query
       required: true
 
+    namespace:
+      name: namespace
+      description: "namespace"
+      schema:
+        type: string
+      in: query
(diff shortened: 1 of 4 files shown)


perun-openapi/openapi.yml
```diff
@@ -2329,6 +2329,87 @@ components:
...
+
+    password:
+      name: password
+      description: password
+      schema:
+        type: string
+      in: query
+      required: true
+
+    sponsor:
```
Vojtech-Sassmann, line 2380:
```suggestion
    sponsorId:
```
I would change the name so it is consistent with other id parameters.

Vojtech-Sassmann (changes requested):

PR #2890, opened 2020-09-09
Core - UUID

Vojtech-Sassmann:
* DEPLOYMENT:
  * New DB version
  * Need to enable postgres plugin with command `CREATE EXTENSION IF NOT
EXISTS "pgcrypto";`
* Added UUID to Group, User, and Resource entities.
* The UUID is of version 4 and is generated by the DB.
* I have also performed manual testing that the Postgres DB can work with the UUIDs.

### DIFF
--- a/perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
+++ b/perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
@@ -13,6 +13,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -277,6 +278,10 @@ public class AuditParser {
 		if(beanAttr==null) return null;
 		User user = new User();
 		user.setId(Integer.valueOf(beanAttr.get("id")));
+		String uuidString = beanAttr.get("uuid");
+		if (uuidString != null && !"null".equals(uuidString)) {
+			user.setUuid(UUID.fromString(uuidString));
+		}
 		user.setTitleBefore(BeansUtils.eraseEscaping(beanAttr.get("titleBefore")));
 		user.setTitleAfter(BeansUtils.eraseEscaping(beanAttr.get("titleAfter")));
 		user.setFirstName(BeansUtils.eraseEscaping(beanAttr.get("firstName")));
@@ -370,6 +375,10 @@ public class AuditParser {
 		if(beanAttr.get("parentGroupId").equals("\\0")) group.setParentGroupId(null);
 		else group.setParentGroupId(Integer.valueOf(beanAttr.get("parentGroupId")));
 		group.setId(Integer.valueOf(beanAttr.get("id")));
+		String uuidString = beanAttr.get("uuid");
+		if (uuidString != null && !"null".equals(uuidString)) {
+			group.setUuid(UUID.fromString(uuidString));
+		}
 		group.setName(BeansUtils.eraseEscaping(beanAttr.get("name")));
 		group.setDescription(BeansUtils.eraseEscaping(beanAttr.get("description")));
 		group.setVoId(Integer.valueOf(beanAttr.get("voId")));
@@ -416,6 +425,10 @@ public class AuditParser {
 		if(beanAttr==null) return null;
 		Resource resource = new Resource();
 		resource.setId(Integer.valueOf(beanAttr.get("id")));
+		String uuidString = beanAttr.get("uuid");
+		if (uuidString != null && !"null".equals(uuidString)) {
+			resource.setUuid(UUID.fromString(uuidString));
+		}
 		resource.setVoId(Integer.valueOf(beanAttr.get("voId")));
 		resource.setFacilityId(Integer.valueOf(beanAttr.get("facilityId")));
 		resource.setName(BeansUtils.eraseEscaping(beanAttr.get("name")));
@@ -715,6 +728,10 @@ public class AuditParser {
 		if(beanAttr==null) return null;
 		RichResource richResource = new RichResource();
 		richResource.setId(Integer.valueOf(beanAttr.get("id")));
+		String uuidString = beanAttr.get("uuid");
+		if (uuidString != null && !"null".equals(uuidString)) {
+			richResource.setUuid(UUID.fromString(uuidString));
+		}
 		richResource.setVoId(Integer.valueOf(beanAttr.get("voId")));
 		richResource.setFacilityId(Integer.valueOf(beanAttr.get("facilityId")));
 		richResource.setName(BeansUtils.eraseEscaping(beanAttr.get("name")));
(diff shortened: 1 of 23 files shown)


perun-core/src/main/resources/hsqldbChangelog.txt
```diff
@@ -3,6 +3,9 @@
 
 -- this update is not supported on hsql since its used only as in-memory db
 
+3.1.66
+update configurations set value='3.1.65' where property='DATABASE VERSION';
```
zlamalp, line 7: Version doesn't matches.

zlamalp (changes requested):
Not sure if we want to, but eg. select for applications in registrar might need to be updated too. It is custom implementation and doesn't retrieve UUIDs for Group and User related to the Application.

zlamalp (commented):
There will be probably deployment issue, since at VŠUP we have Postgres 9.2 and by its docs, its ``pgcrypto`` module version doesn't contain ``gen_random_uuid()`` function.

Vojtech-Sassmann:
> gen_random_uuid

It seems that the function can be used since version 9.5

Vojtech-Sassmann:
> Not sure if we want to, but eg. select for applications in registrar might need to be updated too. It is custom implementation and doesn't retrieve UUIDs for Group and User related to the Application.

I have updated the registrar as well.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -129,21 +131,28 @@ public Group createGroup(PerunSession sess, Vo vo, Group group) throws GroupExis
...
 		}
+		Group newGroup;
+		try {
+			newGroup = getGroupById(sess, newId);
+		} catch (GroupNotExistsException e) {
+			throw new InternalErrorException("Failed to read newly created group with id: " + newId, e);
+		}
+		group.setId(newId);
+		group.setUuid(newGroup.getUuid());
+		group.setVoId(newGroup.getVoId());
```
stavamichal, line 153: I don't really think this is necessary.  You get the whole new group correctly set with all expected attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
```diff
@@ -222,22 +224,28 @@ public Resource getResourceByName(PerunSession sess, Vo vo, Facility facility, S
...
+		Resource newResource;
+		try {
+			newResource = getResourceById(sess, newId);
+		} catch (ResourceNotExistsException e) {
+			throw new InternalErrorException("Failed to read newly created resource with id: " + newId, e);
+		}
+		resource.setId(newId);
+		resource.setVoId(newResource.getVoId());
+		resource.setFacilityId(newResource.getFacilityId());
+		resource.setUuid(newResource.getUuid());
```
stavamichal, line 247: Same as for Group.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -367,18 +368,25 @@ public void deleteUser(PerunSession sess, User user) throws UserAlreadyRemovedEx
...
 		}
+		User newUser;
+		try {
+			newUser = getUserById(sess, newId);
+		} catch (UserNotExistsException e) {
+			throw new InternalErrorException("Failed to read newly created user with id: " + newId, e);
+		}
+		user.setId(newId);
+		user.setUuid(newUser.getUuid());
+		return newUser;
```
stavamichal, line 389: Same as for Group.

stavamichal (changes requested):
I would add information about uuid also to serializing objects in audite_parser. Even if this class will be removed soon.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
Vojtech-Sassmann: Actually, it is. Because a lot of code expects the given object to be modified. Without it, a lot of code would have to be rewritten.

stavamichal: Ok so please just comment it.

Vojtech-Sassmann:
@stavamichal I have made changes based on your review.

perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
```diff
@@ -716,6 +729,10 @@ private static RichResource createRichResource(Map<String, String> beanAttr) {
 		if(beanAttr==null) return null;
 		RichResource richResource = new RichResource();
 		richResource.setId(Integer.valueOf(beanAttr.get("id")));
+		String uuidString = beanAttr.get("uuid");
+		if (uuidString != null && !"null".equals(uuidString)) {
+			richResource.setUuid(UUID.fromString(uuidString));
+		}
```
stavamichal, line 735: There is missing the change for RichGroup.

perun-base/src/main/java/cz/metacentrum/perun/core/api/RichResource.java
```diff
@@ -130,6 +131,7 @@ public String toString() {
 
 		return str.append(getClass().getSimpleName()).append(":["
 			).append("id='").append(getId()
+			).append("', uuid='").append(getUuid()
```
stavamichal, line 134: Setting of UUID is missing in constructor.

stavamichal (changes requested):

perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
Vojtech-Sassmann: It is not needed in rich group because it is reusing the method for creating a group where it is present.

perun-base/src/main/java/cz/metacentrum/perun/core/api/RichResource.java
Vojtech-Sassmann: Fixed.

perun-auditparser/src/main/java/cz/metacentrum/perun/auditparser/AuditParser.java
stavamichal: My bad :)

stavamichal (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
zlamalp: Regarding this, don't we want to then return ``group`` object instead of the ``newGroup`` ?

And just a note, it would be so nice to use Postgres ``INSERT .... RETURNING`` query, but its not supported by HSQLDB which we use for the tests.

zlamalp (approved):

zlamalp (changes requested):
Please rebase it (db version - changelog), so it can be merged. Thank you.

Vojtech-Sassmann:
> Please rebase it (db version - changelog), so it can be merged. Thank you.

Done.

zlamalp (approved):

PR #2902, opened 2020-09-14
New role TrustedFacilityAdmin

HejdaJakub:
* I created new role TrustedFacilityAdmin (in task named TrustedFacilityManager).
* User with this role is able to assign group(s) to resource(s) and remove group(s) from resource(s) for any facility in the specific VO. Trusted facility admin has privileges to set roles like ResourceAdmin and ResourceSelfService.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -79,6 +79,12 @@
 #
 # UNKNOWN exists, but it is not used in Perun.
 #
+# TRUSTEDFACILITYADMIN role can be assign to a user or a group in specific VO. Only VO administrator is able to set this
+# role, but only in his own VO. User with role Trusted facility admin has privileges to assign group(s) to resource(s)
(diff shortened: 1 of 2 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2429,6 +2458,7 @@ perun_policies:
       - VOADMIN: Vo
       - VOOBSERVER: Vo
       - PERUNOBSERVER:
+      - TRUSTEDFACILITYADMIN:
```
balcirakpeter, line 2461:
```suggestion
      - TRUSTEDFACILITYADMIN: Vo
```

balcirakpeter (changes requested):
Please allow for the new role also this policy:
copyGroups_Resource_Resource_policy (need to have TrustedFacility and Facility admin)

Vojtech-Sassmann (approved):
Except for the things mentioned by @balcirakpeter, I think it is ok.

balcirakpeter (approved):

PR #2904, opened 2020-09-17
Removed subgroups as child elements from getDataWithGroups

metodej:
- subgroups are no longer child elements of structure returned by getDataWithGroups, instead there is empty first child element in group part of the structure so current services does not have to be changed

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
@@ -519,7 +519,7 @@ public interface ServicesManager {
 	 * @return attributes in special structure. Facility is in the root, facility children are resources.
 	 *         Resource first child is abstract structure which children are groups.
 	 *         Resource  second chi is abstract structure which children are members.
-	 *         Group first child is abstract structure which children are groups.
+	 *         Group first child is empty structure.
 	 *         Group second chi is abstract structure which children are members.
 	 <pre>
 	 Facility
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
```diff
@@ -549,10 +549,6 @@
 	 |              |       |        +-------Attrs
 	 |              |       |        +-------ChildNodes
 	 |              |       |                   +-------()
-	 |              |       |                   |        +---ChildNodes
```
balcirakpeter, line 552: This part of JavaDoc is needed to be removed also a few lines above.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
```diff
@@ -519,7 +519,7 @@
 	 * @return attributes in special structure. Facility is in the root, facility children are resources.
 	 *         Resource first child is abstract structure which children are groups.
 	 *         Resource  second chi is abstract structure which children are members.
-	 *         Group first child is abstract structure which children are groups.
+	 *         Group first child is empty structure.
```
balcirakpeter, line 522: Please add also some short info about why is the empty structure needed.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ServicesManagerBl.java
```diff
@@ -452,10 +452,8 @@
 	 * @return attributes in special structure. Facility is in the root, facility children are resources.
 	 *         Resource first child is abstract structure which children are groups.
 	 *         Resource  second chi is abstract structure which children are members.
-	 *         Group first child is abstract structure which children are groups.
+	 *         Group first child is empty structure.
```
balcirakpeter, line 455: Same as above

balcirakpeter (changes requested):

stavamichal (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
metodej: Ok, thanks. I have also updated the javadoc of rpc.

balcirakpeter (approved):

PR #2907, opened 2020-09-23
Registrar - name widget

HejdaJakub:
* There was a problem, that we were not able to parse names and titles from displayName always correctly.
* I created a back-end part of a new feature called name widget, that allows user to put names and titles to application manually, but correctly.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -34,6 +34,7 @@ import cz.metacentrum.perun.registrar.ConsolidatorManager;
 import cz.metacentrum.perun.registrar.exceptions.*;
 import cz.metacentrum.perun.registrar.model.Identity;
 import org.apache.commons.lang3.StringUtils;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1702,6 +1703,25 @@ public Application approveApplicationInternal(PerunSession sess, int appId) thro
...
+						User user = usersManager.getUserById(registrarSession, app.getUser().getId());
+						if (item.getValue() != null && !item.getValue().isEmpty()) {
+							// user's attributes in JSON format from name widget
+							JSONObject fullName = new JSONObject(item.getValue());
+							user.setTitleBefore(fullName.getString("titleBefore"));
+							user.setFirstName(fullName.getString("firstName"));
+							user.setMiddleName(fullName.getString("middleName"));
+							user.setLastName(fullName.getString("lastName"));
+							user.setTitleAfter(fullName.getString("titleAfter"));
+							usersManager.updateUser(registrarSession, user);
```
Vojtech-Sassmann, line 1720: You should also set the updated user to the application object.

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1702,6 +1703,26 @@ public Application approveApplicationInternal(PerunSession sess, int appId) thro
 					}
 				}
 
+				// get all app items
+				List<ApplicationFormItemData> items = getApplicationDataById(registrarSession, app.getId());
```
balcirakpeter, line 1707: I do not think that the place where this code is used is right. You want to set names only when you are creating new users.
This way you could re-set them when some user is applying to another Vo.

balcirakpeter (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Vojtech-Sassmann: Actually, I have discussed this with @licehammer, and for this widget, we want it to set the names even to existing users.

HejdaJakub:
We have decided to deal with this problem only in GUI, so back-end changes are not needed.

PR #2908, opened 2020-09-23
GUI - new format of audit messages

Vojtech-Sassmann:
* We will get rid of the field message in the audit messages and
therefor the gui cannot use it anymore.
* The audit messages table has been reimplemented so it uses only the
new audit message fileds.
* Also, the GetUiMessage getter was removed from the perun-base model,
since it is not needed anymore.

### DIFF
--- a/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/model/AuditMessage.java
+++ b/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/model/AuditMessage.java
@@ -29,13 +29,26 @@ public class AuditMessage extends JavaScriptObject {
 			return this.event;
 		}-*/;
 
-	/**
-	 * Get message
-	 * @return message
-	 */
-	public final native String getMessage() /*-{
-		if (!this.uimessage) return "";
-		return this.uimessage;
+	public final native String getEventName() /*-{
+        if (!this.event) return "";
+        return this.event.name;
+    }-*/;
+
+	public final native String getAuditEventObjectsMessage() /*-{
+    	if (!this.event) return "";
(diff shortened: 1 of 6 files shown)


perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/model/AuditMessage.java
```diff
@@ -29,6 +29,29 @@ public final native String getEvent() /*-{
...
+	public final native String getEventName() /*-{
+        if (!this.event) return "";
+        return this.event.name;
+    }-*/;
+
+	public final native String getAuditEventObjectsMessage() /*-{
+    	if (!this.event) return "";
+    	var keys = Object.keys(this.event);
+    	var message = "";
+    	console.log("xxx");
```
zlamalp, line 41: Please remove logging/debug messages.

zlamalp (changes requested):
We can also remove getter ``getUIMessage()`` from the ``AuditMessage`` in core, since it only escapes the message from inside of the event for the purpose of the GUI. Then we remove corresponding ``getMessage()`` in GUI object and its usages. Also check for ``AuditMessageMixIn`` in (de)serializers, since "uimessage" property is now ignored and it won't be necessary after removal.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/model/AuditMessage.java
Vojtech-Sassmann: Done.

zlamalp (approved):

PR #2911, opened 2020-09-24
check quota path syntax acccording to regex, not URI

mvocu:
- The changes concern all quota attributes and methods validating their content.
- The original URI syntax checks allowed for transformations changing the path meaning,
  which is undesirable.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.math.BigDecimal;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.Paths;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
@@ -784,11 +785,10 @@ public class ModulesUtilsBlImpl implements ModulesUtilsBl {
 
 			//testing if path is unique
 			String canonicalPath;
-			try {
-				canonicalPath = new URI(path).normalize().getPath();
-				//path should not end on '/' (problem with some systems as GPFS)
-				if(!canonicalPath.equals("/") && canonicalPath.endsWith("/")) canonicalPath = canonicalPath.substring(0, canonicalPath.length() - 1);
-			} catch (URISyntaxException ex) {
+			canonicalPath = Paths.get(path).normalize().toString();
+			//path should not end on '/' (problem with some systems as GPFS)
+			if(!canonicalPath.equals("/") && canonicalPath.endsWith("/")) canonicalPath = canonicalPath.substring(0, canonicalPath.length() - 1);
+			if(!canonicalPath.matches("^[-a-zA-Z.0-9_/:=,]+$")) {
 				throw new WrongAttributeValueException(quotasAttribute, firstPlaceholder, secondPlaceholder, "Path '" + path + "' is not correct form.");
 			}
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
@@ -35,8 +35,6 @@ import org.slf4j.LoggerFactory;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.math.BigDecimal;
-import java.net.URI;
-import java.net.URISyntaxException;
 import java.nio.file.Paths;
 import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
```diff
@@ -784,11 +785,10 @@ public void checkIfQuotasIsInLimit(Map<String, Pair<BigDecimal, BigDecimal>> quo
...
 			String canonicalPath;
-			try {
-				canonicalPath = new URI(path).normalize().getPath();
-				//path should not end on '/' (problem with some systems as GPFS)
-				if(!canonicalPath.equals("/") && canonicalPath.endsWith("/")) canonicalPath = canonicalPath.substring(0, canonicalPath.length() - 1);
-			} catch (URISyntaxException ex) {
+			canonicalPath = Paths.get(path).normalize().toString();
+			//path should not end on '/' (problem with some systems as GPFS)
+			if(!canonicalPath.equals("/") && canonicalPath.endsWith("/")) canonicalPath = canonicalPath.substring(0, canonicalPath.length() - 1);
+			if(!canonicalPath.matches("^[-a-zA-Z.0-9_/:=,]+$")) {
```
metodej, line 791: The regex could be maybe a constant?

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ModulesUtilsEntryIntegrationTest.java
```diff
@@ -715,6 +715,15 @@ public void checkQuotasErrorInValue() throws Exception {
 		modulesUtilsBl.checkAndTransferQuotas(dataQuotasAttribute, resource, null, true);
 	}
 
+	@Test(expected=WrongAttributeValueException.class)
+	public void checkQuotasErrorInPath() throws Exception {
+		System.out.println(CLASS_NAME + "WrongAttributeValueException");
```
metodej, line 720: I think there is supposed to be name of the test in comment.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
```diff
@@ -784,11 +785,10 @@ public void checkIfQuotasIsInLimit(Map<String, Pair<BigDecimal, BigDecimal>> quo
 
 			//testing if path is unique
 			String canonicalPath;
-			try {
-				canonicalPath = new URI(path).normalize().getPath();
-				//path should not end on '/' (problem with some systems as GPFS)
-				if(!canonicalPath.equals("/") && canonicalPath.endsWith("/")) canonicalPath = canonicalPath.substring(0, canonicalPath.length() - 1);
-			} catch (URISyntaxException ex) {
+			canonicalPath = Paths.get(path).normalize().toString();
```
metodej, line 788: I think the declaration and the assignment of the variable could be on one line.

metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
metodej, line 788: Also I think that the `normalize()` function changes slashes in windows and then the regex does not match there. But I don't know if we care about users with windows...

stavamichal: We are not supporting window's path at this moment so this is ok.

balcirakpeter (approved):
Please add a short description, why you are making this change (mentioned transformations in the ticket). It may help in the future.
Otherwise ok (when you fix things according to the Metodej's review).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
balcirakpeter: You are right Metodej, but even better would be to create a static variable outside of this method and precompile the regex there so it does not have to be compiled each time this method Is called.

mvocu: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ModulesUtilsBlImpl.java
mvocu: Fixed.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ModulesUtilsEntryIntegrationTest.java
mvocu: Fixed in the original method also.

zlamalp (approved):

PR #2912, opened 2020-09-25
Script for finding groups inconsistencies

balcirakpeter:
- In perun db was created new script find_groups_inconsistencies,
  which checks wheter there are groups with some relationship where
  subgroup contains a member which does not exists in the parent group as
  Indirect.

### DIFF
--- a/perun-db/find_group_inconsistencies
+++ b/perun-db/find_group_inconsistencies
@@ -1,0 +1,124 @@
+#!/usr/bin/perl -w
+use strict;
+use warnings FATAL => 'all';
+
+use DBI;
+use POSIX qw(:errno_h);
+use Getopt::Long qw(:config no_ignore_case);
+no if $] >= 5.017011, warnings => 'experimental::smartmatch';
+
+sub help {
+	return qq{
+        Find all groups' members inconsistencies.
+        --------------------------------------
+        Available options:
+
+        --vo        | -v Vo id in which inconsistencies will be looking for (optional)
+        --user      | -u Username for Postgre DB (required)
+        --password  | -p Password for Postgre DB (required)
+        };
+}
+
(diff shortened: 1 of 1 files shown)


balcirakpeter:
I have realized that the script does not work right when there is the same member in two groups which are subgroups of the same parent. Therefore, I have to fix it.

balcirakpeter:
Fixed, ready for review.

stavamichal (approved):

perun-db/find_group_inconsistencies
```diff
@@ -0,0 +1,120 @@
...
+			}
+			push(@all_sub_groups, $sub_group);
+		}
+	}
+	my $next_indirect_members = prepare_structure_of_indirect_members(\@all_sub_groups);
+	push(@already_processed_groups, keys %$current_indirect_members);
+	check_against_next_group_level($next_indirect_members);
+}
+
+sub prepare_structure_of_indirect_members {
```
stavamichal, line 104: Please add documentation of this function.

perun-db/find_group_inconsistencies
```diff
@@ -0,0 +1,111 @@
...
+		find_inconsistencies_in_vo($selected_vo);
+	}
+}
+disconnect $dbh;
+
+#
+# Find group inconsistencies in specific vo
+#
+sub find_inconsistencies_in_vo {
+	my $vo = shift;
```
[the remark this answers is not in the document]
balcirakpeter: Fixed

perun-db/find_group_inconsistencies
balcirakpeter: Added

Vojtech-Sassmann (approved):

PR #2913, opened 2020-09-29
New method for getting sponsored members and their sponsors

HejdaJakub:
* We want to create a new method, which returns sponsored members and their sponsors for specific VO.
* The method returns list of sponsored members and for each of them list of rich users, who sponsor this sponsored member.
* The returned object also have to be filtered - the method returns to the sponsor only members, who are sponsored by this sponsor.


perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -895,6 +896,15 @@
...
 	List<RichMember> convertMembersToRichMembers(PerunSession sess, List<Member> members);
 
+	/**
+	 * Converts users to rich users.
+	 *
+	 * @param sess
+	 * @param users
+	 * @return
+	 */
+	List<RichUser> convertUsersToRichUsers(PerunSession sess, List<User> users);
```
zlamalp, line 906: We already have this method in the ``UsersManagerBl``. Please use it instead of the new implementation in MembersManagerBl.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2331,6 +2353,19 @@ public Member sponsorMember(PerunSession session, Member sponsoredMember, User s
...
+	@Override
+	public Map<Member, List<RichUser>> getSponsoredMembersAndTheirSponsors(PerunSession sess, Vo vo) {
+		Map<Member, List<RichUser>> sponsoredMembersAndTheirSponsors = new HashMap<>();
+		List<Member> sponsoredMembers = getSponsoredMembers(sess, vo);
+		List<User> sponsors;
+		for (Member sponsoredMember : sponsoredMembers) {
+			sponsors = getPerunBl().getUsersManagerBl().getSponsors(sess, sponsoredMember);
+			sponsoredMembersAndTheirSponsors.put(sponsoredMember, convertUsersToRichUsers(sess, sponsors));
+			sponsors.clear();
+		}
```
zlamalp, line 2365: How about simplifying it
```java
for (Member sponsoredMember : sponsoredMembers) {
	sponsoredMembersAndTheirSponsors.put(sponsoredMember, getPerunBl().getUsersManagerBl().convertUsersToRichUsers(sess, getPerunBl().getUsersManagerBl().getSponsors(sess, sponsoredMember)));
}
```

perun-openapi/openapi.yml
```diff
@@ -924,6 +924,14 @@ components:
           items:
             type: integer
 
+    #Map<Member, List<RichUser>>
+    MembersAndTheirSponsors:
+      type: object
+      properties:
+        member: { $ref: '#/components/schemas/Member' }
+        sponsors: { type: array, items: { $ref: '#/components/schemas/RichUser' }}
```
zlamalp, line 932: I don't think that this will work since the returned json doesn't contains properties "member" and "sponsors", but is plain object, where keys are the member objects and their values are plain lists of sponsors. Maybe see the definition above or consult it with @martin-kuba.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -290,6 +292,23 @@ public String call(ApiCaller ac, Deserializer params) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Gets map of sponsored members and lists of their sponsors for each of them.
+	 *
+	 * @param vo int id of virtual organization from which are the sponsored members chosen
+	 * @throw VoNotExistsException if given VO does not exist
+	 * @throw PrivilegeException if not VOADMIN, VOOBSERVER, PERUNOBSERVER or SPONSOR
```
zlamalp, line 300: We do not usually specify ``PrivilegeException`` in the RPC docs, since its added automatically by the docs parsing tool.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
HejdaJakub: I deleted my new method and I used this method from `UserManagerBl`

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
HejdaJakub: We decided to implemet it only in entry layer, so I changed it little bit more. I also simplified it by using stream.

perun-openapi/openapi.yml
HejdaJakub: I created new class `MemberWithSponsors` and I used it as returned object so I completely changed `openapi` specification, too.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
HejdaJakub: I deleted this `PrivilegeException` from the RPC.

zlamalp (changes requested):
Generally you fixed it correctly, but we discussed it over the Slack and this task was not well described. Please change the ``Member`` object in ``MemberWithSponsors`` to ``RichMember``. Then add ``attrNames`` param to the API, so that caller can specify which attributes to get with each RichMember. You can use existing conversion methods in MembersManager to convert Member to RichMember with attributes. Associated sponsors can be simplified to ``Users`` from ``RichUsers``.

~Regarding sponsors - ``RichUsers``, we are currently discussing it. Either simple ``Users`` will be enough, or we will need also get attributes with each RichUser. In such case you would need to add another ``attrNames`` param for sponsors and distinguish them in the API.~
~For now you can fix the first part. I'll let you know about the decision on second part.~

perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberWithSponsors.java
```diff
@@ -0,0 +1,56 @@
...
+
+	public Member getMember() {
+		return member;
+	}
+
+	public void setMember(RichMember member) {
+		this.member = member;
+	}
+
+	public List<User> getListOfSponsors() {
```
zlamalp, line 36: This should be just ``getSponsors()`` since it is a basic getter/setter for property ``List<User> sponsors``.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1271,6 +1271,17 @@
 	 */
 	List<RichMember> getSponsoredMembers(PerunSession sess, Vo vo) throws PrivilegeException, VoNotExistsException;
 
+	/**
+	 * Gets list of sponsored members with sponsors.
+	 *
+	 * @param sess session
+	 * @param vo virtual organization from which are the sponsored members chosen
```
zlamalp, line 1278: Please add ``attrNames`` param to the javadoc.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1341,6 +1341,29 @@ public RichMember sponsorMember(PerunSession session, Member sponsored, User spo
...
+		if(!AuthzResolver.authorizedInternal(sess, "getSponsoredMembersAndTheirSponsors_Vo_policy", vo)) {
+			throw new PrivilegeException(sess, "getSponsoredMembersAndTheirSponsors");
+		}
+
+		List<AttributeDefinition> attrsDef = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attrsDef.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		return membersManagerBl.convertMembersToRichMembersWithAttributes(sess, getSponsoredMembers(sess, vo), attrsDef).stream()
```
zlamalp, line 1361: You should also filter only allowed attributes based on the caller authorization, see ``membersManagerBl.filterOnlyAllowedAttributes(sess, richMembers, null, true);``

zlamalp (changes requested):

perun-base/src/main/java/cz/metacentrum/perun/core/api/MemberWithSponsors.java
HejdaJakub: Changed.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
HejdaJakub: I added this param to javadoc.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
HejdaJakub: I added this filter to the method.

zlamalp (approved):

metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1341,6 +1341,32 @@ public RichMember sponsorMember(PerunSession session, Member sponsored, User spo
...
+			throw new PrivilegeException(sess, "getSponsoredMembersAndTheirSponsors");
+		}
+
+		List<AttributeDefinition> attrsDef = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attrsDef.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		List<RichMember> richMembers = membersManagerBl.convertMembersToRichMembersWithAttributes(sess, getSponsoredMembers(sess, vo), attrsDef);
+		richMembers = membersManagerBl.filterOnlyAllowedAttributes(sess, richMembers, null, true);
```
Vojtech-Sassmann, line 1362: Is there any reason why do you call this version of the method instead of this one?
```suggestion
		richMembers = membersManagerBl.filterOnlyAllowedAttributes(sess, richMembers);
```

Vojtech-Sassmann: @zlamalp why did you suggested this version of the method?

zlamalp: Well, there is no reason to resolve attribute authorization for each RichMember, if you can be sure, they are from the same VO and with the same set of attributes. Hence you can use "context=true" and authorization is resolved only once and results are reused for the next member.

Vojtech-Sassmann: Ok, I was just confused because of the group parameter.

Vojtech-Sassmann (approved):

PR #2914, opened 2020-09-30
Add method getSponsors to OpenApi

HejdaJakub:
* Method getSponsors was missing in OpenApi, so I wrote specificaton for this method.

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -7800,6 +7800,38 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/membersManager/getSponsors/member:
+    get:
+      tags:
+        - MembersManager
+      operationId: getSponsorsByMember
+      summary: Gets sponsors for given member with optional attribute names.
+      parameters:
+        - $ref: '#/components/parameters/memberId'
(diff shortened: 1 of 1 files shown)


perun-openapi/openapi.yml
```diff
@@ -7800,6 +7800,20 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/membersManager/getSponsors:
```
zlamalp, line 7803: Do we want only this version of the method or we want to support passing ``attrNames`` - since we retrieve RichUsers exactly to get their attributes too. And what about combination with ``vo, extSourceName, extLogin`` and again optionally ``attrNames``?
You can add more versions of the same method using same pattern as in below method, where you append ``/someString`` to the URL.

zlamalp (approved):

metodej (approved):

zlamalp (approved):

perun-openapi/openapi.yml
HejdaJakub: Second version of this method was added. For optional parameter `attrNames` I used already defined paramter `attrNamesOptional`.

Vojtech-Sassmann (approved):

metodej (approved):

PR #2916, opened 2020-09-30
Added createSponsoredMembers

metodej:
- added new method to create multiple sponsored members in vo using one sponsor with generated passwords which are returned with logins in map, if there is some exception thrown during creation of members, already created members are in exception message
- also added tests and changed DummyPasswordManagerModule to return some random password

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
@@ -1166,19 +1166,26 @@ public interface MembersManager {
 	RichMember setSponsoredMember(PerunSession session, Vo vo, User userToBeSponsored, String namespace, String password, User sponsor) throws PrivilegeException, AlreadyMemberException, LoginNotExistsException, PasswordCreationFailedException, ExtendMembershipException, WrongAttributeValueException, ExtSourceNotExistsException, WrongReferenceAttributeValueException, UserNotInRoleException, PasswordStrengthException, InvalidLoginException;
 
 	/**
-	 * Creates a new sponsored Members (with random generated passwords) and its User.
+	 * Creates new sponsored Members (with random generated passwords).
+	 *
+	 * Since there may be error while creating some of the members and we cannot simply rollback the transaction and start over,
+	 * exceptions during member creation are not thrown and the returned map has this structure:
+	 *
+	 * name -> {"status" -> "OK" or "Error...", "login" -> login, "password" -> password}
+	 *
+	 * Keys are names given to this method and values are maps containing keys "status", "login" and "password".
+	 * "status" has as its value either "OK" or message of exception which was thrown during creation of the member.
+	 * "login" contains login (e.g. učo) if status is OK, "password" contains password if status is OK.
 	 *
 	 * @param session perun session
 	 * @param vo vo for members
 	 * @param namespace namespace for selecting password module
 	 * @param names a list of names
 	 * @param sponsor sponsoring user or null for the caller
-	 * @return map of logins and passwords of created members
-	 * @throws WrongAttributeAssignmentException
-	 * @throws AttributeNotExistsException
+	 * @return map of names to map of status, login and password
 	 * @throws PrivilegeException
 	 */
-	Map<String, String> createSponsoredMembers(PerunSession session, Vo vo, String namespace, List<String> names, User sponsor) throws PrivilegeException, AttributeNotExistsException, WrongAttributeAssignmentException;
+	Map<String, Map<String, String>> createSponsoredMembers(PerunSession session, Vo vo, String namespace, List<String> names, User sponsor) throws PrivilegeException;
 
 	/**
 	 * Transform non-sponsored member to sponsored one with defined sponsor
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
@@ -27,6 +27,7 @@ import cz.metacentrum.perun.core.api.exceptions.UserExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotInRoleException;
 import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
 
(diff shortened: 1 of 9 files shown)


zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1138,6 +1139,21 @@
 	 */
 	RichMember createSponsoredMember(PerunSession session, Vo vo, String namespace, Map<String, String> name, String password, User sponsor) throws PrivilegeException, AlreadyMemberException, LoginNotExistsException, PasswordCreationFailedException, ExtendMembershipException, WrongAttributeValueException, ExtSourceNotExistsException, WrongReferenceAttributeValueException, UserNotInRoleException, PasswordStrengthException, InvalidLoginException;
 
+	/**
+	 * Creates a new sponsored Members (with random generated passwords) and its User.
```
stavamichal, line 1143: This message is little confusing. You can probably skip the whole part with "and its User".

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2273,6 +2273,31 @@ public Member createSponsoredMember(PerunSession session, Vo vo, String namespac
...
+
+		for (String name : names) {
+			// generate random password
+			String password = module.generateRandomPassword(sess, null);
+			Map<String, String> mapName = new HashMap<>();
+			mapName.put("guestName", name);
+			// create sponsored member
+			RichMember richMember;
+			try {
+				richMember = getRichMember(sess, createSponsoredMember(sess, vo, namespace, mapName, password, sponsor, asyncValidation));
```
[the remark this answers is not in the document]
stavamichal: I think about something like return an error for specific member to the map to the value. Something like:
```
sponsor1Name -> password1
sponsor2Name -> ERROR: error when creating
sponsor3Name -> password3
sponsor4Name -> ERROR: error when creating
```
Then it would be on frondend to show these errors to user. What do you think @zlamalp ?
It is not standard behavior, need to be precisely defined in javadoc, but it should work much better than this.

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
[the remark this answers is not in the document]
zlamalp: We discussed this further and come up with following solution.

You will technically return ``Map<String,Map<String,String>>``.
- Keys will be names of the sponsored members, so that GUI can map each result to the originating data.
- Value will be map again, with keys ``status, login and password``. Then in status you can put "OK" or "Error - ....message...." and in login/password you put generated UČO a password respectively.
  This way we can pass all necessary information to the GUI and extend it in the future if necessary.

Please add detailed description about object structure to the Javadoc.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -156,6 +156,39 @@ public RichMember call(ApiCaller ac, Deserializer params) throws PerunException
...
+			String namespace = params.readString("namespace");
+			User sponsor = null;
+			if(params.contains("sponsor")) {
+				sponsor = ac.getUserById(params.readInt("sponsor"));
+			}
+			List<String> names;
+			if (params.contains("guestNames")) {
+				names = new ArrayList<>(params.readList("guestNames", String.class));
+			} else {
+				throw new RpcException(RpcException.Type.MISSING_VALUE, "Missing value. Either 'guestName' or ('firstName' and 'lastName') must be sent.");
```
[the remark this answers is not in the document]
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1193,6 +1193,27 @@ public RichMember createSponsoredMember(PerunSession session, Vo vo, String name
...
+		Utils.notNull(namespace, "namespace");
+		Utils.notNull(names, "names");
+
+		if (sponsor == null) {
+			//sponsor is the caller
+			sponsor = session.getPerunPrincipal().getUser();
+		} else {
+			//Authorization
+			if (!AuthzResolver.authorizedInternal(session, "createSponsoredMembers_Vo_String_List<String>_User_policy", Arrays.asList(vo, sponsor))) {
+				throw new PrivilegeException(session, "createSponsoredMember");
```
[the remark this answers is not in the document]
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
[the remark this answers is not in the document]
metodej: This looks better, thank you, I've changed the returned type based on this.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2284,15 +2284,21 @@ public Member createSponsoredMember(PerunSession session, Vo vo, String namespac
...
+				String login = perunBl.getAttributesManagerBl().getAttribute(sess, user, PasswordManagerModule.LOGIN_PREFIX + namespace).valueAsString();
+				Map<String, String> statusWithLogin = new HashMap<>();
+				statusWithLogin.put("status", "OK");
+				statusWithLogin.put("login", login);
+				statusWithLogin.put("password", password);
+				result.put(name, statusWithLogin);
 			} catch (Exception e) {
-				throw new InternalErrorException("These accounts had been created before exception was thrown: " + result, e);
+				Map<String, String> status = new HashMap<>();
+				status.put("status", e.getMessage());
```
stavamichal, line 2299: I would start this message like "ERROR: " + message.

metodej: I think that every exception message implicitly starts with "Error <code>: <message>".

zlamalp (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1215,6 +1215,27 @@ public RichMember setSponsoredMember(PerunSession session, Vo vo, User userToBeS
...
 
+	@Override
+	public Map<String, Map<String, String>> createSponsoredMembers(PerunSession session, Vo vo, String namespace, List<String> names, User sponsor) throws PrivilegeException {
+		Utils.checkPerunSession(session);
+		Utils.notNull(vo, "vo");
+		Utils.notNull(namespace, "namespace");
+		Utils.notNull(names, "names");
+
+		if (sponsor == null) {
+			//sponsor is the caller
```
stavamichal, line 1226: You should add the whole information 'sponsor is the caller, authorization is checked in Bl'.

stavamichal (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
metodej: Added.

PR #2918, opened 2020-10-06
Sponsorship ends with last sponsor

balcirakpeter:
- Method removeSponsor was changed.
  Now when a last member's sponsor is removed, also member's
  sponshorship is removed. Therefore, member expires as a normal member.
- Method deleteMember was also changed.
  Now sponsors of member are always removed. It does
  not matter if the member is sponsored or not. It is because when
  member's sponsorship is set to false, his links to sponsors preserves
  (as inactive). Therefore, it would not be possible to delete member.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
@@ -1413,7 +1413,7 @@ public class MembersManagerEntryIntegrationTest extends AbstractPerunIntegration
 		perun.getGroupsManagerBl().removeMember(sess,sponsors,sponsorMember);
 		//refresh from DB
 		sponsoredMember = perun.getMembersManagerBl().getMemberById(sess,sponsoredMember.getId());
-		assertTrue("sponsored member without sponsors should still have flag sponsored",sponsoredMember.isSponsored());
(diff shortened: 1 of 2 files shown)


Vojtech-Sassmann (approved):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -1413,7 +1413,7 @@ public void createSponsoredMember() throws Exception {
 		perun.getGroupsManagerBl().removeMember(sess,sponsors,sponsorMember);
 		//refresh from DB
 		sponsoredMember = perun.getMembersManagerBl().getMemberById(sess,sponsoredMember.getId());
-		assertTrue("sponsored member without sponsors should still have flag sponsored",sponsoredMember.isSponsored());
+		assertFalse("Member's' sponsorship should have been removed.",sponsoredMember.isSponsored());
```
metodej, line 1416: Just a typo:
```suggestion
		assertFalse("Member's sponsorship should have been removed.",sponsoredMember.isSponsored());
```

metodej (approved):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
balcirakpeter: fixed

PR #2922, opened 2020-10-07
Filter eduPersonPrincipalNames in LDAPc

metodej:
- the filter was added in previous task but was commented, its uncommented now
- it filters any values that are not in "login@scope" format


zlamalp (approved):

zlamalp:
It's ok, I just want to merge this after 3.10.0 release.

PR #2925, opened 2020-10-09
fix forgotten tasks in WAITING state

mvocu:
Adds periodic scheduling of tasks in WAITING state (the same way as DONE tasks are scheduled):
  * WAITING tasks stay in that state most often because there are no destinations or service is blocked,
  * these tasks should be scheduled when the service is unblocked or destinations added, as indicated by
    audit message,
  * the same mechanism is used for DONE tasks - flag is set indicating new data and the task is rescheduled,
  * if the task is still not eligible for scheduling, it stays in the WAITING state until switched to ERROR, when time limit expires.


balcirakpeter (approved):
Functionality seems to be ok. Just a small thing, comments on lines 95, 145 could be also updated.

zlamalp (approved):

PR #2928, opened 2020-10-12
fix messages in PrivilegeException 

mvocu:
After failed authorizedInternal() check, make sure that the text in PrivilegeException contains only method name.


balcirakpeter (changes requested):
There are some other occurrences. Most of them do not contain information about the privileged roles so it is not a big deal, but we would like to have these messages unified later.
FacilitiesManager->addHost, addHosts (2 methods)
GroupsManagerEntry-> getSubGroupsCount, getParentGroupMembers, getParentGroupRichMembers, getParentGroupRichMembersWithAttributes, 
ServicesManagerEntry-> addDestination (2 methods)
UsersManagerEntry -> getSponsors

mvocu:
Fixed.

stavamichal (approved):

balcirakpeter (approved):

PR #2933, opened 2020-10-14
Fix of creating application in VSUP registrar module

metodej:
- merging of new application with user already in perun based on birth number in VSUP module now has to be approved manually by vo admin
- at the end of approval of application the user external source is created based on RC for future merging of the user

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -3648,7 +3648,7 @@ public class RegistrarManagerImpl implements RegistrarManager {
 
 		// TODO - add authorization (and add to rpc)
 
-		if (jdbc.update("update application set apptype=? where id=?", application.getType(), application.getId()) > 0) {
+		if (jdbc.update("update application set apptype=? where id=?", application.getType().toString(), application.getId()) > 0) {
 			log.debug("Application type changed to + " + application.getType());
 		}
 	}
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -3643,6 +3643,16 @@ public class RegistrarManagerImpl implements RegistrarManager {
 
 	}
 
+	@Override
+	public void updateApplicationType(PerunSession session, Application application) {
+
+		// TODO - add authorization (and add to rpc)
+
+		if (jdbc.update("update application set apptype=? where id=?", application.getType(), application.getId()) > 0) {
+			log.debug("Application type changed to + " + application.getType());
+		}
+	}
+
 	// ------------------ MAPPERS AND SELECTS -------------------------------------
 
 	// FIXME - we are retrieving GROUP name using only "short_name" so it's not same as getGroupById()
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
@@ -10,11 +10,13 @@ import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.UserExtSource;
 import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.UserExtSourceExistsException;
+import cz.metacentrum.perun.core.api.exceptions.UserExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
@@ -62,9 +64,7 @@ public class Vsup extends DefaultRegistrarModule {
 							User user = ((PerunBl) session.getPerun()).getUsersManagerBl().getUserByExtSourceNameAndExtLogin(session, "RC", rc);
 							throw new CantBeApprovedException("Application has the same birth number " + rc + " as user " + user.getDisplayName() + " with id " + user.getId() +
 								" that is already in Perun and thus would be merged with him.", null, null, null, true);
-						} catch (CantBeApprovedException ex) {
-							throw ex;
-						} catch (Exception ex) {
+						} catch (ExtSourceNotExistsException | UserExtSourceNotExistsException | UserNotExistsException ex) {
 							log.warn("Couldn't find or set user to application {} by RC: {}", app, ex);
 						}
 					}
@@ -109,6 +109,18 @@ public class Vsup extends DefaultRegistrarModule {
 							} catch (UserExtSourceExistsException ex) {
 								// we can ignore, user will be paired with application
 								log.warn("{} already had identity associated from application {}", app.getUser(), app);
+							}
+						} else {
+							try {
+								Member member = ((PerunBl) session.getPerun()).getMembersManagerBl().getMemberByUser(session, app.getVo(), app.getUser());
+								// user is already a member, switch application type
+								if (Application.AppType.INITIAL.equals(app.getType())) {
+									app.setType(Application.AppType.EXTENSION);
+									registrar.updateApplicationType(session, app);
+									log.debug("Updating application type to EXTENSION since we matched user which is VO member!");
+								}
+							} catch (MemberNotExistsException e) {
+								// OK state
 							}
 						}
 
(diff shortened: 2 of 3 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -159,6 +194,31 @@ public Application approveApplication(PerunSession session, Application app) thr
...
+		} catch (RegistrarException e) {
+			// ignore because application's id is not null
+		}
+		for (ApplicationFormItemData item : data) {
+			if (item.getFormItem() != null &&
+				Objects.equals(AttributesManager.NS_USER_ATTR_DEF + ":birthNumber", item.getFormItem().getPerunDestinationAttribute())) {
+
+				String rc = item.getValue();
+				if (rc != null && !rc.isEmpty()) {
+					ExtSource es = perun.getExtSourcesManager().checkOrCreateExtSource(session, "RC", rc);
```
zlamalp, line 210: This would create ExtSource with name RC and type as value of birthNumber. At VŠUP type of RC ExtSource is ``INTERNAL``, we should have constant for it somewhere.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -159,6 +194,31 @@ public Application approveApplication(PerunSession session, Application app) thr
...
+					UserExtSource ues = new UserExtSource(es, app.getExtSourceLoa(), app.getCreatedBy());
```
zlamalp, line 211: You shouldn't pass identity from application into the RC UserExtSource. It should contain login from the rc, like 
```java
new UserExtSource(es, 0, rc));
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -40,55 +43,87 @@
...
+				if (item.getFormItem() != null &&
+					Objects.equals(AttributesManager.NS_USER_ATTR_DEF + ":birthNumber", item.getFormItem().getPerunDestinationAttribute())) {
+
+					// if application contains birth number, try to map to existing user
+					String rc = item.getValue();
+					if (rc != null && !rc.isEmpty()) {
+
+						try {
+							User user = ((PerunBl) session.getPerun()).getUsersManagerBl().getUserByExtSourceNameAndExtLogin(session, "RC", rc);
+							throw new CantBeApprovedException("Application has the same birth number " + rc + " as user " + user + " already in Perun and thus would be merged with him.", null, null, null, true);
```
zlamalp, line 62: Message will contain toString() from the User object, which is not most user friendly way. I would prefer printing just displayName and ID.

zlamalp (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
metodej: Fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
metodej: Fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
metodej: Replaced with `INTERVAL` constant.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -40,55 +44,88 @@
...
+					if (rc != null && !rc.isEmpty()) {
+
+						try {
+							User user = ((PerunBl) session.getPerun()).getUsersManagerBl().getUserByExtSourceNameAndExtLogin(session, "RC", rc);
+							throw new CantBeApprovedException("Application has the same birth number " + rc + " as user " + user.getDisplayName() + " with id " + user.getId() +
+								" that is already in Perun and thus would be merged with him.", null, null, null, true);
+						} catch (CantBeApprovedException ex) {
+							throw ex;
+						} catch (Exception ex) {
+							log.warn("Couldn't find or set user to application {} by RC: {}", app, ex);
```
stavamichal, line 68: Are you sure this should react to any exception? Because if there will be for example "InternalErrorException", this will say that the application can be approved if I understand it right.
@zlamalp should it work this way?

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -40,55 +44,88 @@
...
+							registrar.updateApplicationUser(session, app);
+							log.debug("Existing user found by RC for {}", app);
 						} catch (Exception ex) {
-							log.warn("Couldn't find or set user to application {} by RC: {}", application, ex);
+							log.warn("Couldn't find or set user to application {} by RC: {}", app, ex);
 						}
 
 						// associate existing user with the identity used on registration form
-						if (application.getUser() != null) {
+						if (app.getUser() != null) {
```
zlamalp, line 102: I realized, that this won't handle the cases, when found user is already a member, since after this method, normal approval process takes place. So here, for case when user is not null, please add check, whether user is already a member of VO, eg.:
```java
try {
	Member member = ((PerunBl) session.getPerun()).getMembersManagerBl().getMemberByUser(session, app.getVo(), app.getUser());
	// user is already a member, switch application type
	if (Application.AppType.INITIAL.equals(app.getType())) {
		app.setType(Application.AppType.EXTENSION);
		registrar.updateApplicationType(session, app);
		log.debug("Updating application type to EXTENSION since we matched user which is VO member!");
	}
} catch (MemberNotExistsException e) {
	// OK state
}
```
You must also implement new method in the RegistrarManager to update application type in a DB. It must be public, but there is no need to add it to the RPC, so we don't need authorization in it yet.

zlamalp (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
metodej: Ok, I've change the catched exception only to those that are thrown by `getUserByExtSourceNameAndExtLogin()`.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -159,6 +196,31 @@ public Application approveApplication(PerunSession session, Application app) thr
...
+		} catch (RegistrarException e) {
+			// ignore because application's id is not null
+		}
+		for (ApplicationFormItemData item : data) {
+			if (item.getFormItem() != null &&
+				Objects.equals(AttributesManager.NS_USER_ATTR_DEF + ":birthNumber", item.getFormItem().getPerunDestinationAttribute())) {
+
+				String rc = item.getValue();
+				if (rc != null && !rc.isEmpty()) {
+					ExtSource es = perun.getExtSourcesManager().checkOrCreateExtSource(session, ExtSourcesManager.EXTSOURCE_NAME_INTERNAL, rc);
```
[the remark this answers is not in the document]
metodej: Fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
metodej: Okay, thank you for comment, I've added this check and also the mentioned method.

stavamichal (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
```diff
@@ -40,55 +46,98 @@
...
+								ues = perunBl.getUsersManagerBl().addUserExtSource(session, app.getUser(), ues);
+								log.debug("{} associated with {} from application {}", app.getUser(), ues, app);
 							} catch (UserExtSourceExistsException ex) {
 								// we can ignore, user will be paired with application
-								log.warn("{} already had identity associated from application {}", application.getUser(), application);
+								log.warn("{} already had identity associated from application {}", app.getUser(), app);
+							}
+						} else {
+							try {
+								Member member = ((PerunBl) session.getPerun()).getMembersManagerBl().getMemberByUser(session, app.getVo(), app.getUser());
```
Vojtech-Sassmann, line 115: Isn't here the `app.getUser()` always null? This should probably be located elsewhere.

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Vsup.java
metodej: Yes, it should have been in the other branch of if, thank you.

Vojtech-Sassmann (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3643,6 +3643,16 @@ private int processApplication(PerunSession session, Application application, Li
 
 	}
 
+	@Override
+	public void updateApplicationType(PerunSession session, Application application) {
+
+		// TODO - add authorization (and add to rpc)
+
+		if (jdbc.update("update application set apptype=? where id=?", application.getType(), application.getId()) > 0) {
```
zlamalp, line 3651: Must be ``application.getType().toString()`` otherwise jdbc driver cannot guess the type: ```Can't infer the SQL type to use for an instance of cz.metacentrum.perun.registrar.model.Application$AppType.```

zlamalp (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
metodej: Ok, fixed.

zlamalp (approved):

PR #2934, opened 2020-10-19
NotfScheduler - bug fix

Vojtech-Sassmann:
* When resolving incoming expirations, to calculate expirations in a
month, we need to use some fixed number of days, instead of plus month.
If we use plusMonth, multiple days might resolve to the same day. Or,
some days might not even be resolved and because of that, some users
might not get the one month notification. E.g. 28-1-2000 + month ->
28-2-2000 but 29-1-2000 + month -> 28-2-2000 as well. And for
expirations on 31-3-2000 there will be no notification.


zlamalp:
This waits for the moment, when we check all expiration notification templates on all instances. Template must not "talk about month", but print exact membership expiration of the member. Then it can be merged.

PR #2938, opened 2020-10-22
Fixed regexes in ldapc

balcirakpeter:
- perun-ldapc.xml contained old regexes for seting/unseting roles.
  Therefore, they were rewriten to match new audit messages.
- However, these audit messages had to be also changed. The role is
  converted to upper case, because the whole logic around role seting is
  case independent. Also, some unnecesary information was removed from
  messages.


zlamalp (commented):
I tested all use-cases, it works. Thank you. Can you please also remove all old AuditEvents related to the authorization? They are in ``VosManagerEvents, GroupManagerEvents, SecurityTeamsManagerEvents and FacilityManagerEvents`` packages. They are used only as imports in ``AuthzResolverBlImpl``.

balcirakpeter:
> I tested all use-cases, it works. Thank you. Can you please also remove all old AuditEvents related to the authorization? They are in `VosManagerEvents, GroupManagerEvents, SecurityTeamsManagerEvents and FacilityManagerEvents` packages. They are used only as imports in `AuthzResolverBlImpl`.

Done

zlamalp (approved):

PR #2939, opened 2020-10-22
GetAdminGroups - allow SPONSOR role

Vojtech-Sassmann:
* We need to be able to return groups which have the SPONSOR role set
for given vo. This is a temporary fix as more complex changes will be
done in the future.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
@@ -380,8 +380,9 @@ public class VosManagerEntry implements VosManager {
 		//Role can be only supported one (TopGroupCreator, VoAdmin or VoObserver)
 		if(!role.equals(Role.TOPGROUPCREATOR) &&
 						!role.equals(Role.VOADMIN) &&
+						!role.equals(Role.SPONSOR) &&
 						!role.equals(Role.VOOBSERVER)) {
-			throw new RoleNotSupportedException("Supported roles are VoAdmin, VoObserver and TopGroupCreator.", role);
+			throw new RoleNotSupportedException("Supported roles are VoAdmin, VoObserver, Sponsor and TopGroupCreator.", role);
 		}
 
 		// Authorization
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
```diff
@@ -380,6 +380,7 @@ public void removeAdmin(PerunSession sess, Vo vo, Group group) throws PrivilegeE
 		//Role can be only supported one (TopGroupCreator, VoAdmin or VoObserver)
 		if(!role.equals(Role.TOPGROUPCREATOR) &&
 						!role.equals(Role.VOADMIN) &&
+						!role.equals(Role.SPONSOR) &&
 						!role.equals(Role.VOOBSERVER)) {
 			throw new RoleNotSupportedException("Supported roles are VoAdmin, VoObserver and TopGroupCreator.", role);
```
stavamichal, line 385: You should change the message of the exception too.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
```diff
@@ -380,6 +380,7 @@ public void removeAdmin(PerunSession sess, Vo vo, Group group) throws PrivilegeE
 		//Role can be only supported one (TopGroupCreator, VoAdmin or VoObserver)
```
stavamichal, line 380: There should be mentioned the new supported role too.

balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
Vojtech-Sassmann: Done.

stavamichal (approved):

PR #2945, opened 2020-10-26
Specific users refactorization

Vojtech-Sassmann:
* Some methods working with specific users, were adjusted so it is more
clear, which method to use when working with sponsored and service
members.
* createSpecificMember methods renamed to createServiceMember and now
can create only service members.
* createSpecificMemberSync methods removed since they were used only in
tests and could be replaced by other methods calls.
* Tests were adjusted to the new formats.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -5,7 +5,6 @@ import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AuthzResolver;
 import cz.metacentrum.perun.core.api.Candidate;
 import cz.metacentrum.perun.core.api.Sponsor;
-import cz.metacentrum.perun.core.api.Sponsorship;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.Group;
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -178,7 +177,7 @@ public Member createSpecificMember(PerunSession sess, Vo vo, Candidate candidate
 			}
 		}
 
-		return getMembersManagerBl().createSpecificMember(sess, vo, candidate, specificUserOwners, specificUserType, groups);
+		return getMembersManagerBl().createServiceMember(sess, vo, candidate, specificUserOwners, groups);
```
stavamichal, line 180: You should add exception to the RPC when someone is trying to create specific member which is not service member.

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
Vojtech-Sassmann: There is already an exception thrown on line 156.

stavamichal: My bad, didn't see that.

stavamichal (approved):

PR #2947, opened 2020-10-26
Rewrite isCesnetEligibleLastSeen to user attribute

cuadradek:
- Attribute ues:def:isCesnetEligibleLastSeen has String value representing
timestamp. When the value is set, user version of this attribute should be
updated if ues attribute’s value is more recent timestamp.
- Created module ues:def:isCesnetEligibleLastSeen and implemented its method
changedAttributeHook.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_ues_attribute_def_def_isCesnetEligibleLastSeen.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_ues_attribute_def_def_isCesnetEligibleLastSeen.java
@@ -1,0 +1,101 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.UserExtSource;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_ues_attribute_def_def_isCesnetEligibleLastSeen.java
```diff
@@ -0,0 +1,84 @@
...
+		User user;
+		try {
+			user = session.getPerunBl().getUsersManagerBl().getUserById(session, ues.getUserId());
+		} catch (UserNotExistsException ex) {
+			throw new ConsistencyErrorException(ex);
+		}
+
+		Attribute userIsCesnetEligible;
+		try {
+			userIsCesnetEligible = session.getPerunBl().getAttributesManagerBl().getAttribute(session, user, AttributesManager.NS_USER_ATTR_DEF + ":isCesnetEligibleLastSeen");
```
balcirakpeter, line 53: You can use the constant instead of composing the attribute name again.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_ues_attribute_def_def_isCesnetEligibleLastSeen.java
```diff
@@ -0,0 +1,84 @@
...
+		} catch (AttributeNotExistsException ex) {
+			log.warn("Attribute {} doesn't exist.", A_USER_DEF_IS_CESNET_ELIGIBLE_LAST_SEEN);
+			return;
+		} catch (WrongAttributeAssignmentException ex) {
+			throw new InternalErrorException(ex);
+		}
+
+		Timestamp uesTimestamp = Timestamp.valueOf(attribute.valueAsString());
```
balcirakpeter, line 61: Are we sure that the ues attribute will always contain the correct timestamp? Maybe we should create method checkAttributeSyntax which ensures that there is not some random string. What do you think @zlamalp ?

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_ues_attribute_def_def_isCesnetEligibleLastSeen.java
zlamalp: Good point @balcirakpeter, we should have syntax check method and same should be done for user version of attribute.

balcirakpeter (approved):

zlamalp (approved):

PR #2953, opened 2020-10-28
CreateSponsoredMembers - new name format

Vojtech-Sassmann:
* The createSponsoredMembers now supports a new format of the names -
[firstName];[lastName]. This way, we can decide, which part of the given
value should be used as a first name and as a last name. It also
supports the old behaviour - if there is no comma, it will try to parse
the given value as a common name.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -2250,7 +2250,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 		if (name.containsKey("guestName")) {
 			user = Utils.parseUserFromCommonName(name.get("guestName"), true);
 		} else {
-			user = Utils.createUserFromNameMap(name);
+			user = Utils.createUserFromNameMap(name, true);
 		}
 
 		User sponsoredUser = getPerunBl().getUsersManagerBl().createUser(session, user);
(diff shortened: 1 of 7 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2313,10 +2313,17 @@ public Member setSponsoredMember(PerunSession session, Vo vo, User userToBeSpons
 		PasswordManagerModule module = getPerunBl().getUsersManagerBl().getPasswordManagerModule(sess, namespace);
 
 		for (String name : names) {
-			// generate random password
-			String password = module.generateRandomPassword(sess, null);
 			Map<String, String> mapName = new HashMap<>();
-			mapName.put("guestName", name);
+			if (name.contains(";")) {
```
stavamichal, line 2317: You must also add this way of parsing to the javadoc (Bl, entry, RPC and also openapi).

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: I have updated the documentation.

stavamichal (approved):

balcirakpeter (approved):

PR #2954, opened 2020-10-28
Application auto rejection

HejdaJakub:
* I created two expiration attributes - one for group and one for VO.
* I also created methods for auto rejection of expired applications.
* Now is posible to set number of days to expiration of applications in each group and VO. If application is expired, then will be auto rejected.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -7229,6 +7229,31 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		rights.add(new AttributeRights(-1, Role.VOADMIN, Arrays.asList(ActionType.READ, ActionType.WRITE)));
 		attributes.put(attr, rights);
 
+		//urn:perun:vo:attribute-def:def:applicationExpirationRules
+		attr = new AttributeDefinition();
+		attr.setDisplayName("Application expiration rules");
+		attr.setFriendlyName("applicationExpirationRules");
+		attr.setNamespace(AttributesManager.NS_VO_ATTR_DEF);
+		attr.setDescription("Set of rules to determine date of application (to vo) expiration. If not set, application will not be auto rejected.");
+		attr.setType(LinkedHashMap.class.getName());
+		//set attribute rights (with dummy id of attribute - not known yet)
+		rights = new ArrayList<>();
+		rights.add(new AttributeRights(-1, Role.VOADMIN, Arrays.asList(ActionType.READ, ActionType.WRITE)));
+		attributes.put(attr, rights);
+
+		//urn:perun:group:attribute-def:def:applicationExpirationRules
+		attr = new AttributeDefinition();
+		attr.setDisplayName("Application expiration rules");
+		attr.setFriendlyName("applicationExpirationRules");
+		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
+		attr.setDescription("Set of rules to determine date of application (to group) expiration. If not set, application will not be auto rejected.");
+		attr.setType(LinkedHashMap.class.getName());
+		//set attribute rights (with dummy id of attribute - not known yet)
+		rights = new ArrayList<>();
+		rights.add(new AttributeRights(-1, Role.GROUPADMIN, Arrays.asList(ActionType.READ, ActionType.WRITE)));
+		rights.add(new AttributeRights(-1, Role.VOADMIN, Arrays.asList(ActionType.READ, ActionType.WRITE)));
+		attributes.put(attr, rights);
+
 		//urn:perun:member:group:attribute-def:def:groupMembershipExpiration
 		attr = new AttributeDefinition();
 		attr.setDisplayName("Group membership expiration");
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_applicationExpirationRules.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_applicationExpirationRules.java
@@ -1,0 +1,50 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.AbstractApplicationExpirationRulesModule;
+import cz.metacentrum.perun.core.implApi.modules.attributes.GroupAttributesModuleImplApi;
+
+import java.util.LinkedHashMap;
+
+/**
+ * @author Jakub Hejda <Jakub.Hejda@cesnet.cz>
+ */
+public class urn_perun_group_attribute_def_def_applicationExpirationRules extends AbstractApplicationExpirationRulesModule<Group> implements GroupAttributesModuleImplApi {
+	@Override
+	public Attribute fillAttribute(PerunSessionImpl perunSession, Group group, AttributeDefinition attribute) throws WrongAttributeAssignmentException {
+		return new Attribute(attribute);
+	}
+
+	@Override
+	public void checkAttributeSyntax(PerunSessionImpl perunSession, Group group, Attribute attribute) throws WrongAttributeValueException {
+		super.checkAttributeSyntax(perunSession, group, attribute);
+	}
+
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl perunSession, Group group, Attribute attribute) throws WrongAttributeAssignmentException, WrongReferenceAttributeValueException {
+
+	}
+
+	@Override
+	public void changedAttributeHook(PerunSessionImpl session, Group group, Attribute attribute) throws WrongReferenceAttributeValueException {
+
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
+		attr.setFriendlyName("applicationExpirationRules");
+		attr.setDisplayName("Application expiration rules");
+		attr.setType(LinkedHashMap.class.getName());
+		attr.setDescription("Rules which define when auto reject application to Group.");
+		return attr;
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_vo_attribute_def_def_applicationExpirationRules.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_vo_attribute_def_def_applicationExpirationRules.java
@@ -1,0 +1,49 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.AbstractApplicationExpirationRulesModule;
+import cz.metacentrum.perun.core.implApi.modules.attributes.VoAttributesModuleImplApi;
+
+import java.util.LinkedHashMap;
+
+/**
+ * @author Jakub Hejda <Jakub.Hejda@cesnet.cz>
+ */
+public class urn_perun_vo_attribute_def_def_applicationExpirationRules extends AbstractApplicationExpirationRulesModule<Vo> implements VoAttributesModuleImplApi {
+	@Override
+	public Attribute fillAttribute(PerunSessionImpl perunSession, Vo vo, AttributeDefinition attribute) {
+		return new Attribute(attribute);
+	}
+
+	@Override
+	public void checkAttributeSyntax(PerunSessionImpl perunSession, Vo vo, Attribute attribute) throws WrongAttributeValueException {
+		super.checkAttributeSyntax(perunSession, vo, attribute);
+	}
+
+	@Override
+	public void checkAttributeSemantics(PerunSessionImpl perunSession, Vo vo, Attribute attribute) throws WrongReferenceAttributeValueException {
+
+	}
+
+	@Override
+	public void changedAttributeHook(PerunSessionImpl session, Vo vo, Attribute attribute) {
+
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_VO_ATTR_DEF);
+		attr.setFriendlyName("applicationExpirationRules");
+		attr.setDisplayName("Application expiration rules");
+		attr.setType(LinkedHashMap.class.getName());
+		attr.setDescription("Rules which define when auto reject application to Vo.");
+		return attr;
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/AbstractApplicationExpirationRulesModule.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/AbstractApplicationExpirationRulesModule.java
@@ -1,0 +1,47 @@
+package cz.metacentrum.perun.core.implApi.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.PerunBean;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * @author Jakub Hejda <Jakub.Hejda@cesnet.cz>
+ */
+public abstract class AbstractApplicationExpirationRulesModule<T extends PerunBean> extends AttributesModuleAbstract implements AttributesModuleImplApi {
+	public static final String applicationWaitingForEmailVerificationKeyName = "emailVerification";
+	public static final String applicationIgnoredByAdminKeyName = "ignoredByAdmin";
+	private static final Pattern daysToExpirationPattern = Pattern.compile("^\\d+$");
+
+	public void checkAttributeSyntax(PerunSessionImpl sess, T entity, Attribute attribute) throws WrongAttributeValueException {
+
+		Map<String, String> attrValue = attribute.valueAsMap();
+
+		if (attrValue == null || attrValue.isEmpty()) return;
+
+		// If is not empty, so i will check if all keys and values are correct
+		for (Map.Entry<String, String> entry : attrValue.entrySet()) {
+			if (!isAllowedParameter(entry.getKey()))
+				throw new WrongAttributeValueException(attribute, "There is not allowed parameter value: " + entry.getKey());
+			Matcher daysToExpirationMatcher = daysToExpirationPattern.matcher(entry.getValue());
+			if (!daysToExpirationMatcher.find())
+				throw new WrongAttributeValueException(attribute, "There is not allowed value for parameter '" + entry.getKey() + "': " + entry.getValue());
+		}
+
+	}
+
+	/**
+	 * If parameter (key) is allowed in HashMap for applicationExpirationRules
+	 *
+	 * @param parameter String
+	 * @return true if parameter is allowed, false if is not
+	 */
+	private boolean isAllowedParameter(String parameter) {
+		if(parameter == null) return false;
+		return parameter.equals(applicationWaitingForEmailVerificationKeyName) || parameter.equals(applicationIgnoredByAdminKeyName);
+	}
+}
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
@@ -31,6 +31,7 @@ import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.MemberGroupMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
@@ -39,6 +40,7 @@ import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.impl.Auditer;
 import cz.metacentrum.perun.core.impl.Synchronizer;
 import cz.metacentrum.perun.core.impl.Utils;
+import cz.metacentrum.perun.registrar.RegistrarManager;
 import cz.metacentrum.perun.registrar.model.Application;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -80,9 +82,13 @@ public class ExpirationNotifScheduler {
 	private PerunBl perun;
 	private JdbcPerunTemplate jdbc;
 
+	private RegistrarManager registrarManager;
+
 	private final DateTimeFormatter lastAccessFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
 
 	private static final String A_VO_MEMBERSHIP_EXP_RULES = "urn:perun:vo:attribute-def:def:membershipExpirationRules";
+	private static final String A_VO_APP_EXP_RULES = "urn:perun:vo:attribute-def:def:applicationExpirationRules";
+	private static final String A_GROUP_APP_EXP_RULES = "urn:perun:group:attribute-def:def:applicationExpirationRules";
 
 	@Autowired
 	public void setDataSource(DataSource dataSource) {
@@ -101,6 +107,11 @@ public class ExpirationNotifScheduler {
 
 	public ExpirationNotifScheduler() {
 	}
+
+	public RegistrarManager getRegistrarManager() { return registrarManager; }
+
+	@Autowired
+	public void setRegistrarManager(RegistrarManager registrarManager) { this.registrarManager = registrarManager; }
 
 	/**
 	 * Constructor for unit tests
@@ -321,6 +332,20 @@ public class ExpirationNotifScheduler {
 			log.error("Synchronizer: checkMembersState, attribute name is from wrong namespace.", e);
 		}
 
+		// check applications expiration in vos
+		try {
+			voApplicationsAutoRejection(vos);
+		} catch(InternalErrorException | PerunException e){
+			log.error("Synchronizer: voApplicationsAutoRejection", e);
+		}
+
+		// check applications expiration in groups
+		try {
+			groupApplicationsAutoRejection(vos);
+		} catch (InternalErrorException | PerunException e){
+			log.error("Synchronizer: groupApplicationsAutoRejection", e);
+		}
+
 		log.debug("Processing checkMemberState() on (to be) expired members DONE!");
 
 	}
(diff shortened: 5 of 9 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+	 * @throws PerunException
+	 */
+	private void voApplicationsAutoRejection(List<Vo> vos) throws PerunException {
+		List<String> states = new ArrayList<>();
+		states.add("NEW");
+		states.add("VERIFIED");
+
+		for (Vo vo : vos) {
+			Attribute expiration = perun.getAttributesManagerBl().getAttribute(sess, vo, A_VO_APP_EXP_RULES);
+			if (expiration != null) {
```
Vojtech-Sassmann, line 395: I think the attribute will always be not null. You should instead check its value:
```suggestion
			if (expiration.getValue() != null) {
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+		states.add("VERIFIED");
+
+		List<Group> groups = new ArrayList<>();
+		for (Vo vo : vos) {
+			groups.addAll(perun.getGroupsManagerBl().getGroups(sess, vo));
+		}
+
+		for (Group group : groups) {
+			Attribute expiration = perun.getAttributesManagerBl().getAttribute(sess, group, A_GROUP_APP_EXP_RULES);
+			if (expiration != null) {
```
Vojtech-Sassmann, line 421: Same here:
```suggestion
			if (expiration.getValue() != null) {
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+	 * @param expiration attribute with number of days to application expiration
+	 * @throws PerunException
+	 */
+	private void rejectExpiredApplications (List<Application> applications, Attribute expiration) throws PerunException {
+		Map<String, String> attrValue = expiration.valueAsMap();
+		for(Application application : applications) {
+			String date = application.getCreatedAt();
+			LocalDate createdAt = LocalDate.parse(date.substring(0, 10));
+			LocalDate now = getCurrentLocalDate();
+			if (application.getState().toString().equals("NEW")) {
```
Vojtech-Sassmann, line 442: For enums, you can compare values directly:
```suggestion
			if (application.getState() == AppState.NEW)) {
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+				int expirationAppWaitingForEmail = Integer.parseInt(attrValue.get("emailVerification"));
+				if (now.minusDays(expirationAppWaitingForEmail).isAfter(createdAt)) {
+					registrarManager.rejectApplication(sess, application.getId(), "Application with id " +
```
Vojtech-Sassmann, line 445: I think there might be a problem with the localization of the message. Some applications support also the Czech language. But I am not sure how to exactly solve this. @zlamalp any ideas?

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+						application.getId() + " was auto rejected because of expiration rules set for Vo.");
+				}
+			}
+			int expirationAppIgnoredByAdmin = Integer.parseInt(attrValue.get("ignoredByAdmin"));
+			if (now.minusDays(expirationAppIgnoredByAdmin).isAfter(createdAt)) {
+				registrarManager.rejectApplication(sess ,application.getId(), "Application with id " +
+					application.getId() + " was auto rejected because of expiration rules set for Vo.");
+			}
```
Vojtech-Sassmann, line 453: I think you should put this in an `else` branch for the `if` checking the application state.

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -686,4 +757,62 @@ private AttributeDefinition setUpGroupMembershipExpirationAttribute() throws Exc
...
+	 *
+	 * @param object ApplicationFormItemWithPrefilledValue to convert
+	 * @return converted object ApplicationFormItemData
+	 */
+	private ApplicationFormItemData convertAppFormItemWithPrefValToAppFormItemData(ApplicationFormItemWithPrefilledValue object) {
+		return new ApplicationFormItemData(object.getFormItem(), object.getFormItem().getShortname(), "", "0");
+	}
+
+	/**
+	 *
```
Vojtech-Sassmann, line 772: You are missing here a comment about what the method does.

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -232,6 +239,70 @@ public void checkMembersState() throws Exception {
...
+
+		// check results
+		Application returnedApp = registrarManager.getApplicationById(session, submitApp.getId());
+		assertEquals("Application should be rejected.", returnedApp.getState(), Application.AppState.REJECTED);
+	}
+	@Test
+	public void checkGroupApplicationShouldNotBeAutoRejected() throws Exception {
+		System.out.println(CLASS_NAME + "checkGroupApplicationAutoRejection");
+
+		jdbc.update("DELETE from application_form WHERE vo_id = ?", vo.getId());
```
Vojtech-Sassmann, line 290: This deserves comment on why you have to do it.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/AbstractApplicationExpirationRulesModule.java
```diff
@@ -0,0 +1,54 @@
...
+		Map<String, String> attrValue;
+
+		//For no value is correct (it means no rules)
+		if (attribute.getValue() == null) return;
+
+		//save value to map attrValue
+		attrValue = attribute.valueAsMap();
+
+		//Same for empty HashList
+		if (attrValue.isEmpty()) return;
```
zlamalp, line 31: I believe ``attribute.valueAsMap()`` wouldn't throw a null pointer, since you can cast null to any type. You could simplify it like this:
```java
Map<String, String> attrValue = attribute.valueAsMap();

if (attrValue == null || attrValue.isEmpty()) return;
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -102,6 +106,9 @@ public void setPerun(PerunBl perun) {
 	public ExpirationNotifScheduler() {
 	}
 
+	@Autowired
+	private RegistrarManager registrarManager;
```
zlamalp, line 110: I would prefer having getter/setter and put ``@Autowired`` above the setter. Also please specify it in ``perun-registrar-lib.xml`` in bean definition.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
 	}
 
+	/**
+	 * Checks all applications for given vos and if some application is expired (according to expiration rules set by
+	 * VO Manager), then rejects this application.
+	 *
+	 * @param vos virtual organizations
+	 * @throws PerunException
+	 */
+	private void voApplicationsAutoRejection(List<Vo> vos) throws PerunException {
```
zlamalp, line 388: Generally it would be better to try/catch exceptions on each application rejection rather than the bulk of them or all vos. Like this, one exception will discard processing of all other applications, vos and groups.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
zlamalp, line 445: You seem to provide the same rejection message ending (``rejected because of expiration rules set for Vo.``) for both VO and Group use-case. You can get ``Vo`` and ``Group`` objects from the ``Application`` (group is null for vo applications) and eg. provide entity name in the message, since application ID is generally meaningless to the user, who might even forgot he registered somewhere.
You should also state for the user, if it was rejected, because admin didn't do it in the timely manner, or it was users fault, he forgot to verify his mail address.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+			LocalDate createdAt = LocalDate.parse(date.substring(0, 10));
+			LocalDate now = getCurrentLocalDate();
+			if (application.getState().toString().equals("NEW")) {
+				int expirationAppWaitingForEmail = Integer.parseInt(attrValue.get("emailVerification"));
+				if (now.minusDays(expirationAppWaitingForEmail).isAfter(createdAt)) {
+					registrarManager.rejectApplication(sess, application.getId(), "Application with id " +
+						application.getId() + " was auto rejected because of expiration rules set for Vo.");
+				}
+			}
+			int expirationAppIgnoredByAdmin = Integer.parseInt(attrValue.get("ignoredByAdmin"));
```
zlamalp, line 449: You should "continue" in the cycle after first rejection above, since code here and below might be applied to the same application object you already processed and would fail on "already rejected" kind of exception.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -7229,6 +7229,30 @@ protected void initialize() {
...
+		//urn:perun:group:attribute-def:def:applicationExpirationRules
+		attr = new AttributeDefinition();
+		attr.setDisplayName("Application expiration rules");
+		attr.setFriendlyName("applicationExpirationRules");
+		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
+		attr.setDescription("Set of rules to determine date of application (to group) expiration. If not set, application will not be auto rejected.");
+		attr.setType(LinkedHashMap.class.getName());
+		//set attribute rights (with dummy id of attribute - not known yet)
+		rights = new ArrayList<>();
+		rights.add(new AttributeRights(-1, Role.GROUPADMIN, Arrays.asList(ActionType.READ, ActionType.WRITE)));
```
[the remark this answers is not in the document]
HejdaJakub: Rights have been added

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +378,82 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+			if (now.minusDays(expirationAppIgnoredByAdmin).isAfter(createdAt)) {
```
zlamalp, line 450: We might want to compute the time against ``modified_at`` in this case, since ``created_at`` is about NEW applications, and vo manager can't even approve such applications.
Expected time for admin reaction/approval would be shortened like this by the time it took the user to verify his application.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_applicationExpirationRules.java
```diff
@@ -0,0 +1,50 @@
...
+	@Override
+	public void changedAttributeHook(PerunSessionImpl session, Group group, Attribute attribute) throws WrongReferenceAttributeValueException {
+
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
+		attr.setFriendlyName("groupApplicationAutoExpirationRules");
```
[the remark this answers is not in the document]
HejdaJakub: Changed

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_vo_attribute_def_def_applicationExpirationRules.java
```diff
@@ -0,0 +1,52 @@
...
+	@Override
+	public void changedAttributeHook(PerunSessionImpl session, Vo vo, Attribute attribute) {
+
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_VO_ATTR_DEF);
+		attr.setFriendlyName("voApplicationExpirationRules");
```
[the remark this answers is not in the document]
HejdaJakub: Changed

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -232,6 +239,70 @@ public void checkMembersState() throws Exception {
...
+		// check results
+		Application returnedApp = registrarManager.getApplicationById(session, submitApp.getId());
+		assertEquals("Application shouldn't be rejected.", returnedApp.getState(), Application.AppState.VERIFIED);
+	}
+
+	@Test
+	public void checkGroupApplicationShouldBeAutoRejected() throws Exception {
+		System.out.println(CLASS_NAME + "checkGroupApplicationAutoRejection");
+
+		jdbc.update("DELETE from application_form WHERE vo_id = ?", vo.getId());
```
zlamalp, line 272: I don't get that. This deletes all vo and group applications of that VO. Shouldn't each test have own rollbacked transaction? Why do you do this?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_vo_attribute_def_def_applicationExpirationRules.java
```diff
@@ -0,0 +1,52 @@
...
+import cz.metacentrum.perun.core.implApi.modules.attributes.VoAttributesModuleImplApi;
+
+import java.util.LinkedHashMap;
+
+/**
+ * @author Jakub Hejda <Jakub.Hejda@cesnet.cz>
+ */
+public class urn_perun_vo_attribute_def_def_applicationExpirationRules extends AbstractApplicationExpirationRulesModule<Vo> implements VoAttributesModuleImplApi {
+
+	//public static final String VO_APPLICATION_AUTO_EXPIRATION_RULES_ATTR = AttributesManager.NS_VO_ATTR_DEF + ":applicationAutoExpirationRules";
```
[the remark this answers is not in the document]
HejdaJakub: Comment has been removed

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -232,6 +239,70 @@ public void checkMembersState() throws Exception {
...
+
+		Group group = new Group();
+		group.setName("Group for apply");
+		perun.getGroupsManagerBl().createGroup(session, vo, group);
+
+		Application submitApp = setUpAndSubmitAppForPotentialAutoRejection(50, group, GROUP_APP_EXP_RULES);
+
+		ReflectionTestUtils.invokeMethod(spyScheduler, "groupApplicationsAutoRejection", perun.getVosManagerBl().getVos(session));
```
zlamalp, line 298: I would prefer only passing expected VO in the list. Otherwise it would process all VOs in the DB. We sometimes run tests against non-empty devel DB.

zlamalp (changes requested):

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
Vojtech-Sassmann: This is actually necessary because the HSQLDB has a slightly different schema. In the regular postgres.sql we have: 
```
create unique index applform_u1 on application_form (vo_id) where group_id is null;
create unique index applform_u2 on application_form (vo_id, group_id) where group_id is not null;
```
But in the test-schema.sql:
```
create unique index applform_u1 on application_form (vo_id);
create unique index applform_u2 on application_form (vo_id, group_id);
```
It seems that HSQLDB doesn't support the `where` condition. Because of that, in tests, there can be only a single application form across the whole Vo and its groups.

So, to create an application form in a group, you have to delete the form for its vo.

Vojtech-Sassmann: I couldn't figure out a better solution, so I suggested this workaround for now.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: But I think that the second parameter of attribute shouldn't depend on application state. So I think that should be possible to set up only second value of attribute and then we couldn't check it only in `else` branch.

Vojtech-Sassmann: You are right, I forget that we want this for all states.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Changed

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Changed

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
HejdaJakub: Comment has been added

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
HejdaJakub: Comment has been added (and also removed with group creating to private method )

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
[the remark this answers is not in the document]
HejdaJakub: Done

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_applicationExpirationRules.java
```diff
@@ -0,0 +1,50 @@
...
+		attr.setDisplayName("Group application auto expiration rules");
```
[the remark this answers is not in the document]
HejdaJakub: Done

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/AbstractApplicationExpirationRulesModule.java
HejdaJakub: Simplified

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Ok, I changed it like you write above

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: The rejection process is in method `rejectExpiredApplications` so I added try/catch block here. Is it what you meant?

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Yes, I forgot about it. Thanks! Now it should be ok.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Changed

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
HejdaJakub: Changed to singleton list of expected vo

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
zlamalp: Wow, didn't know that. Maybe this note should be in the test method and it should be called only for HSQLDB, since we sometime run tests against real DB, where this condition doesn't apply.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
zlamalp: Yes, thank you.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
zlamalp: You would have to resolve preferred language of the user (if it is available) or from the application form data. Otherwise we should fallback to the english. Its done for registrar mails in ``MailManagerImpl``.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Done

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +382,102 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+		}
+	}
+
+	/**
+	 * Compares creation day of application to values in expiration attribute and if find expired application, then
+	 * reject it.
+	 *
+	 * @param applications applications
+	 * @param expiration attribute with number of days to application expiration
+	 * @throws PerunException
```
zlamalp, line 438: We no longer throw PerunException.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +382,102 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+			Attribute expiration = perun.getAttributesManagerBl().getAttribute(sess, group, A_GROUP_APP_EXP_RULES);
+			if (expiration.getValue() != null) {
+				List<Application> applications = registrarManager.getApplicationsForGroup(sess, group, states);
+				rejectExpiredApplications(applications, expiration);
+			}
+		}
+	}
+
+	/**
+	 * Compares creation day of application to values in expiration attribute and if find expired application, then
```
zlamalp, line 433: We now check against last modification timestamp, not creation day of application. Only for NEW applications, modified_at and created_at should be equal.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -321,6 +332,20 @@ public void checkMembersState() {
 			log.error("Synchronizer: checkMembersState, attribute name is from wrong namespace.", e);
 		}
 
+		// check applications expiration in vos
+		try {
+			voApplicationsAutoRejection(vos);
```
zlamalp, line 337: Probably not as a part of this pull-request, but we should really implement specific methods to get only such VOs and Groups, where there are expiration rules defined and which have some unprocessed applications.

Like this we are checking too much vos/group (like on BBMRI there is cca 6000 groups, but only a few have applications).

Also I feel like we should split this. We can have own class for application expiration and trigger it separately from the member expiration logic.

zlamalp (changes requested):

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -686,4 +751,81 @@ private AttributeDefinition setUpGroupMembershipExpirationAttribute() throws Exc
...
+	 * In test DB in application_form is created unique index for vo_id without dependency on group_id, so if you
+	 * want to create application form for group, firstly you have to delete automatically created form for Vo.
+	 * Then method creates new group.
+	 *
+	 * @return created group
+	 * @throws GroupExistsException
+	 */
+	private Group deleteApplicationFormAndCreateGroup() throws GroupExistsException {
+		// delete automatically created form for Vo
+		jdbc.update("DELETE from application_form WHERE vo_id = ?", vo.getId());
```
zlamalp, line 765: I forgot to mention, that you should call this only for HSQLDB, since it doesn't apply to the Postgres DB. See ``Compatibility.isHSQLDB()``.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
Vojtech-Sassmann: I agree with both. But I am not sure we can now get attributes for entites where the attribute is set(at least I dont know about such metod?).

Also, there is a lot of things happenening in this class and it would be definitely worth splitting it up.

zlamalp: There isn't such generic method, but if we split this logic, we can simply add new specific query method to get us just what we want - like when we get only members with expiration relative to passed date in the membership expiration logic.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Javadoc has been edited

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
HejdaJakub: Also edited

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
HejdaJakub: It's true! So edited

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -357,6 +382,101 @@ private void checkVoMembersState(List<Vo> vos) throws WrongAttributeAssignmentEx
...
+	}
+
+	/**
+	 * Compares date of last modification of application to values in expiration attribute and if finds expired application, then
+	 * rejects it.
+	 *
+	 * @param applications applications
+	 * @param expiration attribute with number of days to application expiration
+	 */
+	private void rejectExpiredApplications (List<Application> applications, Attribute expiration) {
```
Vojtech-Sassmann, line 439: This doesn't have to be part of this PR, but this method would deserve to be split into multiple methods. With so many indentations, it isn't easy to navigate in it.

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -232,6 +242,62 @@ public void checkMembersState() throws Exception {
...
+
+		ReflectionTestUtils.invokeMethod(spyScheduler, "groupApplicationsAutoRejection", Collections.singletonList(vo));
+
+		// check results
+		Application returnedApp = registrarManager.getApplicationById(session, submitApp.getId());
+		assertEquals("Application should be rejected.", returnedApp.getState(), Application.AppState.REJECTED);
+	}
+	@Test
+	public void checkGroupApplicationShouldNotBeAutoRejected() throws Exception {
+		System.out.println(CLASS_NAME + "checkGroupApplicationAutoRejection");
```
Vojtech-Sassmann, line 287:
```suggestion
		System.out.println(CLASS_NAME + "checkGroupApplicationShouldNotBeAutoRejected");
```

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -232,6 +242,62 @@ public void checkMembersState() throws Exception {
...
+		ReflectionTestUtils.invokeMethod(spyScheduler, "voApplicationsAutoRejection", Collections.singletonList(vo));
+
+		// check results
+		Application returnedApp = registrarManager.getApplicationById(session, submitApp.getId());
+		assertEquals("Application shouldn't be rejected.", returnedApp.getState(), Application.AppState.VERIFIED);
+	}
+
+	@Test
+	public void checkGroupApplicationShouldBeAutoRejected() throws Exception {
+		System.out.println(CLASS_NAME + "checkGroupApplicationAutoRejection");
```
Vojtech-Sassmann, line 273:
```suggestion
		System.out.println(CLASS_NAME + "checkGroupApplicationShouldBeAutoRejected");
```

Vojtech-Sassmann (approved):

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
HejdaJakub: Done

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
HejdaJakub: Done

zlamalp (approved):

zlamalp:
I'll merge this after 3.12.0 release, since I want another pull-request to be merged before this. We will iterate on this functionality in the next sprint.

PR #2955, opened 2020-10-28
New SPONSORSHIP role

Vojtech-Sassmann:
* We needed to able in the policies to be able distinguish sponsorhips
of the principal. Therefore, a new role SPONSORSHIP is added.
* The new role SPONSORSHIP is not saved in DB, but is resolved
from the members_sponsored table.
* Now, in the refreshAuthz method, are also added records to principal
roles for this role. For each sponship of the principal there is
information about who is he sponsoring.


stavamichal (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2280,8 +2283,17 @@ public Member setSponsoredMember(PerunSession session, Vo vo, User userToBeSpons
 	                                 String password, User sponsor, LocalDate validityTo, boolean asyncValidation) throws AlreadyMemberException, ExtendMembershipException, UserNotInRoleException, WrongAttributeValueException, WrongReferenceAttributeValueException, LoginNotExistsException, PasswordCreationFailedException, InvalidLoginException, ExtSourceNotExistsException {
 		//check that sponsoring user has role SPONSOR for the VO
 		if (!getPerunBl().getVosManagerBl().isUserInRoleForVo(session, sponsor, Role.SPONSOR, vo, true)) {
-			throw new UserNotInRoleException("user " + sponsor.getId() + " is not in role SPONSOR for VO " + vo.getId());
+			if (!AuthzResolver.isAuthorizedToManageRole(session, vo, Role.SPONSOR)) {
```
balcirakpeter, line 2286: Please use method `authorizedToManageRole`. This one should be deprecated. Or even better and simpler would be to call the method `setRole` from `AuthzResolver` right away. It will throw `PrivilegeException` if the given principal cannot manage the role in the Vo.

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: I have changed it to the `authorizedToManageRole` method. I don't want to use the AuthzResolver.setRole, since it throws a large amount of unreasonable exception. Also, I don't what to call a higher level component from the bl layer.

balcirakpeter (approved):

PR #2957, opened 2020-10-28
Fix policies for sponsor methods

Vojtech-Sassmann:
* We wanted sponsors to be able call methods `removeSponsor` and
`sponsorMember`. We also don't want VoAdmins to do this, this is
job for sponsors.
* However, the policy for the removeSponsor method has to be duplicated.
One will be resolved with the sponsor and the other one with the sponored
member. Only this way we can define roles which should be relevant only
to one of them.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2471,9 +2471,15 @@ perun_policies:
     include_policies:
       - default_policy
 
-  sponsorMember_Member_User_LocalDate_policy:
+  sponsored-sponsorMember_Member_User_LocalDate_policy:
     policy_roles:
-      - VOADMIN:
+      - SPONSOR: Vo
+    include_policies:
+      - default_policy
+
+  sponsor-sponsorMember_Member_User_LocalDate_policy:
+    policy_roles:
+      - SELF: User
     include_policies:
       - default_policy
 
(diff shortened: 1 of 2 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2471,9 +2471,15 @@ perun_policies:
     include_policies:
       - default_policy
 
-  sponsorMember_Member_User_LocalDate_policy:
+  sponsored-sponsorMember_Member_User_LocalDate_policy:
```
balcirakpeter, line 2474: I do not think you want to split this policy into two. When you call method `sponsorMember` you don't have the SPONSORSHIP role with the member yet. therefore, what you want to use is this:
```
  sponsor-sponsorMember_Member_User_LocalDate_policy:
    policy_roles:
      - SPONSOR: Vo
        SELF: User
    include_policies:
      - default_policy

balcirakpeter (changes requested):

perun-base/src/main/resources/perun-roles.yml
Vojtech-Sassmann: Actually, it has to be split. But the policy is wrong.

balcirakpeter (approved):

stavamichal (approved):

PR #2958, opened 2020-10-28
Added extraction of ues attributes to extractAdditionalUserExtSources

metodej:
- the method extractAdditionalUserExtSources() now returns RichUserExtSources with attributes extracted from subjectData
- also created new type of Candidate which holds RichUserExtSources and will be used during synchronization (in later commit)
- getCandidate now returns this new CandidateSync, every current usage of method was modified to transform to the old Candidate

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/CandidateSync.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/CandidateSync.java
@@ -1,0 +1,83 @@
+package cz.metacentrum.perun.core.api;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+/**
+ * Candidate used in group synchronization.
+ * @author Metodej Klang
+ */
+public class CandidateSync extends User {
+
+	private List<RichUserExtSource> richUserExtSources;
+	private Map<String, String> attributes;
+
+	public CandidateSync() {
+	}
+
+	public CandidateSync(Candidate candidate) {
+		this.setFirstName(candidate.getFirstName());
+		this.setLastName(candidate.getLastName());
+		this.setMiddleName(candidate.getMiddleName());
+		this.setTitleAfter(candidate.getTitleAfter());
+		this.setTitleBefore(candidate.getTitleBefore());
+		this.setServiceUser(candidate.isServiceUser());
+		this.setSponsoredUser(candidate.isSponsoredUser());
+		setAttributes(candidate.getAttributes());
+		setRichUserExtSources(candidate.getUserExtSources().stream().map(extSource ->
+			new RichUserExtSource(extSource, new ArrayList<>())).collect(Collectors.toList()));
+	}
+
+	public Map<String, String> getAttributes() {
+		return attributes;
+	}
+
+	public void setAttributes(Map<String, String> attributes) {
+		this.attributes = attributes;
+	}
+
+	public List<RichUserExtSource> getRichUserExtSources() {
+		return richUserExtSources;
+	}
+
+	public void setRichUserExtSources(List<RichUserExtSource> richUserExtSources) {
+		this.richUserExtSources = richUserExtSources;
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result
+			+ ((attributes == null) ? 0 : attributes.hashCode());
+		result = prime * result
+			+ ((richUserExtSources == null) ? 0 : richUserExtSources.hashCode());
+		return result;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj) {
+			return true;
+		}
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
+		CandidateSync other = (CandidateSync) obj;
+		if (attributes == null) {
+			if (other.attributes != null) {
+				return false;
+			}
+		} else if (!attributes.equals(other.attributes)) {
+			return false;
+		}
+		if (richUserExtSources == null) {
+			return other.richUserExtSources == null;
+		} else return richUserExtSources.equals(other.richUserExtSources);
+	}
+}
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/CandidateSync.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/CandidateSync.java
@@ -1,6 +1,7 @@
 package cz.metacentrum.perun.core.api;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -11,7 +12,8 @@ import java.util.stream.Collectors;
  */
 public class CandidateSync extends User {
 
-	private List<RichUserExtSource> richUserExtSources;
+	private RichUserExtSource richUserExtSource;
+	private List<RichUserExtSource> additionalRichUserExtSources;
 	private Map<String, String> attributes;
 
 	public CandidateSync() {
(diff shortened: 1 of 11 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/CandidateSync.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Candidate used in group synchronization.
+ * @author Metodej Klang
+ */
+public class CandidateSync extends User {
+
+	private List<RichUserExtSource> richUserExtSources;
+	private Map<String, String> attributes;
+
```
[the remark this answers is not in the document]
stavamichal: Also, you will never get the correct object of a Candidate from the CandidateSync without this information.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -206,8 +212,25 @@ public static String join(Iterable<?> collection, String separator) {
 							throw new ConsistencyErrorException("Creating existing extSource: " + additionalExtSourceName);
 						}
 					}
+					UserExtSource userExtSource = new UserExtSource(additionalExtSource, additionalExtLoa, additionalExtLogin);
+
+					List<Attribute> attributes = new ArrayList<>();
+					// Set ues attribute
+					if (additionalExtLoginAndAttribute.length == 2) {
```
stavamichal, line 219: There could be more than 1 attribute after the login. You should parse all of them. So this code should probably work for length>1.

stavamichal (changes requested):

balcirakpeter (commented):
I haven't found any issues except the ones Michal pointed out. I will do the review again when you fix them.

metodej:
Thanks for review, I've fixed all aforementioned issues.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
```diff
@@ -281,11 +279,12 @@ public Candidate getCandidate(PerunSession perunSession, Map<String,String> subj
 				attributes.put(attrName, subjectData.get(attrName));
 			}
 		}
-		List<UserExtSource> additionalUserExtSources = Utils.extractAdditionalUserExtSources(perunSession, subjectData);
-		candidate.setAdditionalUserExtSources(additionalUserExtSources);
-		candidate.setAttributes(attributes);
 
-		return candidate;
+		candidateSync.setRichUserExtSource(new RichUserExtSource(userExtSource, new ArrayList<>()));
```
balcirakpeter, line 283: I am just wondering about a use case where we want to find users by some attribute from the main userExtSource. I think this that the way the code is written now, it will be not possible. What do you think @metodej and @stavamichal ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -206,8 +213,28 @@ public static String join(Iterable<?> collection, String separator) {
...
+					List<Attribute> attributes = new ArrayList<>();
+					// Set ues attribute
+					if (additionalExtLoginAndAttribute.length > 1) {
+						List<String> uesAttributesWithValues =  new LinkedList<>();
+						uesAttributesWithValues.addAll(Arrays.asList(additionalExtLoginAndAttribute));
+						uesAttributesWithValues.remove(0);
+						for (String uesAttributeWithValue : uesAttributesWithValues) {
+							String[] uesAttribute = uesAttributeWithValue.split("=");
+							try {
+								Attribute attribute = new Attribute(sess.getPerun().getAttributesManager().getAttributeDefinition(sess, uesAttribute[0]), uesAttribute[1]);
```
balcirakpeter, line 227: I think this will work only for single value attributes. @stavamichal do we want to support also multivalue attributes?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ExtSourcesManagerBlImpl.java
metodej: I am not sure where would I find the attributes for main ues... Could they be part of the subjectData like the member and user attributes?

stavamichal: At this moment there is no way how to set attributes for main extSource. It would probably need another iteration (other PR) to fix this specific problem. You can't fix it here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: @balcirakpeter Do we already have ues attribute with multi-value as logins? If yes, then we need to fix it too. Which means to change format for the multivalue attribute. For example we can use something like "|login,urn=value1;value2,urn2=value1|1" (at least if these characters are not in the login itself and they are not used for parsing the whole string yet).

stavamichal (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: If you will be changing this one, please after the change send me re-request for review. Thank you.

balcirakpeter: Yes, we have such ues attribute on eduTEAMS instances called additionalIdentifiers.

perun-base/src/main/java/cz/metacentrum/perun/core/api/CandidateSync.java
```diff
@@ -0,0 +1,107 @@
...
+			userExtSources.add(this.richUserExtSource);
+		}
+		if (this.additionalRichUserExtSources != null) {
+			userExtSources.addAll(this.additionalRichUserExtSources);
+		}
+		return Collections.unmodifiableList(userExtSources);
+	}
+
+	@Override
+	public int hashCode() {
```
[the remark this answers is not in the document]
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
metodej: Ok, so I've updated extraction to support multivalue attributes and re-request your reviews.

stavamichal: In such a case please @metodej prepare this code to be able to process multi-value attributes. Don't need to use my suggestion.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -158,8 +163,8 @@ public static String join(Iterable<?> collection, String separator) {
 	 * @return List<UserExtSource> all additional ExtSources from the subject, returned list will never contain null value
 	 * @throws InternalErrorException
 	 */
-	public static List<UserExtSource> extractAdditionalUserExtSources(PerunSession sess, Map<String, String> subjectFromExtSource) {
-		List<UserExtSource> additionalUserExtSources = new ArrayList<>();
+	public static List<RichUserExtSource> extractAdditionalUserExtSources(PerunSession sess, Map<String, String> subjectFromExtSource) {
```
stavamichal, line 166: This whole method should be refactored to smaller parts with better Javadoc. There should be info what is the expected format of additionalUserExtSource in the subject, which part is which and how it will be parsed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -206,14 +213,49 @@ public static String join(Iterable<?> collection, String separator) {
...
 		return additionalUserExtSources;
 	}
 
+	/**
+	 * Splits attribute value by ";" and returns list of string values or single string if no ";" occurres.
+	 *
+	 * @param attributeValue unsplitted attribute value
+	 * @return splitted attribute value to list of string or single string
+	 */
+	private static Object getAttributeValueFromString(String attributeValue) {
```
stavamichal, line 250: The name of this method is very similar to the one parsing attribute value from escaped string (from DB) to the object Attribute in Perun. I would change the name and add more info to the javadoc (because this method has just one purpose and can't be used for other).

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
metodej: Changed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
metodej: Ok, I've splitted the method to smaller parts and added some comments and javadocs.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -152,68 +153,163 @@ public static String join(Iterable<?> collection, String separator) {
...
+	/**
+	 * Parses attributes of additional ues, which are in format: uesAttributeName=value1,value2. These attributes are returned with their values.
+	 * Used in extractAdditionalUserExtSources to get ues attributes.
+	 *
+	 * @param sess perun session
+	 * @param login login of subject
+	 * @param additionalExtLoginAndAttribute array containing extLogin and uesAttributes
+	 * @return List<Attribute> attributes of additional ues
+	 */
+	private static List<Attribute> parseUESAttributes(PerunSession sess, String login, String[] additionalExtLoginAndAttribute) {
```
stavamichal, line 270: You can throw the part with attributes here and then take care of parsing and everything else (same as in the method above).

stavamichal (changes requested):

metodej:
Ok, I've refactored the method a little bit more and also added some tests.

stavamichal (approved):

stavamichal:
It looks much better now, thank you.

balcirakpeter (approved):

PR #2959, opened 2020-10-29
Exceptions for invalid password reset request

cuadradek:
- Method changeNonAuthzPassword() threw InternalErrorExceptions when the
password reset request link has expired or when it never existed.
- Created new exceptions PasswordResetLinkExpired and PasswordResetLinkNotValid
to distuinguish between these two cases.
- Created new method checkPasswordResetRequestIsValid() to check for validity
of the given password reset request, that should be called from GUI before
submitting a new password.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
@@ -15,6 +15,8 @@ import cz.metacentrum.perun.core.api.exceptions.PasswordCreationFailedException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordDeletionFailedException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordDoesntMatchException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordOperationTimeoutException;
+import cz.metacentrum.perun.core.api.exceptions.PasswordResetLinkExpiredException;
+import cz.metacentrum.perun.core.api.exceptions.PasswordResetLinkNotValidException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthFailedException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
@@ -798,6 +800,18 @@ public interface UsersManager {
 	void changePassword(PerunSession sess, User user, String loginNamespace, String oldPassword, String newPassword, boolean checkOldPassword)
 			throws PrivilegeException, UserNotExistsException, LoginNotExistsException, PasswordDoesntMatchException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException;
 
+	/**
+	 * Checks if the password reset request based on encrypted parameters is valid.
+	 *
+	 * @param sess
+	 * @param i encrypted user id
+	 * @param m encrypted request id
+	 * @return
+	 * @throws UserNotExistsException
+	 * @throws PasswordResetLinkExpiredException when the reset link expired
+	 * @throws PasswordResetLinkNotValidException when the reset link was already used or has never existed
+	 */
+	void checkPasswordResetRequestIsValid(PerunSession sess, String i, String m) throws UserNotExistsException, PasswordResetLinkExpiredException, PasswordResetLinkNotValidException;
 
 	/**
 	 * Changes user password in defined login-namespace using encrypted parameters.
@@ -811,9 +825,11 @@ public interface UsersManager {
 	 * @throws UserNotExistsException
 	 * @throws LoginNotExistsException
 	 * @throws PasswordChangeFailedException
+	 * @throws PasswordResetLinkNotValidException
+	 * @throws PasswordResetLinkExpiredException
 	 */
 	void changeNonAuthzPassword(PerunSession sess, String i, String m, String password, String lang)
-			throws UserNotExistsException, LoginNotExistsException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException;
+		throws UserNotExistsException, LoginNotExistsException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException, PasswordResetLinkExpiredException, PasswordResetLinkNotValidException;
 
 	/**
 	 * Reserves random password in external system. User must not exists.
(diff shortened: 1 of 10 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -798,6 +800,18 @@ void changePassword(PerunSession sess, String login, String loginNamespace, Stri
 	void changePassword(PerunSession sess, User user, String loginNamespace, String oldPassword, String newPassword, boolean checkOldPassword)
 			throws PrivilegeException, UserNotExistsException, LoginNotExistsException, PasswordDoesntMatchException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException;
 
+	/**
+	 * Checks if the password reset request based on encrypted parameters is valid.
+	 *
+	 * @param sess
+	 * @param i encrypted user id
```
balcirakpeter, line 807: Please, if it is not too much work, rename these two variables to some meaningful names in both methods `checkPasswordResetRequestIsValid` and `changeNonAuthzPassword`.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -1269,7 +1271,18 @@ public String validatePreferredEmailChange(PerunSession sess, User user, String
 	}
 
 	@Override
-	public void changeNonAuthzPassword(PerunSession sess, String i, String m, String password, String lang) throws UserNotExistsException, LoginNotExistsException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException {
+	public void checkPasswordResetRequestIsValid(PerunSession sess, String i, String m) throws UserNotExistsException, PasswordResetLinkExpiredException, PasswordResetLinkNotValidException {
```
balcirakpeter, line 1274: I think we could add authorization to this method, allowing only SELF and PERUNADMIN to perform the operation. However, I am not sure if it is needed. What do you think @zlamalp ?

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
zlamalp: They supposed to be meaningless - not to give hints in mails, and also it shortens the URL itself. Generally I would do this in new issue, if you want it. We wanted to check request presence, not to break outer API.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
zlamalp: This will be called by non-authorized user, so we can't check the caller against the user inside the request. For now it has to be without authorization.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -1918,14 +1920,17 @@ public User convertUserEmptyStringsInObjectAttributesIntoNull(User user) {
 	}
 
 	@Override
-	public void changeNonAuthzPassword(PerunSession sess, User user, String m, String password, String lang) throws LoginNotExistsException, PasswordChangeFailedException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException {
+	public void checkPasswordResetRequestIsValid(PerunSession sess, User user, String m) throws PasswordResetLinkExpiredException, PasswordResetLinkNotValidException {
+		int requestId = Integer.parseInt(Utils.cipherInput(m, true));
 
-		String requestId = Utils.cipherInput(m, true);
-		Pair<String,String> resetRequest = getUsersManagerImpl().loadPasswordResetRequest(user, Integer.parseInt(requestId));
+		usersManagerImpl.checkPasswordResetRequestIsValid(sess, user, requestId);
```
zlamalp, line 1926: You should use ``getUsersManagerImpl()`` method, so that Spring will use its proxy class and respect transaction config. It would probably worked here, since you only read, but we should be consistent in getting manager classes.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1246,30 +1248,49 @@ public void removeAllPreferredEmailChangeRequests(PerunSession sess, User user)
...
-		int validWindow = BeansUtils.getCoreConfig().getPwdresetValidationWindow();
+	public void checkPasswordResetRequestIsValid(PerunSession sess, User user, int requestId) throws PasswordResetLinkExpiredException, PasswordResetLinkNotValidException {
+		this.checkAndGetPasswordResetRequest(user, requestId);
+	}
 
+	private Pair<String, String> checkAndGetPasswordResetRequest(User user, int requestId) throws PasswordResetLinkExpiredException, PasswordResetLinkNotValidException {
 		try {
+			int numberOfRequests = jdbc.queryForInt("select count(1) from pwdreset where user_id=? and id=?", user.getId(), requestId);
+			if (numberOfRequests == 0) {
+				throw new PasswordResetLinkNotValidException("Password request " + requestId + " doesn't exist for user: " + user);
```
zlamalp, line 1259: I would omit ``user:`` from exception message, since ``user.toString()`` prints also object name. Same for the exception thrown below and at the end.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/UsersManagerImplApi.java
```diff
@@ -691,17 +693,33 @@
...
 	List<String> getPendingPreferredEmailChanges(PerunSession sess, User user);
 
+	/**
+	 * Checks if the password reset request link is valid. The request is valid, if it
+	 * was created, never used and hasn't expired yet.
+	 *
+	 * @param sess PerunSession
+	 * @param user user to check request for
+	 * @param requestId request id to check
+	 * @return
```
zlamalp, line 703: You shouldn't mention empty ``@return``.

perun-openapi/openapi.yml
```diff
@@ -7712,6 +7712,46 @@ paths:
                 candidate: { $ref: '#/components/schemas/Candidate' }
                 specificUserOwners: { type: array, items: { $ref: "#/components/schemas/User" } }
 
+  /json/usersManager/checkPasswordResetRequestIsValid:
+    get:
+      tags:
+        - UsersManager
+      operationId: checkPasswordResetRequestIsValid
+      summary: Checks if the password reset request based on ecnrypted input parameters is valid.
```
zlamalp, line 7720: Typo: ``ecnrypted`` -> ``encrypted``.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
balcirakpeter: ok

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
balcirakpeter: ok

balcirakpeter (approved):

zlamalp (approved):

PR #2967, opened 2020-11-02
Add rights for role sponsor for methods getSponsoredMembers

HejdaJakub:
* I added rights for role sponsor for methods getSponsoredMembers to perun-roles.yml
* I also added filters to this methods to entry layer.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2491,6 +2491,19 @@ perun_policies:
       - PERUNOBSERVER:
       - REGISTRAR:
       - VOADMIN: Vo
+      - SPONSOR: Vo
+    include_policies:
+      - default_policy
+
+  filter-getSponsoredMembers_Vo_User_List<String>_policy:
+    policy_roles:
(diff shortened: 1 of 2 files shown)


balcirakpeter (commented):
Functionality seems to be OK. I've got just a suggestion regarding performance. I would firstly filter given members and then I would convert them to rich members. What do you think @HejdaJakub, @Vojtech-Sassmann?

Vojtech-Sassmann:
> Functionality seems to be OK. I've got just a suggestion regarding performance. I would firstly filter given members and then I would convert them to rich members. What do you think @HejdaJakub, @Vojtech-Sassmann?

Well, it would increase performance, but I think it is not necessary. So it all depends on @HejdaJakub

Vojtech-Sassmann (approved):

HejdaJakub:
> Functionality seems to be OK. I've got just a suggestion regarding performance. I would firstly filter given members and then I would convert them to rich members. What do you think @HejdaJakub, @Vojtech-Sassmann?

OK, so I split it and now firstly filter the members and then convert them.

balcirakpeter (approved):

PR #2969, opened 2020-11-03
Methods for convertion of sponsored users

Vojtech-Sassmann:
* Created two new methods that can be used to convert sponsored users.
* Method `convertSponsoredUsers` can be used to create sponsored members
for all sponsored users who are members of the given vo. All sponsors
are kept, if they were not sponsors in the given, the role is set for
them.
* Method `convertSponsoredUsersWithNewSponsor` can also be used for the
conversion. The difference is, that this method sets a given users as a
sponsor of all members. Old sponsors will be not set.
* Once all conversions are done, we have to consider how to remove all
of the user sponsorships. This is not handled by these two methods.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/VosManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/VosManager.java
@@ -549,4 +549,25 @@ public interface VosManager {
 	 * @throws VoNotExistsException if there is no vo with given id
 	 */
 	List<BanOnVo> getBansForVo(PerunSession sess, int voId) throws PrivilegeException, VoNotExistsException;
+
+	/**
+	 * For the given vo, creates sponsored members for each sponsored user who is a member
+	 * of the given vo. Original sponsors of the users will be set to the sponsored members.
+	 *
+	 * @param sess session
+	 * @param vo vo where members will be converted
+	 */
+	void convertSponsoredUsers(PerunSession sess, Vo vo) throws PrivilegeException;
+
+
+	/**
+	 * For the given vo, creates sponsored members for each sponsored user who is a member
+	 * of the given vo. The sponsored members will be sponsored by the given user, not by its
+	 * original sponsors.
+	 *
+	 * @param sess session
+	 * @param vo vo where members will be converted
+	 * @param newSponsor user, who will be set as a sponsor to the sponsored members
+	 */
+	void convertSponsoredUsersWithNewSponsor(PerunSession sess, Vo vo, User newSponsor) throws PrivilegeException;
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
@@ -23,6 +23,7 @@ import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.VosManager;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
+import cz.metacentrum.perun.core.api.exceptions.AlreadySponsorException;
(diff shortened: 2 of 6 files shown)


balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/api/VosManager.java
```diff
@@ -549,4 +549,25 @@
 	 * @throws VoNotExistsException if there is no vo with given id
 	 */
 	List<BanOnVo> getBansForVo(PerunSession sess, int voId) throws PrivilegeException, VoNotExistsException;
+
+	/**
+	 * For the given vo, creates sponsored members for each sponsored user who is a member
+	 * of the given vo.
```
stavamichal, line 555: You should add info that all original sponors will be processed somehow too.

stavamichal: Same for other javadocs of this method in different layers.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -821,6 +825,92 @@ public boolean isMemberBanned(PerunSession sess, int memberId) {
...
+	 * If the user is not member of the given vo, it is skipped.
+	 *
+	 * @param sess session
+	 * @param user user
+	 * @param vo vo where the given user will be sponsored
+	 */
+	private void convertToSponsoredMember(PerunSession sess, User user, Vo vo) {
+		try {
+			List<User> owners = perunBl.getUsersManagerBl().getUsersBySpecificUser(sess, user);
+			Member member = perunBl.getMembersManagerBl().getMemberByUser(sess, vo, user);
```
stavamichal, line 872: I would skip the order of these two lines. If user is not member of the VO, you don't need to have his owners at all.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -821,6 +825,92 @@ public boolean isMemberBanned(PerunSession sess, int memberId) {
...
+			if (!getPerunBl().getVosManagerBl().isUserInRoleForVo(sess, sponsor, Role.SPONSOR, vo, true)) {
+				AuthzResolverBlImpl.setRole(sess, sponsor, vo, Role.SPONSOR);
+			}
+
+			// we need to refresh information and check if the member is already sponsored
+			member = perunBl.getMembersManagerBl().getMemberById(sess, member.getId());
+
+			if (member.isSponsored()) {
+				// if the member is already sponsored, just add another sponsor
+				perunBl.getMembersManagerBl().sponsorMember(sess, member, sponsor);
```
stavamichal, line 904: Converting could be done more than once on the same VO. For this reason, you should allow to skip situation when sponsor is already sponsoring user's member. In such situation, it will add missing sponsors or fill new one (if needed).

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/VosManager.java
Vojtech-Sassmann: Javadocs updated.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Vojtech-Sassmann: I have added a catch block that should solve this problem.

stavamichal (approved):

PR #2975, opened 2020-11-05
Refactorization of automatic rejection of applications

HejdaJakub:
* Created new class which now handles all functionality about application auto rejection.
* Optimalized methods check only eligible vos and groups now instead of all vos and groups in Perun.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
@@ -1,0 +1,247 @@
+package cz.metacentrum.perun.registrar.impl;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.ExtSourcesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.PerunClient;
+import cz.metacentrum.perun.core.api.PerunPrincipal;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.exceptions.GroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.PerunException;
+import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
+import cz.metacentrum.perun.core.bl.PerunBl;
+import cz.metacentrum.perun.core.bl.SearcherBl;
+import cz.metacentrum.perun.core.impl.Synchronizer;
+import cz.metacentrum.perun.registrar.RegistrarManager;
+import cz.metacentrum.perun.registrar.model.Application;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.jdbc.core.JdbcPerunTemplate;
+
+import javax.sql.DataSource;
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Class handling auto rejection of expired applications for VOs and Groups
+ *
+ * @author Jakub Hejda <Jakub.Hejda@cesnet.cz>
+ */
+public class AppAutoRejectionScheduler {
+
+	private final static Logger log = LoggerFactory.getLogger(Synchronizer.class);
+
+	private JdbcPerunTemplate jdbc;
+	private PerunSession sess;
+	private PerunBl perun;
+	private RegistrarManager registrarManager;
+	private final SearcherBl searcherBl;
+
+	private static final String A_VO_APP_EXP_RULES = "urn:perun:vo:attribute-def:def:applicationExpirationRules";
+	private static final String A_GROUP_APP_EXP_RULES = "urn:perun:group:attribute-def:def:applicationExpirationRules";
+
+	@Autowired
+	public void setDataSource(DataSource dataSource) {
+		this.jdbc = new JdbcPerunTemplate(dataSource);
+		this.jdbc.setQueryTimeout(BeansUtils.getCoreConfig().getQueryTimeout());
+	}
+
+	public PerunBl getPerun() {
+		return perun;
+	}
+
+	@Autowired
+	public void setPerun(PerunBl perun) {
+		this.perun = perun;
+	}
+
+	public RegistrarManager getRegistrarManager() { return registrarManager; }
+
+	@Autowired
+	public void setRegistrarManager(RegistrarManager registrarManager) { this.registrarManager = registrarManager; }
+
+	/**
+	 * Constructor for unit tests
+	 *
+	 * @param perun PerunBl bean
+	 */
+	public AppAutoRejectionScheduler(PerunBl perun, SearcherBl searcherBl) {
+		this.perun = perun;
+		this.searcherBl = searcherBl;
+		initialize();
+	}
+
+	public void initialize() {
+		this.sess = perun.getPerunSession(
+			new PerunPrincipal("perunRegistrar", ExtSourcesManager.EXTSOURCE_NAME_INTERNAL, ExtSourcesManager.EXTSOURCE_INTERNAL),
+			new PerunClient());
+	}
+
+	/**
+	 * Perform check on applications state and expiration attribute and reject them, if it is necessary.
+	 * Rejection is based on current date and their value of application expiration.
+	 *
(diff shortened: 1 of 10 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+
+	}
+
+	/**
+	 * Selects all vos from database, in which could be some expired applications acceptable for auto rejection.
+	 *
+	 * @return list of vos
+	 * @throws VoNotExistsException if vo not exist (it shouldn't happen)
+	 */
+	private List<Vo> getAllEligibleVos() throws VoNotExistsException {
```
Vojtech-Sassmann, line 219: I would suggest changing the return type of these methods to `List<Integer>` and moving them to the SearcherImpl component. Then you can call here the SearcherBl and map the ids to Vos.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+
+	}
+
+	/**
+	 * Selects all groups from database, in which could be some expired applications acceptable for auto rejection.
+	 *
+	 * @return list of groups
+	 * @throws GroupNotExistsException if group not exist (it shouldn't happen)
+	 */
+	private List<Group> getAllEligibleGroups() throws GroupNotExistsException {
```
Vojtech-Sassmann, line 239: Same for this one.

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+	 *
+	 * @param perun PerunBl bean
+	 */
+	public AppAutoRejectionScheduler(PerunBl perun) {
+		this.perun = perun;
+		initialize();
+	}
+
+	public void initialize() {
+		String synchronizerPrincipal = "perunSynchronizer";
```
balcirakpeter, line 78: I think this is not perunSynchronizer :)

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
```diff
@@ -0,0 +1,252 @@
...
+			if (application.getState() == Application.AppState.NEW) {
+				int expirationAppWaitingForEmail = Integer.parseInt(attrValue.get("emailVerification"));
+				if (now.minusDays(expirationAppWaitingForEmail).isAfter(modifiedAt)) {
+					String reasonForVo = "Your application to VO " + application.getVo().getName() + " was auto rejected, because you didn't verify your email address.";
+					String reasonForGroup = application.getGroup() == null ? "" : "Your application to group " + application.getGroup().getName() + " was auto rejected, because you didn't verify your email address.";
+					rejectWithReason(application, reasonForVo, reasonForGroup);
+					continue;
+				}
+			}
+			int expirationAppIgnoredByAdmin = Integer.parseInt(attrValue.get("ignoredByAdmin"));
```
balcirakpeter, line 184: I would add the whole part with "ignoreByAdmin" option to the else branch (or `else if (Application.AppState.VERIFIED)`).

The reason is this (what could happen now):
It can happen that the "ignoredByAdmin" is set to less days than "emailVerification".
Then when the application is in state NEW but the "emailVerification" did not expired yet, it will auto-reject the application with the message saying that admin did not approve the application.
However, this is not true because the application was not even in the approval process yet (it was waiting for the email verification). Therefore, it couldn't (or shouldn't) be approved.

balcirakpeter (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
HejdaJakub: But on the other hand, this is on Vo/Group admin, if he set it like this, isn't it? In the first task (https://homeproj.cesnet.cz/issues/6556) was written, that the second value of attribute ("ignoredByAdmin") shouldn't depend on state, so I think, that this possibility is desirable. Due to this I have the "continue" in the if statement, so if the number of days for both values is the same, then shouldnť happen the double try to reject the same application.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
HejdaJakub: Query moved to the SearcherImpl.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
HejdaJakub: This query also moved to the SearcherImpl.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
HejdaJakub: Changed to perunRegistrar.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
balcirakpeter: I see. I think the best way to do is to ask the task requestor :) How should we solve it @vyskocilpavel ?

vyskocilpavel: It's possible to make a decision on the combination of both attributes?

1. use-case:
* emailVerification - 14 days
* ignoredByAdmin - unset
* then reject only application in state: "Waiting for mail verification" after 14 days

2. use-case:
* emailVerification - unset
* ignoredByAdmin - 30 days
* then reject all applications after 30 days

3. use-case:
* emailVerification - 14 days
* ignoredByAdmin - 30 days
* then reject all unverified applications after 14 days and all other applications after 30 days

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
```diff
@@ -0,0 +1,246 @@
...
+	 * @param expiration attribute with number of days to application expiration
+	 */
+	private void rejectExpiredApplications (List<Application> applications, Attribute expiration) {
+		Map<String, String> attrValue = expiration.valueAsMap();
+		for(Application application : applications) {
+			String date = application.getModifiedAt();
+			LocalDate modifiedAt = LocalDate.parse(date.substring(0, 10));
+			LocalDate now = getCurrentLocalDate();
+			if (application.getState() == Application.AppState.NEW) {
+				int expirationAppWaitingForEmail = Integer.parseInt(attrValue.get("emailVerification"));
```
balcirakpeter, line 178: I think this could cause nullPointerException when the "emailVerification" key is not set. This can happen in all places where you get a value for a key from this attribute.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
balcirakpeter: I have discussed it with Pavel and the result is following. If there are both limits set, we want to separately check NEW applications and separately VERIFIED applications. So the code should be constructed something like this:

```
if (application.getState() == Application.AppState.NEW && attrValue.containsKey("emailVerification")) {
    //process this situation
} else if (attrValue.containsKey("ignoredByAdmin")) {
    //process this situation
}
```

This should cover all possible use cases. Correct me if I am wrong. I could overlook something.

balcirakpeter (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/AppAutoRejectionScheduler.java
HejdaJakub: Good point. Your implementation below should solve this problem and I also think, that it covers all use cases.

balcirakpeter (approved):

Vojtech-Sassmann (approved):

PR #2976, opened 2020-11-10
Unique sponsorship between member and user

cuadradek:
- Sponsorship between a user and a member should be unique.
- Unique constraint was added to table members_sponsored.
- Now before inserting a sponsorship to the table, it is checked if it is not
already there. If it is in the table, it is updated (reactivated) instead of
inserted.


stavamichal (changes requested):
There are missing changes in files: 
- perun-core/src/main/resources/postgresChangelog.txt
- perun-core/src/main/resources/hsqldbChagenlog.txt
You should also increase version of DB in files:
- perun-base/src/test/resources/test-schema.sql
- perun-db/postgres.sql

stavamichal (approved):

balcirakpeter (approved):

PR #2978, opened 2020-11-10
Throw AlreadySponsorException where needed

stavamichal:
- do not mask AlreadySponsorException as InternalErrorException,
 instead we can throw it in all methods where needed

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
@@ -7,12 +7,12 @@ import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.MemberGroupStatus;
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -313,8 +312,9 @@
 	 * @throws AlreadySponsoredMemberException if member was already flagged as sponsored
 	 * @throws UserNotInRoleException if sponsor has not right role in the member's VO
 	 * @throws InternalErrorException if something unexpected happened
+	 * @throws sponsoredMember is already sponsored by User and his sponsorship is still active
```
zlamalp, line 315: Missing ``AlreadySponsorException`` at the start.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
stavamichal: Should be fixed now.

zlamalp (approved):

balcirakpeter (approved):

PR #2981, opened 2020-11-11
Synchronization based on userExtSource attributes

metodej:
- added logic that tries to find correspoding group member of candidate based on ues attribute values, not just extLogins
- at the end of categorization stage of synchronization, eppn attribute is created for every candidate for future synchronizations


stavamichal:
There is a misunderstanding of this PR goal. We need to do it more general, EPPN was just an example of such a situation. We will discuss this with Metodej outside of this thread.

PR #2984, opened 2020-11-13
Method setSponsorshipForMember should validate members

stavamichal:
- when member get new sponsor, his status in VO should be set to VALID
 again if was different

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -2216,6 +2216,19 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 		getPerunBl().getAuditer().log(session, new SponsoredMemberSet(sponsoredMember));
 		getPerunBl().getAuditer().log(session, new SponsorshipEstablished(sponsoredMember, sponsor, validityTo));
 
+		//remove expiration and validate member
+		try {
+			AttributeDefinition expiration = getPerunBl().getAttributesManagerBl().getAttributeDefinition(session, EXPIRATION);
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2216,6 +2216,9 @@ public Member setSponsorshipForMember(PerunSession session, Member sponsoredMemb
 		getPerunBl().getAuditer().log(session, new SponsoredMemberSet(sponsoredMember));
 		getPerunBl().getAuditer().log(session, new SponsorshipEstablished(sponsoredMember, sponsor, validityTo));
 
+		//validate sponsored member in the VO (you can do it asynchronous) - if member was already valid, it will not change anything
+		validateMemberAsync(session, sponsoredMember);
```
zlamalp, line 2220: This is about existing member, right? What about membership expiration? Above methods do nothing with it, so the member would be switched back to expired after midnight if his expiration date is in the past.

Also I'm not sure if we don't want to have also "sync" validation for the purpose of the tests, since it creates thread outside of the test scope (rollback), but tests passed, since they test only sponsorship validity, not members validity. @martin-kuba ?

stavamichal: Right I will check the expiration too.
There is no problem to do it synchronously instead, I just wanted to save the time of sponsor here.

stavamichal: I changed that to copy logic from sponsorMember method.

zlamalp (approved):

balcirakpeter (approved):

PR #2989, opened 2020-11-17
Synchronization based on userExtSource attributes

metodej:
- added logic that tries to find correspoding group member of candidate based on ues attribute values, not just extLogins
- ues attributes are also updated (or created) during sync for every candidate's ues for future synchronizations
- also added tests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -27,6 +27,7 @@ import cz.metacentrum.perun.core.api.AuthzResolver;
 import cz.metacentrum.perun.core.api.BeansUtils;
 import cz.metacentrum.perun.core.api.Candidate;
 import cz.metacentrum.perun.core.api.CandidateGroup;
+import cz.metacentrum.perun.core.api.CandidateSync;
 import cz.metacentrum.perun.core.api.ContactGroup;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.ExtSourcesManager;
@@ -65,7 +66,6 @@ import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
 import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
-import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedFromResourceException;
 import cz.metacentrum.perun.core.api.exceptions.GroupExistsException;
@@ -1689,8 +1689,8 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 			log.debug("Group synchronization {}: using configuration extSource for membership {}, extSource for members {}", group, membersSource, membersSource.getName());
 
 			//Prepare containers for work with group members
-			List<Candidate> candidatesToAdd = new ArrayList<>();
-			Map<Candidate, RichMember> membersToUpdate = new HashMap<>();
+			List<CandidateSync> candidatesToAdd = new ArrayList<>();
+			Map<CandidateSync, RichMember> membersToUpdate = new HashMap<>();
 			List<RichMember> membersToRemove = new ArrayList<>();
 
 			//get all direct members of synchronized group (only direct, because we want to set direct membership with this group by synchronization)
@@ -1702,7 +1702,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 				//Get subjects from extSource
 				List<Map<String, String>> subjects = getSubjectsFromExtSource(sess, source, group);
 				//Convert subjects to candidates
-				List<Candidate> candidates = convertSubjectsToCandidates(sess, subjects, membersSource, source, actualGroupMembers, skippedMembers);
+				List<CandidateSync> candidates = convertSubjectsToCandidates(sess, subjects, membersSource, source, actualGroupMembers, skippedMembers);
 
 				categorizeMembersForSynchronization(sess, actualGroupMembers, candidates, candidatesToAdd, membersToUpdate, membersToRemove);
 			}
@@ -1716,7 +1716,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 			List<AttributeDefinition> attrDefs = new ArrayList<>();
 			//Update members already presented in group
-			for (Candidate candidate : membersToUpdate.keySet()) {
+			for (CandidateSync candidate : membersToUpdate.keySet()) {
 				RichMember memberToUpdate = membersToUpdate.get(candidate);
 				//Load attrDefinitions just once for first candidate
 				if (!candidate.getAttributes().isEmpty() && attrDefs.isEmpty()) {
@@ -1727,7 +1727,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 			//Add not presented candidates to group
 			Collections.sort(candidatesToAdd);
-			for (Candidate candidateToAdd : candidatesToAdd) {
+			for (CandidateSync candidateToAdd : candidatesToAdd) {
 				addMissingMemberWhileSynchronization(sess, group, candidateToAdd, overwriteUserAttributesList, mergeMemberAttributesList, skippedMembers);
 			}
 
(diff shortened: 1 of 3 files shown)


Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3030,22 +3045,104 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
...
+			//check if extSource with login from ues attribute is in the map
+			for (UserExtSource ues : mappingStructure.keySet()) {
+				if (ues.getExtSource().equals(key.asUserExtSource().getExtSource())) {
+					try {
+						Attribute attributeToCompare = perunBl.getAttributesManagerBl().getAttribute(sess, ues, attribute.getName());
+
+						List<String> attributeToCompareValue;
+						if (attributeToCompare.getType().equals(String.class.getName())) {
+							attributeToCompareValue = Arrays.asList(attribute.valueAsString());
+						} else {
```
stavamichal, line 3103: At this moment, you should throw an error if there is something different than String or ArrayList (Integer, boolean, map, etc.) I am afraid this would end with some kind of unexpected Exception.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3030,22 +3045,104 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
...
+	 */
+	private void updateAdditionalUserExtSourceAttributes(PerunSession sess, CandidateSync candidateSync) {
+		for (RichUserExtSource ues : candidateSync.getRichUserExtSources()) {
+			for (Attribute attribute : ues.getAttributes()) {
+				try {
+					Attribute attributeToAdd = perunBl.getAttributesManagerBl().getAttribute(sess, ues.asUserExtSource(), attribute.getName());
+
+					if (attributeToAdd.getValue() == null || !attributeToAdd.getValue().equals(attribute.getValue())) {
+						perunBl.getAttributesManagerBl().setAttribute(sess, ues.asUserExtSource(), attribute);
+					}
```
stavamichal, line 3137: I am afraid this is not ok. We need to respect what is already set in Perun. This means if there is one value attribute (String type), and it is already set with a different value, you need to skip this user and inform about it to the messages (variable 'skippedMembers'). If the value of the attribute is an array list, you can update the attribute by missing values.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3030,22 +3045,104 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
...
+				} catch (WrongAttributeAssignmentException | WrongAttributeValueException | WrongReferenceAttributeValueException e) {
+					log.error("Attribute {} for UserExtSource {} was not updated due to exception: {}", attribute, ues.asUserExtSource(), e);
+				} catch (AttributeNotExistsException e) {
+					try {
+						perunBl.getAttributesManagerBl().setAttribute(sess, ues.asUserExtSource(), attribute);
```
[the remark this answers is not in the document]
stavamichal: Yes, because this means that definition does not exist in Perun. In such a case you can create a new user ext source attribute definition and then set the value again.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3673,11 +3773,16 @@ private void updateMemberStatus(PerunSession sess, RichMember memberToUpdate) th
...
 		Member member;
 		try {
 			// Check if the member is already in the VO (just not in the group)
-			member = getPerunBl().getMembersManagerBl().getMemberByUserExtSources(sess, getPerunBl().getGroupsManagerBl().getVo(sess, group), candidate.getUserExtSources());
+			List<UserExtSource> list = new ArrayList<>();
+			for (RichUserExtSource richUserExtSource : candidate.getRichUserExtSources()) {
+				UserExtSource asUserExtSource = richUserExtSource.asUserExtSource();
+				list.add(asUserExtSource);
+			}
+			member = getPerunBl().getMembersManagerBl().getMemberByUserExtSources(sess, getPerunBl().getGroupsManagerBl().getVo(sess, group), list);
```
stavamichal, line 3785: What if the user can be found only by the user ext source attribute?
Like situation when login in one of the additional user ext-sources is not the same but one of the attributes is?

stavamichal (changes requested):

metodej:
Thank you for reviews, I've fixed the names of variables/methods as suggested. The synchronization should now search the whole vo (not just group) to link candidate to member (in lightweight sync). Also the update of ues attributes was fixed to create new attribute (if it's not in Perun) and also deal with attributes which had different value in Perun.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3030,22 +3045,104 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
...
+
+						List<String> attributeToCompareValue;
+						if (attributeToCompare.getType().equals(String.class.getName())) {
+							attributeToCompareValue = Arrays.asList(attribute.valueAsString());
+						} else {
+							attributeToCompareValue = attribute.valueAsList();
+						}
+
+						//calculate intersection of attribute values
+						List<String> result = attributeToCompareValue.stream().distinct().filter(loginsFromAttribute::contains).collect(Collectors.toList());
```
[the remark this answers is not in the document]
stavamichal: The second one has been used.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3673,11 +3773,16 @@ private void updateMemberStatus(PerunSession sess, RichMember memberToUpdate) th
...
+	public void addMissingMemberWhileSynchronization(PerunSession sess, Group group, CandidateSync candidate, List<String> overwriteUserAttributesList, List<String> mergeMemberAttributesList, List<String> skippedMembers) {
 		Member member;
 		try {
 			// Check if the member is already in the VO (just not in the group)
-			member = getPerunBl().getMembersManagerBl().getMemberByUserExtSources(sess, getPerunBl().getGroupsManagerBl().getVo(sess, group), candidate.getUserExtSources());
+			List<UserExtSource> list = new ArrayList<>();
+			for (RichUserExtSource richUserExtSource : candidate.getRichUserExtSources()) {
+				UserExtSource asUserExtSource = richUserExtSource.asUserExtSource();
+				list.add(asUserExtSource);
+			}
```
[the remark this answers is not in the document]
stavamichal: Old code was changed and now works differently.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3774,28 +3801,29 @@ private void updateMemberStatus(PerunSession sess, RichMember memberToUpdate) th
...
-			try {
-				updateExistingMemberWhileSynchronization(sess, group, candidate, memberToUpdate, overwriteUserAttributesList, mergeMemberAttributesList, new ArrayList<>());
-			} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
-				// if update fails, skip him
-				log.warn("Can't update member from candidate {} due to attribute value exception {}.", candidate, e);
-				skippedMembers.add("MemberEntry:[" + candidate + "] was skipped because there was problem when updating member from candidate: Exception: " + e.getName() + " => '" + e.getMessage() + "'");
-				return;
+		Member member = null;
+		// Check if the member is already in the VO (just not in the group)
+		List<RichMember> richMembers = getPerunBl().getMembersManagerBl().getRichMembers(sess, getPerunBl().getGroupsManagerBl().getVo(sess, group));
```
stavamichal, line 3806: Doing this for every member of the group seems like really expensive operation. I am afraid we need to find better way.

stavamichal (changes requested):

balcirakpeter (commented):
I've got a few comments:
- It is not possible to find members according to attributes from primary userExtSource. I am afraid it may be a problem.
-  As @stavamichal mentioned, the new implementation may be slow.
- I am wondering if it will be a problem when we use different loginSource and memberSource during the synchronization. Meaning when we fetch just logins from one source and the rest of the data from the other. What do you think?
- According to update user extSource attributes, It seems weird that we allow to update lists and throw an error for others attributes. What about use overwriteAttributeList for userExtSource attributes? Or do not update userExtSources at all. Just use them to find users by their attributes. Or just ignore the String attributes when they do not match?

However, I will definitely need to read the PR again.

stavamichal:
> I've got a few comments:
> 
>     * It is not possible to find members according to attributes from primary userExtSource. I am afraid it may be a problem.

I don't think this is really a problem, but we should dicuss this personally. It is complicated.

> 
>     * As @stavamichal mentioned, the new implementation may be slow.
> 
>     * I am wondering if it will be a problem when we use different loginSource and memberSource during the synchronization. Meaning when we fetch just logins from one source and the rest of the data from the other. What do you think?

I don't think this will be possible with just attributes. Not without bigger changes in code.

> 
>     * According to update user extSource attributes, It seems weird that we allow to update lists and throw an error for others attributes. What about use overwriteAttributeList for userExtSource attributes? Or do not update userExtSources at all. Just use them to find users by their attributes. Or just ignore the String attributes when they do not match?

This is a right question I already tried to ask myself. You definitely want to save attributes if they are missing. But you probably don't want to change user's UES attributes if he get them from IDP, proxy etc. So what we can do is to save them only if they are missing.

> 
> 
> However, I will definitely need to read the PR again.

stavamichal:
For this PR we need to discuss if changes in it are not slowing existing synchronization and if is clear how does it work.

balcirakpeter:
Closing and putting on the graveyard. We do not plan to finish this shortly.

PR #2993, opened 2020-11-22
Fix roles management

balcirakpeter:
- getAdminGroups and getRichAdmins use only simple perun bean consisting
  only from its name and id. However, to properly resolve authorization,
  we need also information about connected objects like vo id for group etc.
- Second issue is that methods setRole and unsetRole are composing
  PerunBeans only from information recieved from input. Therefore, someone
  can fake information about the connected objects like fake vo id for
  group. This way it would be possible to setRole on object without proper
  rights.
- Solution for these issues was to fetch the real object from database
  according to bean name and its id. This way the final object will have
  correct all its values. For this purpose were implemented methods
  fetchPerunBean and fetchPerunBeans. They suppport only objects which
  can be managed for now. It will be extended later. These methods are
  now used in RPC methods setRole, unsetRole, getAdminGroups and
  getRichAdmins.
- While fixing these problems, we have realized that it is not possible
  to call set and unset role methods without complementary object.
  However the underlying logic allows to operate with null complementary
  object. Therefore, this option was added to these RPC methods.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
@@ -11,6 +11,7 @@ import cz.metacentrum.perun.cabinet.model.Category;
 import cz.metacentrum.perun.cabinet.model.Publication;
 import cz.metacentrum.perun.cabinet.model.PublicationSystem;
 import cz.metacentrum.perun.cabinet.model.Thanks;
+import cz.metacentrum.perun.core.api.PerunBean;
 import cz.metacentrum.perun.core.api.PerunClient;
 import cz.metacentrum.perun.core.api.TasksManager;
 import cz.metacentrum.perun.registrar.model.Application;
@@ -494,5 +495,61 @@ public class ApiCaller {
 
 		return listOfObjects.subList(fromIndex, toIndex);
 
+	}
+
+	/**
+	 * Fetch PerunBean from the Perun database
+	 *
+	 * IMPORTANT: Actually are supported only objects, that can be connected with some user/authorizedGroup role!!!
+	 *
+	 * @param object according to which will be the real object retrieved
+	 * @return real PerunBean object from the database
+	 * @throws PerunException
+	 */
+	public PerunBean fetchPerunBean(PerunBean object) throws PerunException {
+		if (object == null) throw new InternalErrorException("PerunBean cannot be null while fetching the real object from the database.");
+		return getBeanFromDatabase(object);
+	}
+
+	/**
+	 * Fetch PerunBeans from the Perun database
+	 *
+	 * IMPORTANT: Actually are supported only objects, that can be connected with some user/authorizedGroup role!!!
+	 *
+	 * @param objects according to which will be real objects retrieved
+	 * @return list of real PerunBean objects from the database
+	 * @throws PerunException
+	 */
+	public List<PerunBean> fetchPerunBeans(List<PerunBean> objects) throws PerunException {
+		if (objects == null) throw new InternalErrorException("PerunBeans cannot be null while fetching the real objects from the database.");
+		List<PerunBean> result = new ArrayList<>();
+		for (PerunBean object: objects) {
+			result.add(fetchPerunBean(object));
+		}
+		return result;
+	}
+
+	/**
+	 * Method which fetch the object from the database according to object name and its id
+	 *
+	 * @param object according to which will be real object retrieved
+	 * @return real PerunBean object from the database
+	 * @throws PerunException
+	 */
+	private PerunBean getBeanFromDatabase(PerunBean object) throws PerunException {
+		switch(object.getBeanName()) {
+			case "Group":
+				return getGroupById(object.getId());
+			case "Vo":
+				return getVoById(object.getId());
+			case "Resource":
+				return getResourceById(object.getId());
+			case "Facility":
+				return getFacilityById(object.getId());
+			case "SecurityTeam":
+				return getSecurityTeamById(object.getId());
+			default:
+				throw new RpcException(RpcException.Type.WRONG_PARAMETER, "Object " + object.getBeanName() + " is not supported.");
+		}
 	}
 }
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
@@ -129,6 +129,7 @@ public enum AuthzResolverMethod implements ManagerMethod {
 			try {
 				bean = (PerunBean) Class.forName("cz.metacentrum.perun.core.api." + complementaryObjectName).getConstructor().newInstance();
 				bean.setId(complementaryObjectId);
+				bean = ac.fetchPerunBean(bean);
 			} catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
 				throw new InternalErrorException(e);
 			} catch (ClassNotFoundException e) {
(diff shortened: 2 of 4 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -129,6 +129,7 @@ public AuthzRoles call(ApiCaller ac, Deserializer parms) throws PerunException {
 			try {
 				bean = (PerunBean) Class.forName("cz.metacentrum.perun.core.api." + complementaryObjectName).getConstructor().newInstance();
 				bean.setId(complementaryObjectId);
+				bean = ac.fetchPerunBean(bean);
```
zlamalp, line 132: Can't we simply fetch bean based on "class" and "id? If you would update ``getBeanFromDatabase()`` for that, you could use shared logic, instead of instantiating fake class and then replacing it with the real object anyway.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -250,6 +270,15 @@ public AuthzRoles call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 	 */
+	/*#
+	 * Set role for users.
+	 *
+	 * IMPORTANT: Refresh authz only if user in session is affected.
+	 *
+	 * @param role String Role which will be set for given users ( FACILITYADMIN | GROUPADMIN | PERUNADMIN | RESOURCEADMIN | RESOURCESELFSERVICE | SPONSOR | TOPGROUPCREATOR | VOADMIN | VOOBSERVER | PERUNOBSERVER | SECURITYADMIN | CABINETADMIN )
+	 * @param users int[] <code>ids</code> of users for which is the role set
+	 * @exampleParam role "VOADMIN"
+	 */
```
zlamalp, line 281: So, generally there is no longer any check, that role requires some complementary object and we can expect more "group managers without any group" or what? If there is no check, then it might fail in ``AuthzResolverBlImpl.unsetRole(sess, user, complementaryObject, role)`` for role ``SPONSOR`` where code fetches beanName from complementaryObject. There might be more places like this.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
balcirakpeter: You are right I will change it.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
balcirakpeter: Rules for that are defined in the perun-roles.yml. So for example, it is possible to set/unset role without a complementary object for PERUNADMIN, PERUNOBSERVER, CABINETADMIN. However, it is not possible to set/unset role without a complementary object for GROUPADMIN, VOADMIN, SPONSOR etc.
This is checked in the business logic of set/unset role methods.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
balcirakpeter: Ok, we have decided that we will do it in the future as a refactorization.

balcirakpeter:
Functionality was tested on test instance.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
zlamalp: Ok, that makes sense. It just might be confusing in javadoc, since we do not list correct combinations, but only all possible. But since its technically configurable per-instance we probably can't do anything about it.

zlamalp (approved):

Vojtech-Sassmann (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/ApiCaller.java
```diff
@@ -1,8 +1,12 @@
 package cz.metacentrum.perun.rpc;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.function.Function;
```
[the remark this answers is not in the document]
balcirakpeter: Removed

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -279,63 +317,79 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 				} else if(parms.contains("complementaryObjects")) {
 					cz.metacentrum.perun.core.api.AuthzResolver.setRole(ac.getSession(),
 								ac.getGroupById(parms.readInt("authorizedGroup")),
 								roleName,
-								parms.readListPerunBeans("complementaryObjects"));
+								ac.fetchPerunBeans(parms.readListPerunBeans("complementaryObjects")));
 					return null;
 				} else {
-					throw new RpcException(RpcException.Type.MISSING_VALUE, "list of complementary objects or complementary object");
+					cz.metacentrum.perun.core.api.AuthzResolver.unsetRole(ac.getSession(),
```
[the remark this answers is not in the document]
balcirakpeter: Fixed

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -279,63 +317,79 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
-					}
 					cz.metacentrum.perun.core.api.AuthzResolver.setRole(ac.getSession(),
 								groups,
-								parms.readPerunBean("complementaryObject"),
+								ac.fetchPerunBean(parms.readPerunBean("complementaryObject")),
 								roleName);
 					return null;
 				} else {
-					throw new RpcException(RpcException.Type.MISSING_VALUE, "complementaryObject");
+					cz.metacentrum.perun.core.api.AuthzResolver.unsetRole(ac.getSession(),
```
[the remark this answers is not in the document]
balcirakpeter: Fixed

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -279,63 +317,79 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
-					}
 					cz.metacentrum.perun.core.api.AuthzResolver.setRole(ac.getSession(),
 								users,
 								roleName,
-								parms.readPerunBean("complementaryObject"));
+						ac.fetchPerunBean(parms.readPerunBean("complementaryObject")));
 					return null;
 				} else {
-					throw new RpcException(RpcException.Type.MISSING_VALUE, "complementaryObject");
+					cz.metacentrum.perun.core.api.AuthzResolver.unsetRole(ac.getSession(),
```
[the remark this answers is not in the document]
balcirakpeter: Fixed

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -279,63 +317,79 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 				} else if(parms.contains("complementaryObjects")) {
 					cz.metacentrum.perun.core.api.AuthzResolver.setRole(ac.getSession(),
 								ac.getUserById(parms.readInt("user")),
 								roleName,
-								parms.readListPerunBeans("complementaryObjects"));
+						ac.fetchPerunBeans(parms.readListPerunBeans("complementaryObjects")));
 					return null;
 				} else {
-					throw new RpcException(RpcException.Type.MISSING_VALUE, "list of complementary objects or complementary object");
+					cz.metacentrum.perun.core.api.AuthzResolver.unsetRole(ac.getSession(),
```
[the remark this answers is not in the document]
balcirakpeter: Fixed

Vojtech-Sassmann (approved):

PR #2994, opened 2020-11-23
CORE: Fixed possible null pointer

zlamalp:
- Make sure we return non-null string for stdout/stderr
  when getting it from TaskExecutionException.
- When creating TaskResult handle also null stderr.

### DIFF
--- a/perun-engine/src/main/java/cz/metacentrum/perun/engine/exceptions/TaskExecutionException.java
+++ b/perun-engine/src/main/java/cz/metacentrum/perun/engine/exceptions/TaskExecutionException.java
@@ -82,7 +82,7 @@ public class TaskExecutionException extends EngineException {
 	}
 
 	public String getStdout() {
-		return stdout;
+		return (this.stdout == null) ? "" : this.stdout;
 	}
 
 	public void setStdout(String stdout) {
@@ -90,7 +90,7 @@ public class TaskExecutionException extends EngineException {
 	}
 
 	public String getStderr() {
-		return stderr;
+		return (this.stderr == null) ? "" : this.stderr;
 	}
 
 	public void setStderr(String stderr) {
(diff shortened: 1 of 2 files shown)


perun-engine/src/main/java/cz/metacentrum/perun/engine/exceptions/TaskExecutionException.java
```diff
@@ -82,15 +82,15 @@ public void setReturnCode(int returnCode) {
...
+		return (this.stdout == null) ? "" : this.stdout;
 	}
 
 	public void setStdout(String stdout) {
 		this.stdout = stdout;
 	}
 
 	public String getStderr() {
-		return stderr;
+		return (this.stderr == null) ? "" : this.stderr;
```
stavamichal, line 93: Not sure if empty string is correct result of "getStd*" methods.
IMHO empty string feels like - "There is an error but we don't get any text for it. Text is missing." What do you think @zlamalp ?

zlamalp: @stavamichal Generally, you are probably right.

In this case I followed original implementation. We already do this with the outputs in ``AbstractWorker``, which is responsible for actually running the scripts and taking their output. I'm not sure, where the null value came from, but null pointer was thrown on the row where only possible place is null in ``stderr`` String. So I wanted to prevent this globally, since it shouldn't happen. That is why I modified also ``createTaskResult()`` to take either null or empty string as OK and non-empty as warning.

stavamichal: Ok.

stavamichal (approved):

PR #2995, opened 2020-11-24
Perun core - updateSponsorshipValidity fix

Vojtech-Sassmann:
* The updateSponsorshipValidity method used to force the validityTo
parameter. Not, when not specified, it is used with the null value
(which means that the sponsorship will not expire)
* This was changed because the generated open api library in GUI was not
able to pass the validityTo param with null value.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
@@ -370,7 +370,7 @@ public enum MembersManagerMethod implements ManagerMethod {
 	 * @param member int id of sponsored member, optional
 	 * @param sponsor int id of sponsoring user that is to be removed
 	 * @param validityTo String the last day, when the sponsorship is active, yyyy-mm-dd format.
-	 *                          can be set to null never expire
+	 *                          if it is not passed, or null, it can be set to never expire
 	 * @throw PrivilegeException insufficient permissions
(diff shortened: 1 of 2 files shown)


stavamichal (approved):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -383,7 +383,10 @@ public Void call(ApiCaller ac, Deserializer params) throws PerunException {
 
 			Member sponsoredMember = ac.getMemberById(params.readInt("member"));
 			User sponsor = ac.getUserById(params.readInt("sponsor"));
-			LocalDate newValidity = params.readLocalDate("validityTo");
+			LocalDate newValidity = null;
+			if (params.contains("validityTo")) {
+				newValidity = params.readLocalDate("validityTo");
```
zlamalp, line 388: This will still fail, if the param is present, but the value is null, because of the way we parse date from the input.

zlamalp (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
Vojtech-Sassmann: You are right, I forget about that. It should be fixed now.

zlamalp (approved):

PR #2996, opened 2020-11-24
Extend and refactor sending of validation requests in Utils

stavamichal:
- new methods:
    - sendAccountActivationEmail
    - sendAccountActivationConfirmationEmail
 - logic is same as for password reset methods just with different text
 templates for body and subject of email
 - refactor code of methods:
    - sendPasswordResetEmail
    - sendValidationEmail
    - sendPasswordResetConfirmationEmail
 - to prevent using duplicates of code and make it easier to use in the
 future
 - for purpose of refactoring, these new private methods were created:
    - prepareValidationLinkForEmailChange
    - prepareValidationLinkForPasswordResetAndAccountActivation
    - prepareValidityTo
    - prepareBodyOfEmail
    - prepareSubjectOfEmail
    - sendEmail
 - all methods are there to prepare environment to work with such
 requests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
@@ -9,7 +9,6 @@ import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.GroupsManager;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.Pair;
-import cz.metacentrum.perun.core.api.Perun;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.RichUserExtSource;
 import cz.metacentrum.perun.core.api.User;
@@ -46,7 +45,6 @@ import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
 import org.springframework.mail.MailException;
 import org.springframework.mail.SimpleMailMessage;
 import org.springframework.mail.javamail.JavaMailSender;
-import org.springframework.mail.javamail.JavaMailSenderImpl;
 
 import javax.crypto.Cipher;
 import javax.crypto.Mac;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1009,84 +1007,106 @@ public static String getMessageAuthenticationCode(String input) {
...
-				path = "/ceitec/gui/";
+				linkLocation = "/ceitec/gui/";
 			} else if (urlObject.getPath().contains("/cert/")) {
-				path = "/cert/gui/";
+				linkLocation = "/cert/gui/";
 			}
+		} catch (MalformedURLException ex) {
+			throw new InternalErrorException("Not valid URL of running Perun instance.", ex);
+		}
+		String validationLink = prepareValidationLinkForEmailChange(url, linkLocation, changeId, user);
```
zlamalp, line 1032: ``validationLink`` is unused in the later code. Maybe it get mixed with ``linkLocation`` in the template?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1112,148 +1322,116 @@ public static void sendPasswordResetEmail(User user, String email, String namesp
...
-	 * @param email user's email to send notification to
-	 * @param namespace namespace the password was re-set
-	 * @param login login of user
-	 * @param subject Subject from template or null
-	 * @param content Message from template or null
+	 * @param url base URL of Perun instance
+	 * @param linkLocation location of validation link under specific Perun instance (for example '/non/pwd-reset/')
+	 * @param id ID of request
+	 * @param user user to who link will be send
+	 * @param namespace namespace to processing request for
```
zlamalp, line 1401: Please update javadoc, there is no ``namespace`` and ``id`` in the method params.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1112,148 +1322,116 @@ public static void sendPasswordResetEmail(User user, String email, String namesp
...
-			message.setSubject("[" + instanceName + "] Password reset in namespace: " + namespace);
-		} else {
-			subject = subject.replace("{namespace}", namespace);
-			subject = subject.replace("{instanceName}", instanceName);
-			message.setSubject(subject);
+		// send email
+		try {
+			mailSender.send(message);
+		} catch (MailException ex) {
+			throw new InternalErrorException("Unable to send email.", ex);
```
zlamalp, line 1333: Please do explicit logging in error mode for that exception. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1009,84 +1007,106 @@ public static String getMessageAuthenticationCode(String input) {
...
+				Matcher m2 = namespacePattern.matcher(toSubstitute);
+				if (m2.find()) {
+					// only language "cs", "en",...
+					String lang = m2.group(1);
+					langLink = langLink + "&locale=" + lang;
+				}
+				bodyParametersToReplace.put(toSubstitute, langLink);
+			}
+		} else {
+			bodyParametersToReplace.put("{link}", validationLink);
```
zlamalp, line 1105: We must append ``&activation=true`` so that app for password reset knows, it should behave as an account activation. Or you can handle this in link preparation method.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: Added to preparation of link.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: Right, fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: Yes, I used linkLocation instead of validation link. It should be fixed now.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
zlamalp: Location of the error in the code should be visible from the stacktrace of the passed `ex`. I would shorten the error message like: ``Unable to send mail to "email".`` or something like this.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
zlamalp: It remained in ``defaultBody`` as directly joined strings. Maybe you should just prepare all replacable strings first and then use placeholders (eg. ``{link}``) also in default template body. This might apply to other mail types too.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: You are right, it is still in the body. I have replaced that.
If you are using defaultBody or defaultSubject I would prefer to not doing replacements. It is better to have it ready before using.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: Changed.

zlamalp (approved):

Vojtech-Sassmann (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1103,6 +1123,196 @@ public static void sendValidationEmail(User user, String url, String email, int
...
+		subjectParametersToReplace.put("{instanceName}", instanceName);
+		subjectParametersToReplace.put("{namespace}", namespace);
+		subject = prepareSubjectOfEmail(subject, defaultSubject, subjectParametersToReplace);
+
+		Map<String, String> bodyParametersToReplace = new HashMap<>();
+		bodyParametersToReplace.put("{displayName}", user.getDisplayName());
+		bodyParametersToReplace.put("{namespace}", namespace);
+		bodyParametersToReplace.put("{validity}", validityTo);
+		// allow enforcing per-language links
+		if (messageTemplate.contains("{link-")) {
```
zlamalp, line 1149: This fails if ``messageTemplate`` is null, which is, when there is no pre-defined template stored in the entityless attribute. Same for the account activation method.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
stavamichal: Fixed.

zlamalp (approved):

PR #2999, opened 2020-11-25
Extend role management rules

balcirakpeter:
- We have realized that we need to know what is the primary object for
  the role. Therefore this key was added to the RoleManagementRules.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
@@ -11,6 +11,7 @@ import java.util.Objects;
  * so the database query can be created and executed more generally.
  *
  * roleName is role's unique identification which is used in the configuration file perun-roles.yml
+ * primaryObject serves to determine with which object is the role primarily connected. Other objects are just complementary.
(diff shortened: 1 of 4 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -26,6 +26,7 @@
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
```
zlamalp, line 29: Not sure why added if there is no other change in the tests.

zlamalp (approved):

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
balcirakpeter: Sorry I will remove it. I've just run some tests to be sure that everything works and I forgot to remove the imports after that. I did remove tests because they depended on the actual state of the config file which can change in the future.

PR #3000, opened 2020-11-25
Core - createSponsoredMembersFromCSV

Vojtech-Sassmann:
* Users using the method used for creation of multiple users are used to
provide the data in a csv format. Because of that, we have decided to
create a new method which can work with such input.
* Created method createSponsoredMembersFromCSV which takes data
parameter, representing the content of the csv file; and header
parameter, representing the header of the csv file.
* We use the `;` semicolon as a delimiter for both, the body and the
header.
* Required values for now are - `firstname`, `lastname`,
`urn:perun:user:attribute-def:def:preferredMail`.
* Optional values are - `urn:perun:user:attribute-def:def:note`.
* Created new user:def:note attribute. This attribute can be used to
show some note for users in the users overview in GUI.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2454,6 +2454,13 @@ perun_policies:
     include_policies:
       - default_policy
 
+  createSponsoredMembersFromCSV_Vo_String_List<String>_User_policy:
+    policy_roles:
+      - SPONSOR: Vo
+        SELF: User
+    include_policies:
+      - default_policy
+
   createSponsoredMembers_Vo_String_List<String>_User_policy:
     policy_roles:
       - SPONSOR: Vo
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
@@ -1173,6 +1173,34 @@ public interface MembersManager {
 	RichMember setSponsoredMember(PerunSession session, Vo vo, User userToBeSponsored, String namespace, String password, User sponsor, LocalDate validityTo) throws PrivilegeException, AlreadyMemberException, LoginNotExistsException, PasswordCreationFailedException, ExtendMembershipException, WrongAttributeValueException, ExtSourceNotExistsException, WrongReferenceAttributeValueException, UserNotInRoleException, PasswordStrengthException, InvalidLoginException, AlreadySponsorException;
 
 	/**
+	 * Creates new sponsored members using input from CSV file.
+	 *
+	 * Since there may be error while creating some of the members and we cannot simply rollback the transaction and
+	 * start over, exceptions during member creation are not thrown and the returned map has this structure:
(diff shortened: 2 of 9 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2454,6 +2454,13 @@ perun_policies:
     include_policies:
       - default_policy
 
+  createSponsoredMembersFromCSV_Vo_String_List<String>_User_policy:
+    policy_roles:
+      - SPONSOR: Vo
+        SELF: User
```
metodej, line 2460: Shouldn't here be also REGISTRAR role since it is mentioned in rpc javadoc?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2673,4 +2725,116 @@ private void processMemberAfterRemovingLastSponsor(PerunSession sess, Member spo
...
+		mapName.put("firstName", data.get("firstname"));
+		mapName.put("lastName", data.get("lastname"));
+
+		String email = data.get(A_U_PREF_MAIL);
+
+		PasswordManagerModule module = getPerunBl().getUsersManagerBl().getPasswordManagerModule(sess, namespace);
+		String password = module.generateRandomPassword(sess, null);
+
+		// create sponsored member
+		User user;
```
metodej, line 2774: Is there any reason to declare user here and not later?

stavamichal (approved):

perun-base/src/main/resources/perun-roles.yml
Vojtech-Sassmann: It doesn't really make sense, I will edit the javadoc.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: No, there is not. I have moved it.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1172,6 +1172,34 @@
 	 */
 	RichMember setSponsoredMember(PerunSession session, Vo vo, User userToBeSponsored, String namespace, String password, User sponsor, LocalDate validityTo) throws PrivilegeException, AlreadyMemberException, LoginNotExistsException, PasswordCreationFailedException, ExtendMembershipException, WrongAttributeValueException, ExtSourceNotExistsException, WrongReferenceAttributeValueException, UserNotInRoleException, PasswordStrengthException, InvalidLoginException, AlreadySponsorException;
 
+	/**
+	 * Creates new sponsored members.
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Added.

metodej (approved):

PR #3001, opened 2020-11-25
new perun-auditlogger component

mvocu:
New perun-auditlogger component that reads audit log messages and logs them via syslog.
The pull request contains:
  - sources of the auditlogger component in `perun-auditlogger` module
  - startup scripts for init and systemd in `perun-utils/init.d-scripts` and `perun-utils/systemd`
  - default logback configuration (in jar on classpath)

Auditlogger configuration:
   * `perun-auditlogger.properties` at default location (configurable by _perun.conf.custom_ property), properties:
        - _auditlogger.statefile_ - location of file that contains id of the last successfully sent message (default `/tmp/auditlogger.state`)
        - _auditlogger.consumer_ - name of the consumer passed to `pollConsumerMessages()` (default `auditlogger`)
   * `/etc/default/perun-auditlogger` - read by startup script, may override default setting by including the following variables:
     - _SYSLOG_HOST_ - name of the server with running syslog (default `localhost`)
     - _SYSLOG_FACILITY_ - syslog facility used for the logged messages (default `LOCAL0`)
  * `logback-auditlogger.xml` - logback configuration file, its location may be specified by _logback.configurationFile_ property (**only on java command line**, ie. in startup script)
    - may be used to completely redefine the logging behaviour
    - the `syslog-appender` and `syslog-logger` are used to feed messages into syslog

**IMPORTANT**
 * the host configured as _SYSLOG_HOST_ must have a syslog daemon running and listening on UDP port 514 for incoming messages (usually disabled by default)
 * example of configuration for rsyslogd is here:

> module(load="imudp")
> input(type="imudp" address="127.0.0.1" port="514")
> 
> local0.*        /var/log/perun/auditlogger.log


perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/main/AuditLoggerStarter.java
```diff
@@ -0,0 +1,82 @@
...
+public class AuditLoggerStarter {
+	private final static Logger log = LoggerFactory.getLogger(AuditLoggerStarter.class);
+
+	private AuditLoggerManager auditLoggerManager;
+	private AbstractApplicationContext springCtx;
+	private PerunPrincipal perunPrincipal;
+	private Perun perunBl;
+
+	public AuditLoggerStarter() {
+		this.perunPrincipal = new PerunPrincipal("perunLdapc", ExtSourcesManager.EXTSOURCE_NAME_INTERNAL, ExtSourcesManager.EXTSOURCE_INTERNAL);
```
zlamalp, line 27: I would use own name instead of ``perunLdapc``.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/service/impl/AuditLoggerManagerImpl.java
```diff
@@ -0,0 +1,96 @@
...
+import cz.metacentrum.perun.core.api.PerunPrincipal;
+import cz.metacentrum.perun.core.api.PerunSession;
+
+@org.springframework.stereotype.Service(value = "auditLoggerManager")
+public class AuditLoggerManagerImpl implements AuditLoggerManager {
+
+	private final static Logger log = LoggerFactory.getLogger(AuditLoggerManagerImpl.class);
+
+	private final static String DEFAULT_CONSUMER_NAME = "auditlogger";
+	private final static String DEFAULT_STATE_FILE = "/tmp/auditlogger.state";
```
zlamalp, line 22: I think we want this to be persistent, not disposable on server restart. Yet I'm not sure, how is this handled in dockerized version of LDAPc @martin-kuba ? On instances without docker, we have this property set to ``./[filename]`` which leads to the same folder as jar.
So there are two solutions. We can have default fallback value here and do not set property on each instance or we can explicitly set default in properties file and keep current default value (or any other required by the docker).

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/main/AuditLoggerStarter.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+		try {
+			AuditLoggerStarter auditLoggerStarter = new AuditLoggerStarter();
+
+			// Just for the Spring IoC to exit gracefully...
+			auditLoggerStarter.springCtx.registerShutdownHook();
+
+			// Sets RPC Caller and Perun
+			// ldapcStarter.ldapcManager.setRpcCaller(rpcCaller);
```
zlamalp, line 62: Please remove this line.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
```diff
@@ -0,0 +1,197 @@
...
+
+			//If running is true, then this process will be continuously
+			while (running) {
+
+				int sleepTime = 1000;
+				//Waiting for new messages. If consumer failed in some internal case, waiting until it will be repaired (waiting time is increases by each attempt)
+				while(messages == null) {
+					try {
+						//IMPORTANT STEP1: Get new bulk of messages
+						messages = perun.getAuditMessagesManager().pollConsumerMessages(perunSession, auditLoggerManager.getConsumerName(), lastProcessedIdNumber);
```
zlamalp, line 87: I think we can use BL instead of Entry layer to avoid authorization check, which might be the reason you used ``perunLdapc`` principal in the session, which is by default configured to be perun admin.

zlamalp (changes requested):

perun-utils/systemd/perun-auditlogger.service
```diff
@@ -0,0 +1,18 @@
...
+#
+
+[Unit]
+Description=Perun-auditlogger
+After=perun.service
+
+[Service]
+Type=forking
+ExecStart=/bin/bash /home/perun/perun-auditlogger/start_auditlogger.sh
+ExecStop=/bin/bash /home/perun/perun-engine/stop_auditlogger.sh
```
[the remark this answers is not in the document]
mvocu: Fixed.

perun-utils/init.d-scripts/perun-auditlogger.debian
```diff
@@ -0,0 +1,165 @@
...
+# Short-Description:    Perun AuditLogger
+# Description:          Starts Perun AuditLogger component.
+### END INIT INFO
+#
+# Written by Michal Prochazka <michalp@ics.muni.cz>
+# Modified by Pavel Zlamal <zlamal@cesnet.cz>
+# Modified by Michal Vocu <michal.vocu@gmail.com>
+#
+# 22.06.2017 Use /etc/perun/logback-auditlogger.xml and /var/log/perun/
+# 24.03.2016 Create /var/run/perun if not exists, will work when started on system boot
```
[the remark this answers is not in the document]
mvocu: Fixed.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/main/AuditLoggerStarter.java
mvocu: Fixed. Requires change to the perun.properties.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/service/impl/AuditLoggerManagerImpl.java
mvocu: I would argue that the default value should lead to somewhere guaranteed to exist. Anything else is placing constraint (or rather dependency) on the actual packaging and deployment. Sensible deployment should decide on the actual location of state file and there is a couple of ways to configure it; moreover in my opinion placing state file in the same directory as the binary installation is not a good practice.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/main/AuditLoggerStarter.java
mvocu: Removed and fixed the comment below.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
mvocu: Ok. Which makes the authorization checks rather superficial imho.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/main/AuditLoggerStarter.java
mvocu: Although, in light of the last change, probably not.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
zlamalp: Well, I don't think, that internal component (calling single method) should be perun admin just because the called method requires it. We could probably create new role, but it should be then true also for other internal principals, etc. Also entry generally creates new transaction, which might not be desired in other components (overhead). So we probably should improve all this logic in the future, but it is outside of the scope of this pull-request.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/service/impl/AuditLoggerManagerImpl.java
zlamalp: Ok, I'll leave it to @martin-kuba, since he knows deployment details.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/main/AuditLoggerStarter.java
zlamalp: As you changed usage of entry to bl, it shouldn't. If you wish, you can update ``defaultCoreProperties`` in ``perun-base.xml``.

zlamalp (approved):

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/service/impl/AuditLoggerManagerImpl.java
martin-kuba: Docker containers keep their image between restarts, so the file /tmp/auditlogger.state would survive restart. But it will not survive deployment of a new version of the application. Is it a problem? We have the same situation with perun_ldapc, where the setting is `ldap.stateFile=./perun-ldapc-last-state` and the working directory is _outside_ of the container. So if you want to keep the state between different versions of the application, set it to `./auditerlogger.state`.

perun-auditlogger/src/main/resources/META-INF/spring.handlers
```diff
@@ -0,0 +1,6 @@
+http\://www.springframework.org/schema/tx=org.springframework.transaction.config.TxNamespaceHandler
```
martin-kuba, line 1: Why do we need this file? There is no such file for perun-ldapc. It should be create automatically by maven-shade-plugin.

martin-kuba (changes requested):
- please remove the files perun-auditlogger/src/main/resources/META-INF/spring.* unless you are very sure that they are unavoidable
- do we need to have both SysV init.d file and systemd service file?
- the default location of state file would be better in the current working directory

perun-auditlogger/src/main/resources/META-INF/spring.handlers
mvocu: Removed from git.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/service/impl/AuditLoggerManagerImpl.java
mvocu: OK, changed the default.

martin-kuba (approved):

PR #3008, opened 2020-11-26
Cleaned up TaskSchedulerTest

zlamalp:
- Removed unused imports and tests.
- Fixed usage of recurrenceSpy.


zlamalp:
These test needs to be rewritten completely. Closing.

PR #3009, opened 2020-11-27
Methods for getting list of entities by their IDs

cuadradek:
- New methods "getEntitiesByIds" were created for Facility, Group, Member,
Resource, RichResource, User, UserExtSource and VO.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -497,6 +497,26 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getFacilitiesByIds_List<Integer>_policy:
+    policy_roles:
+      - RPC:
+      - ENGINE:
+      - FACILITYADMIN:
+      - FACILITYOBSERVER:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  filter-getFacilitiesByIds_List<Integer>_policy:
+    policy_roles:
+      - RPC:
+      - ENGINE:
+      - FACILITYADMIN: Facility
+      - FACILITYOBSERVER: Facility
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
   getRichFacilities_policy:
     policy_roles:
       - FACILITYADMIN:
@@ -1162,6 +1182,38 @@ perun_policies:
       - VOOBSERVER: Vo
       - GROUPADMIN: Group
       - GROUPOBSERVER: Group
+      - VOADMIN: Vo
+      - TRUSTEDFACILITYADMIN: Vo
+    include_policies:
+      - default_policy
+
+  getGroupsByIds_List<Integer>_policy:
+    policy_roles:
+#      Temporary fix until a new MEMBERSHIP role is created
+      - SELF:
+      - RESOURCEADMIN:
+      - RESOURCEOBSERVER:
+      - PERUNOBSERVER:
+      - RPC:
+      - GROUPADMIN:
+      - GROUPOBSERVER:
+      - VOOBSERVER:
+      - VOADMIN:
+      - TRUSTEDFACILITYADMIN:
+    include_policies:
+      - default_policy
+
+  filter-getGroupsByIds_List<Integer>_policy:
+    policy_roles:
+#      Temporary fix until a new MEMBERSHIP role is created
+      - SELF:
+      - RESOURCEADMIN: Vo
+      - RESOURCEOBSERVER: Vo
+      - PERUNOBSERVER:
+      - RPC:
+      - GROUPADMIN: Group
+      - GROUPOBSERVER: Group
+      - VOOBSERVER: Vo
       - VOADMIN: Vo
       - TRUSTEDFACILITYADMIN: Vo
     include_policies:
@@ -1973,6 +2025,30 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getMembersByIds_List<Integer>_policy:
+    policy_roles:
+      - GROUPADMIN:
+      - GROUPOBSERVER:
+      - SELF:
+      - RPC:
+      - PERUNOBSERVER:
+      - VOOBSERVER:
+      - VOADMIN:
+    include_policies:
+      - default_policy
+
+  filter-getMembersByIds_List<Integer>_policy:
+    policy_roles:
+      - GROUPADMIN: Vo
+      - GROUPOBSERVER: Vo
+      - SELF: User
+      - RPC:
+      - PERUNOBSERVER:
+      - VOOBSERVER: Vo
+      - VOADMIN: Vo
+    include_policies:
+      - default_policy
+
   getMemberByUser_Vo_User_policy:
     policy_roles:
       - SELF: User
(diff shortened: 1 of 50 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
```diff
@@ -59,6 +59,17 @@
 	 */
 	Facility getFacilityByName(PerunSession perunSession, String name) throws FacilityNotExistsException, PrivilegeException;
 
+	/**
+	 * Searches for the Facilities with specified ids.
+	 *
+	 * @param perunSession
+	 * @param ids
+	 * @return list of Facilities with specified ids
+	 * @throws InternalErrorException
```
metodej, line 68: I don't think there is any reason to include `InternalErrorException` to javadoc, could you please delete it from the javadocs you added?

metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
cuadradek: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/FacilitiesManagerImpl.java
```diff
@@ -336,6 +342,18 @@ public Facility getFacilityByName(PerunSession sess, String name) throws Facilit
 		}
 	}
 
+	@Override
+	public List<Facility> getFacilitiesByIds(PerunSession perunSession, List<Integer> ids) {
+		try {
+			SqlParameterSource parameters = new MapSqlParameterSource("ids", ids);
+
+			return namedParameterJdbcTemplate.query("select " + facilityMappingSelectQuery + " from facilities where id in (:ids)",
```
zlamalp, line 350: Using ``id in (...)`` is not effective for large number of IDs. Please see ``UsersManagerImpl.getUsersByIds()`` implementation. This apply to impl of all methods.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -515,6 +515,20 @@ public UserExtSource getUserExtSourceById(PerunSession sess, int id) throws User
...
+		Utils.checkPerunSession(sess);
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "getUserExtSourcesByIds_List<Integer>_policy")) {
+			throw new PrivilegeException(sess, "getUserExtSourcesByIds");
+		}
+		List<UserExtSource> userExtSources = getUsersManagerBl().getUserExtSourcesByIds(sess, ids);
+		userExtSources.removeIf(ues -> !AuthzResolver.authorizedInternal(sess, "filter-getUserExtSourcesByIds_List<Integer>_policy", ues));
+
+		return userExtSources;
```
HejdaJakub, line 529: Due to changes in perun-roles.yml you should modify this method.

HejdaJakub (changes requested):

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4529,6 +4661,28 @@ perun_policies:
...
+  filter-getUserExtSourcesByIds_List<Integer>_policy:
+    policy_roles:
+      - RPC:
+      - GROUPADMIN:
+      - GROUPOBSERVER:
+      - VOOBSERVER:
+      - VOADMIN:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
```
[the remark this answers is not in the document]
cuadradek: I removed the filter policy.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
cuadradek: Done

perun-core/src/main/java/cz/metacentrum/perun/core/impl/FacilitiesManagerImpl.java
cuadradek: Done

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/GroupsManagerImplApi.java
```diff
@@ -158,6 +158,16 @@
 	 */
 	Group getGroupByName(PerunSession perunSession, Vo vo, String name) throws GroupNotExistsException;
 
+	/**
+	 * Search for the groups with specified ids in all VOs.
```
zlamalp, line 162: I would prefer having similar messages in javadoc for all those impl methods. Its ``get`` method. It doesn't search/find. Also you should note, that it silently skips non-existing entities.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -1149,7 +1172,7 @@ public void getUsersByIds() throws Exception {
 			users.add(user2);
 		}
 
-		assertEquals(users, perun.getUsersManagerBl().getUsersByIds(sess, ids));
+		assertEquals(users, perun.getUsersManager().getUsersByIds(sess, ids));
```
zlamalp, line 1175: When you are at it, can you please fix this test? When we use real DB, there is no guarantee, that returned list has same order as checked one, resulting in false positive test failure.
You can gather users on input to set and you can do the same for the result of getUsersByIds(). Then you can call assertEquals() on two sets, where order doesn't matter. Thank you.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -42,6 +42,20 @@ public Facility call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Searches for the Facilities with specified IDs.
```
zlamalp, line 46: I would use unified javadoc message in RPC for all methods as I suggested for impl layer.

zlamalp (changes requested):

HejdaJakub (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/GroupsManagerImplApi.java
cuadradek: Done

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
cuadradek: Done

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
cuadradek: Done

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
zlamalp: Thank you.

zlamalp (approved):

PR #3014, opened 2020-11-30
REGISTRAR: Support editing submitted form by user

zlamalp:
- Added updateFormItemsData() which will update form items
  data with new values by its IDs. Only application submitter
  can use this.
- Updating form items sets application state to NEW
  and starts verification and auto-approval if possible.
- Method updateFormItemData() used from old GUI is now
  marked as deprecated.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -2545,8 +2545,8 @@ public class RegistrarManagerImpl implements RegistrarManager {
 		Application app = getApplicationById(sess, appId);
 		if (AppState.APPROVED.equals(app.getState()) || AppState.REJECTED.equals(app.getState())) throw new RegistrarException("Form items of once approved or rejected applications can't be modified.");
 
-		ApplicationFormItemData existingData = getFormItemDataById(data.getId());
-		if (existingData == null) throw new RegistrarException("Form item data specified by ID: "+ data.getId() + " not found in the DB.");
+		ApplicationFormItemData existingData = getFormItemDataById(data.getId(), appId);
+		if (existingData == null) throw new RegistrarException("Form item data specified by ID: "+ data.getId() + " not found or doesn't belong to the application "+appId);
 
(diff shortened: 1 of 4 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2552,16 +2552,71 @@ public void updateFormItemData(PerunSession sess, int appId, ApplicationFormItem
...
+	}
+
+	@Transactional(rollbackFor = Exception.class)
+	public void updateFormItemsData(PerunSession sess, int appId, List<ApplicationFormItemData> data) throws PerunException {
+
+		Application app = getApplicationById(appId);
+
+		if (app == null) throw new InternalErrorException("Application with ID="+appId+" doesn't exist.");
+
+		if (!AuthzResolver.selfAuthorizedForApplication(sess, app)) {
```
balcirakpeter, line 2566: Just a reminder that by calling only `selfAuthorizedForApplication`, PERUNADMIN will not be allowed to call this method. I suppose it is ok if the whole method will be used only for perun-wui, but it could be insufficient when for example NGUI would want to use it.

balcirakpeter (approved):

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
zlamalp: Thanks for the insight. Use-cases for admin gui are not yet defined, I'll let this to be decided once its needed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2552,16 +2552,71 @@ public void updateFormItemData(PerunSession sess, int appId, ApplicationFormItem
...
+			throw new PrivilegeException(sess, "updateFormItemsData");
+		}
+
+		if (AppState.APPROVED.equals(app.getState()) || AppState.REJECTED.equals(app.getState())) throw new RegistrarException("Form items of once approved or rejected applications can't be modified.");
+
+		// no data to change
+		if (data == null || data.isEmpty()) return;
+
+		for (ApplicationFormItemData dataItem : data) {
```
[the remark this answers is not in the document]
zlamalp: Should be fixed. Now we pass also application id in order to get form item data from the DB.

Vojtech-Sassmann (approved):

PR #3021, opened 2020-12-02
moved to Java source level 15

martin-kuba:
- changes Java source and target level to 15 in the main pom.xml
- drops GitHub CI testing with JDK 8, it is no longer needed
- changes compilation of release assets to use JDK 15

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -43,7 +43,7 @@
 	<properties>
 		<!-- property used by spring-boot-starter-parent project to define maven.compiler.source and maven.compiler.target
 		     properties that in turn are used by maven-compiler-plugin to specify java source and target version -->
-		<java.version>8</java.version>
+		<java.version>15</java.version>
 
 		<!-- USE UTF-8 in whole project -->
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -43,7 +43,7 @@
 	<properties>
 		<!-- property used by spring-boot-starter-parent project to define maven.compiler.source and maven.compiler.target
 		     properties that in turn are used by maven-compiler-plugin to specify java source and target version -->
-		<java.version>8</java.version>
+		<java.version>15</java.version>
```
zlamalp, line 46: Can you please set java version property to 11 in ``perun-web-gui/pom.xml`` and set ``<sourceLevel>1.11</sourceLevel>`` for gwt-maven plugin in it?
It all seems to work, but settings in IDE are a bit confusing.
* Module language level is displayed as 15.
* GWT compiler used to run developer mode respect GWT facet settings of IDE, which are taken from that ``sourceLevel`` on project import/refresh and if its "auto", it falls back to Java 8.

IDE correctly prevents you to use newer java features, but I believe we would want to use Java 11 language features in whole GWT project, so we should specify it and it will align all IDE settings to Java 11 for that module.

zlamalp (changes requested):

zlamalp (approved):

PR #3026, opened 2020-12-08
Core - extension of hashed data

Vojtech-Sassmann:
* Some services couldn't be rewritten to use the new hashedData method
because they were using getDataWithVos or getFlatData. We have figured
out that we don't actually need to create new "hashed" version of these
methods. Insteady, we can reuse the already existing methods with some
minor modifications.
* The methods getHashedHierarchicalData and getHashedDataWithGroups now
also contains vo id in resource nodes and can return vo attributes.
* In the data provider class, there were some missuses of the emptyList
and singletonList methods that caused trouble, because they are
immutable.
* The old getData methods were annotated as deprecated in the API.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
@@ -453,6 +453,7 @@ public enum ServicesManagerMethod implements ManagerMethod {
 	/*#
 	 * Generates the list of attributes per each member associated with the resource.
 	 *
+	 * @deprecated use getHashedDataWithGroups
 	 * @param service int Service <code>id</code>
 	 * @param facility int Facility <code>id</code>. You will get attributes for this facility, resources associated with it and members assigned to the resources.
 	 * @param filterExpiredMembers if true the method does not take members expired in groups into account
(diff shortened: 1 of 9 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -1107,6 +1112,7 @@ public ServiceAttributes call(ApiCaller ac, Deserializer parms) throws PerunExce
 	        |             |                      +...
 	        +-----Vo ....
 	</pre>
+	* @deprecated - use getHashedDataWithGroups
 	 */
```
stavamichal, line 1116: There is a bad indentation somewhere.

stavamichal (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -479,7 +479,7 @@ public Service call(ApiCaller ac, Deserializer parms) throws PerunException {
 	 |        +-------Attrs
 	 +...
 	 </pre>
-	 *
+	 * @deprecated - use getHashedHierarchicalData
```
zlamalp, line 482: I'm not sure if our docs parser won't go crazy about it, since it already skips all that ``<pre>`` part above and doesnẗ handle more lines of param description.

Also it shows deprecation message on hover, so it will be like ``- use getHashedHierarchicalData``. I would suggest to to remove leading dash.

zlamalp (changes requested):

stavamichal (approved):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
Vojtech-Sassmann: Fixed.

zlamalp (approved):

PR #3030, opened 2020-12-09
New method getAllSponsoredMembersAndTheirSponsors

cuadradek:
- New method getAllSponsoredMembersAndTheirSponsors was added. It is similar to
getSponsoredMembersAndTheirSponsors but doesn’t filter members that are not
sponsored by principal.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1456,12 +1456,27 @@ public class MembersManagerEntry implements MembersManager {
 	public List<MemberWithSponsors> getSponsoredMembersAndTheirSponsors(PerunSession sess, Vo vo, List<String> attrNames) throws VoNotExistsException, PrivilegeException, AttributeNotExistsException {
 		Utils.checkPerunSession(sess);
 		Utils.notNull(vo, "vo");
+		
+		//Authorization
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1454,14 +1454,21 @@ public RichMember sponsorMember(PerunSession session, Member sponsored, User spo
 
 	@Override
 	public List<MemberWithSponsors> getSponsoredMembersAndTheirSponsors(PerunSession sess, Vo vo, List<String> attrNames) throws VoNotExistsException, PrivilegeException, AttributeNotExistsException {
```
Vojtech-Sassmann, line 1456: Please keep the old politic `getSponsoredMembersAndTheirSponsors_Vo_policy` for this method. It is because the GUI needs it to resolve authorization.

Vojtech-Sassmann (changes requested):
Also, please add at least one test for the new method.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
cuadradek: Fixed

HejdaJakub (approved):

Vojtech-Sassmann (approved):

PR #3031, opened 2020-12-09
Updated HashedGenData.pm in CLI to support VO attributes

zlamalp:
- We are now receiving VO id within Resource node and
  we can also get VO attributes from the hashes.
- Removed trailing whitespace, fixed indentation.

### DIFF
--- a/perun-cli/Perun/beans/HashedGenData.pm
+++ b/perun-cli/Perun/beans/HashedGenData.pm
@@ -96,10 +96,10 @@ sub getFacilityId ()
 	my @facilityIds = keys %{$self->{_hierarchy}};
 
 	if(@facilityIds == 1) {
-    return $facilityIds[0];
-  } else {
-    die "There are more than 1 facility inside of hashed data!\n";
-  }
+		return $facilityIds[0];
+	} else {
+		die "There are more than 1 facility inside of hashed data!\n";
+	}
 }
 
 sub getMemberIdsForFacility ()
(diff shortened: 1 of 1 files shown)


perun-cli/Perun/beans/HashedGenData.pm
```diff
@@ -124,11 +124,32 @@ sub getResourceIds ()
 	return @resourceIds;
 }
 
+sub getVosIds ()
```
stavamichal, line 127: Even if this is probably the better way, I would change that to 'getVoIds' because other methods are using a singular object for the name instead of plural.

perun-cli/Perun/beans/HashedGenData.pm
```diff
@@ -124,11 +124,32 @@ sub getResourceIds ()
...
+{
+	my $self = shift;
+	my @vosIds = ();
+	my $facilityId = $self->getFacilityId;
+	foreach my $resourceId (sort keys %{$self->{_hierarchy}->{$facilityId}->{c}}) {
```
stavamichal, line 132: You could probably call getResourceIds instead.

stavamichal (changes requested):

perun-cli/Perun/beans/HashedGenData.pm
zlamalp: Fixed.

perun-cli/Perun/beans/HashedGenData.pm
zlamalp: Done.

stavamichal (approved):

PR #3036, opened 2020-12-14
Edited finding firstName and lastName in displayName

HejdaJakub:
* If the name placed at the first position in displayName (it can be firstName or lastName) is the same as middle name, then the parsing was incorrect.
* This use case has been fixed and all tests run successfully.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -3506,9 +3506,9 @@ public class RegistrarManagerImpl implements RegistrarManager {
 
 	private void parseTitlesAndMiddleName(Candidate candidate, String displayName, String firstName, String lastName) {
 		Pattern pattern = getNamesPattern(firstName, lastName);
-		if (!tryToParseTitlesAndMiddleNameFromPattern(candidate, displayName, pattern)) {
+		if (!tryToParseTitlesAndMiddleNameFromPattern(candidate, displayName, pattern, firstName)) {
 			Pattern reversePattern = getNamesPattern(lastName, firstName);
-			tryToParseTitlesAndMiddleNameFromPattern(candidate, displayName, reversePattern);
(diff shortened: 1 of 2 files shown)


zlamalp:
Please rebase this pull-request. Thank you.

HejdaJakub:
> Please rebase this pull-request. Thank you.

Pull request rebased.

zlamalp (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3533,8 +3533,14 @@ private boolean tryToParseTitlesAndMiddleNameFromPattern(Candidate candidate, St
 					"titles after, but get " + matcher.groupCount() + " groups." );
 		}
 
-		parseTitlesBefore(candidate, matcher.group(1).trim());
-		parseMiddleName(candidate, matcher.group(2).trim());
+		// if the middle name equals to the first name placed in displayName (it can be firstName or lastName too)
+		if (matcher.group(1).contains(firstName)) {
+			candidate.setTitleBefore(matcher.group(1).split(firstName)[0].trim());
```
mvocu, line 3538: It would be better to use parseTitlesBefore()

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3533,8 +3533,14 @@ private boolean tryToParseTitlesAndMiddleNameFromPattern(Candidate candidate, St
...
+			candidate.setMiddleName(firstName);
```
mvocu, line 3539: Likewise for parseMiddleName()

mvocu (commented):
Please use the appropriate methods for parsing other parts instead of setting the value directly.

Overall the whole idea of parsing the displayName by creating regular expression from firstName and lastName is little bit off in my opinion, and there may be other corner cases whrer it fails, but it is probably out of scope of this issue.

PR #3041, opened 2020-12-16
allow to call getPerunStatus on read-only instances

martin-kuba:
Updates lastAccess only if the database is writable (it is not read-only).

This change is need to be able to call methods like getPerunStatus or getPerunStatistics on instances with read-only database.


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -151,7 +151,9 @@ public PerunSession getPerunSession(PerunPrincipal principal, PerunClient client
 						usersManagerBl.updateUserExtSource(internalSession, ues);
 					}
 					// Update last access for userExtSource
-					usersManagerBl.updateUserExtSourceLastAccess(internalSession, ues);
+					if(!BeansUtils.isPerunReadOnly()) {
```
zlamalp, line 154: I would prefer to skip all update methods in read-only mode, so that we support also other ways of authentication and basically we allow reading from replica.

zlamalp (changes requested):

zlamalp (approved):

PR #3045, opened 2021-01-11
OpenApi fixed

xkureck:
* registrar method "sendMessage" was taking parameter "applicationId" but the parameter somehow changed to "appId" and somebody just didn't updated opeanapi

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -13074,10 +13074,10 @@ paths:
               type: object
               required:
                 - mailType
-                - applicationId
+                - appId
               properties:
                 mailType: { $ref: '#/components/schemas/MailType' }
-                applicationId: { type: integer, description: "application id" }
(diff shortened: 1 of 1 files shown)


perun-openapi/openapi.yml
```diff
@@ -13074,10 +13074,10 @@ paths:
               type: object
               required:
                 - mailType
-                - applicationId
+                - appId
               properties:
                 mailType: { $ref: '#/components/schemas/MailType' }
-                applicationId: { type: integer, description: "application id" }
+                appId: { type: integer, description: "application id" }
```
zlamalp, line 13080: Parameter didn't change, it was wrong here all the time. But there is a generic (now unused) parameter ``applicationId`` at the top of the file. Maybe there is switched ``appId`` and ``applicationId`` in it. If fixed we could use its definition from here and also other methods.

zlamalp (approved):

PR #3053, opened 2021-01-18
Remove support for Large attributes

balcirakpeter:
- large attributes are no longer needed. Therefore, they were removed
  from the whole perun-core code.
- DB changelog updated. All large attribute types were converted to a
  normal type.

### DIFF
--- a/perun-cli/createAttribute
+++ b/perun-cli/createAttribute
@@ -14,13 +14,11 @@ sub help {
 	--attributeName       | -N attribute friendly name
 	--attributeNameSpace  | -n namespace (e.g.: urn:perun:user:attribute-def:def)
 	--attributeDisplayName| -e attribute display name
-	--attributeType       | -t attribute type (integer/string/array/hash/boolean/largestring/largearray)*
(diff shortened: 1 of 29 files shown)


perun-cli/createAttribute
```diff
@@ -14,14 +14,12 @@ sub help {
 	--attributeName       | -N attribute friendly name
 	--attributeNameSpace  | -n namespace (e.g.: urn:perun:user:attribute-def:def)
 	--attributeDisplayName| -e attribute display name
-	--attributeType       | -t attribute type (integer/string/array/hash/boolean/largestring/largearray)*
+	--attributeType       | -t attribute type (integer/string/array/hash/boolean)*
```
zlamalp, line 17: Since there is no comment, you should remove also ending ``*``.

zlamalp (changes requested):
Seems OK, just one more thing to clean-up.

perun-cli/createAttribute
balcirakpeter: Fixed

zlamalp (approved):

PR #3064, opened 2021-01-26
Fixed async validation when creating sponsored member

zlamalp:
- Async validation should be always the last step in any
  member creation method.
  Since methods createSponsoredMembers() and
  createSponsoredMembersFromCSV() iterate over
  the creation of single sponsored member, we must provide
  the means to skip validation until all iteration is done.
- For this purpose new Validation enum was introduced.
  It can be used to either perform SYNC or ASYNC validation
  or to skip validation using the NONE value.
- Methods to create/set sponsored member(s) which were using
  boolean param "asyncValidation" are now using "validation"
  enum.
- Return value of Map<String, Map<String, String>> was changed
  to Map<String, Map<String, Object>> for inner CSV method,
  since we need to get also the Member object in order
  to call validation afterwards.
  Outer API (RPC/OpenAPI) is not modified.
- Fixed tests.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -135,6 +135,12 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 
 	private static final String NO_REPLY_EMAIL = "no-reply@muni.cz";
 
+	private static final String OK = "OK";
+	private static final String LOGIN = "login";
+	private static final String PASSWORD = "password";
+	private static final String STATUS = "status";
+	private static final String MEMBER = "member";
+
 	public static final List<String> SPONSORED_MEMBER_REQUIRED_FIELDS = Arrays.asList(
 			"firstname",
 			"lastname",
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2464,19 +2468,42 @@ public Member setSponsoredMember(PerunSession session, Vo vo, User userToBeSpons
...
-					sponsor, validityTo, sendActivationLink, url, asyncValidation);
-			totalResult.put(data.get(processedCounter++), singleResult);
+			// async validation must be performed at the end, not directly during member creation
+			Validation localValidation = (Objects.equals(Validation.ASYNC, validation)) ? Validation.NONE : validation;
+			Map<String, Object> originalResult = createSingleSponsoredMemberFromCSV(sess, vo, namespace, singleRow,
+					sponsor, validityTo, sendActivationLink, url, localValidation);
+
+			// convert result to expected "type" for outer API
+			Map<String, String> newResult = new HashMap<>();
+			if ("OK".equals(originalResult.get("status"))) {
```
stavamichal, line 2478: All mentioned text representations of the input should be used as constants (to prevent a mistake in copy+paste).

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
zlamalp: It should be fixed.

stavamichal (approved):

PR #3066, opened 2021-01-27
Changed modules for login umbrellaID

balcirakpeter:
- The persistent shadow module now generates uuid instead of
  hash@domainName. Property domain name from configuration is not used
  anymore. Extsource name is still used for creating user extSource.
- Attribute filling will be triggered only when there is no value or
  when there is just an empty value in the shadow attribute.
- UserExtSource creation will be triggered only when a non-empty value
  was set to shadow attribute.
- Use case for this changes is that when proxy sends us umbrellaID
  login during the registration, then Perun will store it to the shadow
  attribute. When the login won't be sent, Perun will generate uuid with
  the first call of the persistent virtual attribute.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
@@ -1,6 +1,24 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
-import cz.metacentrum.perun.core.implApi.modules.attributes.UserPersistentShadowAttributeWithConfig;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.ExtSource;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.UserExtSource;
+import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.UserExtSourceExistsException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.impl.modules.ModulesConfigLoader;
+import cz.metacentrum.perun.core.impl.modules.ModulesYamlConfigLoader;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
```diff
@@ -1,48 +1,137 @@
...
-		return CONFIG_EXT_SOURCE_NAME_UMBRELLA_ID;
+	public Attribute fillAttribute(PerunSessionImpl perunSession, User user, AttributeDefinition attribute) {
+
+		Attribute filledAttribute = new Attribute(attribute);
+
+		if (attribute.getFriendlyName().equals(FRIENDLY_NAME)) {
+			long most64SigBits = get64MostSignificantBits();
+			long least64SigBits = get64LeastSignificantBits();
+
+			filledAttribute.setValue(new UUID(most64SigBits, least64SigBits));
```
zlamalp, line 54: I don't think you should set ``UUID`` object as attribute value. We support specific java types only and attribute is marked as ``String``, so maybe ``UUID.toString()`` ?

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
balcirakpeter: Copy-paste mistake. Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
```diff
@@ -1,48 +1,138 @@
...
-	private final static String CONFIG_DOMAIN_NAME_UMBRELLA_ID = "domainNameUmbrellaID";
+	private final static String FRIENDLY_NAME = "login-namespace:umbrellaid-persistent-shadow";
+	private final static String FRIENDLY_NAME_PARAMETER = "umbrellaid-persistent-shadow";
+
+	private final ModulesConfigLoader loader = new ModulesYamlConfigLoader();
 
+	/**
+	 * fillAttribute will generate a UUID based on the current timestamp,
+	 * measured in units of 100 nanoseconds from the 15th of October 1582,
+	 * concatenated with the MAC address of the device where the UUID is created.
```
zlamalp, line 43: I don't think that we use MAC address anywhere in UUID generation logic.

zlamalp (approved):

PR #3067, opened 2021-01-27
Remove support for HSQL database

balcirakpeter:
PR contains various commits that completely remove HSQL database support in Perun. For testing will be used PostgreSQL run in a docker container.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -914,10 +914,6 @@ public void removeAllUserExtSources(PerunSession sess, User user) {
 				return jdbc.query("select " + userMappingSelectQuery + "  from users " +
```
zlamalp, line 914: We don't plan to have a different DB backend for Perun, so you can probably remove whole ``if / else`` and simply use postgres SQL code.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -978,7 +974,7 @@ public void removeAllUserExtSources(PerunSession sess, User user) {
 
 		// the searchString is already lower cased
 		try {
-			if (Compatibility.isPostgreSql() || Compatibility.isHSQLDB()) {
+			if (Compatibility.isPostgreSql()) {
```
zlamalp, line 977: Same here, no need to check DB type, since there will be only one.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -522,9 +520,6 @@ public static int getNewId(JdbcTemplate jdbc, String sequenceName) {
 			case "postgresql":
 				query = "select nextval('" + sequenceName + "')";
 				break;
-			case "hsqldb":
```
zlamalp, line 525: Again, we can probably stop distinguish DBs, unless this method is useful in some ExtSource-like logic.

perun-dispatcher/src/test/java/cz/metacentrum/perun/dispatcher/unit/EventServiceResolverTest.java
```diff
@@ -29,10 +33,39 @@
...
+		// assign the group to resource
+		perun.getResourcesManagerBl().assignGroupToResource(sess, group1, resource1);
+
+	}
+
 	@Test
 	public void parseEventTest() throws ServiceNotExistsException, InvalidEventMessageException, PrivilegeException {
 		System.out.println("EventServiceResolver.parseEventTest()");
 
+		// FIXME - přesunout setup sem, jinde member není potřeba
```
zlamalp, line 67: Comments shouldn't be in Czech.

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/AppAutoRejectionSchedulerTest.java
```diff
@@ -180,11 +180,6 @@ private void setUpVo() throws Exception {
 	 * @throws GroupExistsException if group already exists
 	 */
 	private Group deleteApplicationFormAndCreateGroup() throws GroupExistsException {
```
zlamalp, line 182: We should probably rename this method and update javadoc, since its no longer deleting application form to fix weird behavior of HSQLDB.

perun-dispatcher/src/test/java/cz/metacentrum/perun/dispatcher/unit/TaskSchedulerTest.java
```diff
@@ -107,11 +81,16 @@ public void simpleRunTest() throws InterruptedException, ExecutionException, Tim
 
 		simpleFutureTask.run();
 		assertFalse(simpleSpy.testFailed);
+		try {
+			// FIXME - make every second test not to fail on interrupted exception
+			Thread.sleep(1000);
```
zlamalp, line 86: While this seem to work, success or failure of the test is still depending on async behaviour (system it runs on, workload etc.) and it might randomly fail on automatic testing. I would prefer annotating the tests with ``@Ignore`` for now.

perun-base/src/main/resources/perun-base.xml
```diff
@@ -90,7 +90,7 @@
 					perunRegistrar, perunSynchronizer, perunCabinet, perunNotifications, perunRpc, perunLdapc,
 					perunLdapInitializer
 				</prop>
-				<prop key="perun.db.type">hsqldb</prop>
+				<prop key="perun.db.type">postgresql</prop>
```
zlamalp, line 93: We should probably re-think necessity of this property/config option, since Perun supports only one DB backend now.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
balcirakpeter: Fixed

perun-dispatcher/src/test/java/cz/metacentrum/perun/dispatcher/unit/EventServiceResolverTest.java
balcirakpeter: I've removed the comment. I do not think it is bad to have set up in a separate method. It can be used in the future.

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/AppAutoRejectionSchedulerTest.java
balcirakpeter: Done

perun-dispatcher/src/test/java/cz/metacentrum/perun/dispatcher/unit/TaskSchedulerTest.java
balcirakpeter: @ignore put for the whole class.

perun-base/src/main/resources/perun-base.xml
balcirakpeter: Not removed at this PR.

zlamalp (approved):

PR #3072, opened 2021-02-02
umbrellaIDIdentifier fix

balcirakpeter:
- Changed filling of attribute umbrellaid-persistent-shadow. It now look
  to umbrellaIDIdentifier as default. It will use the value if it is set.
  It will generate a new value otherwise.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
@@ -6,9 +6,11 @@ import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.UserExtSource;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.UserExtSourceExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.impl.modules.ModulesConfigLoader;
 import cz.metacentrum.perun.core.impl.modules.ModulesYamlConfigLoader;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
```diff
@@ -31,26 +33,42 @@
...
+	private static final String A_U_UMBRELLAID_IDENTIFIER = AttributesManager.NS_USER_ATTR_DEF + ":umbrellaIDIdentifier";
 	private final static String CONFIG_EXT_SOURCE_NAME_UMBRELLA_ID = "extSourceNameUmbrellaID";
 	private final static String FRIENDLY_NAME = "login-namespace:umbrellaid-persistent-shadow";
 	private final static String FRIENDLY_NAME_PARAMETER = "umbrellaid-persistent-shadow";
 
 	private final ModulesConfigLoader loader = new ModulesYamlConfigLoader();
 
 	/**
-	 * fillAttribute will generate a version 1 UUID
+	 * fillAttribute will set umbrellaIDIdentifier attribute or generate a version 1 UUID if there is no value set for the attribute
```
zlamalp, line 44: ....will **set value from** umbrellaIDIdentifier...

Right now it seems, that it would fill sourcing attribute instead of itself.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
```diff
@@ -31,26 +33,42 @@
...
 
-			filledAttribute.setValue(uuid.toString());
+			Attribute umbrellaIDIdentifier = null;
+
+			try {
+				umbrellaIDIdentifier  = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, A_U_UMBRELLAID_IDENTIFIER);
+			} catch (WrongAttributeAssignmentException e) {
+				throw new InternalErrorException(e);
+			} catch (AttributeNotExistsException e) {
+				//this is ok, attribute will be generated.
```
zlamalp, line 60: I would at least ``log.warn()``, since it means, that attribute definition of sourcing attribute value doesn't exists. It is not mandatory, but otherwise you might not notice, that attribute is missing at all.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_umbrellaid_persistent_shadow.java
balcirakpeter: Fixed

zlamalp (approved):

PR #3073, opened 2021-02-02
User anonymization

cuadradek:
- New method anonymizeUser() was implemented. It works as replacement for
deleteUser() when it is not possible to delete the user.
- New config property ‘attributesToAnonymize’ was added, it contains list of
attributes names, which can’t be deleted and should be anonymized instead.
- New config property ‘attributesToKeep’ was added, it contains list of
attributes names, which can’t be deleted and should be kept untouched.

Difference from deleteUser():
- User attributes configured in property ‘attributesToAnonymize’ are not
deleted but anonymized instead. The anonymized value of an attribute is
determined from the attr module's method getAnonymizedValue().
- User attributes configured in property ‘attributesToKeep’ are not
deleted but kept untouched instead.

### DIFF
--- a/perun-base/src/main/resources/perun-base.xml
+++ b/perun-base/src/main/resources/perun-base.xml
@@ -66,6 +66,8 @@
 		<property name="queryTimeout" value="${perun.queryTimeout}" />
 		<property name="defaultLoaIdP" value="${perun.defaultLoa.idp}"/>
 		<property name="attributesToSearchUsersAndMembersBy" value="#{'${perun.attributesToSearchUsersAndMembersBy}'.split('\s*,\s*')}"/>
+		<property name="attributesToAnonymize" value="#{'${perun.attributesToAnonymize}'.split('\s*,\s*')}"/>
+		<property name="attributesToKeep" value="#{'${perun.attributesToKeep}'.split('\s*,\s*')}"/>
 	</bean>
 
 
@@ -131,6 +133,8 @@
 				<prop key="perun.queryTimeout">-1</prop>
 				<prop key="perun.defaultLoa.idp">2</prop>
 				<prop key="perun.attributesToSearchUsersAndMembersBy">urn:perun:user:attribute-def:def:preferredMail, urn:perun:member:attribute-def:def:mail</prop>
+				<prop key="perun.attributesToAnonymize"></prop>
+				<prop key="perun.attributesToKeep">urn:perun:user:attribute-def:def:login-namespace:*, urn:perun:user:attribute-def:def:uid-namespace:*</prop>
 				<!--
 				   this creates a map from OIDC issuer to user extsources that are used for looking up a user identified by "sub" claim
 				-->
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
@@ -1,5 +1,6 @@
 package cz.metacentrum.perun.core.api;
 
+import cz.metacentrum.perun.core.api.exceptions.AnonymizationNotSupportedException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.FacilityNotExistsException;
@@ -345,6 +346,21 @@ public interface UsersManager {
 	 * @throws SpecificUserAlreadyRemovedException
 	 */
 	void deleteUser(PerunSession perunSession, User user, boolean forceDelete) throws UserNotExistsException, PrivilegeException, RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException;
+
+	/**
+	 * Anonymizes user - according to configuration, each of user's attributes is either
+	 * anonymized, kept untouched or deleted. Also deletes other user's related data, e.g.
+	 * authorships of users publications, mail change and password reset requests, bans...
+	 *
+	 * @param perunSession
+	 * @param user
+	 * @throws InternalErrorException if an internal error has occurred
+	 * @throws UserNotExistsException if the user doesn't exist
+	 * @throws PrivilegeException if the method isn't called by perun admin
+	 * @throws RelationExistsException if the user has some members assigned
+	 * @throws AnonymizationNotSupportedException if an attribute should be anonymized but its module doesn't specify the anonymization process
+	 */
+	void anonymizeUser(PerunSession perunSession, User user) throws UserNotExistsException, PrivilegeException, RelationExistsException, AnonymizationNotSupportedException;
 
 	/**
 	 *  Updates users data in DB.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -41,6 +41,7 @@ import cz.metacentrum.perun.core.api.UserExtSource;
 import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyReservedLoginException;
+import cz.metacentrum.perun.core.api.exceptions.AnonymizationNotSupportedException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.BanNotExistsException;
@@ -460,6 +461,15 @@ public class UsersManagerBlImpl implements UsersManagerBl {
 
 	@Override
 	public void deleteUser(PerunSession sess, User user, boolean forceDelete) throws RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException {
+		try {
+			this.deleteUser(sess, user, forceDelete, false);
+		} catch (AnonymizationNotSupportedException ex) {
+			//this shouldn't happen with 'anonymizedInstead' set to false
+			throw new InternalErrorException(ex);
+		}
+	}
+
+	private void deleteUser(PerunSession sess, User user, boolean forceDelete, boolean anonymizeInstead) throws RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException, AnonymizationNotSupportedException {
 		List<Member> members = getPerunBl().getMembersManagerBl().getMembersByUser(sess, user);
 
 		if (members != null && (members.size() > 0)) {
@@ -526,8 +536,6 @@ public class UsersManagerBlImpl implements UsersManagerBl {
 		// delete them from DB
 		getUsersManagerImpl().deleteUsersReservedLogins(user);
 
-		// all users applications and submitted data are deleted on cascade when "deleteUser()"
-
 		// Remove all possible passwords associated with logins (stored in attributes)
 		for (Attribute loginAttribute: getPerunBl().getAttributesManagerBl().getLogins(sess, user)) {
 			try {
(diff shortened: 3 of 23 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -547,14 +549,40 @@ public void deleteUser(PerunSession sess, User user, boolean forceDelete) throws
...
+
+			// Users one
+			if (anonymizeInstead) {
+				List<String> attrNames = BeansUtils.getCoreConfig().getAttributesToAnonymize();
+				List<Attribute> userAttributes = getPerunBl().getAttributesManagerBl().getAttributes(sess, user);
+				for (Attribute attribute : userAttributes) {
+					// Skip core attributes
+					if (getPerunBl().getAttributesManager().isCoreAttribute(sess, attribute)) {
+						continue;
+					}
```
stavamichal, line 565: I think you should also skip virtual attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -547,14 +549,40 @@ public void deleteUser(PerunSession sess, User user, boolean forceDelete) throws
...
+					// Anonymize configured attributes
+					if (attrNames.contains(attribute.getName()) ||
+						// Attributes like 'login-namespace:mu' are configured as 'login-namespace:*'
+						(!attribute.getFriendlyName().equals(attribute.getBaseFriendlyName()) &&
```
stavamichal, line 569: You can probably use:
```
!attribute.getFriendlyNameParameter().isEmpty() &&
attrNames.contains(attribute.getNamespace() + ":" + attribute.getBaseFriendlyName() + ":*")
```

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -391,6 +392,45 @@ public void deleteUserWhenUserNotExists() throws Exception {
 		// shouldn't find user
 	}
 
+	@Test
+	public void anonymizeUser() throws Exception {
```
stavamichal, line 396: If the default set of anonymized attributes would be changed (anonymized only uids and logins by default) you need to also change the test accordingly.

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -3438,6 +3438,17 @@ public void checkAttributeSyntax(PerunSession sess, String key, Attribute attrib
 		attributeModule.checkAttributeSyntax((PerunSessionImpl) sess, key, attribute);
 	}
 
+	@Override
+	public Attribute getAnonymizedValue(PerunSession sess, User user, Attribute attribute) {
+		UserAttributesModuleImplApi attributeModule = getUserAttributeModule(sess, attribute);
+		if (attributeModule == null) {
+			log.debug("Cannot get anonymized attribute value. There is no module for: {}.", attribute.getName());
+			return attribute;
```
mvocu, line 3446: I think we should be able to anonymize also attributes without attribute modules.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserAttributesModuleAbstract.java
```diff
@@ -35,4 +35,11 @@ public Attribute fillAttribute(PerunSessionImpl session, User user, AttributeDef
 	public void changedAttributeHook(PerunSessionImpl session, User user, Attribute attribute) throws WrongReferenceAttributeValueException {
 
 	}
+
+	public Attribute getAnonymizedValue(PerunSessionImpl session, User user, Attribute attribute) {
+		Attribute anonymizedAttribute = new Attribute(attribute);
+		anonymizedAttribute.setValue(null);
```
mvocu, line 41: I think 'null' is not appropriate as anonymized value of attribute, it may have unforeseen consequences later. I suggest using some hash and possibly check the attribute syntax too (is the situtation when you return value which is subsequently refused in setValue(), excluded?)

mvocu (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
cuadradek: I don't think it is necessary. Every configured attribute should have specified its anonymization process in method getAnonymizedValue() - if it is not specified, then the attribute shouldn't be anonymized. What do you think @stavamichal ?

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserAttributesModuleAbstract.java
cuadradek: Maybe we could use the current attribute value as default value instead of 'null'. What do you think @stavamichal ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
stavamichal: I agree with @cuadradek If we need to anonymize specific attributes, we need to create a module for it that specifies how to do that. (the only question here are the core attributes, but we will not anonymize them for now at all - no reason for that so far)
I just changed my mind what is the default. The default shouldn't be null (like it is now) but it should be an old value without change (like in this specific situation). Can you change that? Because if you try to save the same value, we will do nothing (which is perfectly fine here). So if there is no module or the method uses the default setting, you should get the attribute without changed value and you can call setAttribute without any problem.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserAttributesModuleAbstract.java
stavamichal: The current value is the best option.

perun-base/src/main/resources/perun-base.xml
```diff
@@ -131,6 +132,7 @@
 				<prop key="perun.queryTimeout">-1</prop>
 				<prop key="perun.defaultLoa.idp">2</prop>
 				<prop key="perun.attributesToSearchUsersAndMembersBy">urn:perun:user:attribute-def:def:preferredMail, urn:perun:member:attribute-def:def:mail</prop>
+				<prop key="perun.attributesToAnonymize">urn:perun:user:attribute-def:def:login-namespace:*, urn:perun:user:attribute-def:def:uid-namespace:*, urn:perun:user:attribute-def:def:kerberosLogins, urn:perun:user:attribute-def:def:preferredMail, urn:perun:user:attribute-def:def:phone, urn:perun:user:attribute-def:def:mail, urn:perun:user:attribute-def:def:sshPublicAdminKey</prop>
```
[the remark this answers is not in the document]
cuadradek: Fixed.

perun-db/postgres.sql
```diff
@@ -30,6 +30,7 @@ create table users (
 	modified_by varchar default user not null,
 	service_acc boolean default false not null, --is it service account?
 	sponsored_acc boolean default false not null, --is it sponsored account?
+    anonymized boolean default false not null, --was user anonymized?
```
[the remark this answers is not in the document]
cuadradek: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -346,6 +346,20 @@
...
+	 * Anonymizes user - anonymizes user's attributes, which are configured to be anonymized,
+	 * and deletes remaining ones. Also deletes other user's related data, e.g. authorships
+	 * of users publications, mail change and password reset requests, bans...
+	 *
+	 * @param perunSession
+	 * @param user
+	 * @throws InternalErrorException
+	 * @throws UserNotExistsException
+	 * @throws PrivilegeException
+	 * @throws RelationExistsException
```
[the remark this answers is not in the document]
cuadradek: Added.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -463,6 +463,23 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 	},
 
+	/*#
+	 * Anonymizes user - anonymizes user's attributes, which are configured to be anonymized,
+	 * and deletes remaining ones. Also deletes other user's related data, e.g. authorships
+	 * of users publications, mail change and password reset requests, bans...
+	 *
+	 * @param user int User <code>id</code>
+	 */
+	anonymizeUser {
```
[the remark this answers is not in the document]
cuadradek: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
cuadradek: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
cuadradek: Fixed.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
cuadradek: The attributesToAnonymize property is changed in the test to contain only mail attribute. After anonymizing the user in the test, the property is set back to the original attribute set.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -346,6 +346,36 @@
...
+	 *
+	 * @param perunSession
+	 * @param user
+	 * @param forceDelete
+	 * @throws InternalErrorException
+	 * @throws UserNotExistsException
+	 * @throws PrivilegeException
+	 * @throws RelationExistsException
+	 */
+	void anonymizeUser(PerunSession perunSession, User user, boolean forceDelete) throws UserNotExistsException, PrivilegeException, RelationExistsException;
```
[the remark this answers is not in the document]
cuadradek: This part was outdated before.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserAttributesModuleAbstract.java
cuadradek: I changed default value to the current value.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
mvocu: Or you could provide default anonymization method for attributes without module here. You are creating a situation where it is possible to specify (in instance configuration) that certain attributes should be anonymized, but they are not (when called by admins) and the only error message is buried within log file, leaving the admin wondering what has just happened.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserAttributesModuleAbstract.java
mvocu: Should not it be an error if you want to anonymize attribute that is not capable of anonymization?

stavamichal: I feel that it is confusing, but what you want is to define a set of attributes you can't remove, and for these attributes, you want to ask Perun what you should set here as a value instead of the existing one. And if Perun will return the same value, it means there is no way how this value could be anonymized. Maybe the problem is in the name of the method, but the logic is right in my opinion.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
stavamichal: That's it and the default so far is "no change". It is the same question we are talking about in the comment below this one. The default situation is that there is no way how to the value should be anonymized. And if is there any, you will specify it in the module of the attribute.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/attributes/UserAttributesModuleAbstract.java
mvocu: OK, then where is the logic that prevents deletion of these attributes if someone calls deleteUser()? 
Also the configuration suggests that the attributes should be anonymized, not kept untouched. 

Is there a way of saying "this attribute must be kept only in a form that enables comparison, but not reading the actual value"?

stavamichal: We discussed this outside of this thread. The goal should be to create two arrays of processed attributes:
- attributes skipped by the removing/anonymization process
- attributes processed and anonymized (throwing an exception if there is no support in the module)

mvocu (approved):

stavamichal (approved):

PR #3075, opened 2021-02-03
Register Attribute Modules only for attributes with definitions

metodej:
- not all attributes are register during initialize method, but only such attributes which are already defined
- when new attribute is being created, the method also checks whether its module needs to be initialized (and registered for auditer listening) and does so
- also added tests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -2398,7 +2398,7 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		AttributeDefinition attributeToReturn = createAttribute(sess, attribute, true);
 
 		// try to initialize and register module
-		AttributesModuleImplApi module = getAttributesManagerImpl().getUninitiatedAttributesModule(sess, attributeToReturn);
+		AttributesModuleImplApi module = getAttributesManagerImpl().getUninitializedAttributesModule(sess, attributeToReturn);
 		if (module != null) {
 			getAttributesManagerImpl().initAttributeModule(module);
 			getAttributesManagerImpl().registerAttributeModule(module);
@@ -2565,6 +2565,13 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			} else {
 				log.warn("Inverse strong dependencies inconsistency. Inverse strong dependencies should contain information about {}. ", attributeDef);
 			}
+		}
+
+		// try to remove and unregister attribute module
+		AttributesModuleImplApi module = (AttributesModuleImplApi) getAttributesManagerImpl().getAttributesModule(sess, attribute);
+		if (module != null) {
+			getAttributesManagerImpl().removeAttributeModule(module);
+			getAttributesManagerImpl().unregisterAttributeModule(module);
 		}
 	}
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -2395,7 +2395,16 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 
 	@Override
 	public AttributeDefinition createAttribute(PerunSession sess, AttributeDefinition attribute) throws AttributeDefinitionExistsException {
-		return createAttribute(sess, attribute, true);
+		AttributeDefinition attributeToReturn = createAttribute(sess, attribute, true);
+
+		// try to initialize and register module
+		AttributesModuleImplApi module = getAttributesManagerImpl().getUninitiatedAttributesModule(sess, attributeToReturn);
+		if (module != null) {
+			getAttributesManagerImpl().initAttributeModule(module);
+			getAttributesManagerImpl().registerAttributeModule(module);
+		}
+
+		return attributeToReturn;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -7838,6 +7842,10 @@ protected void initialize() {
 		//Prepare all attribute definition from system perun
 		Set<AttributeDefinition> allAttributesDef = new HashSet<>(this.getAttributesDefinition(sess));
```
zlamalp, line 7843: Using ``allAttributesDef`` probably doesn't ensure, that all definitions are created, since some attributes are created by registrar and cabinet components and they are probably initialized later, since they depend on perun bean. Please move attribute definitions creation from registrar and cabinet into this init method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4508,21 +4509,52 @@ public Object getAttributesModule(PerunSession sess, AttributeDefinition attribu
 	}
 
 	@Override
-	public void initAttributeModules(ServiceLoader<AttributesModuleImplApi> modules) {
-		for (AttributesModuleImplApi module : modules) {
-			attributesModulesMap.put(module.getClass().getName(), module);
-			log.debug("Module {} loaded.", module.getClass().getSimpleName());
+	public AttributesModuleImplApi getUninitiatedAttributesModule(PerunSession sess, AttributeDefinition attributeDefinition) {
```
zlamalp, line 4512: There is a mixing of ``uninitialized`` and ``uninitiated``, first should be correct.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4508,21 +4509,52 @@ public Object getAttributesModule(PerunSession sess, AttributeDefinition attribu
...
-			Auditer.registerAttributeModule(module);
-			log.debug("Module {} was registered for audit message listening.", module.getClass().getName());
+			uninitializedAttributesModulesMap.put(module.getClass().getName(), module);
+		}
+
+		for (AttributeDefinition attributeDefinition : allAttributesDef) {
+			AttributesModuleImplApi module = getUninitiatedAttributesModule(sess, attributeDefinition);
+
+			if (module != null) {
+				attributesModulesMap.put(module.getClass().getName(), module);
```
zlamalp, line 4538: I would rather call ``initAttributeModule()``, which you added below. Also you can move logging messages in those methods.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -2394,7 +2394,16 @@ public void setAttributeInNestedTransaction(PerunSession sess, UserExtSource use
...
+
+		// try to initialize and register module
+		AttributesModuleImplApi module = getAttributesManagerImpl().getUninitiatedAttributesModule(sess, attributeToReturn);
+		if (module != null) {
+			getAttributesManagerImpl().initAttributeModule(module);
+			getAttributesManagerImpl().registerAttributeModule(module);
+		}
+
+		return attributeToReturn;
 	}
```
zlamalp, line 2407: We should have also opposite logic, which unregister module when attribute definition is deleted, otherwise, we could have messages about eg. virtual attribute changes even when such attribute doesn't exist anymore (at least until restart). We already do the same for attribute dependencies.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/AttributesManagerImplIntegrationTest.java
```diff
@@ -124,6 +126,34 @@ public void getRequiredAttributes_services_resource_group_ForOneService() {
 		assertThat(attributes).hasSize(1);
 	}
 
+	@Test
+	public void getUninitiatedAttributesModule() throws Exception {
```
zlamalp, line 130: Once method name is fixed to ``uninitialized``, you should also update tests names.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
metodej: Since the method `createAttribute()` was changed and now it also tries to initialize the module. And the cabinet/registrar attributes are created through this method, I think it should be fine.

metodej: Although it may be faster if they were moved into this init method... Should I move them?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
metodej: Added.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/AttributesManagerImplIntegrationTest.java
metodej: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
zlamalp: No, you are right, I didn't realize that. It should be fine.

zlamalp (commented):
Generally OK, I will just require a review also from someone else. I'm worried about thread safety, since Auditer is iterating over the registered modules when comiting transaction and we might be modifying that set from outside by creating attribute in another transaction.

zlamalp (approved):

balcirakpeter (approved):

PR #3076, opened 2021-02-04
OpenAPI - RichDestination type fix

xkostka2:
* Added missing property 'blocked' to RichDestination entity OpenAPI specification(RPC already returns it)

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -576,6 +576,7 @@ components:
       allOf:
         - $ref: '#/components/schemas/Destination'
         - properties:
+            blocked: {type: boolean}
             service: { $ref: '#/components/schemas/Service' }
             facility: { $ref: '#/components/schemas/Facility' }
       discriminator:


perun-openapi/openapi.yml
```diff
@@ -566,6 +566,7 @@ components:
       allOf:
         - $ref: '#/components/schemas/Auditable'
         - properties:
+            blocked: {type: boolean}
```
zlamalp, line 569: Shouldn't this be a part of ``RichDestination`` instead? I don't see such property or method in ``Destination`` object in perun core.

zlamalp (changes requested):

perun-openapi/openapi.yml
martin-kuba: It is in RichDestination, see https://github.com/CESNET/perun/blob/master/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichDestination.java#L55

zlamalp (approved):

PR #3079, opened 2021-02-08
Created method getRichMembersByIds

balcirakpeter:
- method fetch all rich members according to the provided list of ids.
  Attributes are fetched according to the list of attributes names.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
@@ -1447,4 +1447,14 @@ public interface MembersManager {
 	 * @return all namespaces rules
 	 */
 	List<NamespaceRules> getAllNamespacesRules();
+
+	/**
+	 * Returns rich members by their IDs with specific attributes.
+	 *
+	 * @param sess Perun session
+	 * @param ids List<Integer> list of members IDs
+	 * @param attrsNames List<String> Attribute names. If the list is null or empty, no attributes will be provided.
+	 * @return List<RichMember> RichMembers with specified IDs and attributes
+	 */
+	List<RichMember> getRichMembersByIds(PerunSession sess, List<Integer> ids, List<String> attrsNames) throws PrivilegeException, AttributeNotExistsException;
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1596,6 +1596,30 @@ public class MembersManagerEntry implements MembersManager {
 		getMembersManagerBl().moveMembership(sess, vo, sourceUser, targetUser);
 	}
 
+	@Override
+	public List<RichMember> getRichMembersByIds(PerunSession sess, List<Integer> ids, List<String> attrNames) throws PrivilegeException, AttributeNotExistsException {
+		Utils.checkPerunSession(sess);
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "getRichMembersByIds_List<Integer>_List<String>_policy")) {
+			throw new PrivilegeException(sess, "getRichMembersByIds");
+		}
+
+		List<AttributeDefinition> attributeDefinitions = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attributeDefinitions.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		List<Member> members = getMembersManagerBl().getMembersByIds(sess, ids);
+		members.removeIf(member -> !AuthzResolver.authorizedInternal(sess, "filter-getRichMembersByIds_List<Integer>_List<String>_policy", member));
+
+		List<RichMember> richMembers = membersManagerBl.convertMembersToRichMembers(sess, members);
+		//Enriched rich members with attributes by list of attributes
+		richMembers = membersManagerBl.convertMembersToRichMembersWithAttributes(sess, richMembers, attributeDefinitions);
+		//RichMembers with filtered attributes by rights from session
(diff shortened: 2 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1420,4 +1420,15 @@
 	 * @throws UserNotExistsException if there is no such user
 	 */
 	void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember, User sponsor, LocalDate newValidity) throws PrivilegeException, SponsorshipDoesNotExistException, MemberNotExistsException, UserNotExistsException;
+
+
+	/**
+	 * Returns rich members with specific attributes by their IDs.
```
stavamichal, line 1426: The name is a little confusing. It looks like you would like to get attributes by their IDs.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1420,4 +1420,15 @@
...
+	 *
+	 * @param sess Perun session
+	 * @param ids List<Integer> list of members IDs
+	 * @param attrsNames List<String> Attribute names
```
stavamichal, line 1430: Add here what happens if I send an empty list or null instead of this parameter. The same for the RPC and the open API if it is not already there.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1578,6 +1578,32 @@ public void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember,
...
+		List<AttributeDefinition> attributeDefinitions = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attributeDefinitions.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		List<RichMember> richMembers = membersManagerBl.convertMembersToRichMembers(sess, members);
+		//Enriched rich members with attributes by list of attributes
+		richMembers = membersManagerBl.convertMembersToRichMembersWithAttributes(sess, richMembers, attributeDefinitions);
+		//RichMembers with filtered attributes by rights from session
+		richMembers = membersManagerBl.filterOnlyAllowedAttributes(sess, richMembers);
```
stavamichal, line 1602: There is no need to save the output of filtering to the variable. You can return the whole output of the filtering instead.

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1578,6 +1578,30 @@ public void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember,
...
+		if (!AuthzResolver.authorizedInternal(sess, "getRichMembersByIds_List<Integer>_List<String>_policy")) {
+			throw new PrivilegeException(sess, "getRichMembersByIds");
+		}
+
+		List<Member> members = getMembersManagerBl().getMembersByIds(sess, ids);
+		members.removeIf(member -> !AuthzResolver.authorizedInternal(sess, "filter-getRichMembersByIds_List<Integer>_List<String>_policy", member));
+
+		List<AttributeDefinition> attributeDefinitions = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attributeDefinitions.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
```
zlamalp, line 1595: I would probably move this before getting members, since if there is non existent attribute definition in the list, we save up DB resources by failing before getting more data.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -812,6 +812,30 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Returns rich members by their IDs with specific attributes.
+	 *
+	 * @param ids List<Integer> list of members IDs
+	 * @param attrsNames List<String> Attribute names. If the list is null or empty, no attributes will be provided.
```
zlamalp, line 819: If parameter `attrNames` is optional, we create docs for each version of the method - with and without the parameter.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
balcirakpeter: Fixed

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
balcirakpeter: Fixed

zlamalp (approved):

stavamichal (approved):

PR #3083, opened 2021-02-11
New method getEnrichedFacilities

cuadradek:
- New class EnrichedFacility was created, it represents facility with its hosts
and destinations.
- New method getEnrichedFacilities was created and tested.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
@@ -115,6 +115,19 @@ public interface FacilitiesManager {
 	List<Facility> getFacilities(PerunSession perunSession) throws PrivilegeException;
 
 	/**
+	 * Gets all enriched facilities user has access rights to.
+	 * If User is:
+	 * - PERUNADMIN : all facilities
+	 * - FACILITYADMIN : only facilities where user is facility admin
+	 * - FACILITYOBSERVER: only facilities where user is facility observer
+	 *
+	 * @param perunSession
+	 * @return List of all EnrichedFacilities
+	 * @throws PrivilegeException insufficient permissions
+	 */
+	List<EnrichedFacility> getEnrichedFacilities(PerunSession perunSession) throws PrivilegeException;
+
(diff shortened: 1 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
```diff
@@ -114,6 +114,15 @@
 	 */
 	List<Facility> getFacilities(PerunSession perunSession) throws PrivilegeException;
 
+	/**
+	 * Gets all enriched facilities.
```
stavamichal, line 118: It should be more like "Gets all enriched facilities user has access rights to." Similar to getFacilities javadoc.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -161,6 +161,19 @@ public Integer call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Gets all enriched facilities.
```
stavamichal, line 165: The same change as for Entry.

stavamichal (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/FacilitiesManager.java
cuadradek: Fixed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
cuadradek: Fixed.

stavamichal (approved):

mvocu (approved):
Looks good.

Personally I do not like the concept of calling these classes, that serve only as messages in RPC, "Enriched", and mixing them with other classes that actually model something, but I understand this is a custom here.

PR #3086, opened 2021-02-15
Configuration support for sponsorship

balcirakpeter:
- Added support to configure, which attributes are forbidden, optional
  and required for specific namespace. It will be used in core logic to
  check wheter given parameters for a namespace are allowed/demanded.
- Method getAllNamespacesRules() will be used in NGUI to determine which
  namespaces to show and which attributes need to be filled in the
  frontend.
- Method getNamespaceRule("") will be used just in core logic.
- Added simple tests with simple config file.

### DIFF
--- a/perun-base/src/test/resources/sponsored-accounts-config.yml
+++ b/perun-base/src/test/resources/sponsored-accounts-config.yml
@@ -1,0 +1,9 @@
+---
+namespaces:
+
+  dummy_namespace:
+    required_attributes:
+      - login
+    optional_attributes:
+      - password
+...
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
@@ -7,6 +7,7 @@ import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.MemberGroupStatus;
+import cz.metacentrum.perun.core.api.NamespaceRules;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.RichMember;
@@ -33,6 +34,7 @@ import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotSponsoredException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotValidYetException;
 import cz.metacentrum.perun.core.api.exceptions.MemberResourceMismatchException;
+import cz.metacentrum.perun.core.api.exceptions.NamespaceRulesNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordCreationFailedException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthException;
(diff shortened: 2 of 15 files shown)


perun-base/src/test/resources/sponsored-accounts-config.yml
```diff
@@ -0,0 +1,8 @@
+---
+namespaces:
+
+  dummy_namespace:
+    attributes_rules:
+      login: REQUIRED
+      password: OPTIONAL
+...
```
metodej, line 8: I think there should be a new line at the end of the file.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/SponsoredAccountsConfigContainer.java
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * Get Rules for a specific namespace from the SponsoredAccountsConfigContainer
+	 *
+	 * @param namespace for which will be rules fetched
+	 * @return NamespaceRules for the namespace name
+	 * @throws NamespaceRulesNotExistsException of there are no rules for the namespace
+	 */
+	public NamespaceRules getNamespaceRules(String namespace) throws NamespaceRulesNotExistsException {
+		if (namespacesRules.get(namespace) == null) {
+			throw new NamespaceRulesNotExistsException("Namespace with name "+ namespace + "does not exists in the SponsoredAccountsConfigContainer.");
```
metodej, line 31: Just a typo:
```suggestion
			throw new NamespaceRulesNotExistsException("Namespace with name "+ namespace + "does not exist in the SponsoredAccountsConfigContainer.");
```

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -1767,4 +1769,20 @@
...
+	 * @return all namespaces rules
+	 */
+	List<NamespaceRules> getAllNamespacesRules();
+
+	/**
+	 * Get NamespaceRules for the namespace from the SponsoredAccountsConfigContainer
+	 *
+	 * @param namespace for which will be the rules fetched
+	 * @return NamespaceRules for the namespace
+	 * @throws NamespaceRulesNotExistsException of there are no rules for the namespace
```
metodej, line 1785: Just a typo 
```suggestion
	 * @throws NamespaceRulesNotExistsException if there are no rules for the namespace
```

perun-openapi/openapi.yml
```diff
@@ -8812,6 +8833,17 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/membersManager/getAllNamespacesRules:
+    get:
+      tags:
+        - MembersManager
+      operationId: getAllNamespacesRules
+      summary: Return all loaded namesapces rules.
```
metodej, line 8841: Just a typo:
```suggestion
      summary: Return all loaded namespaces rules.
```

metodej (approved):

perun-base/src/test/resources/sponsored-accounts-config.yml
```diff
@@ -0,0 +1,8 @@
+---
+namespaces:
+
+  dummy_namespace:
+    attributes_rules:
+      login: REQUIRED
+      password: OPTIONAL
```
Vojtech-Sassmann, line 7: I think it would make it easier to work with if the format was a bit different:
```suggestion
  dummy_namespace:
    required_attributes:
    - login
    optional_attributes:
    - password
```
Also, in the NamespaceRules class, it would be nice to have two Set<String> of attributes, one for optional and one for required attributes. Using just a Map<String, String> makes it much harder to decode the actual structure of the data. @balcirakpeter what do you think?

perun-base/src/test/resources/sponsored-accounts-config.yml
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/impl/SponsoredAccountsConfigContainer.java
balcirakpeter: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
balcirakpeter: Fixed

perun-openapi/openapi.yml
balcirakpeter: Fixed

perun-base/src/test/resources/sponsored-accounts-config.yml
balcirakpeter: Reworked

Vojtech-Sassmann (approved):

PR #3093, opened 2021-02-17
New method moveMembership

cuadradek:
- New method moveMembership was added, it moves membership in VO from source user
to target user.
- Source member is deleted and his attributes, sponsorships and bans are moved
to new target member. Target member is added only to non-synchronized groups of
source member.
- Method was added to CLI as well.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -53,6 +53,7 @@ import cz.metacentrum.perun.core.api.exceptions.AlreadySponsorException;
 import cz.metacentrum.perun.core.api.exceptions.AlreadySponsoredMemberException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.BanAlreadyExistsException;
 import cz.metacentrum.perun.core.api.exceptions.BanNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.CandidateNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
@@ -119,6 +120,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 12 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -2207,6 +2208,70 @@ public void getAllSponsoredMembers() throws Exception {
...
+		System.out.println(CLASS_NAME + "moveMembership");
+
+		User sourceUser = perun.getUsersManagerBl().getUserByMember(sess, createdMember);
+
+		Member member = setUpMember(createdVo);
+		User targetUser = perun.getUsersManagerBl().getUserByMember(sess, member);
+		// delete member - member was created just to create target user
+		perun.getMembersManager().deleteMember(sess, member);
+		// save user for deletion after test
+		usersForDeletion.add(targetUser);
```
Vojtech-Sassmann, line 2222: I think this is not necessary and it actually doesn't do anything.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -2207,6 +2208,70 @@ public void getAllSponsoredMembers() throws Exception {
...
+		BanOnVo banOnVo = new BanOnVo(-1, createdMember.getId(), createdMember.getVoId(), null, "ban");
+		perun.getVosManagerBl().setBan(sess, banOnVo);
+
+		perun.getMembersManager().moveMembership(sess, createdVo, sourceUser, targetUser);
+
+		Member targetMember = perun.getMembersManagerBl().getMemberByUser(sess, createdVo, targetUser);
+		List<Attribute> targetMemberAttrs = perun.getAttributesManagerBl().getAttributes(sess, targetMember);
+		List<Attribute> targetMemberGroupAttrs = perun.getAttributesManagerBl().getAttributes(sess, targetMember, createdGroup);
+
+		assertEquals("The source and target member attributes should have the same size.",
```
Vojtech-Sassmann, line 2245: Nit: using assertJ, you can simplify these asserts e.g. like this:
`assertThat(fooA).containsExactlyInAnyOrderElementsOf(fooB);`

AssertJ provides a lot of usefull assertions, that make tests easier to read and also write.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -2207,6 +2208,70 @@ public void getAllSponsoredMembers() throws Exception {
...
+
+		assertTrue("Ban on VO should be moved from the source member to the target member.",
+			perun.getVosManagerBl().getBanForMember(sess, targetMember.getId()).isPresent());
+
+		assertThatExceptionOfType(MemberNotExistsException.class)
+			.isThrownBy(() -> membersManagerEntry.getMemberById(sess, createdMember.getId()));
+	}
+
+	@Test(expected=AlreadyMemberException.class)
+	public void moveMembershipWhenAlreadyMember() throws Exception {
```
Vojtech-Sassmann, line 2266: Also, these tests don't verify all of the functionality provided by the method. Ideally, you would create a single test to validate a single functionality (e.g. one of coping of attributes, one for group memberships, etc.). However, it would be at least nice to test also the sponsorship copying functionality, even if it should be also part of a single giant test.

To create a single test per single functionality is a good practice because it makes the tests easier to read and also, if some of the tests fail, you know exactly what is wrong. This way, you might need to dig through a lot of code to find out, what caused the problem. However, you don't need to change it here, it is just something to think about in future.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2718,6 +2720,155 @@ public void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember,
...
+		getMembersManagerImpl().moveMembersApplications(sess, sourceMember, targetMember);
+
+		// move sponsorships in the VO
+		// move sponsorships where sourceUser is sponsor
+		List<Member> sponsoredMembers = this.getSponsoredMembers(sess, vo, sourceUser);
+		for (Member sponsoredMember : sponsoredMembers) {
+			try {
+				Sponsorship sponsorship = this.getSponsorship(sess, sponsoredMember, sourceUser);
+				// first add sponsorship to targetUser
+				this.sponsorMember(sess, sponsoredMember, targetUser, sponsorship.getValidityTo());
```
Vojtech-Sassmann, line 2810: If the targetUser doesn't have the SPONSOR role in the given Vo, we want to assign it to him so this would not fail.

Vojtech-Sassmann (changes requested):
Also please, add the new method to the OpenAPI specification.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2718,6 +2720,155 @@ public void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember,
 		return membersManagerImpl.getSponsorshipsExpiringInRange(sess, from, to);
 	}
 
+	@Override
+	public void moveMembership(PerunSession sess, Vo vo, User sourceUser, User targetUser) throws MemberNotExistsException, AlreadyMemberException, ExtendMembershipException, UserNotInRoleException {
```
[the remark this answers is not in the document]
cuadradek: Done.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
cuadradek: Removed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
cuadradek: Done.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
cuadradek: I created a single test per single functionality.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2718,6 +2720,208 @@ public void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember,
...
+			}
+		} catch (WrongAttributeAssignmentException | AttributeNotExistsException e) {
+			throw new InternalErrorException(e);
+		}
+
+		Member targetMember;
+		try {
+			// create targetMember and add him to all direct groups that are not synchronized
+			directGroups.removeAll(synchronizedGroups);
+			targetMember = this.createMember(sess, vo, targetUser, directGroups);
```
mvocu, line 2744: Is it possible that targetUser already is a member of this vo?

mvocu (commented):
Is it possible that targetUser is already a member of given VO? Just asking, not sure about the use-case here.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
cuadradek: We don't want to merge two existing memberships for now, therefore AlreadyMemberException is thrown in such a case.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2718,6 +2720,208 @@ public void updateSponsorshipValidity(PerunSession sess, Member sponsoredMember,
...
+		// move all applications
+		getMembersManagerImpl().moveMembersApplications(sess, sourceMember, targetMember);
+
+		// move sponsorships in the VO
+		moveMembersSponsorships(sess, vo, sourceUser, targetUser, sourceMember, targetMember);
+
+		// move members bans
+		moveMembersBans(sess, sourceMember, targetMember);
+
+		// finally delete sourceMember
```
Vojtech-Sassmann, line 2782: This is a nice example, why these inline comments don't (usually) make sense. They are usually used in long methods (as in this example) and this actually hints, that one should split the method into multiple smaller methods. In general, the code itself should be self explonatary (with a help of javadoc). After you split it, these comments are just a copy of the methods' names. It is better to actually omit these comments since, after some time, they might diverge from the code, and after that, it might not be clear, whether the comment is invalid or the actual code is invalid. Again, you don't need to change it here, it is just a tip :wink:

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
Vojtech-Sassmann: Well done! :tada:

Vojtech-Sassmann (approved):

mvocu (approved):

PR #3096, opened 2021-02-18
Added check to Destination and Host creation

metodej:
- destination name is checked by type when creating, new exception InvalidDestinationException is thrown if check fails
- hostname is checked when adding host, new exception InvalidHostnameException is thrown if check fails
- some hostnames in tests needed to be changed to match FQDN pattern

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
@@ -9,6 +9,7 @@ import cz.metacentrum.perun.core.api.exceptions.DestinationAlreadyRemovedExcepti
 import cz.metacentrum.perun.core.api.exceptions.DestinationNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.FacilityNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidDestinationException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.RelationExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ServiceAlreadyAssignedException;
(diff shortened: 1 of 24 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
```diff
@@ -849,13 +850,13 @@
 	 * @param destination (Id of this destination doesn't need to be filled.)
 	 * @return destination with it's id set.
 	 * @throws PrivilegeException
-	 * @throws InternalErrorException
+	 * @throws InvalidDestinationException if destination has invalid name
```
zlamalp, line 853: Hmm, its not exactly "name". Its more like invalid value (since param name is same as the object itself). Maybe also specify, that invalid value depends on type.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
```diff
@@ -980,7 +981,11 @@ public void setPerunBl(PerunBl perunBl) {
 				Destination destination = new Destination();
 				destination.setDestination(host.getHostname());
 				destination.setType(Destination.DESTINATIONHOSTTYPE);
-				destinations.add(this.addDestination(perunSession, service, facility, destination));
+				try {
+					destinations.add(this.addDestination(perunSession, service, facility, destination));
+				} catch (InvalidDestinationException e) {
+					throw new ConsistencyErrorException("Destination created out of hostname has invalid name.", e);
```
zlamalp, line 987: Not sure, if we should cover this up. `ConsistencyErrorException` is not expected to be fixed by user, but should be reported to support. In this case, user should be able to fix it by simply fixing Host name, so I would prefer simply throwing `InvalidDestinationException` again with updated text, that invalid "value" was taken from host.

Same for the other changed method.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/UtilsIntegrationTest.java
```diff
@@ -334,6 +338,110 @@ public void validateGroupNameThrowsACorrectExceptionForInvalidRegex() {
...
+		destination.setDestination("192.168.1.1");
+		Utils.checkDestination(destination);
+
+		destination.setDestination("hostname.test");
+		Utils.checkDestination(destination);
+	}
+
+	@Test
+	public void checkMailDestination() throws Exception {
+		System.out.println("Utils.checkHostDestination");
```
zlamalp, line 357: This should be the same as each test method name, but is same as first test in the most of these tests.

zlamalp (changes requested):
Please update also RPC javadoc for all changed addHost/addDestination methods and if it already exists, also update OpenAPI. Thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
metodej: But hosts are now checked when creating by the same pattern as destination of host type, so if we were able to create host with this hostname, but unable to create destination it is weird and it shouldn't happen...

zlamalp: You are probably right. I just thought about existing Hosts, which were created before the check was deployed, but from this point of view you are right. It shouldn't happen. So keep it as it is.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ServicesManager.java
metodej: Ok, changed to `if destination has invalid value in given destination type`.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/UtilsIntegrationTest.java
metodej: Fixed.

metodej:
> 
> 
> Please update also RPC javadoc for all changed addHost/addDestination methods and if it already exists, also update OpenAPI. Thank you.

I've updated the javadoc and description of these method to show when the new exceptions are thrown.

zlamalp (approved):

PR #3101, opened 2021-02-22
Core - generate password for sponsored user

Vojtech-Sassmann:
* If the password provided for the created sponsored user is blank, and
the sendActivationLink is set to true, we will now generate a random
password for the specified namespace.
* The problem was when users didn't want to specify the password for the
created user and wanted to let him choose it. However, some exteral
systems still requires a password, when creating an account. Therefore,
we have to generate some password in such cases.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
@@ -1445,7 +1445,8 @@ public interface MembersManagerBl {
 	 * @param vo virtual organization
 	 * @param namespace used for selecting external system in which guest user account will be created
 	 * @param name a map containing the full name or its parts (mandatory: firstName, lastName; optionally: titleBefore, titleAfter)
-	 * @param password password
+	 * @param password password, if the password is empty, and the `sendActivationLink` is set to true, this method will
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
```diff
@@ -1474,7 +1474,8 @@
 	 * @param vo virtual organization
 	 * @param namespace used for selecting external system in which guest user account will be created
 	 * @param name a map containing the full name or its parts (mandatory: firstName, lastName; optionally: titleBefore, titleAfter)
-	 * @param password password
+	 * @param password password, if the password is empty, and the `sendActivationLink` is set to true, this method will
```
zlamalp, line 1477: Can you please also update the javadoc for the method above and in above call hierarchy? Thank you.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/bl/MembersManagerBl.java
Vojtech-Sassmann: Done.

zlamalp (approved):

PR #3104, opened 2021-02-25
Created method for easier deletion of service from facility

HejdaJakub:
* Created method getAssignedResourcesByAssignedService which returns list of resources.
* This list will be displayed in GUI in remove service from facility dialog.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
@@ -63,7 +63,6 @@ import cz.metacentrum.perun.core.api.exceptions.WrongPatternException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
 import cz.metacentrum.perun.core.bl.FacilitiesManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
-import cz.metacentrum.perun.core.blImpl.AuthzResolverBlImpl;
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
```diff
@@ -398,6 +396,29 @@ public void copyOwners(PerunSession sess, Facility sourceFacility, Facility dest
...
+		List<Resource> allResources = getFacilitiesManagerBl().getAssignedResources(sess, facility);
+		List<Resource> resources = new ArrayList<>();
+		List<Service> services;
+
+		for (Resource resource: allResources) {
+			services = getPerunBl().getResourcesManagerBl().getAssignedServices(sess, resource);
+			if(services.contains(service)) {
+				resources.add(resource);
+			}
+		}
```
Vojtech-Sassmann, line 417: I have noticed, there is a method that can be used for this purpose.
```
List<Resource> getAssignedResources(PerunSession perunSession, Facility facility, Vo specificVo, Service specificService);
```
Just put `null` to the specificVo parameter and it should work.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/FacilitiesManagerEntry.java
HejdaJakub: Good point. I replaced the block of code by this method. I also added some checks for facility and service.

Vojtech-Sassmann (approved):

PR #3106, opened 2021-02-25
Core - general sponsorship

Vojtech-Sassmann:
* There are changes in the API! These changes require changes to be made
in the gui, so this will work. These changes are introduced in https://github.com/CESNET/perun-web-apps/pull/623
* Until now, it was impossible to sponsor users (or create sponsored
users) in any other namespace than MU. From now, it is possible to
sponsor users for other namespaces, and even without a namespace.
* It is possible to configure namespaces, that can be used to sponsor
members. This should be done in a
/etc/perun/sponsored-accounts-config.yml file.
* The code has been restructured, so it is can be used for such
purposes.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -66,6 +66,7 @@ import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.IllegalArgumentException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.InvalidLoginException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidSponsoredUserDataException;
 import cz.metacentrum.perun.core.api.exceptions.LoginNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MemberAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.MemberGroupMismatchException;
@@ -100,10 +101,14 @@ import cz.metacentrum.perun.core.implApi.ExtSourceSimpleApi;
 import cz.metacentrum.perun.core.implApi.MembersManagerImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.AbstractMembershipExpirationRulesModule;
 import cz.metacentrum.perun.core.implApi.modules.pwdmgr.PasswordManagerModule;
+import cz.metacentrum.perun.core.api.SponsoredUserData;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.beans.IntrospectionException;
+import java.beans.PropertyDescriptor;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDate;
@@ -129,15 +134,18 @@ import java.util.regex.Pattern;
 
 import static cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_vo_attribute_def_def_membershipExpirationRules.VO_EXPIRATION_RULES_ATTR;
 import static cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_vo_attribute_def_def_membershipExpirationRules.expireSponsoredMembers;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class MembersManagerBlImpl implements MembersManagerBl {
 
 	final static Logger log = LoggerFactory.getLogger(MembersManagerBlImpl.class);
 
+	private static final String ATTR_PREFIX = "urn:perun:";
 	private static final String EXPIRATION = AttributesManager.NS_MEMBER_ATTR_DEF + ":membershipExpiration";
 	private static final String A_U_PREF_MAIL = AttributesManager.NS_USER_ATTR_DEF + ":preferredMail";
 
-	private static final String NO_REPLY_EMAIL = "no-reply@muni.cz";
+	private static final String DEFAULT_NO_REPLY_EMAIL = "no-reply@perun-aai.org";
 
 	private static final String OK = "OK";
 	private static final String LOGIN = "login";
@@ -152,6 +160,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	);
 
 	public static final List<String> SPONSORED_MEMBER_ADDITIONAL_FIELDS = Arrays.asList(
+			"login",
 			AttributesManager.NS_USER_ATTR_DEF + ":note"
 	);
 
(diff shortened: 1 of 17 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
```diff
@@ -67,6 +67,27 @@ public EinfraPasswordManagerModule() {
 
 	}
 
+	@Override
+	public Map<String, String> generateAccount(PerunSession sess, Map<String, String> parameters) {
+		String password = parameters.get(PASSWORD_KEY);
+		String login = parameters.get(LOGIN_PREFIX + "einfra");
+
+		if (password == null || password.isBlank()) {
```
zlamalp, line 75: Both login and password are checked within ``reservePassword()`` method and with better debug/descriptions. If you insist on using `generateAccount()` method for purpose of reservation, I would rather add exceptions to API, so that caller knows what happened, instead of wrapping them to generic `InternalErrorException`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2451,7 +2430,9 @@ public Member setSponsoredMember(PerunSession session, Vo vo, User userToBeSpons
 			validateMember(session, sponsoredMember);
 		}
 
-		getPerunBl().getUsersManagerBl().validatePassword(session, userToBeSponsored, namespace);
+		if (data.getNamespace() != null && !data.getNamespace().isBlank()) {
```
zlamalp, line 2433: You can use library method `StringUtils.isNotBlank()`, it should handle also null input. This can be fixed on multiple places.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2998,41 +3014,51 @@ public void setPerunBl(PerunBl perunBl) {
...
+		if (data.getPassword() != null) {
+			p.put(PasswordManagerModule.PASSWORD_KEY, data.getPassword());
+		}
+		if (data.getLogin() != null) {
+			p.put(loginAttributeName, data.getLogin());
 		}
+
+		Map<String, String> r;
+		try {
+			r = getPerunBl().getUsersManagerBl().generateAccount(session, data.getNamespace(), p);
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Ok, let me make sure I understand it correctly. So you would like a general behavior for each namespace module, that would be named something like `handleSponsoredUserCreation`. The mu implementation would call the `generateAccount` method and the einfra implementation would call `reservePassword` ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
Vojtech-Sassmann: This is also checked in the `checkSponsoredUserData` method, which is used in the `MembersManagerBl`, and there is a special Exception thrown, with a more reasonable message. This was meant to be an internal check, more than a validation of the user's input.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: Thanks for the tip. I have used this method and also the `isBlank()` as it makes the code much more readable.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -217,6 +282,7 @@ public RichMember call(ApiCaller ac, Deserializer params) throws PerunException
 		public RichMember call(ApiCaller ac, Deserializer params) throws PerunException {
 			params.stateChangingCheck();
 			String password = params.readString("password");
+			String login = params.readString("login");
```
zlamalp, line 285: I think this way of reading makes the API param required, while it should actually depend on the namespace. You should read it like a "validityTo" param later in the code.

Also how this (in the call hierarchy) handles the case, when to be sponsored user already have login in namespace and we provided new login from input?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
```diff
@@ -67,6 +68,24 @@ public EinfraPasswordManagerModule() {
...
+	public String handleSponsorship(PerunSession sess, SponsoredUserData userData) {
+		if (StringUtils.isBlank(userData.getPassword())) {
+			throw new InternalErrorException("Einfra namespace requires a password.");
+		}
+		if (StringUtils.isBlank(userData.getNamespace())) {
+			throw new InternalErrorException("Einfra requires a login.");
+		}
+
+		try {
+			reservePassword(sess, userData.getLogin(), userData.getPassword());
```
zlamalp, line 81: I still think, that you should throw those exceptions outside and not wrap them.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
```diff
@@ -67,6 +68,24 @@ public EinfraPasswordManagerModule() {
 
 	}
 
+	@Override
+	public String handleSponsorship(PerunSession sess, SponsoredUserData userData) {
+		if (StringUtils.isBlank(userData.getPassword())) {
+			throw new InternalErrorException("Einfra namespace requires a password.");
+		}
+		if (StringUtils.isBlank(userData.getNamespace())) {
```
zlamalp, line 76: `reservePassword` itself performs checks on login and password based on namespace requirements, we shouldn't do it twice. Also you check namespace here but exception message mentions login.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2584,6 +2574,32 @@ private String getAttributeValueAsString (PerunSession session, User userToBeSpo
...
+	/**
+	 * Checks, if there is already a user with given login. If there is, throws the
+	 * InvalidSponsoredUserDataException.
+	 *
+	 * @param sess session
+	 * @param login login value
+	 * @param loginAttributeName login attribute urn
+	 * @throws InvalidSponsoredUserDataException if there is already a user with given login
+	 */
+	private void checkIfLoginIsFree(PerunSession sess, String login, String loginAttributeName) throws InvalidSponsoredUserDataException {
```
zlamalp, line 2586: I think we have similar logic somewhere (users manager?). If not, it should handle (check) also reserved logins from registrations, because they are not stored in attributes.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2996,43 +3043,48 @@ public void setPerunBl(PerunBl perunBl) {
 	}
 
 	/**
-	 * Creates a user account in external system if does not exist yet
+	 * Handles sponsorship for the given user using the given userData. This usually means that
+	 * an account in an external system is created, if needed.
```
zlamalp, line 3047: Please describe all use-cases, like when user already have login in specified namespace, input is ignored and no change is done to the users login/password.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/EinfraPasswordManagerModule.java
Vojtech-Sassmann: Good point. I have removed it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: Good point, for some reason, I overlooked it.

zlamalp (approved):

PR #3107, opened 2021-02-26
Inserting VO using PostgreSQL's insert-returning

cuadradek:
- Inserting VO is now done using PostgreSQL's insert-returning. Inserted row
is mapped to VO and returned.
- Until now, VO object passed to createVo() had its ID set in Impl layer and
this object was returned from Impl. Now, the passed VO object doesn't have
its ID set in Impl layer and new VO object is returned. Some tests were
changed because they expected the passed VO object to have its ID set.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/VosManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/VosManagerImpl.java
@@ -158,20 +158,13 @@ public class VosManagerImpl implements VosManagerImplApi {
 		if (this.voExists(sess, vo)) throw new VoExistsException(vo.toString());
 		if (this.shortNameForVoExists(sess, vo)) throw new VoExistsException(vo.toString());
 
-		// Get VO ID
-		int voId;
 		try {
-			voId = Utils.getNewId(jdbc, "vos_id_seq");
-			jdbc.update("insert into vos(id, name, short_name, created_by,modified_by, created_by_uid, modified_by_uid) values (?,?,?,?,?,?,?)",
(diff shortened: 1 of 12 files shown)


Vojtech-Sassmann (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/VosManagerImpl.java
```diff
@@ -158,20 +158,14 @@ public Vo createVo(PerunSession sess, Vo vo) throws VoExistsException {
...
 
-		// Get VO ID
-		int voId;
 		try {
-			voId = Utils.getNewId(jdbc, "vos_id_seq");
-			jdbc.update("insert into vos(id, name, short_name, created_by,modified_by, created_by_uid, modified_by_uid) values (?,?,?,?,?,?,?)",
-					voId, vo.getName(), vo.getShortName(), sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getUserId(), sess.getPerunPrincipal().getUserId());
+			// Get VO ID
+			int voId = Utils.getNewId(jdbc, "vos_id_seq");
+			return jdbc.queryForObject("insert into vos(id, name, short_name, created_by,modified_by, created_by_uid, modified_by_uid) values (?,?,?,?,?,?,?) returning " + voMappingSelectQuery,
```
zlamalp, line 164: We should replace usage of `Utils.getNewId(jdbc,"seq_name")` with direct SQL call within insert statement, like: `...values (nextval('vos_id_seq'),?,?,.....`

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/VosManagerImpl.java
cuadradek: Done

zlamalp (approved):

PR #3108, opened 2021-03-01
Core - fix getRichUsersFromListOfUsers methods

Vojtech-Sassmann:
* There were 2 methods that used to take a list of users as an input and
just fetched for these users their attributes. However, our
deserializers are forbiden to work with UUIDs (for some reason) and thus
when the rich users were missing their UUIDs.
* These methods were rewritten so they just used the ids from the given
list of users and fetch them from DB. This way, we can ensure that the
returned data are up to date and valid.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
@@ -319,7 +319,13 @@ public class UsersManagerEntry implements UsersManager {
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


stavamichal (approved):

zlamalp (approved):
It works, so OK. I'm just not sure, why we read whole User objects on input anyway. It's not used anywhere from the core (so that we would pass trusted user objects), just outer API. I think, this should be replaced with something like `getRichUsersByIds()`, since we introduced methods like this for all entities and we still retrieve users by IDs in implementation anyway.

PR #3109, opened 2021-03-02
researcher-access persistent-shadow module change

balcirakpeter:
- this module generated a hash@domain login. However, we were asked to
  change it to uuid@domain where uuid is a randomly generated (version 4).
  It is just a cosmetic change applied because uudi is shorter and better to read.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_researcher_access_persistent_shadow.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_researcher_access_persistent_shadow.java
@@ -1,6 +1,21 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
-import cz.metacentrum.perun.core.implApi.modules.attributes.UserPersistentShadowAttributeWithConfig;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_researcher_access_persistent_shadow.java
```diff
@@ -1,48 +1,113 @@
...
+	private final static String FRIENDLY_NAME_PARAMETER = "researcher-access-persistent-shadow";
 
-	@Override
-	public String getExtSourceConfigName() {
-		return CONFIG_EXT_SOURCE_NAME_RESEARCHER_ACCESS;
-	}
+	private final ModulesConfigLoader loader = new ModulesYamlConfigLoader();
 
+	/**
+	 * Format is: "[uuid]@{getDomainName()}" where [hash] represents a version 1 UUID (randomly generated).
```
zlamalp, line 39: In the commit you mention its version 4 UUID, but javadoc states version 1, so which is it?

zlamalp (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_researcher_access_persistent_shadow.java
balcirakpeter: Fixed

PR #3110, opened 2021-03-04
Core - createSponsoredMembersFromCsv, groups

Vojtech-Sassmann:
* From now, it is possible to specify, to which groups from the given vo
should be the created users assigned.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -845,7 +845,8 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 	@Override
 	public void addMember(PerunSession sess, List<Group> groups, Member member) throws WrongReferenceAttributeValueException, AlreadyMemberException, WrongAttributeValueException, GroupNotExistsException {
-		Collections.sort(groups, Collections.reverseOrder());
+		groups = new ArrayList<>(groups);
+		groups.sort(Collections.reverseOrder());
 		for (Group group : groups) {
 			// Check if the group is NOT members or administrators group
 			if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -151,6 +151,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	private static final String LOGIN = "login";
 	private static final String PASSWORD = "password";
 	private static final String STATUS = "status";
+	private static final String GROUP_ADDING_ERRORS = "group_adding_errors";
 	private static final String MEMBER = "member";
 
 	public static final List<String> SPONSORED_MEMBER_REQUIRED_FIELDS = Arrays.asList(
@@ -2444,7 +2445,8 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 
 	@Override
 	public Map<String, Map<String, String>> createSponsoredMembersFromCSV(PerunSession sess, Vo vo, String namespace,
-			List<String> data, String header, User sponsor, LocalDate validityTo, boolean sendActivationLink, String url, Validation validation) {
+			List<String> data, String header, User sponsor, LocalDate validityTo, boolean sendActivationLink,
+			String url, Validation validation, List<Group> groups) {
 
 		Map<String, Map<String, String>> totalResult = new HashMap<>();
 		Set<Member> createdMembers = new HashSet<>();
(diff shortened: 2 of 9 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -845,7 +845,8 @@ public boolean isDirectGroupMember(PerunSession sess, Group group, Member member
 
 	@Override
 	public void addMember(PerunSession sess, List<Group> groups, Member member) throws WrongReferenceAttributeValueException, AlreadyMemberException, WrongAttributeValueException, GroupNotExistsException {
-		Collections.sort(groups, Collections.reverseOrder());
+		groups = new ArrayList<>(groups);
+		groups.sort(Collections.reverseOrder());
```
zlamalp, line 849: Does sorting of input list breaks something or is it just some kind of precaution?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -3183,12 +3185,15 @@ private void processMemberAfterRemovingLastSponsor(PerunSession sess, Member spo
 
 			// we must pass member back for the purpose of validation
 			status.put(MEMBER, member);
-
+			if (groups != null) {
+				perunBl.getGroupsManagerBl().addMember(sess, groups, member);
```
zlamalp, line 3189: I would move this to the end to own try/catch, because it shouldn't be logged as error during sponsored user creation. 

Also when you iterate over the groups here, then we know which groups failed and which were ok and can report this back to GUI. Right now you also stop adding member to group on first error, which might later require more work for admin to get to expected state (find out and add members to all expected groups manually).

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
Vojtech-Sassmann: Both. As a precaution, this should usually be done, instead of modifying the input objects, since its unspecified side-effect. Also, without this, this method cannot work with unmodifiable lists.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: I have implemented these changes. In the returned results, for each member, there is a map of failed group assignments, where the key is the group id and the value is the error message. This can be used in GUI to show information about the assignments that failed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
zlamalp: OK. Thank you for explanation.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -3183,12 +3187,28 @@ private void processMemberAfterRemovingLastSponsor(PerunSession sess, Member spo
...
 			log.error("Failed to create a sponsored user.", e);
 			status.put(STATUS, e.getMessage());
 		}
+
+		if (groups != null && !groups.isEmpty()) {
+			Map<Integer, String> groupAssignmentErrors = new HashMap<>();
+			if (member != null) {
+				for (Group group : groups) {
+					try {
+						perunBl.getGroupsManagerBl().addMember(sess, groups, member);
```
zlamalp, line 3201: You should pass `group` and not `groups`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -3183,12 +3187,28 @@ private void processMemberAfterRemovingLastSponsor(PerunSession sess, Member spo
...
+					} catch (Exception e) {
+						groupAssignmentErrors.put(group.getId(), e.getMessage());
+						log.error("Failed to assign member to a group. Member: {}, Group: {}", member, group, e);
```
zlamalp, line 3204: We should probably refer to it as adding (member to group), not assignment.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -151,6 +151,7 @@
 	private static final String LOGIN = "login";
 	private static final String PASSWORD = "password";
 	private static final String STATUS = "status";
+	private static final String GROUP_ASSIGNMENT_ERRORS = "group_assigned_errors";
```
zlamalp, line 154: I think you should unify constant name and value (`ASSIGNMENT` vs. `assigned`), but I would prefer using `add`, so something like `group_adding_errors`?

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
Vojtech-Sassmann: Fixed.

zlamalp (approved):

PR #3111, opened 2021-03-04
Deleting unused destinations

cuadradek:
- Destinations were never deleted from DB, not even when they didn't have
any services assigned.
- Now, a destination is deleted when the last assigned service is removed
from the destination. Service denials and task results of the destination
are deleted as well.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
@@ -39,6 +39,7 @@ import cz.metacentrum.perun.core.provisioning.GroupsHashedDataGenerator;
 import cz.metacentrum.perun.core.provisioning.HierarchicalHashedDataGenerator;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.provisioning.HashedDataGenerator;
+import cz.metacentrum.perun.taskslib.model.TaskResult;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import cz.metacentrum.perun.core.api.Attribute;
@@ -85,6 +86,7 @@ import cz.metacentrum.perun.core.bl.ServicesManagerBl;
 import cz.metacentrum.perun.core.impl.Utils;
 import cz.metacentrum.perun.core.implApi.ServicesManagerImplApi;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
```diff
@@ -1032,4 +1063,23 @@ public void setPerunBl(PerunBl perunBl) {
 	public int getDestinationsCount(PerunSession sess) {
 		return getServicesManagerImpl().getDestinationsCount(sess);
 	}
+
+	@Override
+	public void deleteDestination(PerunSession sess, Destination destination) throws DestinationAlreadyRemovedException, RelationExistsException {
+		List<Service> services = getServicesManagerImpl().getServicesFromDestination(destination.getId());
+		if (!services.isEmpty()) {
+			throw new RelationExistsException("Destination has some services assigned.");
```
zlamalp, line 1071: I would probably state, that Destination is used by some services and facilities. Its not like its assigned to the service only and I don't think we refer to it as "assignment" when creating.
Same goes for the javadoc.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
```diff
@@ -899,4 +897,14 @@ public int getDestinationsCount(PerunSession sess) {
 		}
 	}
 
+	@Override
+	public void deleteDestination(PerunSession sess, Destination destination) throws DestinationAlreadyRemovedException {
+		try {
+			if (0 == jdbc.update("delete from destinations where id = ?", destination.getId())) {
```
zlamalp, line 903: I wonder, if we could catch `DataIntegrityViolationException` to handle the case of existing relations during the delete operation. Then we could also throw `RelationExistsException` from here. But I'm not sure, its actually thrown by Spring.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
cuadradek: Fixed

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
cuadradek: Yes, `DataIntegrityViolationException` is thrown there by Spring. So I added catch clause for it and `RelationExistsException` is thrown from there.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
```diff
@@ -899,4 +899,16 @@ public int getDestinationsCount(PerunSession sess) {
...
 	}
 
+	@Override
+	public void deleteDestination(PerunSession sess, Destination destination) throws DestinationAlreadyRemovedException, RelationExistsException {
+		try {
+			if (0 == jdbc.update("delete from destinations where id = ?", destination.getId())) {
+				throw new DestinationAlreadyRemovedException("Destination: " + destination);
+			}
+		} catch (DataIntegrityViolationException ex) {
+			throw new RelationExistsException("Destination: " + destination + " has existing relations.");
```
zlamalp, line 909: Please pass also original exception into the `RelationExistsException`.
```java
throw new RelationExistsException("Destination: " + destination + " has existing relations.", ex);
```

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ServicesManagerImpl.java
cuadradek: Done

zlamalp (approved):

PR #3114, opened 2021-03-05
DISPATCHER: Fix WAITING tasks re-scheduling

zlamalp:
- Do not re-schedule WAITING tasks when source data hasn't
  changed in rescheduleDoneTasks(), since they got more
  properly re-scheduled by endStuckTask() later
  or by TaskScheduler or on dispatcher restart.
- Use shorter and positive condition in IF when
  rescheduling all tasks on dispatcher restart.
- Improved comments to be more explanatory.
- This fixes the rare case of unlimited re-scheduling of
  WAITING tasks, when engine is not started and they can't
  be switched to PLANNED. It could fill up the whole waiting queue.
  Also it was logged as "task hasn't run for X hours", which is
  not true, since WAITING tasks have 'end time' NULL by design.


mvocu (approved):
Looks good to me.

PR #3117, opened 2021-03-09
Core - checkPasswordStrength

Vojtech-Sassmann:
* Added new method to the RPC API, that can be used to check the
strength of the given password, for the given namespace.
* For the MU namespace, the method is now calling the IS and returning
any errors received by the call.
* It is necessary to set the `perun.muPasswordStrengthTestLogin` with
the uco of a specific user created for the password test purpose.
* The logic responsible for the communication with the IS has been moved to a separate service.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -5174,6 +5174,12 @@ perun_policies:
     include_policies:
       - default_policy
 
+  checkPasswordStrength_String_String:
+    policy_roles:
+      - SELF:
+    include_policies:
+      - default_policy
+
   getGroupsWhereUserIsActive_Resource_User_policy:
     policy_roles:
       - VOOBSERVER: Vo
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
@@ -1225,6 +1225,17 @@ public interface UsersManager {
 	String changePasswordRandom(PerunSession sess, User user, String loginNamespace) throws PrivilegeException, PasswordOperationTimeoutException, LoginNotExistsException, PasswordChangeFailedException, InvalidLoginException, PasswordStrengthException;
 
 	/**
+	 * Check password strength for the given namespace. If the password is too weak,
+	 * the PasswordStrengthException is thrown
+	 *
+	 * @param password password, that will be checked
+	 * @param namespace namespace, that will be used to check the strength of the password
+	 *
+	 * @throws PasswordStrengthException When password doesn't match expected strength by namespace configuration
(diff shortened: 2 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ISServiceCallerImpl.java
```diff
@@ -0,0 +1,262 @@
...
+import java.io.StringReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
+ */
+public class ISServiceCallerImpl implements ISServiceCaller {
```
zlamalp, line 41: I understand, why you moved this code outside of the module to own class. On the other hand, no other component will be using it, since `ExtSourceISXML` is using own specific configuration and xml template.
I would probably move this class at least to the password managers modules package, since its not really shared core logic.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/ISResponseData.java
```diff
@@ -0,0 +1,36 @@
+package cz.metacentrum.perun.core.implApi;
+
+import org.w3c.dom.Document;
+
+/**
+ * @author Vojtech Sassmann <vojtech.sassmann@gmail.com>
```
zlamalp, line 6: Please describe class purpose in javadoc.

zlamalp (changes requested):

perun-base/src/main/resources/perun-base.xml
```diff
@@ -29,6 +29,7 @@
 		<property name="mailchangeBackupFrom" value="${perun.mailchange.backupFrom}"/>
 		<property name="mailchangeSecretKey" value="${perun.mailchange.secretKey}"/>
 		<property name="mailchangeValidationWindow" value="${perun.mailchange.validationWindow}"/>
+		<property name="muPasswordStrengthTestLogin" value="${perun.muPasswordStrengthTestLogin}"/>
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ISServiceCallerImpl.java
Vojtech-Sassmann: I have moved the classes.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/ISResponseData.java
Vojtech-Sassmann: Done.

zlamalp (approved):

PR #3123, opened 2021-03-12
Core - identity alerts

Vojtech-Sassmann:
* Implemented logic that can be used to sned alerts to users about an
identity being added or removed.
* This behaviour can be enabled by setting the perun.sendIdentityAlerts
property to true.
* Created entityless attribute that can be used to define templates for
these alerts.
* There can be 4 types of alerts: identityAdded(to preferred mail / to
the new identity email), and identityRemoved(to preferred mail / to the
removed identity email)
* In the templates, it is possible to user placeholders {organization}
and {login} where the actual data will be placed.
* These alerts should be send only if they were caused by the user
itself, not by admins for example.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
@@ -3,6 +3,7 @@ package cz.metacentrum.perun.core.entry;
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.AuthzResolver;
+import cz.metacentrum.perun.core.api.BeansUtils;
 import cz.metacentrum.perun.core.api.Candidate;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.Facility;
@@ -63,6 +64,8 @@ import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.bl.UsersManagerBl;
 import cz.metacentrum.perun.core.impl.Utils;
 import cz.metacentrum.perun.core.implApi.UsersManagerImplApi;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -77,6 +80,8 @@ import java.util.stream.Collectors;
  * @author Sona Mastrakova
  */
 public class UsersManagerEntry implements UsersManager {
+
+	private final static Logger log = LoggerFactory.getLogger(UsersManagerEntry.class);
 
 	private UsersManagerBl usersManagerBl;
 	private PerunBl perunBl;
@@ -601,7 +606,18 @@ public class UsersManagerEntry implements UsersManager {
 					"It means it can not be removed.");
 		}
 
+		List<Attribute> attrs = perunBl.getAttributesManagerBl().getAttributes(sess, userExtSource);
+
 		getUsersManagerBl().removeUserExtSource(sess, user, userExtSource);
+
+		if (BeansUtils.getCoreConfig().isSendIdentityAlerts() &&
+				user.getId() == sess.getPerunPrincipal().getUserId()) {
+			try {
+				Utils.sendIdentityRemovedAlerts(sess, userExtSource, attrs);
+			} catch (Exception e) {
+				log.error("Failed to send identity removed alerts.", e);
+			}
+		}
 	}
 
 	@Override
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
@@ -1,6 +1,5 @@
 package cz.metacentrum.perun.core.impl;
 
-import com.zaxxer.hikari.HikariDataSource;
 import cz.metacentrum.perun.core.api.Attribute;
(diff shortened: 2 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -2021,4 +2049,279 @@ public static MapSqlParameterSource getMapSqlParameterSourceToSearchUsersOrMembe
...
+	 * @param templateName name of the template for the message
+	 * @param defaultTemplate default template for the message
+	 * @param subjectName name of the subject template
+	 * @param defaultSubject default template for the subject
+	 */
+	private static void sendIdentityRemovedAlert(PerunSession sess, UserExtSource removedUes, List<Attribute> uesAttrs,
+	                                             String email, String templateName, String defaultTemplate,
+	                                             String subjectName, String defaultSubject) {
+		String message = getAlertTemplate(sess, templateName, defaultTemplate);
+		message = message.replaceAll(LOGIN_PLACEHOLDER, removedUes.getLogin());
```
zlamalp, line 2176: I don't think that you need to use `replaceAll()`, just use standard `replace()`. Then you can simplify the constants to plain text form by removing `\\` in them. This apply to all new occurrences.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_identityAlertsTemplates.java
```diff
@@ -0,0 +1,133 @@
...
+
+	public static final String ORG_UNKNOWN_TEXT = "<unknown>";
+	public static final DateTimeFormatter ALERT_TIME_FORMATTER = DateTimeFormatter.ofPattern("MM/dd/yyyy - HH:mm:ss z");
+
+	public static final String KEY_EN = "en";
+
+	public static final String DEFAULT_IDENTITY_ADDED_PREF_MAIL_SUBJECT = "New identity added";
+	public static final String DEFAULT_IDENTITY_ADDED_UES_MAIL_SUBJECT = "Account linked";
+	public static final String DEFAULT_IDENTITY_ADDED_PREF_MAIL = """
+			A new identity has been added to your account. If you don't recognize this activity, please contact our support at perun@cesnet.cz.
```
zlamalp, line 45: Just a note: I'd like to have this mail address configurable too, but we probably don't have any shared mail config (from/support mail). It`s partly divided between core and gui config. We should unify those settings where possible and then re-use it. I'll plan this for the upcoming cycles.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -2021,4 +2049,279 @@ public static MapSqlParameterSource getMapSqlParameterSourceToSearchUsersOrMembe
...
+	private static void sendIdentityAddedAlert(PerunSession sess, UserExtSource newUes, String email, String templateName,
+	                                           String defaultTemplate, String subjectName, String defaultSubject) {
+		String message = getAlertTemplate(sess, templateName, defaultTemplate);
+		message = message.replaceAll(LOGIN_PLACEHOLDER, newUes.getLogin());
+		message = insertOrganization(sess, message, newUes);
+		message = insertCurrentTime(message);
+
+		String subject = getAlertTemplate(sess, subjectName, defaultSubject);
+
+		Utils.sendEmail(subject, message, email);
```
zlamalp, line 2205: Whole action of adding/removing identity can now fail if mail can't be sent. I'm not sure if it`s desired. We should discuss it with requestor and probably internally and unify this behavior on other actions too.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -601,7 +602,14 @@ public void removeUserExtSource(PerunSession sess, User user, UserExtSource user
 					"It means it can not be removed.");
 		}
 
+		List<Attribute> attrs = perunBl.getAttributesManagerBl().getAttributes(sess, userExtSource);
+
 		getUsersManagerBl().removeUserExtSource(sess, user, userExtSource);
+
+		if (BeansUtils.getCoreConfig().isSendIdentityAlerts() &&
+				user.getId() == sess.getPerunPrincipal().getUser().getId()) {
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
Vojtech-Sassmann: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
Vojtech-Sassmann: Good point, however, I would suggest just catching such exceptions and logging them. I don't think we would like the whole process to fail when some email couldn't be sent.

However, I will ask the requestor, if such behavior is OK.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_entityless_attribute_def_def_identityAlertsTemplates.java
Vojtech-Sassmann: Ok, for now, the email could be just changed with the whole template in the entityless attribute.

zlamalp (approved):

PR #3126, opened 2021-03-16
Reducing debug logs in AuditLogger

Johaney-s:
- Not logging reads where no messages are obtained
- Wait for 3 seconds for new bulk of messages

### DIFF
--- a/perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
+++ b/perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
@@ -82,19 +82,20 @@ public class EventLoggerImpl implements EventLogger, Runnable {
 
 				int sleepTime = 1000;
 				//Waiting for new messages. If consumer failed in some internal case, waiting until it will be repaired (waiting time is increases by each attempt)
-				while(messages == null) {
+				while(messages == null || messages.isEmpty()) {
 					try {
 						//IMPORTANT STEP1: Get new bulk of messages
+						log.debug("Waiting for audit messages.");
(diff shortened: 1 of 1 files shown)


perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
```diff
@@ -81,20 +81,21 @@ public void run() {
...
 					} catch (InternalErrorException ex) {
 						log.error("Consumer failed due to {}. Sleeping for {} ms.", ex, sleepTime);
 						Thread.sleep(sleepTime);
 						sleepTime += sleepTime;
 					}
 
-					//If there are no messages, sleep for 1 sec and then try it again
-					if (messages == null) Thread.sleep(1000);
+					//If there are no messages, sleep for 3 sec and then try it again
+					if (messages == null) Thread.sleep(3000);
```
zlamalp, line 98: `messages` are actually `null` only when this process is started. Method filling it with data returns empty list if there are no new messages. We should probably switch the initial `while` condition. Then we would keep this cycle going until new message arrives. Right now it still logs 3 messages each 5 seconds.
Also maybe use 5 seconds here, since outer cycle (`while (running)`) is doing the same.

perun-auditlogger/src/main/java/cz/metacentrum/perun/auditlogger/logger/impl/EventLoggerImpl.java
```diff
@@ -81,20 +81,21 @@ public void run() {
 			while (running) {
 
 				int sleepTime = 1000;
+				log.debug("Waiting for audit messages.");
```
zlamalp, line 84: Regarding my other comment, this logging should be probably moved into inner `while` cycle.

zlamalp (changes requested):

zlamalp (approved):

PR #3134, opened 2021-03-24
Registrar - visibility and disabled fields

Vojtech-Sassmann:
* For some use-cases, users need to be able to specify fields, that
should be hidden or disabled in some cases. Therefore, new
applicationFormItem fields were created - `hidden` and `disabled`. There
is also a new flag called `updatable`. This flag can be used to
determine if some application field can be updated after submission.
However, some item types cannot still be updated, even if this flag is
set to true (e.g. username).
* It is possible to specify 4 situations, when items should be hidden or
disabled - ALWAYS, NEVER, IF_PREFILLED, IF_EMPTY.
* It is also possible to reference a different item in the form, which
is then used to evaluate the conditions.
* The application types FROM_FEDERATION_SHOW and FROM_FEDERATION_HIDDEN
were merged into a FROM_FEDERATION type. In the dbchangelog, there are
commands, that will transform the old types into the new one, with also
setting the `hidden` and `disabled` columns accordingly.

### DIFF
--- a/perun-core/src/main/resources/postgresChangelog.txt
+++ b/perun-core/src/main/resources/postgresChangelog.txt
@@ -6,6 +6,21 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.77
+create type app_item_disabled as enum ('NEVER','ALWAYS','IF_PREFILLED','IF_EMPTY');
+create type app_item_hidden as enum ('NEVER','ALWAYS','IF_PREFILLED','IF_EMPTY');
+ALTER TABLE application_form_items ADD COLUMN hidden app_item_hidden not null default 'NEVER';
+ALTER TABLE application_form_items ADD COLUMN disabled app_item_disabled not null default 'NEVER';
+ALTER TABLE application_form_items ADD COLUMN hidden_dependency_item_id integer;
+ALTER TABLE application_form_items ADD COLUMN disabled_dependency_item_id integer;
+ALTER TABLE application_form_items ADD COLUMN updatable boolean default true not null;
+ALTER TABLE application_form_items ADD CONSTRAINT applfrmit_hd foreign key (hidden_dependency_item_id) references application_form_items(id) ON DELETE SET NULL;
+ALTER TABLE application_form_items ADD CONSTRAINT applfrmit_dd foreign key (disabled_dependency_item_id) references application_form_items(id) ON DELETE SET NULL;
+UPDATE application_form_items SET hidden = 'ALWAYS' where type = 'FROM_FEDERATION_HIDDEN';
+UPDATE application_form_items SET disabled = 'ALWAYS' where type = 'FROM_FEDERATION_SHOW';
+UPDATE application_form_items SET updatable = false, type = 'TEXTFIELD' where type = 'FROM_FEDERATION_SHOW' OR type = 'FROM_FEDERATION_HIDDEN';
+UPDATE configurations SET value='3.1.77' WHERE property='DATABASE VERSION';
+
 3.1.76
 ALTER TABLE pwdreset ADD COLUMN uu_id uuid not null default gen_random_uuid();
 UPDATE configurations SET value='3.1.76' WHERE property='DATABASE VERSION';
--- a/perun-db/postgres.sql
+++ b/perun-db/postgres.sql
@@ -1,4 +1,4 @@
--- database version 3.1.76 (don't forget to update insert statement at the end of file)
+-- database version 3.1.77 (don't forget to update insert statement at the end of file)
 
 -- VOS - virtual organizations
 create table vos (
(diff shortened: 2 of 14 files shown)


Vojtech-Sassmann:
We should probably merge this PR after we test these changes in the Registrar WUI.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2627,7 +2657,7 @@ public void updateFormItemData(PerunSession sess, int appId, ApplicationFormItem
 		ApplicationFormItemData existingData = getFormItemDataById(data.getId(), appId);
 		if (existingData == null) throw new RegistrarException("Form item data specified by ID: "+ data.getId() + " not found or doesn't belong to the application "+appId);
 
-		List<Type> notAllowed = Arrays.asList(FROM_FEDERATION_HIDDEN, FROM_FEDERATION_SHOW, USERNAME, PASSWORD, HEADING, HTML_COMMENT, SUBMIT_BUTTON, AUTO_SUBMIT_BUTTON);
+		List<Type> notAllowed = Arrays.asList(FROM_FEDERATION, USERNAME, PASSWORD, HEADING, HTML_COMMENT, SUBMIT_BUTTON, AUTO_SUBMIT_BUTTON);
```
balcirakpeter, line 2660: This is not correct. Users can change the prefilled value from fed. He cannot change the value only id the field is Hidden/Disabled. Also if it is possible, add some checks to the submitApplication method, so user does not change disabled/hidden value.

balcirakpeter (changes requested):

perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,18 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.76
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Fixed.

perun-db/postgres.sql
```diff
@@ -486,10 +486,28 @@ create table application_form (
...
+	'ALWAYS',
+	'IF_PREFILLED',
+	'IF_EMPTY'
+	);
+
 -- APPLICATION_FORM_ITEMS - items of application form
 create table application_form_items (
 	id integer not null,
 	form_id integer not null,  --identifier of form (application_form.id)
+	hidden app_item_hidden not null default 'NEVER',
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2354,7 +2383,7 @@ public void updateFormItem(PerunSession sess, ApplicationFormItem item) throws P
 
 				// We do require value from IDP (federation) if attribute is supposed to be pre-filled and item is required and not editable to users
 				if ((itemW.getPrefilledValue() == null || itemW.getPrefilledValue().isEmpty()) && itemW.getFormItem().isRequired() &&
-						(Type.FROM_FEDERATION_HIDDEN.equals(itemW.getFormItem().getType()) || Type.FROM_FEDERATION_SHOW.equals(itemW.getFormItem().getType()))) {
+						(Type.FROM_FEDERATION == itemW.getFormItem().getType())) {
```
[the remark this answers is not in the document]
Vojtech-Sassmann: Fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Vojtech-Sassmann: Fixed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2352,17 +2387,20 @@ public void updateFormItem(PerunSession sess, ApplicationFormItem item) throws P
...
-						log.error("Couldn't resolve displayName from: {}, parsedNames were: {}", federValues, parsedName);
-					}
+		Map<Integer, ApplicationFormItemWithPrefilledValue> allItemsByIds = itemsWithValues.stream()
+				.collect(toMap(item -> item.getFormItem().getId(), Function.identity()));
 
-					itemsWithMissingData.add(itemW);
+		for (ApplicationFormItemWithPrefilledValue itemW : itemsWithValues) {
+			// We do require value from IDP (federation) if attribute is supposed to be pre-filled and item is required and not editable to users
+			if ((itemW.getPrefilledValue() == null || itemW.getPrefilledValue().isEmpty()) &&
+					itemW.getFormItem().isRequired() && isItemHidden(itemW, allItemsByIds)) {
```
balcirakpeter, line 2399: What about disabled items?

balcirakpeter (changes requested):
Please change also the commit message.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Vojtech-Sassmann: Good point, I added a check for that as well.

balcirakpeter (approved):

Johaney-s (approved):

PR #3137, opened 2021-03-26
Suspending propagating tasks to engine

Johaney-s:
- From CLI allows stopping sending waiting tasks to the engine

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/TasksManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/TasksManager.java
@@ -244,5 +244,13 @@ public interface TasksManager {
 	 */
 	List<Task> listAllTasksInState(PerunSession perunSession, Task.TaskStatus state) throws PrivilegeException;
 
+	/**
+	 * Suspends tasks propagation to engine.
+	 *
+	 * @param perunSession
+	 * @param suspend True to suspend propagation, false to resume propagation
+	 * @throws PrivilegeException
+	 */
+	void suspendTasksPropagation(PerunSession perunSession, boolean suspend) throws PrivilegeException;
 
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/TasksManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/TasksManagerBlImpl.java
@@ -41,6 +41,8 @@ public class TasksManagerBlImpl implements TasksManagerBl {
 	@Autowired
 	protected TasksManagerImplApi tasksManagerImpl;
 
+	private static boolean suspendedTasksPropagation = false; //are tasks stopped from propagating to engine
+
 	// -------------- constructors
 	
 	public TasksManagerBlImpl(TasksManagerImplApi tasksManagerImpl) {
@@ -398,6 +400,20 @@ public class TasksManagerBlImpl implements TasksManagerBl {
 	@Override
 	public void updateTask(PerunSession sess, Task task) {
 		getTasksManagerImpl().updateTask(task);
+	}
+
+	@Override
+	public void suspendTasksPropagation(PerunSession perunSession, boolean suspend) {
+		synchronized(TasksManagerBlImpl.class) {
+			suspendedTasksPropagation = suspend;
+		}
+	}
+
+	@Override
+	public boolean isSuspendedTasksPropagation() {
+		synchronized(TasksManagerBlImpl.class) {
+			return suspendedTasksPropagation;
+		}
 	}
 
 }
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
@@ -148,6 +148,10 @@ public class TaskScheduler extends AbstractRunner {
 		TaskSchedule schedule;
 		while (!shouldStop()) {
 			try {
+				if (tasksManagerBl.isSuspendedTasksPropagation()) {
+					// do not continue sending tasks to the engine until propagation is set to resume
+					waitForResumingPropagation();
+				}
 				schedule = getWaitingTaskSchedule();
 			} catch (InterruptedException e) {
 				String message = "Thread was interrupted, cannot continue.";
@@ -212,6 +216,22 @@ public class TaskScheduler extends AbstractRunner {
 		}
 		log.trace("[{}] Returning Task schedule {}.", taskSchedule.getTask().getId(), taskSchedule);
 		return taskSchedule;
+	}
+
+	/**
+	 * Method waiting for propagation of tasks to the engine to be resumed.
+	 * Called when propagation was suspended.
+	 * @throws InterruptedException Waiting thread was interrupted.
+	 */
+	private void waitForResumingPropagation() throws InterruptedException {
+		int sleepTime = 10000;
+		while (tasksManagerBl.isSuspendedTasksPropagation()) {
+			log.debug("Propagation of tasks is suspended.");
+			log.debug(schedulingPool.getReport());
+			log.debug("WaitingTasksQueue has {} normal Tasks and {} forced Tasks.", waitingTasksQueue.size(), waitingForcedTasksQueue.size());
+			Thread.sleep(sleepTime);
+		}
+		log.debug("Propagation of tasks is resumed.");
 	}
 
 	/**
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -15339,6 +15339,20 @@ paths:
                 destinationName: { type: string, description: "destination name", nullable: false }
                 destinationType: { type: string, description: "destination type", nullable: false }
 
+  /json/tasksManager/suspendTasksPropagation:
+    post:
+      tags:
+        - TasksManager
+      operationId: suspendTasksPropagation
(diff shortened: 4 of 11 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/TasksManagerMethod.java
```diff
@@ -326,6 +326,31 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 */
+	resumeTasksPropagation {
+		public Integer call(ApiCaller ac, Deserializer parms) throws PerunException {
+			ac.getTasksManager().suspendTasksPropagation(
+				ac.getSession(),
+				false);
+			return null;
+		}
 	};
+	;
```
cuadradek, line 354: unnecessary semicolon

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
```diff
@@ -147,6 +147,10 @@ public void run() {
 		log.debug("Pool contains {} tasks in total", schedulingPool.getSize());
 		TaskSchedule schedule;
 		while (!shouldStop()) {
+			if (tasksManagerBl.isSuspendedTasksPropagation()) {
+				log.debug("Propagation of tasks is suspended.");
+				break;
```
cuadradek, line 152: I would maybe use continue instead of break, so the propagation of tasks can be resumed later.

cuadradek (changes requested):
Also, please add methods `suspendTasksPropagation` and `resumeTasksPropagation` to OpenAPI.

perun-cli/resumePropagation
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/perl
```
zlamalp, line 1: I would prefer to have same naming scheme for both tools. Please rename this file to `resumeTasksPropagation`.

perun-core/src/main/java/cz/metacentrum/perun/core/api/TasksManager.java
```diff
@@ -244,5 +244,13 @@
...
 	List<Task> listAllTasksInState(PerunSession perunSession, Task.TaskStatus state) throws PrivilegeException;
 
+	/**
+	 * Suspends tasks propagation to engine.
+	 *
+	 * @param perunSession
+	 * @param val True to suspend propagation, false to resume propagation
+	 * @throws PrivilegeException
+	 */
+	void suspendTasksPropagation(PerunSession perunSession, boolean val) throws PrivilegeException;
```
zlamalp, line 254: Using something like `suspend` instead of `val` for variable name will be generally better. Please update it in a whole method chain.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/TasksManagerBlImpl.java
```diff
@@ -41,6 +41,8 @@
 	@Autowired
 	protected TasksManagerImplApi tasksManagerImpl;
 
+	private boolean suspendedTasksPropagation = false; //are tasks stopped from propagating to engine
```
zlamalp, line 44: TaskManager should be singleton in a Spring config, but you probably want this to be static variable.

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
zlamalp: I agree, since if you stop this whole thread, there is nothing to start it again, just server restart, which is not desired.

Also be careful, since there is no waiting in the loop, if you wont get to the `getWaitingTaskSchedule()` which internally polls queue with 10s timeout, you will hype one CPU core to max and fill logs with that debug message. Maybe create some simple private method to do the waiting, so that code here remains clear.

Your solution seems to be good otherwise, since you keep tasks in the queue and they will be processed once propagation is resumed.

zlamalp (changes requested):

Johaney-s:
Thanks for the reviews,
> Also, please add methods `suspendTasksPropagation` and `resumeTasksPropagation` to OpenAPI.

should `resumeTasksPropagation` be added to OpenAPI too, if the TasksManager contains only `suspendTasksPropagation`? Or is it overall a bad idea to make the Perl scripts for both _suspend_ and _resume_ and have only one method with boolean in the `TasksManager`?

zlamalp:
> should `resumeTasksPropagation` be added to OpenAPI too, if the TasksManager contains only `suspendTasksPropagation`? Or is it overall a bad idea to make the Perl scripts for both _suspend_ and _resume_ and have only one method with boolean in the `TasksManager`?

OpenAPI should describe the API, so it should contain only single method with boolean parameter as it is. While it`s not generally true for all API methods, I think it`s ok to have two CLI tools, because it is like user interface - it can be "frinedly" to the user. We also have tools for specific purpose, which call multiple API methods.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/TasksManagerMethod.java
```diff
@@ -326,6 +326,31 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 			}
 			return null;
 		}
+	},
+
+	/*#
+	 * Stops dispatcher from propagating waiting tasks to the engine.
+	 * Tasks which were sent to the engine before won't be affected and will be finished.
+	 */
+	suspendTasksPropagation {
```
zlamalp, line 335: Please add `parms.stateChangingCheck();` to bothe methods, since we want to be sure they are not called using GET.

perun-openapi/openapi.yml
```diff
@@ -15339,6 +15339,20 @@ paths:
                 destinationName: { type: string, description: "destination name", nullable: false }
                 destinationType: { type: string, description: "destination type", nullable: false }
 
+  /json/tasksManager/suspendTasksPropagation:
+    get:
```
zlamalp, line 15343: Please use `post:` instead.

zlamalp (changes requested):

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
```diff
@@ -214,6 +218,22 @@ private TaskSchedule getWaitingTaskSchedule() throws InterruptedException {
...
 	}
 
+	/**
+	 * Method waiting for propagation of tasks to the engine to be resumed.
+	 * Called when propagation was suspended.
+	 */
+	private void waitForResumingPropagation() {
+		int sleepTime = 10000;
+		while (tasksManagerBl.isSuspendedTasksPropagation()) {
+			try {
```
zlamalp, line 228: If Tasks propagation is suspended, we idle here in a cycle. Hence we won't get some debug messages about the pool and queue. I would like to have them logged in every cycle as waiting for TaskSchedule does.
```java
log.debug(schedulingPool.getReport());
log.debug("WaitingTasksQueue has {} normal Tasks and {} forced Tasks.", waitingTasksQueue.size(), waitingForcedTasksQueue.size());
```
Also, following log message is logged only once (before the cycle), so it might not be visible from the log, that propagation is currently suspended. Please repeat this message also in each cycle.
```java 
log.debug("Propagation of tasks is suspended.");
```

perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
```diff
@@ -147,6 +147,10 @@ public void run() {
 		log.debug("Pool contains {} tasks in total", schedulingPool.getSize());
 		TaskSchedule schedule;
 		while (!shouldStop()) {
+			if (tasksManagerBl.isSuspendedTasksPropagation()) {
```
zlamalp, line 150: You could probably move this to the `try/catch` below and make method `waitForResumingPropagation()` to throw `InterruptedException`, so you don't have to repeat the catch implementation (similar to `getWaitingTaskSchedule()`)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/TasksManagerBlImpl.java
```diff
@@ -41,6 +41,8 @@
 	@Autowired
 	protected TasksManagerImplApi tasksManagerImpl;
 
+	private static boolean suspendedTasksPropagation = false; //are tasks stopped from propagating to engine
```
zlamalp, line 44: We should have safe concurrent access here. You can use `AtomicBoolean` or you can synchronize all access to the variable using `synchronized(TasksManagerBlImpl.class) { ...read/write code.... }` in getter/setter.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/TasksManagerBlImpl.java
```diff
@@ -28,6 +28,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
```
zlamalp, line 31: Since you don't use `AtomicBoolean` in the end, you can remove the import.

zlamalp (changes requested):

zlamalp (approved):

PR #3151, opened 2021-04-07
Force delete methods added to OpenAPI

Johaney-s:
- Already implemented force delete methods for facility, owner and user added to OpenAPI

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -2682,6 +2682,14 @@ components:
       in: query
       required: true
 
+    forceDelete:
+      name: force
+      description: "If true, delete entity forcefully."
+      schema:
+        type: boolean
+      in: query
+      required: false
+
 
 
 #################################################
@@ -7863,7 +7871,7 @@ paths:
       parameters:
         - $ref: '#/components/parameters/userId'
         - $ref: '#/components/parameters/userExtSourceId'
-        - { name: force, description: "If true, use force deletion.", schema: { type: boolean },  in: query, required: false }
+        - $ref: '#/components/parameters/forceDelete'
       responses:
         '200':
           $ref: '#/components/responses/VoidResponse'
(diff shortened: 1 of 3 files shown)


perun-openapi/openapi.yml
```diff
@@ -8205,6 +8205,21 @@ paths:
...
 
+  /urlinjsonout/usersManager/deleteUser:
+    post:
+      tags:
+        - UsersManager
+      operationId: deleteUser
+      summary: Deletes a user.
+      parameters:
+        - $ref: '#/components/parameters/userId'
+        - { name: force, description: "If true, use force deletion.", schema: { type: boolean },  in: query, required: false }
```
Vojtech-Sassmann, line 8216: It might be worth defining the force parameters in the parameters schema since we can reuse it in multiple methods (like the `userId` for example).

perun-openapi/openapi.yml
```diff
@@ -10876,6 +10892,7 @@ paths:
           $ref: '#/components/responses/ExceptionResponse'
       parameters:
         - $ref: '#/components/parameters/ownerId'
+        - { name: force, description: "If true, use force deletion.", schema: { type: boolean },  in: query, required: false }
```
Vojtech-Sassmann, line 10895: The `deleteOwner` method doesn't support the `force` flag. You should probably also implement it in the `OwnersManagerMethod`.

Vojtech-Sassmann (changes requested):

perun-openapi/openapi.yml
```diff
@@ -2682,6 +2682,14 @@ components:
       in: query
       required: true
 
+    forceDelete:
+      name: forceDelete
```
Vojtech-Sassmann, line 2686: This name has to be exactly the same as the string that is parsed in the RPC API, otherwise, it would not be used.
```suggestion
      name: force
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/OwnersManagerMethod.java
```diff
@@ -58,13 +58,25 @@ public Owner call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 			parms.stateChangingCheck();
 
-			ac.getOwnersManager().deleteOwner(ac.getSession(),
+			if (parms.contains("force")) {
+				ac.getOwnersManager().deleteOwner(ac.getSession(),
+					ac.getOwnerById(parms.readInt("owner")), parms.readBoolean("force"));
+			} else {
+				ac.getOwnersManager().deleteOwner(ac.getSession(),
 					ac.getOwnerById(parms.readInt("owner")));
+			}
```
Vojtech-Sassmann, line 79: To avoid the code duplication, you can do something like this:
```suggestion
			boolean force = false;
			if (parms.contains("force")) {
			    force = parms.readBoolean("force");
			}
		    ac.getOwnersManager().deleteOwner(ac.getSession(),
					ac.getOwnerById(parms.readInt("owner")), force);
```

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (approved):

PR #3154, opened 2021-04-08
Added filter for members by list of group statuses

HejdaJakub:
* Added filter to existing methods.
* Edited OpenAPI according to implemented changes.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2259,7 +2259,7 @@ perun_policies:
     include_policies:
       - default_policy
 
-  getCompleteRichMembers_Group_List<String>_List<String>_boolean_policy:
+  getCompleteRichMembers_Group_List<String>_List<String>_List<String>_boolean_policy:
     policy_roles:
       - GROUPADMIN: Group
       - GROUPOBSERVER: Group
@@ -2328,7 +2328,7 @@ perun_policies:
     include_policies:
       - default_policy
 
-  findCompleteRichMembers_Group_List<String>_List<String>_String_boolean_policy:
+  findCompleteRichMembers_Group_List<String>_List<String>_List<String>_String_boolean_policy:
     policy_roles:
       - GROUPADMIN: Group
       - GROUPOBSERVER: Group
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -785,8 +785,8 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	}
 
 	@Override
-	public List<RichMember> getCompleteRichMembers(PerunSession sess, Group group, List<String> attrsNames, List<String> allowedStatuses, boolean lookingInParentGroup) throws AttributeNotExistsException, ParentGroupNotExistsException {
-		return getOnlyRichMembersWithAllowedStatuses(sess, this.getCompleteRichMembers(sess, group, attrsNames, lookingInParentGroup), allowedStatuses);
+	public List<RichMember> getCompleteRichMembers(PerunSession sess, Group group, List<String> attrsNames, List<String> allowedStatuses, List<String> allowedGroupStatuses, boolean lookingInParentGroup) throws AttributeNotExistsException, ParentGroupNotExistsException {
+		return getOnlyRichMembersWithAllowedStatusesAndGroupStatuses(sess, this.getCompleteRichMembers(sess, group, attrsNames, lookingInParentGroup), allowedStatuses, allowedGroupStatuses);
 	}
 
 	@Override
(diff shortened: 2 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -848,6 +848,34 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
...
+	 * @param sess perun session
+	 * @param richMembers rich members
+	 * @param allowedStatuses allowed statuses
+	 * @param allowedGroupStatuses allowed group statuses
+	 * @return list of allowed richMembers
+	 */
+	private List<RichMember> getOnlyRichMembersWithAllowedStatusesAndGroupStatuses(PerunSession sess, List<RichMember> richMembers, List<String> allowedStatuses, List<String> allowedGroupStatuses) {
+		List<RichMember> allowedRichMembers = new ArrayList<>();
+		if(richMembers == null || richMembers.isEmpty()) return allowedRichMembers;
+		if(allowedGroupStatuses == null || allowedGroupStatuses.isEmpty()) return richMembers;
```
Vojtech-Sassmann, line 864: This ignores the allowedStatuses if they are set.

perun-openapi/openapi.yml
```diff
@@ -2374,6 +2374,16 @@ components:
...
+    listOfAllowedGroupStatuses:
+      name: "allowedGroupStatuses[]"
+      description: "list of allowed group statuses List<String>"
+      in: query
+      required: false
+      schema:
+        type: array
+        items:
+          type: string
+
```
Vojtech-Sassmann, line 2386: You can also add the `enum` and specify the allowed values.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -1541,6 +1545,7 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 								ac.getGroupById(parms.readInt("group")),
 								parms.readList("attrsNames", String.class),
 								parms.readList("allowedStatuses", String.class),
+								parms.readList("allowedGroupStatuses", String.class),
```
Vojtech-Sassmann, line 1548: I would suggest making the `allowedGroupStatuses` parameter optional. This method might be already used in multiple GUI applications and this would require all of them to be changed. First, check if the `allowedGroupStatuses` is present in the params and if so, read it. If not, use some empty list.

Please make this change for the other method as well.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
HejdaJakub: Edited.

perun-openapi/openapi.yml
HejdaJakub: `Enum` added.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
HejdaJakub: The parameter `allowedGroupStatuses` is now optional.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -1128,19 +1129,39 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
...
-							return ac.getMembersManager().getCompleteRichMembers(ac.getSession(),
+							if(parms.contains("allowedGroupStatuses")) {
+								return ac.getMembersManager().getCompleteRichMembers(ac.getSession(),
+									ac.getGroupById(parms.readInt("group")),
+									parms.readList("attrsNames", String.class),
+									parms.readList("allowedStatuses", String.class),
+									parms.readList("allowedGroupStatuses", String.class),
+									parms.readBoolean("lookingInParentGroup"));
+							} else {
+								return ac.getMembersManager().getCompleteRichMembers(ac.getSession(),
```
Vojtech-Sassmann, line 1140: Just a suggestion of a minor readability improvement.

You don't need to duplicate the whole method call and params reading, you can do something like this:
```Java
List<String> allowedGroupStatuses = Collections.emptyList();
if (parms.contains("allowedGroupStatuses")) {
    allowedGroupStatuses = parms.readList("allowedGroupStatuses", String.class);
}
```
This reduces the duplicated code and makes it more clear, how the two calls differ.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2328,7 +2328,7 @@ perun_policies:
     include_policies:
       - default_policy
 
-  findCompleteRichMembers_Group_List<String>_List<String>_String_boolean_policy:
+  findCompleteRichMembers_Group_List<String>_List<String>_List<String>_String_boolean_policy:
```
Vojtech-Sassmann, line 2331: Also, please check if any of these policies are used in GUI. We need to make sure that the GUI uses the correct policy names.

Vojtech-Sassmann (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
HejdaJakub: Code has been simplified.

Vojtech-Sassmann (approved):

PR #3159, opened 2021-04-12
Show unsaved form items in application form dependency options

Johaney-s:
- Allows setting dependency on items which were not saved yet
- Also do not show warning if deleting unsaved item


Vojtech-Sassmann (changes requested):
Just a minor problem. Otherwise, it seems to be working well.

PR #3163, opened 2021-04-14
GUI - alert to new GUI

Vojtech-Sassmann:
* Added alert which might contain a link to new GUI.
* The alert appears if the property `newAdminGuiAlert` is specified in
the `perun-web-gui.properties` file.


balcirakpeter (approved):
Do not forget to add the property also to Ansible templates so it is possible to set it through Ansible.

PR #3164, opened 2021-04-15
Modules for adGroupName

balcirakpeter:
- module for def adGroupName checks the correct regex.
- module for virt adGroupName compose full name of the group in ad.
  It creates a string which consists of adGroupName def attr values of
  all parents of the given group together with the value of the given
  group itself. Values are sorted from the root group to the leaf group
  (the given one) and "-" is used as a delimiter.
- basic tests included.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_virt_adGroupName.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_virt_adGroupName.java
@@ -1,0 +1,81 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
+import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.bl.AttributesManagerBl;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.GroupVirtualAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.GroupVirtualAttributesModuleImplApi;
+
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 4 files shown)


rosinec (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_virt_adGroupName.java
```diff
@@ -0,0 +1,81 @@
...
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.GroupVirtualAttributesModuleAbstract;
+import cz.metacentrum.perun.core.implApi.modules.attributes.GroupVirtualAttributesModuleImplApi;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class urn_perun_group_attribute_def_virt_adGroupName extends GroupVirtualAttributesModuleAbstract implements GroupVirtualAttributesModuleImplApi {
+
+	private static final String SOURCE_ATTR_NAME = AttributesManager.NS_GROUP_ATTR + ":adGroupName";
```
Vojtech-Sassmann, line 21: I think this is invalid
```suggestion
	private static final String SOURCE_ATTR_NAME = AttributesManager.NS_GROUP_ATTR_DEF + ":adGroupName";
```

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_virt_adGroupNameTest.java
```diff
@@ -0,0 +1,92 @@
...
+			.thenReturn(am);
+		when(perunBl.getGroupsManagerBl())
+			.thenReturn(gm);
+
+		when(gm.getParentGroup(sess, groupB))
+			.thenReturn(groupA);
+		when(gm.getParentGroup(sess, groupC))
+			.thenReturn(groupB);
+
+		when(am.getAttribute(eq(sess), eq(groupA), any(String.class)))
```
Vojtech-Sassmann, line 61: You should probably test here that the given attribute name is correct.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_virt_adGroupName.java
balcirakpeter: Fixed

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_virt_adGroupNameTest.java
balcirakpeter: Fixed

Vojtech-Sassmann (approved):

PR #3169, opened 2021-04-21
Add table of groups for auto registration

metodej:
- for deployment: new db version with changelog
- this table should represent groups which user can register into during vo registration
- created basic methods to add/delete/get groups from this table in perun and in openapi

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2032,6 +2032,26 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getGroupsForAutoRegistration_Vo_policy:
+    policy_roles:
+      - VOADMIN: Vo
+    include_policies:
+      - default_policy
+
+  deleteGroupsFromAutoRegistration_Vo_List<Group>_policy:
+    policy_roles:
+      - GROUPADMIN: Group
+      - VOADMIN: Vo
+    include_policies:
+      - default_policy
+
+  addGroupsToAutoRegistration_Vo_List<Group>_policy:
+    policy_roles:
+      - GROUPADMIN: Group
+      - VOADMIN: Vo
+    include_policies:
+      - default_policy
+
   #MembersManagerEntry
   deleteMember_Member_policy:
     policy_roles:
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2035,17 +2035,18 @@ perun_policies:
   getGroupsForAutoRegistration_Vo_policy:
     policy_roles:
       - VOADMIN: Vo
+      - VOOBSERVER: Vo
     include_policies:
       - default_policy
 
-  deleteGroupsFromAutoRegistration_Vo_List<Group>_policy:
+  deleteGroupsFromAutoRegistration_List<Group>_policy:
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
     include_policies:
       - default_policy
 
-  addGroupsToAutoRegistration_Vo_List<Group>_policy:
+  addGroupsToAutoRegistration_List<Group>_policy:
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
@@ -1027,4 +1027,34 @@ public class GroupsManagerImpl implements GroupsManagerImplApi {
 			throw new InternalErrorException(e);
 		}
 	}
+
+	@Override
+	public List<Group> getGroupsForAutoRegistration(PerunSession sess, Vo vo) {
+		try {
+			return jdbc.query("select " + groupMappingSelectQuery + " from groups join groups_to_register on groups.id=groups_to_register.group_id " +
+				"where groups_to_register.vo_id=?", GROUP_MAPPER, vo.getId());
+		} catch (RuntimeException ex) {
+			throw new InternalErrorException(ex);
+		}
+	}
+
+	@Override
+	public void deleteGroupFromAutoRegistration(PerunSession sess, Vo vo, Group group) throws GroupAlreadyRemovedException {
+		try {
+			int rowAffected = jdbc.update("delete from groups_to_register where group_id=? and vo_id=?", group.getId(), vo.getId());
+
+			if(rowAffected == 0) throw new GroupAlreadyRemovedException("Group: " + group + " , Vo: " + vo);
+		} catch (RuntimeException err) {
+			throw new InternalErrorException(err);
+		}
+	}
+
+	@Override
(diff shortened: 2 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -1027,4 +1027,34 @@ public MemberGroupStatus getTotalMemberGroupStatus(PerunSession session, Member
 			throw new InternalErrorException(e);
 		}
 	}
+
+	@Override
+	public List<Group> getGroupsForAutoRegistration(PerunSession sess, Vo vo) {
+		try {
+			return jdbc.query("select " + groupMappingSelectQuery + " from groups join groups_to_register on groups.id=groups_to_register.group_id " +
+				"where groups_to_register.vo_id=?", GROUP_MAPPER, vo.getId());
```
Vojtech-Sassmann, line 1035: I don't think we need to save the voId in the `group_to_register` table. You can then modify the SQL something like this:
```suggestion
			return jdbc.query("select " + groupMappingSelectQuery + " from groups where vo_id=? and id IN (select group_id from groups_to_register)", GROUP_MAPPER, vo.getId());
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1759,5 +1758,78 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			parms.stateChangingCheck();
+			List<Integer> groupsInts = parms.readList("groups", Integer.class);
+
+			List<Group> groups = new ArrayList<>();
+			for (Integer groupInt : groupsInts) {
+				groups.add(ac.getGroupById(groupInt));
+			}
+
+			ac.getGroupsManager().deleteGroupsFromAutoRegistration(ac.getSession(),
+				ac.getVoById(parms.readInt("vo")),
```
Vojtech-Sassmann, line 1801: It shouldn't be necessary to specify vo in this method since each group has exactly one vo assigned, therefore, we can get it ourselves.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1759,5 +1758,78 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			parms.stateChangingCheck();
+			List<Integer> groupsInts = parms.readList("groups", Integer.class);
+
+			List<Group> groups = new ArrayList<>();
+			for (Integer groupInt : groupsInts) {
+				groups.add(ac.getGroupById(groupInt));
+			}
+
+			ac.getGroupsManager().addGroupsToAutoRegistration(ac.getSession(),
+				ac.getVoById(parms.readInt("vo")),
```
Vojtech-Sassmann, line 1829: Same here.

Vojtech-Sassmann (changes requested):

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2032,6 +2032,26 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getGroupsForAutoRegistration_Vo_policy:
+    policy_roles:
+      - VOADMIN: Vo
```
Vojtech-Sassmann, line 2037: Please, add here the `VOOBSERVER` role.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
metodej: Fixed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
metodej: Ok, I did not realise that, it should be removed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
metodej: Fixed.

perun-base/src/main/resources/perun-roles.yml
metodej: Added

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -1027,4 +1027,33 @@ public MemberGroupStatus getTotalMemberGroupStatus(PerunSession session, Member
...
+			throw new InternalErrorException(ex);
+		}
+	}
+
+	@Override
+	public void deleteGroupFromAutoRegistration(PerunSession sess, Group group) throws GroupAlreadyRemovedException {
+		try {
+			int rowAffected = jdbc.update("delete from groups_to_register where group_id=?", group.getId());
+
+			if(rowAffected == 0) throw new GroupAlreadyRemovedException("Group: " + group);
```
Vojtech-Sassmann, line 1045: This exception is meant for a different purpose. I think it is not necessary to throw an exception when this happens. You can just silently skip it.

Vojtech-Sassmann (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1759,5 +1758,72 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+
+	/*#
+	 * Deletes groups from a list of groups which can be registered into during vo registration.
+	 *
+	 * @throw GroupNotExistsException When the group doesn't exist
+	 * @throw GroupAlreadyRemovedException When deleting group which is already removed
+	 *
+	 * @param ids List<Integer> list of groups IDs
+	 */
+	deleteGroupsToAutoRegistration {
```
Vojtech-Sassmann, line 1787: I think this is a typo.
```suggestion
	deleteGroupsFromAutoRegistration {
```

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
metodej: Ok, removed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
metodej: Fixed, thanks.

Vojtech-Sassmann (approved):

PR #3180, opened 2021-04-23
Added check for adding groups to auto registration

metodej:
- member groups and groups with synchronization enabled cannot be added to auto registration now


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -5532,8 +5533,21 @@ public void deleteGroupsFromAutoRegistration(PerunSession sess, List<Group> grou
...
-	public void addGroupsToAutoRegistration(PerunSession sess, List<Group> groups) {
+	public void addGroupsToAutoRegistration(PerunSession sess, List<Group> groups) throws GroupNotAllowedToAutoRegistrationException {
 		for (Group group : groups) {
+			if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
+				throw new GroupNotAllowedToAutoRegistrationException("Members group cannot be added to auto registration.", group);
+			}
+
+			try {
+				Attribute syncEnabledAttr = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUPSYNCHROENABLED_ATTRNAME);
+				if (syncEnabledAttr.valueAsString() != null && syncEnabledAttr.valueAsString().equals("true")) {
```
cuadradek, line 5544: This could be simplified: `if ("true".equals(syncEnabledAttr.valueAsString())) {`

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupNotAllowedToAutoRegistrationException.java
```diff
@@ -0,0 +1,39 @@
...
+	public GroupNotAllowedToAutoRegistrationException(String message, Throwable cause) {
+		super(message, cause);
+	}
+
+	/**
+	 * Constructor with the group that is not allowed to be added to auto registration
+	 * @param group the group
+	 */
+	public GroupNotAllowedToAutoRegistrationException(String message, Group group) {
+		super(message);
```
cuadradek, line 36: I would include the group to the message, something like: `super(message + " " + group.toString())`

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupNotAllowedToAutoRegistrationException.java
```diff
@@ -0,0 +1,39 @@
...
+import cz.metacentrum.perun.core.api.Group;
+
+/**
+ * This exception is thrown when group is not allowed to be added to auto registration.
+ *
+ * @author Metodej Klang
+ */
+public class GroupNotAllowedToAutoRegistrationException extends PerunException {
+
+	private Group group;
```
cuadradek, line 12: Getter for this field might be useful

cuadradek (changes requested):
Looks good, just a couple of nit-picks.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: Fixed.

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupNotAllowedToAutoRegistrationException.java
metodej: Fixed.

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/GroupNotAllowedToAutoRegistrationException.java
metodej: Added.

cuadradek (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -5532,8 +5533,21 @@ public void deleteGroupsFromAutoRegistration(PerunSession sess, List<Group> grou
...
 	@Override
-	public void addGroupsToAutoRegistration(PerunSession sess, List<Group> groups) {
+	public void addGroupsToAutoRegistration(PerunSession sess, List<Group> groups) throws GroupNotAllowedToAutoRegistrationException {
 		for (Group group : groups) {
+			if (group.getName().equals(VosManager.MEMBERS_GROUP)) {
+				throw new GroupNotAllowedToAutoRegistrationException("Members group cannot be added to auto registration.", group);
+			}
+
+			try {
+				Attribute syncEnabledAttr = getPerunBl().getAttributesManagerBl().getAttribute(sess, group, GroupsManager.GROUPSYNCHROENABLED_ATTRNAME);
```
balcirakpeter, line 5543: There is also a group structure synchronization. It should be considered as well.
It will be probably a little more difficult because you will need to search also in the group tree. Check methods like `addMember`, specifically the use of method `isGroupInStructureSynchronizationTree`  in it.

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: Ok, I've added groups with structure sync to the list of not allowed groups.

zlamalp:
Just a note: We should also add opposite check - when setting synchronization "enabled", we should prevent it if group is setup for registrations. This can be probably implemented in separate task.

metodej:
> 
> 
> Just a note: We should also add opposite check - when setting synchronization "enabled", we should prevent it if group is setup for registrations. This can be probably implemented in separate task.

Ok, good point, I've made PR with this check: https://github.com/CESNET/perun/pull/3186

balcirakpeter (approved):

PR #3182, opened 2021-04-27
Recover connection between dispatcher and engine

HejdaJakub:
1. Now the auto reconnection of JMS queue is working, when the server is restarted.
* In the `JMSQueueManager` class in method `registerForReceivingMessages()` is if condition, which is sending the registration message only if the variable `receivingMessages` is `false`. But if the server is restarted, this variable stays in `true` state, so the registration message couldn't be sent again.
* In the initial method in `JMSQueueManager` class I set this variable to `false`, so the registration message can be sent after restarting the server.

2. Logging was fixed too.
* I added the `Theard.sleep()` to catch block, so the next potential message will be sent after the wait time, which is mentioned in log error message.

### DIFF
--- a/perun-engine/src/main/java/cz/metacentrum/perun/engine/jms/MessageReceiver.java
+++ b/perun-engine/src/main/java/cz/metacentrum/perun/engine/jms/MessageReceiver.java
@@ -82,17 +82,23 @@ public class MessageReceiver implements Runnable {
 					waitTime = 0;
 				} catch (InvalidDestinationException e) {
 					queueAcquired = false;
-					waitTime = waitTime + 5000;
+					waitTime = setWaitTime(waitTime);
 					log.error("Queue doesn't exist yet. We gonna wait a bit ({} s) and try it again.",
 							(waitTime / 1000), e);
+					// wait for a time mentioned in the error message before try it again
+					try {
+						Thread.sleep(waitTime);
+					} catch (InterruptedException interrupted) {
+						log.error(interrupted.toString(), interrupted);
+					}
 				} catch (JMSException e) {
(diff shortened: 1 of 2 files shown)


Vojtech-Sassmann (changes requested):
Please, provide a commit message explaining, what was the original problem and how this was fixed.

perun-engine/src/main/java/cz/metacentrum/perun/engine/jms/MessageReceiver.java
```diff
@@ -85,6 +85,11 @@ public void run() {
 					waitTime = waitTime + 5000;
 					log.error("Queue doesn't exist yet. We gonna wait a bit ({} s) and try it again.",
 							(waitTime / 1000), e);
+					try {
+						Thread.sleep(waitTime);
+					} catch (InterruptedException interrupted) {
+						log.error(interrupted.toString(), interrupted);
+					}
```
Vojtech-Sassmann, line 92: It might be also worth adding an inline comment describing the reason for this wait.

Vojtech-Sassmann (changes requested):

perun-engine/src/main/java/cz/metacentrum/perun/engine/jms/MessageReceiver.java
HejdaJakub: Inline comment added before try-catch block.

Vojtech-Sassmann (changes requested):
Please, also update the commit message, not only the PR description.

Vojtech-Sassmann (approved):

PR #3193, opened 2021-05-04
Check number of embedded groups form items

Johaney-s:
* Maximum one application form item for embedded groups is allowed in VO application
* Exception is thrown if multiple definitions are found when updating application


Vojtech-Sassmann (changes requested):
These changes look good, but we should add the check to the `addFormItem` method as well. Also, please check if any other methods can add form items.

Johaney-s:
> These changes look good, but we should add the check to the `addFormItem` method as well. Also, please check if any other methods can add form items.

Thanks, I added the check also to `addFormItem` and `copyItems`.

balcirakpeter (approved):

Vojtech-Sassmann (approved):

PR #3194, opened 2021-05-04
Core - perun apps config

Vojtech-Sassmann:
* Added config for domains of gui applications.
* The configuration file should be located at
`/etc/perun/perun-apps-config.yml`.


Vojtech-Sassmann:
Before merging this PR, we need to update the config files on the devel instance.

balcirakpeter (approved):

PR #3195, opened 2021-05-04
Reject member open applications during removal

balcirakpeter:
- We want to reject all member open applications before we remove him
  from VO. Therefore, functionality for that was added.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
@@ -22,11 +22,16 @@ import cz.metacentrum.perun.core.api.exceptions.AlreadyMemberException;
 import cz.metacentrum.perun.core.api.exceptions.AlreadySponsorException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidLoginException;
+import cz.metacentrum.perun.core.api.exceptions.LoginNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MemberAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.NamespaceRulesNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PasswordDeletionFailedException;
+import cz.metacentrum.perun.core.api.exceptions.PasswordOperationTimeoutException;
 import cz.metacentrum.perun.core.api.exceptions.SponsorshipDoesNotExistException;
 import cz.metacentrum.perun.core.bl.DatabaseManagerBl;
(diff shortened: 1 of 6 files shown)


balcirakpeter:
Code was manually tested.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -57,6 +57,9 @@
 
 public class MembersManagerImpl implements MembersManagerImplApi {
 
+	private static String rejected = "REJECTED";
```
Vojtech-Sassmann, line 60: Maybe it would be better to use the enum value of `AppState`.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/MembersManagerImplApi.java
```diff
@@ -457,4 +457,12 @@
...
 	 */
 	NamespaceRules getNamespaceRules(String namespace) throws NamespaceRulesNotExistsException;
+
+	/**
+	 * Reject all member applications which are not Approved or already rejected
+	 *
+	 * @param sess perunSession
+	 * @param member which applications will be removed
+	 */
+	void rejectAllMemberOpenApplications(PerunSession sess, Member member);
```
Vojtech-Sassmann, line 467: It might be worth mentioning that no notifications are sent when using this method.

Vojtech-Sassmann: @balcirakpeter Hi, could you please take a look at this so we can merge this PR?

balcirakpeter:
Changes not tested yet.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -802,4 +811,53 @@ public void setSponsoredAccountsConfigLoader(SponsoredAccountsConfigLoader spons
...
 		return sponsoredAccountsConfigContainer.getNamespaceRules(namespace);
 	}
+
+	@Override
+	public void rejectAllMemberOpenApplications(PerunSession sess, Member member) {
+		try {
+			List<Integer> ids = jdbc.query(
+				"select id from application " +
+					"where user_id=? and vo_id=? and state not in (?, ?)",
+				new SingleColumnRowMapper<>(Integer.class),  member.getUserId(), member.getVoId(), rejected, approved);
```
Vojtech-Sassmann, line 821: I was testing this on devel and I have discovered a problem. If a user wants to submit a group application, and he is not in the Perun yet, the group application doesn't contain `user_id`. Therefore, this won't work, because the application will not be found. :disappointed: 

I'm not sure how to exactly solve this. We might set the `user_id` for all user applications, when the object user is created. Or, you would need to search the applications here based on extsources (which, I guess, would be much difficult). What do you think? :thinking:

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann:
@balcirakpeter I also think, we should add some tests for this logic in the code.

PR #3198, opened 2021-05-05
Additional empty attribute check in application

Johaney-s:
* attribute dst_attr was checked for null value
* it needs to be checked for empty value too as it is passed from GUI

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -3815,7 +3815,7 @@ public class RegistrarManagerImpl implements RegistrarManager {
 		// put application data into Candidate
 		final Map<String, String> attributes = new HashMap<>();
 		jdbc.query("select dst_attr,value from application_data d, application_form_items i where d.item_id=i.id "
-				+ "and i.dst_attr is not null and d.value is not null and app_id=?",
+				+ "and i.dst_attr is not null and i.dst_attr <> '' and d.value is not null and app_id=?",
 			(resultSet, i) -> {
 				attributes.put(resultSet.getString("dst_attr"), resultSet.getString("value"));
 				return null;


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3815,7 +3815,7 @@ private Candidate createCandidateFromApplicationData(Application app) {
 		// put application data into Candidate
 		final Map<String, String> attributes = new HashMap<>();
 		jdbc.query("select dst_attr,value from application_data d, application_form_items i where d.item_id=i.id "
-				+ "and i.dst_attr is not null and d.value is not null and app_id=?",
+				+ "and i.dst_attr is not null and i.dst_attr <> '' and d.value is not null and app_id=?",
```
zlamalp, line 3818: Hmmm, the question is, if its not a bug also somewhere else. Like you probably shouldn't store "empty" destination attribute in the first place. It makes no sense to have two value (null / empty string) to represent the fact, that value shouldn't be stored in attribute in perun.

Vojtech-Sassmann: The empty string values were sent from the new gui. Our RPC API allows the empty strings and it could already happen that they got into DBs. So this is just an easier fix. Otherwise, we would need to fix all methods that update the form items, so that they would map the empty strings to nulls, and we would also need to check production DBs. 

@zlamalp Do you think that these empty strings could cause some other problems?

zlamalp: @Vojtech-Sassmann Well, in this case probably not and its probably OK as a quick-fix.

Generally it adds complexity and you may encounter problems in the future. In the long term it would be better to fix clients/API and clean up DB data and have single representation of "empty" state.
We do the same for empty attribute values. It was not originally a problem, since Oracle DB saved empty strings as nulls, but postgres stores empty strings as they are.

Vojtech-Sassmann (approved):

PR #3201, opened 2021-05-06
Core - bugfix identities alerts

Vojtech-Sassmann:
* There was a bug with sending alerts when an identity was added by a
user. The problem was with an invalid if condition. The condition should
check that the attribute's value is not null but actually the attribute
itself was checked.
* There was also another bug. When trying to retreive attribute with
notifications templates, invalid attribute name was used -
`identityAlerts`. The correct name is `identityAlertsTemplates`.


Vojtech-Sassmann:
@balcirakpeter we should test this on the devel instance, if it actually works.

balcirakpeter (approved):

PR #3202, opened 2021-05-06
API method for apps config

Johaney-s:
* Added openApi definition for retrieving apps configuration

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -1076,6 +1076,26 @@ components:
         password: { type: string }
         login: { type: string }
 
+    NewApps:
+      type: object
+      properties:
+        admin: { type: string }
+        profile: { type: string }
+        pwdReset: { type: string }
+        publications: { type: string }
+
+    Brand:
+      type: object
+      properties:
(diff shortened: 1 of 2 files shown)


Vojtech-Sassmann (approved):

perun-openapi/openapi.yml
```diff
@@ -2895,6 +2922,19 @@ paths:
...
+  /json/utils/getAppsConfig:
+    get:
+      tags:
+        - Utils
+      operationId: getAppsConfig
+      summary: Gets configuration of perun apps brandings and apps' domains
+      description: Returns object of type PerunAppsConfig
+      responses:
+        '200':
+          $ref: '#/components/responses/StringResponse'
```
balcirakpeter, line 2934: I think this should be `PerunAppsConfigResponse`.

balcirakpeter (changes requested):

perun-openapi/openapi.yml
Johaney-s: Right, I think I forgot to commit this change. Fixed.

PR #3203, opened 2021-05-06
Added searchString and status parameters to getMembersPage

metodej:
- method getMembersPage was extended to support searchString and a list of vo statuses
- these parameters are optional, if searchString is empty, the method returns all members, if statuses are empty, members are returned regardless of their status
- also added tests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
@@ -54,6 +54,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
+import java.util.stream.Collectors;
 
 public class MembersManagerImpl implements MembersManagerImplApi {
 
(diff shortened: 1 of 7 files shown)


balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -689,21 +690,63 @@ public MemberGroupStatus getUnifiedMemberGroupStatus(PerunSession sess, User use
...
+				" ) " +
 				" ORDER BY " + query.getSortColumn().getSqlOrderBy(query) +
-				" OFFSET ?" +
-				" LIMIT ?;",
-				getPaginatedMembersExtractor(query),
-				vo.getId(),
-				query.getOffset(),
-				query.getPageSize());
+				" OFFSET (:offset)" +
+				" LIMIT (:limit);",
```
zlamalp, line 746: I don't think we end queries with `;`, since its done by the DB connection logic.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -1884,17 +1885,33 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 	 * @param vo vo
 	 * @param query query with page information
 	 * @param attrNames attribute names
+	 * @param searchString (optional) get only members based on this string and specific parameters in DB,
```
zlamalp, line 1888: I would enhance this message, since its documentation for external API. Its useful to tell, that by default it searches through names, emails, logins of member or other attributes specified by the perun configuration.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -1884,17 +1885,33 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			}
+
+			String searchString = parms.contains("searchString") ? parms.readString("searchString") : "";
+
 			return ac.getMembersManager().getMembersPage(ac.getSession(),
 					ac.getVoById(parms.readInt("vo")),
 					parms.read("query", MembersPageQuery.class),
-					parms.readList("attrNames", String.class));
+					parms.readList("attrNames", String.class),
+					searchString,
```
Vojtech-Sassmann, line 1913: Please, don't add new parameters for this. Add these values to the MembersPageQuery.

Vojtech-Sassmann (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
metodej: Ok, I've added the new parameters to MembersPageQuery.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
metodej: Removed.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
metodej: Ok, since I have moved the parameters to the MembersPageQuery, the message was removed. But I have added a note about query parameter to the method description.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -690,20 +691,62 @@ public MemberGroupStatus getUnifiedMemberGroupStatus(PerunSession sess, User use
...
+			int id = Integer.parseInt(query.getSearchString());
+			idQueryString = " members.user_id=" + id + " or members.id=" + id + " or ";
+		} catch (NumberFormatException e) {
+			// IGNORE wrong format of ID
+		}
+
+		Paginated<Member> members = namedParameterJdbcTemplate.query(
+				"SELECT DISTINCT " + memberMappingSelectQuery +
 				query.getSortColumn().getSqlSelect() +
 				", count(*) OVER() AS total_count" +
```
Vojtech-Sassmann, line 726: With the current implementation, the count is not working, probably because of the added `DISTINCT` keyword.

Vojtech-Sassmann (changes requested):
Also, please update the schema for the MembersPageQuery in the OpenAPI.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -1879,6 +1879,9 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 
 	/*#
 	 * Get page of members from the given vo, with the given attributes.
+	 * Query parameter specifies offset, page size, sorting order, sorting column, statuses in vo and string to search
+	 * members by (by default it searches in names, emails, logins of member or other attributes based on perun
+	 * configuration), last two parameters are optional and by default it finds all members.
```
zlamalp, line 1884: `@param query query` won't generate anything meaningful in the API docs. We need to specify usage of `MembersPageQuery` and you need to add support for its definition in docs generating script in `perun-utils`.

zlamalp (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
metodej: Thank you, it should be fixed now.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
metodej: Ok, I've edited parameters and added `MembersPageQuery` and `Paginated<RichMember>` to `perun-utils`.

Vojtech-Sassmann (changes requested):
I just thought of a minor improvement. Otherwise, it seems to be working as expected, and also, it is not that much slower. Loading time of a single page is something around 3s in a larger vo.

Vojtech-Sassmann (approved):
I think this works well, at least for now. However, we should definitely consider rewriting the SQL, since this might be a bit unresponsive for users.

PR #3211, opened 2021-05-10
Check apps configuration when generating url

Johaney-s:
* When generating url for mail notification about changing preferred email, the domain must be included in the config file
* For new GUI it shouldn't add /krb/gui/ part in the url as that would invalidate it

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
@@ -90,6 +90,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 
+import static cz.metacentrum.perun.core.impl.PerunAppsConfig.getBrandContainingDomain;
 import static cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_entityless_attribute_def_def_identityAlertsTemplates.ALERT_TIME_FORMATTER;
 import static cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_entityless_attribute_def_def_identityAlertsTemplates.DEFAULT_IDENTITY_ADDED_PREF_MAIL;
 import static cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_entityless_attribute_def_def_identityAlertsTemplates.DEFAULT_IDENTITY_ADDED_PREF_MAIL_SUBJECT;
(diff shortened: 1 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -2318,4 +2328,21 @@ private static String getUserPreferredEmail(PerunSession sess) {
 		}
 	}
 
+	/**
+	 * Returns configuration's brand in which the provided domain is specified
+	 * @param domain
+	 * @return brand or null if domain is not present in the configuration
+	 */
+	private static PerunAppsConfig.Brand getBrandContainingDomain(String domain) {
```
zlamalp, line 2336: This kind of search function could be part of `PerunAppsConfig` itself, what do you think @Vojtech-Sassmann ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -1016,9 +1016,19 @@ public static void sendValidationEmail(User user, String url, String email, int
 		String linkLocation = "/gui/";
 		try {
 			URL urlObject = new URL(url);
+			String domain = urlObject.getProtocol()+"://"+urlObject.getHost();
```
zlamalp, line 1019: This solves the issue for now. But we still rely on `referer` header to be sent. In the end of this dev cycle we want GUI to send us domain and authorization filter to the API, so just keep that in mind. I'm not sure, if it should or shouldn't be part of this pull-request.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
Vojtech-Sassmann: Yeah, definitely.

Johaney-s: Okay, moved to `PerunAppsConfig`.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
Johaney-s: So now it should be possible to either pass referer or verification link url, authentication can be passed as query parameter too, the old way of appending it to path is still possible for now, might want to delete later?

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (approved):
I think this should be working. We can test this with the changes that have to be done to the GUI applications.

PR #3217, opened 2021-05-19
Create group form when added to VO's embedded groups

Johaney-s:
* Requires change in GUI (openApi specs changed)
* Empty form was created when submitting application to embedded group
* Newly this happens already when group is allowed for auto registration


balcirakpeter (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1660,11 +1660,6 @@ private void submitEmbeddedGroupApplications(PerunSession sess, List<Group> grou
 				groupApplication.setExtSourceName(app.getExtSourceName());
 				groupApplication.setExtSourceType(app.getExtSourceType());
 				groupApplication.setCreatedBy("Automatically generated");
-				try {
-					getFormForGroup(group);
-				} catch (FormNotExistsException e) {
-					createApplicationFormInGroup(sess, group);
-				}
```
Vojtech-Sassmann, line 1667: Since this logic with autoregistration was already present in the last release, we need to make sure there would not occur any inconsistencies. If we remove this, it might happen that some group was already added to the list of allowed groups for auto registrations and the form had not yet been created. We should either keep this in the code for a while, or we would have to manually create application forms for such groups.

 I'd suggest keeping this code here with a comment that this could be removed in the future. What do you think?

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3161,6 +3156,65 @@ else if (extSourceName.equals(application.getExtSourceName()) && actor.equals(ap
...
+		// Create application form if non exists
+		for (Group group : groups) {
+			try {
+				getFormForGroup(group);
+			} catch (FormNotExistsException e) {
+				createApplicationFormInGroup(sess, group);
+			}
+		}
+
+		perun.getGroupsManagerBl().addGroupsToAutoRegistration(sess, groups);
```
Vojtech-Sassmann, line 3215: I like this solution since it required fewer changes to be made in the code. However, we should denote that this method `addGroupsToAutoRegistration` in the groupsManagerBl does not create the application form and that it has to be done. It should be at least mentioned in the javadoc.

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Johaney-s: Right, I added this information to `GroupsManagerBl` javadoc.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Johaney-s: Thanks, I didn't think of that... I agree we can remove it later, I commented it in the code and kept it at the original place also.

Vojtech-Sassmann (approved):

PR #3219, opened 2021-05-20
fix(core): fix invalid pattern for URL validation

### LINKED ISSUE #3197: Buggy regex for URL dest type
https://github.com/CESNET/perun/blob/fed8403aea502ed89744c0dc9c84c67e0c3f570d/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java#L137

- this does not allow the use of dashes in the hostname, i.e. as in "https://denbi-pka.cloud.embl.de/perun/upload"

Vojtech-Sassmann:
* There was an invalid pattern that allowed valid urls to be detected as invalids.

fix #3197

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
@@ -135,7 +135,7 @@ public class Utils {
 	private static final Pattern lastNamePattern = Pattern.compile("^(([\\p{L}-']+)|([\\p{L}][.]))$");
 
 	public static final Pattern hostPattern = Pattern.compile("^(?!:\\/\\/)(?=.{1,255}$)((.{1,63}\\.){1,127}(?![0-9]*$)[a-z0-9-]+\\.?)$|^(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)(\\.(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)){3}$");
-	public static final Pattern urlPattern = Pattern.compile("[(http(s)?):\\/\\/(www\\.)?a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)");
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -135,7 +135,7 @@
 	private static final Pattern lastNamePattern = Pattern.compile("^(([\\p{L}-']+)|([\\p{L}][.]))$");
 
 	public static final Pattern hostPattern = Pattern.compile("^(?!:\\/\\/)(?=.{1,255}$)((.{1,63}\\.){1,127}(?![0-9]*$)[a-z0-9-]+\\.?)$|^(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)(\\.(25[0-5]|2[0-4]\\d|[0-1]?\\d?\\d)){3}$");
-	public static final Pattern urlPattern = Pattern.compile("[(http(s)?):\\/\\/(www\\.)?a-zA-Z0-9@:%._\\+~#=]{2,256}\\.[a-z]{2,6}\\b([-a-zA-Z0-9@:%_\\+.~#?&//=]*)");
+	public static final Pattern urlPattern = Pattern.compile("^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]");
```
metodej, line 138: I think parentheses "()" should be also ok in url.
```suggestion
	public static final Pattern urlPattern = Pattern.compile("^(https?|ftp|file)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;()]*[-a-zA-Z0-9+&@#/%=~_|()]");
```

metodej (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
Johaney-s: I also found some other characters in the [list of URI reserved characters](https://tools.ietf.org/html/rfc3986#section-2.2): `*$'`, maybe they would be worth adding too. What do you guys think?

Vojtech-Sassmann: Thanks for the suggestions. I think it makes sense to add all of them.

metodej (approved):

PR #3220, opened 2021-05-20
Added group and groupStatus parameters to MembersPageQuery

metodej:
- method getMembersPage was extended to support group and group statuses for searching in specified group
- these parameters are optional, if not included, the method works as before
- also added tests

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1569,7 +1569,8 @@ public class MembersManagerEntry implements MembersManager {
 		Utils.checkPerunSession(sess);
 		perunBl.getVosManagerBl().checkVoExists(sess, vo);
 
-		if (!AuthzResolver.authorizedInternal(sess, "getMembersPage_Vo_MembersPageQuery_List<String>_policy", vo)) {
+		if (!AuthzResolver.authorizedInternal(sess, "getMembersPage_Vo_MembersPageQuery_List<String>_policy", vo)
+			&& (query.getGroup() != null && !AuthzResolver.authorizedInternal(sess, "group-getMembersPage_Vo_MembersPageQuery_List<String>_policy", query.getGroup()))) {
 			throw new PrivilegeException(sess, "getMembersPage");
 		}
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1570,7 +1570,7 @@ public class MembersManagerEntry implements MembersManager {
 		perunBl.getVosManagerBl().checkVoExists(sess, vo);
 
 		if (!AuthzResolver.authorizedInternal(sess, "getMembersPage_Vo_MembersPageQuery_List<String>_policy", vo)
-			&& (query.getGroup() != null && !AuthzResolver.authorizedInternal(sess, "group-getMembersPage_Vo_MembersPageQuery_List<String>_policy", query.getGroup()))) {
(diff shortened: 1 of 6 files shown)


Johaney-s (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1569,7 +1569,8 @@ public void removeSponsor(PerunSession sess, Member sponsoredMember, User sponso
 		Utils.checkPerunSession(sess);
 		perunBl.getVosManagerBl().checkVoExists(sess, vo);
 
-		if (!AuthzResolver.authorizedInternal(sess, "getMembersPage_Vo_MembersPageQuery_List<String>_policy", vo)) {
+		if (!AuthzResolver.authorizedInternal(sess, "getMembersPage_Vo_MembersPageQuery_List<String>_policy", vo)
+			&& (query.getGroup() != null && !AuthzResolver.authorizedInternal(sess, "group-getMembersPage_Vo_MembersPageQuery_List<String>_policy", query.getGroup()))) {
```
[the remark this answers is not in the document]
metodej: Yes, thank you, it should be fixed now.

Johaney-s (approved):

Vojtech-Sassmann:
I would like to manually test these changes first, I will try to do it today.

balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -739,6 +739,23 @@ public MemberGroupStatus getUnifiedMemberGroupStatus(PerunSession sess, User use
...
+			groupQueryJoin = " LEFT JOIN groups_members ON members.id=groups_members.member_id ";
+			groupQueryWhere = " AND groups_members.group_id = (:groupId) ";
+			namedParams.addValue("groupId", query.getGroup().getId());
+
+			if (query.getGroupStatuses() != null && !query.getGroupStatuses().isEmpty()) {
+				groupStatusesQueryString = " AND groups_members.source_group_status in (:groupStatuses) ";
+				List<Integer> groupStatusCodes = query.getGroupStatuses().stream()
+					.map(MemberGroupStatus::getCode)
+					.collect(Collectors.toList());
+				namedParams.addValue("groupStatuses", groupStatusCodes);
```
Vojtech-Sassmann, line 755: I have realized that the support for the group filtering will not be that easy. The group `group_members` table can contain multiple records for a single member because of indirect group memberships. We probably need to `merge` such records in the SQL and evaluate the overall member status. I think this might require vast changes to be made to the SQL.

An example of the problem. Suppose we have a member in groups A, A:B. In the table we would have something like this:
|member_id | group_id  |source_group_id|source_group_status|
--- | --- | --- | ---
|22|3|3|VALID|
|22|3|4|EXPIRED|

With the current SQL, this member would be returned if we asked for all expired members. However, he is not expired.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -739,6 +739,23 @@ public MemberGroupStatus getUnifiedMemberGroupStatus(PerunSession sess, User use
 			namedParams.addValue("statuses", statusCodes);
 		}
 
+		String groupQueryJoin = "";
+		String groupQueryWhere = "";
+		String groupStatusesQueryString = "";
+		if (query.getGroup() != null) {
+			groupQueryJoin = " LEFT JOIN groups_members ON members.id=groups_members.member_id ";
```
Vojtech-Sassmann, line 746: I think you don't want a `LEFT` join here. You don't want to keep records for members without groups. However, if you will rewrite the SQL, this commend might be irrelevant.

Vojtech-Sassmann (changes requested):

PR #3221, opened 2021-05-20
fix(core): fix generating of email validation links

Vojtech-Sassmann:
* When the client provided to the API customUrl together with customPath, the customPath was ignored
and the path was taken from the url.


Vojtech-Sassmann:
We have agreed that we don't want support customUrl + customPath so this PR can be closed.

PR #3222, opened 2021-05-20
Check linkPath passed to preferred mail change method

Johaney-s:
* linkPath cannot change domain of the verification link (for example by passing ".evil.com" as path)

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
@@ -1312,7 +1312,7 @@ public enum UsersManagerMethod implements ManagerMethod {
 				throw new RpcException(RpcException.Type.MISSING_VALUE, "Missing \"Referer\" header in HTTP request and no custom verification link specified.");
 			}
 
-			if (customUrl != null) {
+			if (customUrl != null) { // customUrl option
 				URL url = null;
 				try {
 					url = new URL(customUrl);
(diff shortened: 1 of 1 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -1321,6 +1321,19 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+					try {
+						URL refUrl = new URL(referer);
+						URL refDomainWithPath = new URL(refUrl.getProtocol()+"://"+refUrl.getHost()+customPath);
+						if (!refUrl.getHost().equals(refDomainWithPath.getHost())) {
+							throw new RpcException(RpcException.Type.INVALID_URL, "Invalid verification link - path changes domain: " + refDomainWithPath);
+						}
+					} catch (MalformedURLException e) {
+						throw new RpcException(RpcException.Type.INVALID_URL, "Invalid referer or path.");
+					}
+				}
```
Vojtech-Sassmann, line 1336: I really like this solution. However, we should also support the combination of `customUrl` and `customPath` being provided at the same time. I have discovered this problem today and I have created a PR with a fix: https://github.com/CESNET/perun/pull/3221

Could you please review it, so I can merge it? Then you probably should rebase this PR.

Also, I think you can move this check out of the first `if`. You can validate this regardless of the `customUrl` being provided or not.

Vojtech-Sassmann (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
Johaney-s: After agreeing to further keep only customUrl and no linkPath, I only edited the else - if part.

balcirakpeter (approved):

Vojtech-Sassmann (approved):

PR #3225, opened 2021-05-21
Added tests for url destination

metodej:
- added new tests for check of url destination


Johaney-s:
I think these tests are sufficient, I can't think of anything more to be tested. I commented to the [original PR](https://github.com/CESNET/perun/pull/3219) about the usage of reserved characters.

Johaney-s (approved):

balcirakpeter (commented):
Shouldn't we add tests for `www.` in the URL? I am not sure how we treat it but I do not think `https://www.com/` is valid.

Vojtech-Sassmann:
> Shouldn't we add tests for `www.` in the URL? I am not sure how we treat it but I do not think `https://www.com/` is valid.

Actually, it is valid. :smile: 
image

PR #3231, opened 2021-05-27
Fix: Cannot force delete vo

HejdaJakub:
* In some cases GUI threw an internal error excepion when user tried to force delete vo. This error occurred when some group had subroups.
* On backend we were trying to delete all vo groups one by one (including subgroups), so if some parent group was force deleted, then the subroups couldn't be deleted again.
* I changed the business logic, so now we are trying to force delete only parent groups (it correctly deletes all subgroups).


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -524,7 +524,10 @@ private void logTotallyRemovedMembers(PerunSession sess, Integer parentGroupId,
 
 	@Override
 	public void deleteAllGroups(PerunSession sess, Vo vo) throws GroupAlreadyRemovedException, GroupAlreadyRemovedFromResourceException, GroupRelationDoesNotExist, GroupRelationCannotBeRemoved {
-		for(Group group: getGroupsManagerImpl().getGroups(sess, vo)) {
+		// get parent groups
+		List<Group> groups = getGroupsManagerImpl().getGroups(sess, vo).stream()
+			.filter(group -> !group.getName().contains(":")).collect(Collectors.toList());
```
Vojtech-Sassmann, line 529: This is just about conventions, when using Java streams it is a good practice to put each call on a separate line. It improves readability because it is a convention.
```suggestion
		List<Group> groups = getGroupsManagerImpl().getGroups(sess, vo).stream()
			.filter(group -> !group.getName().contains(":"))
			.collect(Collectors.toList());
```

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -411,6 +411,21 @@ public void deleteVoWhichNotExists() throws Exception {
 		vosManagerEntry.deleteVo(sess, new Vo());
 	}
 
+	@Test(expected = VoNotExistsException.class)
```
Vojtech-Sassmann, line 414: This is not needed to change, but I prefer using the `assertThatExceptionOfType()` method from the assertJ library, instead of this JUnit functionality. It is more clear, where the exception should be thrown and you can also easily extend the check with methods that can check, for example, the message of the thrown exception. If you want, you can try to use this method here (without the message check, that was just an example, what can be achieved by using it). :slightly_smiling_face: 

This method is also used in this test file, so you can take a look, how to use it.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
balcirakpeter, line 529: Just a small thing that does not have to be changed:
It would be better to filter according to `getParentGroupId == null` rather than `group.getName().contains(":")`. It would be more readable.

balcirakpeter (approved):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
HejdaJakub: Thanks for hints! I edit the streams and rewrote the test. :)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
HejdaJakub: Yes, it looks better. I changed it. :+1:

Vojtech-Sassmann (approved):

PR #3232, opened 2021-05-27
Fixed subgroup statuses in group members page 

cuadradek:
- This is a fix for CESNET/perun/#3220 - page of group members with
selected group status could contain members with wrong status if they
have different status in subgroups. 
- When getting page of group members with selected group status, all
records with the same group_id and member_id in group_members table are
merged into one row. The group status is evaluated as MIN of all the
group statuses since the member is valid (0) in the group if and only if
he is valid in at least one of the groups. Otherwise, he is expired (1).
- Using group members mapping so membership types can be set to returned
members.
- Added test and modified existing tests to check membership types of
members.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersPageQuery.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersPageQuery.java
@@ -14,6 +14,8 @@ public class MembersPageQuery {
 	private MembersOrderColumn sortColumn;
 	private String searchString = "";
 	private List<Status> statuses;
+	private Group group;
+	private List<MemberGroupStatus> groupStatuses;
 
 	public MembersPageQuery() { }
 	public MembersPageQuery(int pageSize, int offset, SortingOrder sortingOrder, MembersOrderColumn sortColumn) {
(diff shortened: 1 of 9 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersPageQuery.java
```diff
@@ -96,6 +109,22 @@ public void setStatuses(List<Status> statuses) {
 		this.statuses = statuses;
 	}
 
+	public Group getGroup() {
+		return group;
+	}
```
Vojtech-Sassmann, line 114: I didn't notice this earlier we we definitely want to pass here only groupId, not the whole group object. Please, change this to integer and also, don't forget to update the openapi.

perun-openapi/openapi.yml
```diff
@@ -194,6 +201,8 @@ components:
         sortColumn: { $ref: '#/components/schemas/MembersOrderColumn' }
         searchString : { type: string }
         statuses: { type: array, items: { $ref: '#/components/schemas/VoMemberStatuses' } }
+        group: { $ref: '#/components/schemas/Group' }
+        groupStatuses: { type: array, items: { $ref: '#/components/schemas/MemberGroupStatus' } }
```
Vojtech-Sassmann, line 205: Also, there is a file `perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl` which contains example parameters for our API methods. There is a `$objectExamples{"MembersPageQuery"}`. Please, update this example as well.

Vojtech-Sassmann (changes requested):

perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersPageQuery.java
cuadradek: Fixed.

perun-openapi/openapi.yml
cuadradek: Updated.

Vojtech-Sassmann (approved):

PR #3238, opened 2021-06-02
New column status in groups_resources

cuadradek:
- New column 'status' of type integer was added to table
groups_resources. It represents status of group-resource assignment.
- New enum class GroupResourceStatus was created to define possible
values of group-resource assignment status - ACTIVE, INACTIVE, FAILED
or PROCESSING.
- Default status value of all existing assignments is ACTIVE.
- Every SQL query on group_resource table was changed so it queries
only assignments with ACTIVE status.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
@@ -6,6 +6,7 @@ import cz.metacentrum.perun.core.api.BeansUtils;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.GroupResourceStatus;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.PerunSession;
@@ -1396,11 +1397,11 @@ public class UsersManagerImpl implements UsersManagerImplApi {
 	public List<Resource> getAssignedResources(PerunSession sess, User user) {
 		try  {
 			return jdbc.query("select distinct " + resourceMappingSelectQuery + " from resources" +
-					" join groups_resources on resources.id=groups_resources.resource_id " +
+					" join groups_resources on resources.id=groups_resources.resource_id and groups_resources.status=?::group_resource_status " +
 					" join groups on groups_resources.group_id=groups.id" +
 					" join groups_members on groups.id=groups_members.group_id " +
 					" join members on groups_members.member_id=members.id " +
-					" where members.user_id=?", RESOURCE_MAPPER, user.getId());
+					" where members.user_id=?", RESOURCE_MAPPER, GroupResourceStatus.ACTIVE.toString(), user.getId());
 		} catch (RuntimeException e) {
 			throw new InternalErrorException(e);
 		}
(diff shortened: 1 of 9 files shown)


perun-db/postgres.sql
```diff
@@ -762,6 +762,7 @@ create table groups_members (
 create table groups_resources (
 	group_id integer not null,     --identifier of group (groups.id)
 	resource_id integer not null,  --identifier of resource (resources.id)
+	status integer not null default 0,
```
Vojtech-Sassmann, line 765: Did you think about creating the enum also in DB? I think it might be a better option since you don't need to do some custom mapping and you can be always sure that the DB will contain only the specified enum values, not something else.

Vojtech-Sassmann: Also, I'm not sure if we want the `ACTIVE` status to be the default value. I think it would make sense to use the `PROCESSING` as a default value. What do you think?

balcirakpeter (commented):
Code seems to be ok but I agree with Vojta that we do not want to have Active status as default.

perun-db/postgres.sql
cuadradek: You are right. I created enum in DB and set the default value as PROCESSING.

cuadradek:
There are a few uses cases in methods `deleteGroup` or `deleteResource` which could be affected by this change. They should be changed in another PR, so they don't query existing methods like `getAssignedGroups` but new methods like `getGroupAssignments` instead - they need to query every existing group-resource assignment, not just the ACTIVE ones.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1396,11 +1397,11 @@ public void deleteSponsorLinks(PerunSession sess, User sponsor) {
 	public List<Resource> getAssignedResources(PerunSession sess, User user) {
 		try  {
 			return jdbc.query("select distinct " + resourceMappingSelectQuery + " from resources" +
-					" join groups_resources on resources.id=groups_resources.resource_id " +
+					" join groups_resources on resources.id=groups_resources.resource_id and groups_resources.status=?::group_resource_status::group_resource_status " +
```
balcirakpeter, line 1400: There seems to be an extra `::group_resource_status`.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1442,13 +1443,13 @@ public void deleteSponsorLinks(PerunSession sess, User sponsor) {
 					FacilitiesManagerImpl.facilityMappingSelectQuery + ", "+resourceTagMappingSelectQuery+" from resources" +
 					" join vos on resources.vo_id=vos.id " +
 					" join facilities on resources.facility_id=facilities.id " +
-					" join groups_resources on resources.id=groups_resources.resource_id " +
+					" join groups_resources on resources.id=groups_resources.resource_id and groups_resources.status=?::group_resource_status::group_resource_status " +
```
balcirakpeter, line 1446: Same here

balcirakpeter (changes requested):

Vojtech-Sassmann (approved):
I think this is ok, except for the unneeded double cast, mentioned by Peter.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
cuadradek: Fixed, thank you.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
cuadradek: Fixed

balcirakpeter (approved):

PR #3241, opened 2021-06-07
Disabled and hidden icons added also to old GUI

HejdaJakub:
* To the form configuration were added two columns which display hidden and disabled item properties. In columns are also displayed hidden and disabled dependencies.
* To the form preview were added icons which refer to hidden and disabled item properties. In the icon tooltip is dependency info.
* If some item has property disabled='ALWAYS', then is shown as disabled on the form preview.
* If some item has property hidden='ALWAYS', then is not displayed on the form preview.

### DIFF
--- a/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/json/registrarManager/GetFormItems.java
+++ b/perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/json/registrarManager/GetFormItems.java
@@ -28,6 +28,7 @@ import cz.metacentrum.perun.webgui.widgets.Confirm;
 import cz.metacentrum.perun.webgui.widgets.CustomButton;
 
 import java.util.ArrayList;
+import java.util.stream.Collectors;
 
 /**
  * Returns the form elements in application form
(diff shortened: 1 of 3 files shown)


balcirakpeter (commented):
Will the code work also when some form item has set IF_PREFILLED or IF_EMPTY but without the dependency? Meaning when it is the form item itself prefilled or empty.

Vojtech-Sassmann (approved):

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/json/registrarManager/GetFormItems.java
```diff
@@ -261,21 +270,43 @@ public void onFinished(JavaScriptObject jso){
 			type_label.setTitle(item.getType());
 			ft.setWidget(i, 1, type_label);
 
-			// 2 = preview
+			// 2 = disabled
+			if(item.getDisabled() != "NEVER") {
+				Image lockIcon =  new Image(SmallIcons.INSTANCE.lockIcon());
+				String dependency = item.getDisabledDependencyItemId() == 0 ? "" : items.stream()
```
Vojtech-Sassmann, line 276: This might fix the label when the dependency is not selected.
```suggestion
				String dependency = item.getDisabledDependencyItemId() == 0 ? "it" : items.stream()
```

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/json/registrarManager/GetFormItems.java
```diff
@@ -261,21 +270,43 @@ public void onFinished(JavaScriptObject jso){
...
+					.filter(it -> it.getId() == item.getDisabledDependencyItemId())
+					.collect(Collectors.toList()).get(0).getShortname();
+				VerticalPanel lockPanel = createVPForIcon(lockIcon, item.getDisabled(), "disabled", dependency);
+				ft.setWidget(i, 2, lockPanel);
+			}
+
+			// 3 = hidden
+			if(item.getHidden() != "NEVER") {
+				Image eyeIcon =  new Image(SmallIcons.INSTANCE.eyeIcon());
+				String dependency = item.getHiddenDependencyItemId() == 0 ? "" : items.stream()
```
Vojtech-Sassmann, line 286: Same here.
```suggestion
				String dependency = item.getHiddenDependencyItemId() == 0 ? "it" : items.stream()
```

Vojtech-Sassmann (changes requested):

HejdaJakub:
> Will the code work also when some form item has set IF_PREFILLED or IF_EMPTY but without the dependency? Meaning when it is the form item itself prefilled or empty.

The code works correctly, but the lables could be better in this case, so I edited them.

Vojtech-Sassmann (approved):

PR #3242, opened 2021-06-07
Password-related methods

Johaney-s:
* Implementation of loginExist and deletePassword methods
* Added documentation and refactored other password-related methods

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -5218,6 +5218,7 @@ perun_policies:
 
   loginExist_User_String_policy:
     policy_roles:
+      - PERUNOBSERVER:
       - REGISTRAR:
       - SELF: User
     include_policies:
@@ -5225,6 +5226,8 @@ perun_policies:
 
   service_user-loginExist_User_String_policy:
     policy_roles:
+      - PERUNOBSERVER:
+      - VOOBSERVER:
       - VOADMIN:
     include_policies:
       - default_policy
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -5203,6 +5203,32 @@ perun_policies:
     include_policies:
       - default_policy
 
+  deletePassword_User_String_policy:
+    policy_roles:
+      - REGISTRAR:
+      - SELF: User
+    include_policies:
+      - default_policy
+
+  service_user-deletePassword_User_String_policy:
+    policy_roles:
+      - VOADMIN:
+    include_policies:
+      - default_policy
+
+  loginExist_User_String_policy:
+    policy_roles:
+      - REGISTRAR:
+      - SELF: User
+    include_policies:
+      - default_policy
+
+  service_user-loginExist_User_String_policy:
+    policy_roles:
+      - VOADMIN:
+    include_policies:
+      - default_policy
+
   createAlternativePassword_User_String_String_String_policy:
     policy_roles:
       - SELF: User
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
@@ -830,16 +830,16 @@ public interface UsersManager {
 	/**
 	 * Changes user password in defined login-namespace. If checkOldPassword is true, then ask authentication system if old password is correct.
 	 *
-	 * @param sess
-	 * @param login
-	 * @param loginNamespace
-	 * @param oldPassword
-	 * @param newPassword
-	 * @param checkOldPassword
+	 * @param sess Perun session
+	 * @param login String representation of the userLogin
+	 * @param loginNamespace Login-namespace to change password in
+	 * @param oldPassword Old password
+	 * @param newPassword New password
+	 * @param checkOldPassword If true, validates old password
 	 * @throws InternalErrorException
 	 * @throws PrivilegeException
-	 * @throws LoginNotExistsException
-	 * @throws PasswordDoesntMatchException
+	 * @throws LoginNotExistsException When user doesn't have login in specified namespace
+	 * @throws PasswordDoesntMatchException When old password does not match
 	 * @throws PasswordChangeFailedException
 	 */
 	void changePassword(PerunSession sess, String login, String loginNamespace, String oldPassword, String newPassword, boolean checkOldPassword)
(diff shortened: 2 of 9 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5203,6 +5203,32 @@ perun_policies:
...
+    include_policies:
+      - default_policy
+
+  service_user-deletePassword_User_String_policy:
+    policy_roles:
+      - VOADMIN:
+    include_policies:
+      - default_policy
+
+  loginExist_User_String_policy:
```
balcirakpeter, line 5219: Please add also privileges for PERUNOBSERVER.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5203,6 +5203,32 @@ perun_policies:
...
+    policy_roles:
+      - REGISTRAR:
+      - SELF: User
+    include_policies:
+      - default_policy
+
+  service_user-loginExist_User_String_policy:
```
balcirakpeter, line 5226: Please add privileges for PERUNOBSERVER and VOOBSERVER.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -895,92 +895,121 @@ void changePassword(PerunSession sess, User user, String loginNamespace, String
...
 	 * @throws PasswordCreationFailedException
-	 * @throws InvalidLoginException
+	 * @throws InvalidLoginException When login of user has invalid syntax (is not allowed)
 	 */
 	void reservePassword(PerunSession sess, String userLogin, String loginNamespace, String password)
 			throws PasswordCreationFailedException, PrivilegeException, PasswordOperationTimeoutException, PasswordStrengthFailedException, InvalidLoginException, PasswordStrengthException;
 
 	/**
-	 * Reserves the password in external system. User must exists.
+	 * Reserves the password in external system. User must exist.
```
balcirakpeter, line 925: Please add also that login for that user in the specified namespace must exist in Perun when calling this method.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -895,92 +895,121 @@ void changePassword(PerunSession sess, User user, String loginNamespace, String
...
-	 * @throws InvalidLoginException
+	 * @throws InvalidLoginException When login of user has invalid syntax (is not allowed)
 	 */
 	void validatePassword(PerunSession sess, String userLogin, String loginNamespace)
 			throws PasswordCreationFailedException, PrivilegeException, InvalidLoginException;
 
 	/**
 	 * Validates the password in external system and sets user extSources and extSource related attributes.
-	 * User must exists.
+	 * User must exist.
```
balcirakpeter, line 956: Same here. Please add that login in Perun must exist as well.

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -895,92 +895,121 @@ void changePassword(PerunSession sess, User user, String loginNamespace, String
...
-	 * @throws InvalidLoginException
+	 * @throws LoginNotExistsException When user doesn't have login in specified namespace
+	 * @throws InvalidLoginException When login of user has invalid syntax (is not allowed)
+	 * @throws PasswordOperationTimeoutException When password change timed out
 	 */
 	void deletePassword(PerunSession sess, String userLogin, String loginNamespace)
 			throws PasswordDeletionFailedException, PrivilegeException, LoginNotExistsException, PasswordOperationTimeoutException, InvalidLoginException;
 
+	/**
+	 * Deletes password in external system. User must exist.
```
balcirakpeter, line 986: Same here.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/pwdmgr/PasswordManagerModule.java
```diff
@@ -70,4 +69,8 @@ default String handleSponsorship(PerunSession sess, SponsoredUserData userData)
 
 	String generateRandomPassword(PerunSession sess, String login);
 
+	default boolean loginExist(PerunSession sess, String login) {
+		return false;
```
balcirakpeter, line 73: Maybe it would be better to throw an exception instead of false. Something like `OperationNotSupportedException`.
We have this method in OpenAPI so anybody can call it for various namespaces and we do not want to give them false information.

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
Johaney-s: Added for all

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/modules/pwdmgr/PasswordManagerModule.java
Johaney-s: Makes sense, changed for exception.

balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -1148,6 +1146,37 @@ public void deletePassword(PerunSession sess, String userLogin, String loginName
...
+	@Override
+	public void deletePassword(PerunSession sess, User user, String loginNamespace) throws
+			PrivilegeException, UserNotExistsException, LoginNotExistsException, InvalidLoginException, PasswordOperationTimeoutException, PasswordDeletionFailedException {
+		Utils.checkPerunSession(sess);
+
+		// Authorization
+		if(!AuthzResolver.authorizedInternal(sess, "deletePassword_User_String_policy", user)
+			&& (!(AuthzResolver.authorizedInternal(sess, "service_user-deletePassword_User_String_policy", user)) && user.isServiceUser())) {
+			throw new PrivilegeException(sess, "deletePassword");
+		}
```
Vojtech-Sassmann, line 1158: This is a pretty complex condition. It is a bit confusing to understand. I also think there are two problems with it: for the serviceUser, also the first authorizedInternal is evaluated. This way, the service_user policy cannot be more restrictive.

Second, I think there are wrong brackets. The second negation is now only used for the authorizedInternal call. This way, I think that all users, which are not privileged to call this method and also are not service users, would be allowed.

I think this whole thing should be simplified like this:
```suggestion
		boolean authorized;
		if (!user.isServiceUser()) {
			authorized = AuthzResolver.authorizedInternal(sess, "deletePassword_User_String_policy", user);
		} else {
			authorized = AuthzResolver.authorizedInternal(sess, "service_user-deletePassword_User_String_policy", user);
		}
		if(!authorized) {
			throw new PrivilegeException(sess, "deletePassword");
		}
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -1148,6 +1146,37 @@ public void deletePassword(PerunSession sess, String userLogin, String loginName
...
+
+	@Override
+	public boolean loginExist(PerunSession sess, User user, String loginNamespace) throws PrivilegeException, UserNotExistsException {
+		Utils.checkPerunSession(sess);
+
+		// Authorization
+		if(!AuthzResolver.authorizedInternal(sess, "loginExist_User_String_policy", user)
+			&& (!(AuthzResolver.authorizedInternal(sess, "service_user-loginExist_User_String_policy", user)) && user.isServiceUser())) {
+			throw new PrivilegeException(sess, "loginExist");
+		}
```
Vojtech-Sassmann, line 1173: Same here.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
Johaney-s: I put this format to all the methods that used it the same way - `reserveRandomPassword`, `reservePassword`, `validatePassword`, `deletePassword` and `loginExist`.

Vojtech-Sassmann (approved):

PR #3244, opened 2021-06-08
Core - new methods to list group-resource assignments

cuadradek:
- New classes AssignedGroup and AssignedResource were added, they
represents group-resource assignments along with its status.
- New methods getGroupAssignments and getResourceAssignments were added,
they list all group-resource assignments for the given resource or group
respectively. A couple of tests were added too.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -28,6 +28,7 @@ import cz.metacentrum.perun.core.api.BeansUtils;
 import cz.metacentrum.perun.core.api.Candidate;
 import cz.metacentrum.perun.core.api.CandidateGroup;
 import cz.metacentrum.perun.core.api.ContactGroup;
+import cz.metacentrum.perun.core.api.EnrichedGroup;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.ExtSourcesManager;
 import cz.metacentrum.perun.core.api.Facility;
@@ -2622,6 +2623,13 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		}
 		return filteredRichGroups;
 
+	}
+
+	@Override
+	public EnrichedGroup filterOnlyAllowedAttributes(PerunSession sess, EnrichedGroup enrichedGroup) {
+		enrichedGroup.setAttributes(AuthzResolverBlImpl
+			.filterNotAllowedAttributes(sess, enrichedGroup.getGroup(), enrichedGroup.getAttributes()));
+		return enrichedGroup;
 	}
 
 	public void setPerunBl(PerunBl perunBl) {
@@ -5599,5 +5607,16 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 	@Override
 	public boolean isGroupForAutoRegistration(PerunSession sess, Group group) {
 		return this.getGroupsManagerImpl().isGroupForAutoRegistration(sess, group);
+	}
+
+	@Override
+	public EnrichedGroup convertToEnrichedGroup(PerunSession sess, Group group, List<String> attrNames) {
+		List<Attribute> attributes;
+		if (attrNames == null) {
+			attributes = perunBl.getAttributesManagerBl().getAttributes(sess, group);
+		} else {
+			attributes = perunBl.getAttributesManagerBl().getAttributes(sess, group, attrNames);
+		}
+		return new EnrichedGroup(group, attributes);
 	}
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
@@ -1,8 +1,11 @@
 package cz.metacentrum.perun.core.impl;
 
+import cz.metacentrum.perun.core.api.AssignedGroup;
+import cz.metacentrum.perun.core.api.AssignedResource;
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.BanOnResource;
 import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.EnrichedResource;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.GroupResourceStatus;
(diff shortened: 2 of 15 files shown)


Johaney-s (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
```diff
@@ -1183,6 +1194,32 @@ public void removeBan(PerunSession sess, int memberId, int resourceId) throws Ba
...
 	}
 
+	@Override
+	public List<AssignedResource> getResourceAssignments(PerunSession sess, Group group) {
+		try {
+			return jdbc.query("select " + assignedResourceMappingSelectQuery + " from resources" +
+					" join groups_resources on resources.id=groups_resources.resource_id" +
+					" where groups_resources.group_id=?", ASSIGNED_RESOURCE_MAPPER, group.getId());
+		} catch (EmptyResultDataAccessException e) {
+			return new ArrayList<>();
```
Vojtech-Sassmann, line 1204: This exception cannot be actually thrown by the `jdbc.query` method. It can be thrown by `jdbc.queryForObject` for example. Same for other methods.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -2075,6 +2079,64 @@ public void getEnrichedResourcesForFacilityWithGivenAttributes() throws Exceptio
...
+		AssignedResource expectedResource = new AssignedResource(new EnrichedResource(resource, null), GroupResourceStatus.ACTIVE);
+
+		assertThat(resources.size()).isEqualTo(1);
+		assertThat(resources).containsExactly(expectedResource);
+		assertThat(resources.get(0).getEnrichedResource().getAttributes())
+			.containsExactlyInAnyOrderElementsOf(perun.getAttributesManager().getAttributes(sess, resource));
+	}
+
+	@Test (expected=GroupNotExistsException.class)
+	public void getResourceAssignmentsWhenGroupNotExists() throws Exception {
```
Vojtech-Sassmann, line 2104: Instead of this, you can use the `assertThatExceptionOfType().isThrownBy()` from the assertj library. It makes it more clear where the exception is thrown and it provides several useful methods.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -1375,5 +1375,47 @@ public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	},
+
+	/*#
+	 * Lists all of the resource assignments for the given group. Also, returns specified attributes
+	 * for the resources. If attrNames are null or empty, all resource attributes are returned.
+	 *
+	 * @param group int Group <code>id</code>
+	 * @param attrNames List<String> names of attributes to return
+	 * @return List<AssignedResource> resources for given group with specified attributes
+	 */
```
Vojtech-Sassmann, line 1387: We also mention here which exact exceptions can be thrown by this call. These exceptions are taken from the method at the entry layer. Just the PrivilegedException doesn't need to be mentioned, it is added to all methods automatically.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -1375,5 +1375,47 @@ public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	},
+
+	/*#
+	 * Lists all of the assigned groups for the given resource. Also, returns specified attributes
+	 * for the groups. If attrNames are null or empty, all group attributes are returned.
+	 *
+	 * @param resource int Resource <code>id</code>
+	 * @param attrNames List<String> names of attributes to return
+	 * @return List<AssignedGroup> groups for given resource with specified attributes
+	 */
```
Vojtech-Sassmann, line 1408: Same here.

Vojtech-Sassmann (changes requested):
Good job, the implementation looks good. I have just found some minor problems or possible improvements, you can consider.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -5600,4 +5608,15 @@ public void addGroupsToAutoRegistration(PerunSession sess, List<Group> groups) t
...
 	}
+
+	@Override
+	public EnrichedGroup convertToEnrichedGroup(PerunSession sess, Group group, List<String> attrNames) {
+		List<Attribute> attributes;
+		if (attrNames == null || attrNames.isEmpty() ) {
+			attributes = perunBl.getAttributesManagerBl().getAttributes(sess, group);
+		} else {
+			attributes = perunBl.getAttributesManagerBl().getAttributes(sess, group, attrNames);
+		}
```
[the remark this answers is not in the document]
cuadradek: I was thinking we could create method like `getGroupAssignmentsWithoutAttributes` for that use case like we do with RichMembers. But I like your idea better. On the other hand `convertToEnrichedGroup` would behave differently than `convertToEnrichedResource` so this method should be changed too. Another option could be calling these convert methods from `getAssignments` BL only when the attrNames list is null or non empty. What do you think?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
cuadradek: Removed

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
cuadradek: Replaced

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
cuadradek: Added

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
cuadradek: Added

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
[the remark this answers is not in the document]
cuadradek: Ok, so I changed the behavior in `convertToEnrichedGroup` and also updated the java doc of `getGroupAssignments` in all the layers accordingly.

Vojtech-Sassmann (approved):

PR #3245, opened 2021-06-09
Password manager for mu-adm namespace

bodnara:
* created password manager module for mu-adm namespace with simple login and password check

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
@@ -1,0 +1,52 @@
+package cz.metacentrum.perun.core.impl.modules.pwdmgr;
+
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.exceptions.InvalidLoginException;
+import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthException;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+	}
+
+	@Override
+	public void checkPasswordStrength(PerunSession sess, String login, String password) throws PasswordStrengthException {
+		if (StringUtils.isBlank(password)) {
+			log.warn("Password for {}:{} cannot be empty.", "mu-adm", login);
+			throw new PasswordStrengthException("Password for " + "mu-adm" + ":" + login + " cannot be empty.");
+		}
```
Vojtech-Sassmann, line 32: I think we should add here something more complex. At least force the length of the password to be higher than 10. You can take a look at the `EinfraPasswordManagerModule` for inspiration.

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann:
@bodnara Please, also add this constructor to the module: 

```Java
	public MuadmPasswordManagerModule() {
		actualLoginNamespace = "mu-adm";

		if (!binTrue.equals(this.passwordManagerProgram)) {
			passwordManagerProgram += "." + actualLoginNamespace;
		}
	}
```

balcirakpeter (approved):

Vojtech-Sassmann (approved):

PR #3246, opened 2021-06-10
Core - replace old group-resource methods

Johaney-s:
* Use new methods getGroupAssignments and getResourceAssignments where suitable
* The original methods work only with ACTIVE relationships (potential conflicts in DB)
* Handle INACTIVE relationships separately when copying so that they stay inactive in new relationship


Johaney-s:
Uses currently not implemented `allowGroupsToResources()`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -374,8 +381,17 @@ public void assignGroupToResource(PerunSession sess, Group group, Resource resou
...
+	public void assignGroupToResourceAsInactive(PerunSession perunSession, Group group, Resource resource) throws GroupResourceMismatchException, WrongReferenceAttributeValueException, WrongAttributeValueException {
+		assignGroupsToResource(perunSession, List.of(group), resource, GroupResourceStatus.INACTIVE);
+	}
+
 	@Override
 	public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> groups, Resource resource) throws WrongAttributeValueException, WrongReferenceAttributeValueException, GroupResourceMismatchException {
+		assignGroupsToResource(perunSession, groups, resource, GroupResourceStatus.ACTIVE);
+	}
+
+	public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> groups, Resource resource, GroupResourceStatus status) throws WrongAttributeValueException, WrongReferenceAttributeValueException, GroupResourceMismatchException {
```
Vojtech-Sassmann, line 394: This method that takes the status should be available only at the impl layer. We don't want to allow anyone to change the status to 'FAILED' and 'ACTIVE' directly.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -440,8 +460,11 @@ public void removeGroupFromResource(PerunSession sess, Group group, Resource res
 			throw new InternalErrorException("Group " + group + " and resource " + resource + " belongs to the different VOs");
 		}
 
-		// Check if the group was defined on the resource
-		if (!this.getAssignedGroups(sess, resource).contains(group)) {
+		// Check if the group was defined on the resource (with any GroupResource status)
+		List<Group> groupsOnResource = getGroupAssignments(sess, resource, List.of()).stream()
+			.map(g -> g.getEnrichedGroup().getGroup())
+			.collect(Collectors.toList());
+		if (!groupsOnResource.contains(group)) {
```
Vojtech-Sassmann, line 467: I don't think we need to change this method, it will be replaced with a new one.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
```diff
@@ -530,6 +530,22 @@ public void assignGroupToResource(PerunSession sess, Group group, Resource resou
 		}
 	}
 
+	@Override
+	public void assignGroupToResourceAsInactive(PerunSession sess, Group group, Resource resource) throws GroupAlreadyAssignedException {
```
Vojtech-Sassmann, line 534: This method is the one where you should be able to specify the status as an argument.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -832,9 +855,14 @@ public void copyServices(PerunSession sess, Resource sourceResource, Resource de
...
 	public void copyGroups(PerunSession sess, Resource sourceResource, Resource destinationResource) {
-		for (Group group: getAssignedGroups(sess, sourceResource)) {
+		for (AssignedGroup group: getGroupAssignments(sess, sourceResource, List.of())) {
 			try {
-				assignGroupToResource(sess, group, destinationResource);
+				// try to assign groups to new resource, directly set INACTIVE ones as INACTIVE in destinationResource
+				if (group.getStatus() == GroupResourceStatus.INACTIVE) {
+					assignGroupToResourceAsInactive(sess, group.getEnrichedGroup().getGroup(), destinationResource);
+				} else {
+					allowGroupsToResources(sess, group.getEnrichedGroup().getGroup(), destinationResource, false);
```
Vojtech-Sassmann, line 864: For now, use here the old method to assign these groups to the resource.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
Johaney-s: For now set ACTIVE as default call to impl and removed the status attribute.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
Johaney-s: Merged to the original `assignGroupToResource` method (impl layer) with status attribute

Vojtech-Sassmann (approved):

cuadradek (approved):

PR #3248, opened 2021-06-10
Core - new methods for activating group-resource assignments

cuadradek:
- New methods were added - activateGroupResourceAssignment and
deactivateGroupResourceAssignment with simple tests.
- New public method processGroupResourceActivation was also added to BL
 layer. This method runs in separete transaction so the activation of an assignment can be done asynchronously.
- New methods for getting and setting the group-resource status in DB
were added to impl layer.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
@@ -22,6 +22,7 @@ import cz.metacentrum.perun.core.api.BanOnResource;
 import cz.metacentrum.perun.core.api.EnrichedResource;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.GroupResourceStatus;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
@@ -47,6 +48,7 @@ import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedFromResourceE
 import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotDefinedOnResourceException;
 import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
+import cz.metacentrum.perun.core.api.exceptions.GroupResourceStatusException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.MemberResourceMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceAlreadyRemovedException;
@@ -70,6 +72,7 @@ import cz.metacentrum.perun.core.bl.ResourcesManagerBl;
 import cz.metacentrum.perun.core.implApi.ResourcesManagerImplApi;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Async;
 
 import java.util.ArrayList;
 import java.util.HashSet;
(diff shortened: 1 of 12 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -1079,6 +1081,74 @@ public EnrichedResource filterOnlyAllowedAttributes(PerunSession sess, EnrichedR
 		return assignedGroups;
 	}
 
+	@Override
+	public void activateGroupResourceAssignment(PerunSession sess, Group group, Resource resource, boolean async) throws WrongReferenceAttributeValueException, GroupResourceMismatchException, WrongAttributeValueException, GroupNotDefinedOnResourceException {
+		GroupResourceStatus status = getResourcesManagerImpl().getGroupResourceStatus(sess, group, resource);
+		if (status == GroupResourceStatus.ACTIVE || status == GroupResourceStatus.PROCESSING) {
+			// silently skip
```
Vojtech-Sassmann, line 1088: I think in the case of `PROCESSING` we don't want to skip the validation. It might not have even started and we might want to run it as soon as possible.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -1079,6 +1081,74 @@ public EnrichedResource filterOnlyAllowedAttributes(PerunSession sess, EnrichedR
...
+		if (async) {
+			getResourcesManagerImpl().setGroupResourceStatus(sess, group, resource, GroupResourceStatus.PROCESSING);
+			new Thread(() -> {
+				try {
+					getPerunBl().getResourcesManagerBl().processGroupResourceActivation(sess, group, resource);
+				} catch (Exception e) {
+					log.error("Activation of group-resource assignment failed.", e);
+					getResourcesManagerImpl().setGroupResourceStatus(sess, group, resource, GroupResourceStatus.FAILED);
+				}
+			}, "activateGroupResourceAssignment").start();
```
Vojtech-Sassmann, line 1101: This async validation might be more complicated. I think we will need to implement a separate component for it. I would suggest leaving a comment here and in the API, forbid any async calls. We will have to do this for all of the new methods. They won't be completely ready for a release so we should prevent any problems that might happen. Just to be sure, you can also throw here some appropriate runtime exception, e.g. `UnsupportedOperationException`

perun-core/src/main/resources/perun-core.xml
```diff
@@ -30,6 +30,7 @@ http://www.springframework.org/schema/context http://www.springframework.org/sch
 		<aop:advisor advice-ref="txAdviceNestedTransaction" pointcut="execution(* cz.metacentrum.perun.core.blImpl.MembersManagerBlImpl.validateMember(..))"/>
 		<aop:advisor advice-ref="txAdviceNestedTransaction" pointcut="execution(* cz.metacentrum.perun.core.blImpl.MembersManagerBlImpl.expireMember(..))"/>
 		<aop:advisor advice-ref="txAdviceNestedTransaction" pointcut="execution(* cz.metacentrum.perun.core.blImpl.MembersManagerBlImpl.createMemberSync(..))"/>
+		<aop:advisor advice-ref="txAdviceNestedTransaction" pointcut="execution(* cz.metacentrum.perun.core.blImpl.ResourcesManagerBlImpl.processGroupResourceActivation(..))"/>
```
Vojtech-Sassmann, line 33: I think this will not work properly. I'm not sure but I think that the nested transaction will be finished together with the main transaction. I think we need for this a completely new transaction. However, I think we are not able to finish this during this sprint.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -1419,5 +1419,55 @@ public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @throw ResourceNotExistsException when the resource doesn't exist
+	 * @throw WrongAttributeValueException when an attribute value has wrong/illegal syntax
+	 * @throw WrongReferenceAttributeValueException when an attribute value has wrong/illegal semantics
+	 * @throw GroupResourceMismatchException when the given group and resource are not from the same VO
+	 * @throw GroupNotDefinedOnResourceException when the group-resource assignment doesn't exist
+	 */
+	activateGroupResourceAssignment {
+		@Override
+		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
+			ac.getResourcesManager().activateGroupResourceAssignment(ac.getSession(),
```
Vojtech-Sassmann, line 1443: Please, force here the async parameter to be always false, or throw some exception.

Vojtech-Sassmann: Also, mention in the javadoc that this method shouldn't be used, yet.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
```diff
@@ -1216,6 +1217,30 @@ public void removeBan(PerunSession sess, int memberId, int resourceId) throws Ba
...
+		} catch (RuntimeException e) {
+			throw new InternalErrorException(e);
+		}
+	}
+
+	@Override
+	public void setGroupResourceStatus(PerunSession sess, Group group, Resource resource, GroupResourceStatus status) {
+		try {
+			jdbc.update("update groups_resources set status=?::group_resource_status, modified_by=?, modified_by_uid=?, " +
+					" modified_at=" + Compatibility.getSysdate() + " where group_id=? and resource_id=?", status.toString(),
```
Vojtech-Sassmann, line 1237: Since we no longer support multiple DB types, you can simplify this:
```suggestion
					" modified_at=statement_timestamp() where group_id=? and resource_id=?", status.toString(),
```

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann:
@cuadradek I have also just realized, that we don't want to add these new methods during this sprint. It wasn't part of the issue's specification. This is probably my fault since I have declared all of these methods in the OpenAPI and I have also forgotten about that. Therefor, you don't need to rush with these changes and we can finish this during the next sprint.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
cuadradek: I removed `PROCESSING` from the if condition.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
cuadradek: Replaced.

perun-core/src/main/resources/perun-core.xml
cuadradek: Nested propagation creates a completely new transaction in this case since it is invoked from a new thread.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
cuadradek: We decided to implement the async validation using Spring's `@Async` execution.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -1342,6 +1343,34 @@ public void removeResourceSelfServiceGroup(PerunSession sess, Resource resource,
...
 	}
 
+	@Override
+	public void activateGroupResourceAssignment(PerunSession sess, Group group, Resource resource, boolean async) throws ResourceNotExistsException, GroupNotExistsException, PrivilegeException, WrongReferenceAttributeValueException, GroupNotDefinedOnResourceException, GroupResourceMismatchException, WrongAttributeValueException {
+		Utils.checkPerunSession(sess);
+		getResourcesManagerBl().checkResourceExists(sess, resource);
+		getPerunBl().getGroupsManagerBl().checkGroupExists(sess, group);
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "activateGroupResourceAssignment_Group_Resource_boolean_policy", Arrays.asList(group, resource))) {
```
Vojtech-Sassmann, line 1353: I think we have created a new version of this method that can simplify this usage:
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "activateGroupResourceAssignment_Group_Resource_boolean_policy", group, resource)) {
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
```diff
@@ -1342,6 +1343,34 @@ public void removeResourceSelfServiceGroup(PerunSession sess, Resource resource,
...
+	}
+
+	@Override
+	public void deactivateGroupResourceAssignment(PerunSession sess, Group group, Resource resource) throws PrivilegeException, ResourceNotExistsException, GroupNotExistsException, GroupNotDefinedOnResourceException, GroupResourceStatusException {
+		Utils.checkPerunSession(sess);
+		getResourcesManagerBl().checkResourceExists(sess, resource);
+		getPerunBl().getGroupsManagerBl().checkGroupExists(sess, group);
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "deactivateGroupResourceAssignment_Group_Resource_policy", Arrays.asList(group, resource))) {
```
Vojtech-Sassmann, line 1367: Same here
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "deactivateGroupResourceAssignment_Group_Resource_policy", group, resource)) {
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -1419,5 +1419,55 @@ public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+
+	/*#
+	 * Try to activate the group-resource status. If the async is set to false, the validation is performed
+	 * synchronously. The assignment will be either ACTIVE, in case of a successful synchronous call, or it will be
+	 * PROCESSING in case of an asynchronous call. After the async validation, the state can be either ACTIVE or
+	 * FAILED.
+	 *
+	 * @param group int Group <code>id</code>
+	 * @param resource int Resource <code>id</code>
+	 * @param async boolean if true the validation is performed asynchronously
```
Vojtech-Sassmann, line 1432: I think we could specify here the default value.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
```diff
@@ -1216,6 +1217,30 @@ public void removeBan(PerunSession sess, int memberId, int resourceId) throws Ba
...
+	}
+
+	@Override
+	public void setGroupResourceStatus(PerunSession sess, Group group, Resource resource, GroupResourceStatus status) {
+		try {
+			jdbc.update("update groups_resources set status=?::group_resource_status, modified_by=?, modified_by_uid=?, " +
+					" modified_at=statement_timestamp() where group_id=? and resource_id=?", status.toString(),
+				sess.getPerunPrincipal().getActor(), sess.getPerunPrincipal().getUserId(), group.getId(), resource.getId());
+		} catch (RuntimeException e) {
+			throw new InternalErrorException(e);
```
Vojtech-Sassmann, line 1240: I think we should check here the number of affected rows. If the number is 0, we should throw some exception, that the group is not assigned to the resource.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
cuadradek: Done.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntry.java
cuadradek: Done.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
cuadradek: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourcesManagerImpl.java
cuadradek: Done, `GroupNotDefinedOnResourceException` is thrown.

Vojtech-Sassmann (approved):

PR #3254, opened 2021-06-15
Facility in AssignedResource class

bodnara:
* added facility to AssignedResource class


perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -2093,12 +2093,14 @@ public void getResourceAssignments() throws Exception {
 		resourcesManager.assignGroupToResource(sess, group, resource);
 
 		List<AssignedResource> resources = resourcesManager.getResourceAssignments(sess, group, null);
-		AssignedResource expectedResource = new AssignedResource(new EnrichedResource(resource, null), GroupResourceStatus.ACTIVE);
+		AssignedResource expectedResource = new AssignedResource(new EnrichedResource(resource, null), GroupResourceStatus.ACTIVE, facility);
+		Facility expectedFacility = new Facility(-1, "ResourcesManagerTestFacility");
```
Vojtech-Sassmann, line 2097: Instead of the `expectedFacility` you can use the `facility` variable. I think it is the expected facility.

Vojtech-Sassmann (changes requested):

PR #3255, opened 2021-06-15
LDAP attribute rpLoginURL

balcirakpeter:
- Facility attribute rpLoginURL added to LDAP schema and to ldapc
  mapping.


Vojtech-Sassmann (changes requested):
The changes look good but I don't understand what is the 'perun' submodule that has been added in perun-utils? We should probably get rid of it.

zlamalp (approved):

zlamalp:
Manually tested on devel, works.

PR #3257, opened 2021-06-17
Mu-adm password strenght

bodnara:
* added multiple checks for password strength in namespace mu-adm
* added tests for password strength

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
@@ -8,9 +8,16 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.regex.Pattern;
+
 public class MuadmPasswordManagerModule extends GenericPasswordManagerModule {
 
 	private final static Logger log = LoggerFactory.getLogger(MuadmPasswordManagerModule.class);
+
+	private static final Pattern digitPattern = Pattern.compile(".*[0-9].*");
+	private static final Pattern lowerCasePattern = Pattern.compile(".*[a-z].*");
+	private static final Pattern upperCasePattern = Pattern.compile(".*[A-Z].*");
+	private static final Pattern specialCharPattern = Pattern.compile(".*[\\x20-\\x2F\\x3A-\\x40\\x5B-\\x60\\x7B-\\x7E].*");
 
 	public MuadmPasswordManagerModule() {
 		this.actualLoginNamespace = "mu-adm";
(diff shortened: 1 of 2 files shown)


metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
```diff
@@ -8,10 +8,17 @@
...
+import java.util.regex.Pattern;
+
 public class MuadmPasswordManagerModule extends GenericPasswordManagerModule {
 
 	private final static Logger log = LoggerFactory.getLogger(MuadmPasswordManagerModule.class);
 
+	protected final Pattern muadmPasswordContainsDigit = Pattern.compile(".*[0-9].*");
+	protected final Pattern muadmPasswordContainsLower = Pattern.compile(".*[a-z].*");
+	protected final Pattern muadmPasswordContainsUpper = Pattern.compile(".*[A-Z].*");
+	protected final Pattern muadmPasswordContainsSpec = Pattern.compile(".*[\\x20-\\x2F\\x3A-\\x40\\x5B-\\x60\\x7B-\\x7E].*");
```
Vojtech-Sassmann, line 20: If you use these values only in this class, you can make them private. Also, you can make them static.

Also, the names can be simplified. You don't need to name them with muadmPassword, since this is already present in the class name.
```suggestion
	private static final Pattern digitPattern = Pattern.compile(".*[0-9].*");
	private static final Pattern lowerPattern = Pattern.compile(".*[a-z].*");
	private static final Pattern upperPattern = Pattern.compile(".*[A-Z].*");
	private static final Pattern specCharsPattern = Pattern.compile(".*[\\x20-\\x2F\\x3A-\\x40\\x5B-\\x60\\x7B-\\x7E].*");
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModule.java
```diff
@@ -39,14 +46,29 @@ public void checkPasswordStrength(PerunSession sess, String login, String passwo
...
-			throw new PasswordStrengthException("Password for " + actualLoginNamespace + ":" + login + " is too short. At least 10 characters is required.");
+		if (password.length() < 14) {
+			log.warn("Password for {}:{} is too short. At least 14 characters are required.", actualLoginNamespace, login);
+			throw new PasswordStrengthException("Password for " + actualLoginNamespace + ":" + login + " is too short. At least 14 characters is required.");
 		}
 
-		if (!StringUtils.isAsciiPrintable(password)) {
-			log.warn("Password for {}:{} must contain only printable characters.", actualLoginNamespace, login);
-			throw new PasswordStrengthException("Password for " + actualLoginNamespace + ":" + login + " must contain only printable characters.");
+		if (login != null) {
```
Vojtech-Sassmann, line 54: Just to be sure, this check should be performed only if the longer is 3 or more chars longer (this is taken from the specifiaction in AD).
```suggestion
		if (login != null && login.length() > 2) {
```

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModuleTest.java
```diff
@@ -0,0 +1,62 @@
...
+		List<String> weakPasswords = List.of(
+			"eakfjklsjweiml",
+			"LOSFNXHRKJANCS",
+			"18746982365414",
+			"!@#$%^&*_-+=?/",
+			"AkoHCNqrXMsuqm",
+			"afgr487xiy2m63",
+			"!lop@#twoep&*/",
+			"RT672HWZ1068KK",
+			"A!@#KDSAJ?/NMO",
```
Vojtech-Sassmann, line 35: This test definitely covers a lot of cases. However, it is a good practice to test only a single case, which might fail, in a single test. It is also a good practice to put the name of the tested case to the test name e.g.: `loginInPasswordFails` or `reverseLoginInPasswordFails`. 

This is because it is much clearer, what the purpose of a single test is. Also, when some of these tests fail, you immediately know what has failed. With this kind of test, you would need to debug it first to see what failed.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModuleTest.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	@Test
+	public void strongPasswordsTest() {
+		List<String> strongPasswords = List.of(
+			"dskglJDOK46854",
+			"bcvjoEWIQO!@#$",
+			"viwuq24736/?+-",
+			"WMKCO03877@#/*",
+			"jfsUOIW5798+-*"
```
Vojtech-Sassmann, line 51: Same for these.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/pwdmgr/MuadmPasswordManagerModuleTest.java
```diff
@@ -0,0 +1,62 @@
...
+		);
+
+		strongPasswords.forEach(password -> {
+			try {
+				module.checkPasswordStrength(sess, "login", password);
+			} catch (PasswordStrengthException e) {
+				fail();
+			}
```
Vojtech-Sassmann, line 59: If you want to test that some exception is thrown, I suggest using the `assertThatExceptionOfType()` method from the assertJ library. For inspiration, you can take a look at `urn_perun_vo_attribute_def_def_applicationExpirationRulesTest` where this method is already used.

Vojtech-Sassmann (changes requested):

Vojtech-Sassmann (changes requested):
Good job with rewriting the test, I think it looks much better now. I have found just some minor problems with some of the tests.

metodej (approved):
I think the tests look great now.

PR #3260, opened 2021-06-18
Registrar - new module for mu-adm

cuadradek:
- Added new registrar module for mu-adm. It is used to pre-generate
login on application form. Format of generated login is {uco}adm.


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Muadm.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Registration module for mu-adm. It is used to pre-generate login on application form.
+ *
+ * @author Radoslav Čerhák <r.cerhak@gmail.com>
+ */
+public class Muadm extends DefaultRegistrarModule {
+
+	private final static Logger log = LoggerFactory.getLogger(Muadm.class);
+
+	private static final String URN_USER_D_LOGIN_MU = AttributesManager.NS_USER_ATTR_DEF + ":" + "login-namespace:mu";
```
Vojtech-Sassmann, line 31:
```suggestion
	private static final String URN_USER_D_LOGIN_MU = AttributesManager.NS_USER_ATTR_DEF + ":login-namespace:mu";
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Muadm.java
```diff
@@ -0,0 +1,80 @@
...
+				// do not generate login if destination attribute is not set (won't be stored)
+				if (StringUtils.isEmpty(item.getFormItem().getPerunDestinationAttribute())) continue;
+
+				// set new generated value
+				item.setPrefilledValue(generateLogin(session));
+
+				// mark value as generated so the GUI allows editing and on submit server process new login
+				item.setGenerated(true);
+			}
+		}
```
Vojtech-Sassmann, line 54: I have some general tips. If you don't want, you don't need to implement this.
```suggestion
		formItems.stream()
			.filter(item -> item.getFormItem().getType() == USERNAME)
			.filter(item -> isEmpty(item.getPrefilledValue()))
			.filter(item -> isNotEmpty(item.getFormItem().getPerunDestinationAttribute()))
			.forEach(item -> item.setPrefilledValue(generateLogin(session)));
```

* Avoid inline commenting - the code itself should be self-explaining, if you feel like you need to write some inline comment, most of the time this only suggests that you should create a well-named private method (with a proper JavaDoc). The very bad thing with inline comments is that with some future changes, the comments might not correspond to the code anymore (some1 changed the code but forget to update the inline comments). Also, it makes the code much longer if you comment on each line. 
* I have also posted a suggestion of rewriting this method using the Stream API, you can take a look at it. If you like it, you can use it.
* When using methods like the `isEmpty` its usually better to statically import these methods, so the code is cleaner. It doesn't really matter, where the method comes from. 
* When comparing Enum values, use `==` instead of equals. This makes it much clear that you are using constant enum values and not just some strange object instances.

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/Muadm.java
cuadradek: Thanks for the feedback, I used the solution with Stream API.

Vojtech-Sassmann (approved):

PR #3264, opened 2021-06-22
feat(cli): Add getAllAttributes methods

Vojtech-Sassmann:
* In some scripts we need to access all attributes for given objects. These methods were missing and are now added.


Johaney-s (approved):
Tested all possibilities and everything works fine.

zlamalp (approved):

PR #3265, opened 2021-06-22
GUI: Fixed change of application approval mode

zlamalp:
- Do not read and store content of hidden select,
  for embedded applications since its not initialized
  with values.
  This is quickfix. We should check if it stores correct
  value for embedded type this way and maybe we could
  initialize the values anyway.


Vojtech-Sassmann (approved):
I think this should be correct.

Vojtech-Sassmann:
I think that the `form` is the object we received from the API, so it should have set the embedded value.

PR #3271, opened 2021-06-29
Method returning RichMembers from group by memberIds

Johaney-s:
* getGroupRichMembersByIds returns RichMembers from given group including only specified members
* RichMembers include only requested attributes
* supported attribute types are member, user and member-group attributes

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
@@ -1458,4 +1458,35 @@ public class GroupsManagerEntry implements GroupsManager {
 
 		return getGroupsManagerBl().getIndirectMembershipPaths(sess, member, group);
 	}
+
+	@Override
+	public List<RichMember> getGroupRichMembersByIds(PerunSession sess, int groupId, List<Integer> memberIds, List<String> attrNames) throws GroupNotExistsException, PrivilegeException, AttributeNotExistsException  {
+		Utils.checkPerunSession(sess);
+
+		Group group = groupsManagerBl.getGroupById(sess, groupId);
+		getGroupsManagerBl().checkGroupExists(sess, group);
+
+		// Authorization
+		if ( !AuthzResolver.authorizedInternal(sess, "getGroupRichMembersByIds_int_List<Integer>_List<String>_policy", group)) {
+			throw new PrivilegeException(sess, "getGroupRichMembersByIds");
+		}
+
+		List<AttributeDefinition> attributeDefinitions = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attributeDefinitions.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		List<Member> members = new ArrayList<>();
+		for (Integer memberId : memberIds) {
+			try {
+				members.add(groupsManagerBl.getGroupMemberById(sess, group, memberId));
+			} catch (NotGroupMemberException e) {
+				// silently skip
+			}
+		}
+		members.removeIf(member -> !AuthzResolver.authorizedInternal(sess, "filter-getGroupRichMembersByIds_int_List<Integer>_List<String>_policy", member));
+
+		List<RichMember> richMembers = perunBl.getMembersManagerBl().getRichMembersWithAttributes(sess, group, members, attributeDefinitions);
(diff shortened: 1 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -1458,4 +1458,39 @@ public boolean canExtendMembershipInGroupWithReason(PerunSession sess, Member me
...
+
+		List<Member> members = perunBl.getMembersManagerBl().getMembersByIds(sess, memberIds);
+		members.removeIf(member -> !AuthzResolver.authorizedInternal(sess, "filter-getGroupRichMembersByIds_int_List<Integer>_List<String>_policy", member));
+		members.removeIf(member -> {
+			try {
+				return !isGroupMember(sess, group, member);
+			} catch (PrivilegeException e) {
+				// if can't check membership, remove from requested members
+				return true;
+			} catch (MemberNotExistsException | GroupNotExistsException e) {
```
Vojtech-Sassmann, line 1487: These catch blocks can be collapsed into a single one.
```suggestion
			} catch (MemberNotExistsException | GroupNotExistsException | PrivilegeException e) {
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -1458,4 +1458,39 @@ public boolean canExtendMembershipInGroupWithReason(PerunSession sess, Member me
...
+		if ( !AuthzResolver.authorizedInternal(sess, "getGroupRichMembersByIds_int_List<Integer>_List<String>_policy", group)) {
+			throw new PrivilegeException(sess, "getGroupRichMembersByIds");
+		}
+
+		List<AttributeDefinition> attributeDefinitions = new ArrayList<>();
+		for (String attrName : attrNames) {
+			attributeDefinitions.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrName));
+		}
+
+		List<Member> members = perunBl.getMembersManagerBl().getMembersByIds(sess, memberIds);
```
Vojtech-Sassmann, line 1479: I'm afraid this won't return the group statuses of the members, which is the main reason why we need this new method. 

If you want you can use the `getGroupMemberById` method. This method returns the searched member with all of its group statuses. 

If you want to return all of the members by a single SQL into the DB, you would need to create a new method on the bl and impl layer as well. You could modify the already present method `getGroupMembersById` so it would take a list of ids and use the SQL `IN` clause. 

The choice is yours.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -5251,6 +5251,55 @@ public void isGroupForAutoRegistration() throws Exception {
...
+
+		//only requested user attribute is included
+		assertTrue(userAttributes.size() == 1);
+		assertTrue(userAttributes.contains(userAttribute1));
+		assertFalse(userAttributes.contains(userAttribute2));
+
+		//member and member-group attributes are included
+		assertTrue(memberAttributes.size() == 3);
+		assertTrue(memberAttributes.containsAll(List.of(memberAttribute1, memberGroupAttributes.get(0), memberGroupAttributes.get(1))));
+	}
```
Vojtech-Sassmann, line 5301: It is usually better to create multiple tests, each focusing on a single use case. It makes the code more readable and also the debugging. (e.g. if a test named `getGroupRichMembersByIds_returnsSpecifiedAttributes` fails, it is more clear what went wrong and is not working.)

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
Johaney-s: Thanks, I opted for the first choice.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
Johaney-s: Split tests into separate cases.

Vojtech-Sassmann (approved):

PR #3277, opened 2021-06-29
perf(core): Optimize member search sql

Vojtech-Sassmann:
* The SQL which reads members by pages was rewritten to achieve a greater performance. The original implementation used DISTINCT SELECT which has a poor performance.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
@@ -63,6 +63,8 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import static org.apache.commons.lang3.ObjectUtils.isEmpty;
+
 public class MembersManagerImpl implements MembersManagerImplApi {
 
 	private static String rejected = "REJECTED";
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -705,98 +707,32 @@ public MemberGroupStatus getUnifiedMemberGroupStatus(PerunSession sess, User use
...
+		return namedParameterJdbcTemplate.query(
+			    select +
+				" WHERE members.vo_id = (:voId)" +
 				statusesQueryString +
-				groupQueryWhere +
 				groupStatusesQueryString +
-				searchStringQueryWhere +
-				" ORDER BY " + query.getSortColumn().getSqlOrderBy(query) +
+					searchQuery +
+			    " ORDER BY " + query.getSortColumn().getSqlOrderBy(query) +
```
cuadradek, line 732: Wrong indentation

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
Vojtech-Sassmann: Fixed.

Vojtech-Sassmann:
@cuadradek Hi, I have also rebased the PR and added the search by UUID.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
cuadradek: It is still the same :)

Vojtech-Sassmann: Oh, I see. You meant spaces instead of a tab. That should be fixed now.

cuadradek (approved):

PR #3285, opened 2021-07-14
feat(core): Add ResourceAssignmentActivator component

cuadradek:
- New class GroupResourceAssignment was added, it represents a
group-resource assignment with its status.
- New method getGroupResourceAssignments was added and tested, it lists
all group-resource assignments with given statuses.
- ResourceAssignmentActivator component was added. After Perun startup,
the component tries to activate all group-resource assignments in
PROCESSING or FAILED state.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentActivator.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentActivator.java
@@ -1,0 +1,82 @@
+package cz.metacentrum.perun.core.impl;
+
+import cz.metacentrum.perun.core.api.ExtSourcesManager;
+import cz.metacentrum.perun.core.api.GroupResourceAssignment;
+import cz.metacentrum.perun.core.api.GroupResourceStatus;
+import cz.metacentrum.perun.core.api.PerunClient;
+import cz.metacentrum.perun.core.api.PerunPrincipal;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.bl.PerunBl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextRefreshedEvent;
+import org.springframework.lang.NonNull;
+import org.springframework.scheduling.annotation.Async;
+
+import java.util.List;
+
+/**
+ * Component responsible for activating group-resource assignments in PROCESSING or
+ * FAILED state after Perun startup.
+ *
+ * @author Radoslav Čerhák <r.cerhak@gmail.com>
+ */
+public class ResourceAssignmentActivator implements ApplicationListener<ContextRefreshedEvent> {
+
+	private final static Logger log = LoggerFactory.getLogger(ResourceAssignmentActivator.class);
+
+	private final PerunSession sess;
+	private PerunBl perunBl;
+
+	public ResourceAssignmentActivator(PerunBl perunBl) {
+		this.perunBl = perunBl;
+		this.sess = perunBl.getPerunSession(
+			new PerunPrincipal("perunResourceAssignmentActivator", ExtSourcesManager.EXTSOURCE_NAME_INTERNAL, ExtSourcesManager.EXTSOURCE_INTERNAL),
+			new PerunClient());
+	}
+
+	public PerunBl getPerunBl() {
+		return perunBl;
+	}
+
+	public void setPerunBl(PerunBl perunBl) {
+		this.perunBl = perunBl;
+	}
+
+	/**
+	 * Tries to activate all group-resource assignments in PROCESSING or FAILED state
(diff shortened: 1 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/Activator.java
```diff
@@ -0,0 +1,82 @@
...
+			log.debug("Activator starting to activate group-resource assignments in PROCESSING or FAILED state.");
+
+			List<GroupResourceAssignment> assignments = perunBl.getResourcesManagerBl()
+				.getGroupResourceAssignments(sess, List.of(GroupResourceStatus.PROCESSING, GroupResourceStatus.FAILED));
+
+			for (GroupResourceAssignment assignment : assignments) {
+				try {
+					perunBl.getResourcesManagerBl()
+						.activateGroupResourceAssignment(sess, assignment.getGroup(), assignment.getResource(), false);
+				} catch (Throwable e) {
```
Vojtech-Sassmann, line 74: In general, you should never catch a `Throwable` since it also covers `Errors` which represent unrecoverable problems. You should just catch here `Exception`.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Activator.java
```diff
@@ -0,0 +1,82 @@
...
+					log.error("Cannot activate group-resource assignment: " + assignment, e);
+				}
+			}
+		} catch (Throwable e) {
```
Vojtech-Sassmann, line 78: Same here.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Activator.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.List;
+
+/**
+ * Component responsible for activating group-resource assignments in PROCESSING or
+ * FAILED state after Perun startup.
+ *
+ * @author Radoslav Čerhák <r.cerhak@gmail.com>
+ */
+public class Activator implements ApplicationListener<ContextRefreshedEvent> {
```
Vojtech-Sassmann, line 25: The `Activator` name could be a bit confusing since it doesn't have an additional context about what is being activated.  You can consider renaming it, maybe to something like `ResourceAssignmentActivator` ? If you do so, also please update the commit message.

Vojtech-Sassmann (changes requested):
I have just found minor improvements, otherwise, it looks pretty good. After the merge, we should test these changes on the devel instance.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Activator.java
cuadradek: Fixed, thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Activator.java
cuadradek: Fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Activator.java
cuadradek: Renamed to ResourceAssignmentActivator, commit message was also updated.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentActivator.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.util.List;
+
+/**
+ * Component responsible for activating group-resource assignments in PROCESSING or
+ * FAILED state after Perun startup.
+ *
+ * @author Radoslav Čerhák <r.cerhak@gmail.com>
+ */
+public class ResourceAssignmentActivator implements ApplicationListener<ContextRefreshedEvent> {
```
Vojtech-Sassmann, line 25: I have also done some digging and found out that this event could be actually called twice on startup. I think we could use the `ApplicationReadyEvent` instead. What do you think? Was there a specific reason why to use the `ContextRefreshedEvent`?

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentActivator.java
cuadradek: I didn't know about `ApplicationReadyEvent` but it is a Spring Boot event, so I don't think it would work. Another option might be invoking the activation using `init-method` but it can't run asynchronously - we would have to inject a bean to the activator component and invoke the async call on the bean from `init-method`.

Vojtech-Sassmann: I see, I didn't realize that. I think we can keep it this way and we can test it on the devel instance and see, how it behaves.

Vojtech-Sassmann (approved):

PR #3286, opened 2021-07-15
fix(core): Unnecessary audit messages

metodej:
* Audit messages are now generated during member group status calculation only if member status changed.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -661,10 +661,13 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 			throw new GroupMoveNotAllowedException("It is not possible to move Members group.", movingGroup, destinationGroup);
 		}
 
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses = new HashMap<>();
+
 		// check if had parent group
 		Group previousParent;
 		try {
 			previousParent = getParentGroup(sess, movingGroup);
+			previousStatuses = getPreviousStatuses(sess, movingGroup, getGroupMembers(sess, movingGroup));
 		} catch (ParentGroupNotExistsException e) {
 			previousParent = null;
 		}
@@ -793,7 +796,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 			if (previousParent != null) {
 
 				// calculate new member-group statuses for members from previous moving group parent
-				recalculateMemberGroupStatusRecursively(sess, member, previousParent);
+				recalculateMemberGroupStatusRecursively(sess, member, previousParent, previousStatuses);
 			}
 		}
 
@@ -908,6 +911,8 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 		boolean memberWasIndirectInGroup = this.isGroupMember(sess, group, member);
 
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses = getPreviousStatuses(sess, group, List.of(member));
+
 		member = getGroupsManagerImpl().addMember(sess, group, member, MembershipType.DIRECT, group.getId());
 		getPerunBl().getAuditer().log(sess, new DirectMemberAddedToGroup(member, group));
 
@@ -930,7 +935,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		if (!VosManager.MEMBERS_GROUP.equals(group.getName())) {
 
 			// recalculate member group state
-			recalculateMemberGroupStatusRecursively(sess, member, group);
+			recalculateMemberGroupStatusRecursively(sess, member, group, previousStatuses);
 		}
 	}
 
@@ -1091,6 +1096,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 		lockGroupMembership(group, Collections.singletonList(member));
 
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses = getPreviousStatuses(sess, group, List.of(member));
 		member.setSourceGroupId(group.getId());
 		getGroupsManagerImpl().removeMember(sess, group, member);
 		if (this.getGroupsManagerImpl().isGroupMember(sess, group, member)) {
@@ -1115,7 +1121,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		}
 
 		if (!VosManager.MEMBERS_GROUP.equals(group.getName())) {
-			recalculateMemberGroupStatusRecursively(sess, member, group);
+			recalculateMemberGroupStatusRecursively(sess, member, group, previousStatuses);
 		}
 
 		if (!getGroupsManagerImpl().isGroupMember(sess, group, member)) {
@@ -4702,16 +4708,20 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		// save group relation
 		groupsManagerImpl.saveGroupRelation(sess, resultGroup, operandGroup, parentFlag);
 
+		List<Member> affectedMembers = getGroupMembers(sess, operandGroup);
+
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses = getPreviousStatuses(sess, operandGroup, affectedMembers);
+
 		// do the operation logic
 		try {
-			addRelationMembers(sess, resultGroup, getGroupMembers(sess, operandGroup), operandGroup.getId());
+			addRelationMembers(sess, resultGroup, affectedMembers, operandGroup.getId());
 		} catch(AlreadyMemberException ex) {
 			throw new ConsistencyErrorException("AlreadyMemberException caused by DB inconsistency.",ex);
 		}
 
 		// calculate new member-group statuses
-		for (Member member : getGroupMembers(sess, operandGroup)) {
-			recalculateMemberGroupStatusRecursively(sess, member, operandGroup);
+		for (Member member : affectedMembers) {
+			recalculateMemberGroupStatusRecursively(sess, member, operandGroup, previousStatuses);
 		}
 
 		return resultGroup;
@@ -4722,6 +4732,8 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		if (!groupsManagerImpl.isOneWayRelationBetweenGroups(resultGroup, operandGroup)) {
 			throw new GroupRelationDoesNotExist("Union does not exist between result group " + resultGroup + " and operand group" + operandGroup + ".");
 		}
+
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses = getPreviousStatuses(sess, operandGroup, getGroupMembers(sess, resultGroup));
 
 		if (parentFlag || groupsManagerImpl.isRelationRemovable(sess, resultGroup, operandGroup)) {
 			try {
@@ -4740,8 +4752,57 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 
 		// recalculates statuses of members in result group
 		for (Member member : getGroupMembers(sess, resultGroup)) {
-			recalculateMemberGroupStatusRecursively(sess, member, resultGroup);
+			recalculateMemberGroupStatusRecursively(sess, member, resultGroup, previousStatuses);
 		}
+	}
+
+	/**
+	 * Returns previous total member-group statuses of members in all result groups of the given group.
+	 *
+	 * The returned value is a map of groupIds to a map of memberId to MemberGroupStatus of member with this id.
+	 *
+	 * @param sess perun session
+	 * @param group group
+	 * @param members members
+	 * @return previous statuses of members in all result groups of the given group
+	 */
+	private Map<Integer, Map<Integer, MemberGroupStatus>> getPreviousStatuses(PerunSession sess, Group group, List<Member> members) {
+		return getPreviousStatuses(sess, group, members, new HashMap<>());
+	}
+
+
+	/**
+	 * Returns previous total member-group statuses of members in all result groups of the given group.
+	 *
+	 * The returned value is a map of groupIds to a map of memberId to MemberGroupStatus of member with this id.
+	 *
+	 * @param sess perun session
+	 * @param group group
+	 * @param members members
+	 * @param previousStatuses previous statuses, empty at the beginning
+	 * @return previous statuses of members in all result groups of the given group
+	 */
+	private Map<Integer, Map<Integer, MemberGroupStatus>> getPreviousStatuses(PerunSession sess, Group group, List<Member> members, Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses) {
+
+		if (members.isEmpty()) {
+			return previousStatuses;
+		}
+
+		previousStatuses.put(group.getId(), getTotalGroupStatusForMembers(sess, group, members));
+
+		List<Group> affectedGroups = groupsManagerImpl.getResultGroups(sess, group.getId());
+
+		for (Group affectedGroup : affectedGroups) {
+			Map<Integer, Map<Integer, MemberGroupStatus>> newStatuses = getPreviousStatuses(sess, affectedGroup, members, previousStatuses);
+
+			for (Integer key : newStatuses.keySet()) {
+				if (previousStatuses.get(key) == null) {
+					previousStatuses.put(key, newStatuses.get(key));
+				}
+			}
+		}
+
+		return previousStatuses;
 	}
 
 	@Override
@@ -4935,10 +4996,12 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 			log.warn("Expiring member in group where is already expired. Member: {}, Group: {}", member, group);
 		}
 
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousIndirectStatuses = getPreviousStatuses(sess, group, List.of(member));
+
 		// expire in given group
 		groupsManagerImpl.setDirectGroupStatus(sess, member, group, MemberGroupStatus.EXPIRED);
 
-		recalculateMemberGroupStatusRecursively(sess, member, group);
+		recalculateMemberGroupStatusRecursively(sess, member, group, previousIndirectStatuses);
 	}
 
 	@Override
@@ -4962,10 +5025,12 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 			log.warn("Validating member in group where is already validated. Member: {}, Group: {}", member, group);
 		}
 
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousIndirectStatuses = getPreviousStatuses(sess, group, List.of(member));
+
 		// validate member in given group
 		groupsManagerImpl.setDirectGroupStatus(sess, member, group, MemberGroupStatus.VALID);
 
-		recalculateMemberGroupStatusRecursively(sess, member, group);
+		recalculateMemberGroupStatusRecursively(sess, member, group, previousIndirectStatuses);
 	}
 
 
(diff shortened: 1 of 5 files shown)


Johaney-s (approved):
It looks good to me, I have only one cosmetic suggestion.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4990,7 +4996,7 @@ public MemberGroupStatus getTotalMemberGroupStatus(PerunSession session, Member
 	 * @throws InternalErrorException internal error
 	 */
 	@Override
-	public void recalculateMemberGroupStatusRecursively(PerunSession sess, Member member, Group group) {
+	public void recalculateMemberGroupStatusRecursively(PerunSession sess, Member member, Group group, MemberGroupStatus previousStatus) {
```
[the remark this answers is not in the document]
metodej: Fixed, thanks!

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -5021,16 +5028,20 @@ public void recalculateMemberGroupStatusRecursively(PerunSession sess, Member me
...
+				} else if (newStatus.equals(MemberGroupStatus.VALID)) {
+					getPerunBl().getAuditer().log(sess, new MemberValidatedInGroup(member, group));
+				}
 			}
 		}
 
 		// check recursively all parent groups
 		for (Group affectedGroup : affectedGroups) {
-			recalculateMemberGroupStatusRecursively(sess, member, affectedGroup);
+			MemberGroupStatus previousStatusInAffectedGroup = getDirectMemberGroupStatus(sess, member, affectedGroup);
```
cuadradek, line 5042: I am not sure, but I think that `previousStatusInAffectedGroup` should be calculated as `getTotalMemberGroupStatus` before `setIndirectGroupStatus` at line 5029, because we care about member's overall status in the group, not just his direct status. What do you think, @metodej @Johaney-s?

metodej: Yes, you are correct, it should, thanks! Fixed.

cuadradek: I think the previous state should be calculated before `setIndirectGroupStatus` because that's where the status might change. Also, it is worth thinking about calculating the previous state similar way in methods `moveGroup`, `createGroupUnion` and `removeGroupUnion`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4703,16 +4710,25 @@ public Group createGroupUnion(PerunSession sess, Group resultGroup, Group operan
...
 			throw new ConsistencyErrorException("AlreadyMemberException caused by DB inconsistency.",ex);
 		}
 
 		// calculate new member-group statuses
-		for (Member member : getGroupMembers(sess, operandGroup)) {
-			recalculateMemberGroupStatusRecursively(sess, member, operandGroup);
+		for (Member member : affectedMembers) {
+			MemberGroupStatus previousStatus = previousStatuses.get(member);
+
+			recalculateMemberGroupStatusRecursively(sess, member, operandGroup, previousStatus);
```
cuadradek, line 4731:
```suggestion
			recalculateMemberGroupStatusRecursively(sess, member, resultGroup, previousStatus);
```
I think it should be sufficient to recalculate statuses in resultGroup, because statuses could change only there. And `previousStatuses` variable contains statuses for `resultGroup` as well.

metodej: I don't think that would be correct, I think that there is a mistake in `previousStatuses`, where I need statuses from `operandGroup` (it should be fixed now). Although the groupStatus won't change for `operandGroup`, we need to change the status in `resultGroup` to match the status in `operandGroup` and in `recalculateMemberGroupStatusRecursively` we propage status of member to all its `resultGroups`, so we change status in `resultGroup` based on `operandGroup`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: Good point, It's fixed. Thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
cuadradek: I think you are right. We need to call `recalculateMemberGroupStatusRecursively` on `operandGroup` because:
1. `addRelationMembers` in `createGroupUnion` inserts new indirect members to DB with default VALID status.
2. then in `recalculateMemberGroupStatusRecursively` the right status for these indirect members is set in `setIndirectGroupStatus`

But in this case, previousStatuses in `recalculateMemberGroupStatusRecursively` are wrong, because they are calculated before `setIndirectGroupStatus` and they calculate with the default VALID status inserted in `addRelationMembers`. Therefore, previousStatuses would always be VALID.

And calculating previousStatuses after `setIndirectGroupStatus` wouldn't be correct either, because they would be actually new statuses.

I am not sure about this, what do you think?

metodej: It's again a good point, thanks. I have thought about it and I haven't come up with anything better than to send all related `MemberGroupStatuses` of the group structure to the recursive method at the beginning (as is in the new commit). I think it should work correctly now although I do not really like the solution. What do you think? Can you think of anything better?

cuadradek: I can't think of anything better, but it should work correctly now. Nevertheless, I would like if some else (@balcirakpeter or @Vojtech-Sassmann) also reviewed this PR to be sure the changes won't break anything.

cuadradek (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4741,8 +4753,39 @@ public void removeGroupUnion(PerunSession sess, Group resultGroup, Group operand
...
+	 * @param group group
+	 * @param members members
+	 * @param previousStatuses previous statuses, empty in at the beginning
+	 * @return previous statuses of members in all result groups of the given group
+	 */
+	private Map<Pair<Group, Member>, MemberGroupStatus> getPreviousStatuses(PerunSession sess, Group group, List<Member> members, Map<Pair<Group, Member>, MemberGroupStatus> previousStatuses) {
+
+		for (Member member : members) {
+			Pair<Group, Member> groupMemberPair = new Pair<>(group, member);
+			previousStatuses.put(groupMemberPair, getTotalMemberGroupStatus(sess, member, group));
```
Vojtech-Sassmann, line 4773: I think this could be a performance problem. If we consider that some groups have thousands of users, this would take a lot of time. I think we could use here some method that would retrieve statuses of all members of this group (create method `getTotalGroupStatusForMembers`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4741,8 +4753,39 @@ public void removeGroupUnion(PerunSession sess, Group resultGroup, Group operand
...
+	/**
+	 * Returns previous total member-group statuses of members in all result groups of the given group.
+	 *
+	 * @param sess perun session
+	 * @param group group
+	 * @param members members
+	 * @param previousStatuses previous statuses, empty in at the beginning
+	 * @return previous statuses of members in all result groups of the given group
+	 */
+	private Map<Pair<Group, Member>, MemberGroupStatus> getPreviousStatuses(PerunSession sess, Group group, List<Member> members, Map<Pair<Group, Member>, MemberGroupStatus> previousStatuses) {
```
Vojtech-Sassmann, line 4769: It would be nice to create an overloaded version of this method, so you don't have to pass an empty map every time you want to call it: 
```Java
private Map<Pair<Group, Member>, MemberGroupStatus> getPreviousStatuses(PerunSession sess, Group group, List<Member> members, Map<Pair<Group, Member>, MemberGroupStatus> previousStatuses) {
    return getPreviousStatuses(sess, group, members, new HashMap<>);
}
```

Vojtech-Sassmann (changes requested):
In general, I cannot think of a better solution for this problem. However, I think we should optimize the method for retrieving statues)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: Good idea, added, thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4980,17 +5039,23 @@ public MemberGroupStatus getTotalMemberGroupStatus(PerunSession session, Member
...
 	 * for all parent groups.
 	 *
 	 * @param member member
 	 * @param group group
+	 * @param previousStatus previousStatus
 	 * @throws InternalErrorException internal error
 	 */
 	@Override
-	public void recalculateMemberGroupStatusRecursively(PerunSession sess, Member member, Group group) {
+	public void recalculateMemberGroupStatusRecursively(PerunSession sess, Member member, Group group, Map<Pair<Group, Member>, MemberGroupStatus> previousStatus) {
```
Vojtech-Sassmann, line 5058: When I'm thinking about this again, I think it might be better to use 
```Java
Map<Integer, Map<Integer, MemberGroupStatus>>
```
to store the statuses (where the first integer is groupId and the second one is memberId).
The problem that I worry about is with our inheritance. Someone could place a `RichGroup` or a `RichMember` to the map together with `Group` or `Map`. That would lead to inconsistent behaviour of the `hashcode()` and `equals()` methods that might cause some bugs that are very hard to find and debug.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -111,6 +114,20 @@
...
 	};
 
+	private static final ResultSetExtractor<Map<Member, List<Integer>>> MEMBER_MEMBERGROUPSTATUS_EXTRACTOR = resultSet -> {
+		Map<Member, List<Integer>> map = new HashMap<>();
+		while (resultSet.next()) {
+			Member member = MembersManagerImpl.MEMBER_MAPPER.mapRow(resultSet, resultSet.getRow());
+			List<Integer> list = new ArrayList<>();
+			if (map.get(member) != null) {
+				list = map.get(member);
+			}
```
Vojtech-Sassmann, line 124: This might help improve performance.
```suggestion
			List<Integer> list = list = map.get(member);
			if (list == null) {
			    list = new ArrayList<>();
			}
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -977,6 +994,40 @@ public MemberGroupStatus getTotalMemberGroupStatus(PerunSession session, Member
...
+		parameters.addValue("groupId", group.getId());
+		parameters.addValue("memberIds", memberIds);
+
+		try {
+			Map<Member, List<Integer>> map = namedParameterJdbcTemplate.query("select " + MembersManagerImpl.memberMappingSelectQuery + ", source_group_status FROM groups_members" +
+				" join members on groups_members.member_id=members.id where group_id=(:groupId) and member_id in (:memberIds)", parameters, MEMBER_MEMBERGROUPSTATUS_EXTRACTOR);
+
+			if (map == null) {
+				return null;
+			}
```
Vojtech-Sassmann, line 1012: I think the result can never be `null` since you return an empty map from the results extractor.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -5324,6 +5318,53 @@ public void getGroupRichMembersByIds_containsGroupStatuses() throws Exception {
 		assertFalse(richMembers.get(0).getGroupStatuses().isEmpty());
 	}
 
+	@Test
```
Vojtech-Sassmann, line 5321: I think you can include one more test where you would test some use-case with subgroups. Somethink like this:

```
  A
  |
  B
```
Where a member would be expired in the group `A` and valid in the group `B`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4741,10 +4753,53 @@ public void removeGroupUnion(PerunSession sess, Group resultGroup, Group operand
 
 		// recalculates statuses of members in result group
 		for (Member member : getGroupMembers(sess, resultGroup)) {
-			recalculateMemberGroupStatusRecursively(sess, member, resultGroup);
+			recalculateMemberGroupStatusRecursively(sess, member, resultGroup, previousStatuses);
 		}
 	}
 
+	private Map<Pair<Group, Member>, MemberGroupStatus> getPreviousStatuses(PerunSession sess, Group group, List<Member> members) {
```
Vojtech-Sassmann, line 4760: This method should also have a javadoc.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: I haven't thought and I think the Map looks better, changed, thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
metodej: Fixed.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
metodej: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
metodej: I think so, but since the method `query` is annotated as `nullable`, the check from idea does not like it... I guess I could assert it's not null or something like that?

Vojtech-Sassmann: I see. If that is the case, you can keep it that way.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -5021,16 +5088,18 @@ public void recalculateMemberGroupStatusRecursively(PerunSession sess, Member me
 		if (saveStatuses) {
 			groupsManagerImpl.setIndirectGroupStatus(sess, member, group, newStatus);
 
-			if (newStatus.equals(MemberGroupStatus.EXPIRED)) {
-				getPerunBl().getAuditer().log(sess, new MemberExpiredInGroup(member, group));
-			} else if (newStatus.equals(MemberGroupStatus.VALID)) {
-				getPerunBl().getAuditer().log(sess, new MemberValidatedInGroup(member, group));
+			if (!newStatus.equals(previousStatus.get(group.getId()).get(member.getId()))) {
```
Vojtech-Sassmann, line 5091: I think it might be worth checking if the previousStatues contain a map for the group. It might happen due to a race condition. If there are no previous statuses for the groupId, it should also audit a message.

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4741,10 +4753,55 @@ public void removeGroupUnion(PerunSession sess, Group resultGroup, Group operand
...
 		}
 	}
 
+	/**
+	 * Returns previous total member-group statuses of members in all result groups of the given group.
+	 *
+	 * @param sess perun session
+	 * @param group group
+	 * @param members members
+	 * @return previous statuses of members in all result groups of the given group
```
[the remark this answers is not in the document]
metodej: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4741,10 +4753,55 @@ public void removeGroupUnion(PerunSession sess, Group resultGroup, Group operand
...
+
+
+	/**
+	 * Returns previous total member-group statuses of members in all result groups of the given group.
+	 *
+	 * @param sess perun session
+	 * @param group group
+	 * @param members members
+	 * @param previousStatuses previous statuses, empty at the beginning
+	 * @return previous statuses of members in all result groups of the given group
```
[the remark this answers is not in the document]
metodej: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
metodej: Good point, added, thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -4702,16 +4708,20 @@ public Group createGroupUnion(PerunSession sess, Group resultGroup, Group operan
 		// save group relation
 		groupsManagerImpl.saveGroupRelation(sess, resultGroup, operandGroup, parentFlag);
 
+		List<Member> affectedMembers = getGroupMembers(sess, operandGroup);
+
+		Map<Integer, Map<Integer, MemberGroupStatus>> previousStatuses = getPreviousStatuses(sess, operandGroup, affectedMembers);
```
balcirakpeter, line 4713: For all other methods, you retrieve previous statuses before the changing operation. However, for this one, you do it after the operation `saveGroupRelation`. Is it ok? Have you tested it?

metodej: I hope so, in this case, if we got previous statuses before `saveGroupRelation`, we wouldn't have necessarily got the previous status of member who was already in the `resultGroup` and we would possibly generate unnecessary audit message. The `saveGroupRelation` does not change statuses of members or members of groups but it makes sure that `resultGroup` is returned in `getResultGroups` of `operandGroup` when retrieving previous statuses. On the other hand during `removeGroupUnion` we want to get previous statuses before removing the link between groups. In `moveGroup` it's a bit more complicated because there we call `createGroupUnion` or `removeGroupUnion` and we need to retrive the previous statuses before that call, otherwise the statuses could have been changed.

balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
balcirakpeter: thanks for the explanation :)

PR #3287, opened 2021-07-15
feat: method assignGroupToResource() and its variants are sync/async

Kuliak:
RPC level checks presence of "async" parameter in provided params and reads it
(in case of missing parameter, default false value is used).
New parameter added into assignGroupToResource() method signatures in entry and bl level,
where it is promoted from rpc level and promoted higher.
BL level ensures sync/async mode itself -- GroupResourceStatus value combined with activateGroupResourceAssignment() method.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
@@ -387,7 +387,6 @@ public class ResourcesManagerBlImpl implements ResourcesManagerBl {
 
 			//first we must assign group
 			try {
-
 				getResourcesManagerImpl().assignGroupToResource(perunSession, g, resource, GroupResourceStatus.PROCESSING);
 				activateGroupResourceAssignment(perunSession, g, resource, async);
 			} catch (GroupAlreadyAssignedException e) {
@@ -395,6 +394,8 @@ public class ResourcesManagerBlImpl implements ResourcesManagerBl {
 			} catch (GroupNotDefinedOnResourceException ex) {
 				throw new ConsistencyErrorException(ex);
 			}
+		}
+
 	}
 
 	@Override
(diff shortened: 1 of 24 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -426,13 +430,21 @@ public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> gr
 			}
 		}
 
+		// activate each group-resource assignment
+		for (Group g : groups) {
+			try {
+				activateGroupResourceAssignment(perunSession, g, resource, async);
```
cuadradek, line 436: I think it is a good idea to reuse `activateGroupResourceAssignment` method, but now the same activation process (setting attributes,..) would be actually performed twice - once in `assignGroupsToResource` before this for cycle and once in `activateGroupResourceAssignment`. So consider using just `activateGroupResourceAssignment`. Also, be aware that this method silently skips activating assignments in `ACTIVE` state.

cuadradek (changes requested):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -626,7 +626,7 @@ public void removeGroupFromResourceWhichIsSubGroup() throws Exception {
 		subGroup = setUpSubGroup(group);
 		facility = setUpFacility();
 		resource = setUpResource();
-		resourcesManager.assignGroupToResource(sess, group, resource);
+		resourcesManager.assignGroupToResource(sess, group, resource, false);
```
Johaney-s, line 629: The build is not passing because of line 659, where the same parameter needs to be added.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -294,15 +293,19 @@ public Vo call(ApiCaller ac, Deserializer parms) throws PerunException {
 		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 			parms.stateChangingCheck();
 
+			boolean async = parms.contains("async") ? parms.readBoolean("async") : false;
```
Johaney-s, line 296: I think the new parameter should be also added to these methods in `openapi` so that we can call it with the parameter from GUI.

Kuliak:
Changes moved to https://github.com/CESNET/perun/pull/3291

PR #3293, opened 2021-07-22
feat(registrar): prefill group application with vo app values

Johaney-s:
* If user has pending application (NEW or VERIFIED) to vo, values that were filled in were not
prefilled in the follow-up group application, but federation values were used instead. Now vo
application values will be used.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -2405,6 +2405,9 @@ public class RegistrarManagerImpl implements RegistrarManager {
 			itemsWithValues.add(new ApplicationFormItemWithPrefilledValue(item, null));
 		}
 
+		// data from pending app to group's VO, use values from attributes which destination in VO application matches source attribute in group application
+		List<ApplicationFormItemData> pendingVoApplicationData = new ArrayList<>();
+
 		// get user and member attributes from DB for existing users
 		if (user != null) {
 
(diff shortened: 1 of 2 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2543,6 +2557,20 @@ public void updateFormItem(PerunSession sess, ApplicationFormItem item) throws P
...
 			}
+
+			// if pending vo application data can be used for group application, overwrite value with it
+			for (ApplicationFormItemData appFormItemData : pendingVoApplicationData) {
+				String vosAppDestinationAttribute = appFormItemData.getFormItem().getPerunDestinationAttribute();
+				// remove password field
+				if (itemW.getFormItem().getType() == PASSWORD) {
+					it.remove();
+					continue;
+				}
```
Vojtech-Sassmann, line 2568: I'm not sure if this should be here. This would probably remove all password fields from all applications if the vo application has been submitted. I think you should remove all of this code.

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Johaney-s: You're right, and the previous checks should cover the necessary cases when we want to remove the password. Removed this piece of code.

Vojtech-Sassmann (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2459,6 +2462,17 @@ public void updateFormItem(PerunSession sess, ApplicationFormItem item) throws P
 			}
 		}
 
+		// pending vo data for current user (if known to Perun) or session principal
+		if (group != null) {
+			Optional<Application> pendingVoApplication = getApplicationsForUser(sess)
+				.stream()
+				.filter(voApp -> voApp.getGroup() == null)
+				.findFirst();
```
balcirakpeter, line 2470: I think you are missing a filter for states NEW and VERIFIED. Now it could return REJECTED/DELETED app

balcirakpeter (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
Johaney-s: For sure, thanks. Added.

balcirakpeter (approved):

PR #3297, opened 2021-07-23
fix(core): Quick fix failing async group-resource assignments

cuadradek:
- Async group-resource assignments used to fail on
GroupNotDefinedOnResourceException because after inserting the
assignment to DB, the transaction didn't commit before new transaction
for async activation needed the assignment.
- Now, if the assignment is not committed yet, the activation is
repeated 5 times after 5 seconds.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
@@ -1093,7 +1093,22 @@ public class ResourcesManagerBlImpl implements ResourcesManagerBl {
 	@Override
 	public void processGroupResourceActivationAsync(PerunSession sess, Group group, Resource resource) {
 		try {
-			processGroupResourceActivation(sess, group, resource);
+			// try activation for 5 times after 5 seconds because the assignment
+			// might not be committed in DB yet
+			boolean successful = false;
+			for (int i = 0; i < 5; ++i) {
+				try {
+					processGroupResourceActivation(sess, group, resource);
+					successful = true;
+					break;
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -1093,7 +1093,22 @@ public void activateGroupResourceAssignment(PerunSession sess, Group group, Reso
...
+					successful = true;
+					break;
+				} catch (GroupNotDefinedOnResourceException ex) {
+					Thread.sleep(5000);
+				}
+			}
+
+			if (!successful) {
+				processGroupResourceActivation(sess, group, resource);
+			}
```
Vojtech-Sassmann, line 1111: Why is here this extra activation? Is it somehow different from the ones in the for-cycle? This way, if the exception was not GroupNotDefinedOnResourceException, it will try to activate it again, which is not something we want to do (at least I think).

Vojtech-Sassmann (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
cuadradek: If the exception was not GroupNotDefinedOnResourceException, it will be catched in outer try-catch block. The extra activation is there because if GroupNotDefinedOnResourceException was catched 5 times in for cycle, it will be now catched in outer try-catch block, logged and FAILED status will be set.

Vojtech-Sassmann (approved):

PR #3301, opened 2021-07-27
feat(registrar): VO and group application notifications adjustment

Kuliak:
Notifications for VO application are based on toEmail attribute (of given VO), if it is null, preferred email is used to notify all VO admins.
Notifications for Group application are based on toEmail attribute of given group, if it is null, preferred email is used to notify all group admins (if there are any), otherwise VO admins are notified.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -1302,6 +1302,19 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 		return getPerunBl().getUsersManagerBl().filterOnlyAllowedAttributes(sess, richAdminsWithAttributes);
 	}
 
+	/**
+	 * Get all richUser administrators for complementary object and role without any attributes.
+	 *
+	 * @param sess perun session
+	 * @param complementaryObject for which we will get administrator
+	 * @param role expected role to filter managers by
+	 *
+	 * @return list of richUser administrators for complementary object and role.
+	 */
+	public static List<RichUser> getRichAdmins(PerunSession sess, PerunBean complementaryObject, String role) throws RoleCannotBeManagedException {
+		return getRichAdmins(sess, complementaryObject, Collections.emptyList(), role, false, false);
+	}
+
 	public String toString() {
 		return getClass().getSimpleName() + ":[]";
 	}
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
@@ -31,6 +31,7 @@ import cz.metacentrum.perun.core.bl.AttributesManagerBl;
 import cz.metacentrum.perun.core.bl.GroupsManagerBl;
 import cz.metacentrum.perun.core.bl.MembersManagerBl;
 import cz.metacentrum.perun.core.bl.UsersManagerBl;
+import cz.metacentrum.perun.core.blImpl.AuthzResolverBlImpl;
 import cz.metacentrum.perun.core.impl.Compatibility;
 import cz.metacentrum.perun.registrar.exceptions.ApplicationMailAlreadyRemovedException;
 import cz.metacentrum.perun.registrar.exceptions.ApplicationMailExistsException;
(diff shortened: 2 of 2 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -852,29 +852,54 @@ private void setFromAndReplyTo(MimeMessage message, String fromName, String from
...
+				return getMailsFromVo(app, URN_VO_TO_EMAIL);
+			}
+
+			attrSenderEmail = attrManager.getAttribute(registrarSession, app.getGroup(), URN_GROUP_TO_EMAIL);
+
+			if (attrSenderEmail == null || attrSenderEmail.getValue() == null) {
+				// not specified toEmail group attribute
+
+				// try to use all group admins with specified preferred emails
+				List<User> admins = groupsManager.getAdmins(registrarSession, app.getGroup(), true);
```
Vojtech-Sassmann, line 880: This is an old method that was used to get admins. We have a new general method in `AuthzResolverBlImpl.getRichAdmins(...)`. You could use the method here like this:
```suggestion
				List<RichUser> admins = AuthzResolverBlImpl.getRichAdmins(registrarSession, app.getGroup(), Collections.emptyList(), "GROUPADMIN", false, false);
```
This method takes a lot of arguments. If you want, you can add an overloaded version of this method:
```Java
public static List<RichUser> getRichAdmins(PerunSession sess, PerunBean complementaryObject, String role) throws RoleCannotBeManagedException {
	return getRichAdmins(sess, complementaryObject, Collections.emptyList(), role, false, false);
}
```
This would simplify the call here and also the one you will need to get Vo admins.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -2115,4 +2140,46 @@ private MimeMessage getInvitationMessage(Vo vo, Group group, String language, St
...
+				if (adr != null && !adr.isEmpty()) {
+					emails.add(adr);
+				}
+			}
+			return emails;
+		}
+
+		// in case no pre-defined attribute was found, use all VO admins with specified preferred email
+		try {
+			List<User> admins = perun.getVosManager().getAdmins(registrarSession, app.getVo(), Role.VOADMIN, true);
```
Vojtech-Sassmann, line 2166: Similar thing here. Please, use the new method from `AuthzResolverBlImpl`.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -852,29 +852,54 @@ private void setFromAndReplyTo(MimeMessage message, String fromName, String from
...
+				for (User admin : admins) {
+					attrSenderEmail = attrManager.getAttribute(registrarSession, admin, URN_USER_PREFERRED_MAIL);
 
-			if (attrToEmail != null && attrToEmail.getValue() != null) {
-				ArrayList<String> value = attrToEmail.valueAsList();
-				for (String adr : value) {
-					if (adr != null && !adr.isEmpty()) {
-						result.add(adr);
+					if (attrSenderEmail != null && attrSenderEmail.getValue() != null) {
+						result.add(BeansUtils.attributeValueToString(attrSenderEmail));
```
Vojtech-Sassmann, line 885: Please, use this instead.
```suggestion
						result.add(attrSenderEmail.valueAsString());
```

Vojtech-Sassmann (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -2115,4 +2141,44 @@ private MimeMessage getInvitationMessage(Vo vo, Group group, String language, St
 		return message;
 	}
 
+	/**
+	 * Get notification emails for given VO. If toEmail is not specified, all VO admins are used.
```
balcirakpeter, line 2145: It is better to write full JavaDoc with parameters, return values, and so on, when when it is only a private method.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -852,29 +853,54 @@ private void setFromAndReplyTo(MimeMessage message, String fromName, String from
 
 	/**
 	 * Get proper values "TO" for mail message based on VO or GROUP attribute "toEmail"
```
balcirakpeter, line 855: This JavaDoc should be updated so it corresponds to the changed functionality.

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1302,6 +1302,19 @@ public static void makeUserPerunAdmin(PerunSession sess, User user) throws Alrea
 		return getPerunBl().getUsersManagerBl().filterOnlyAllowedAttributes(sess, richAdminsWithAttributes);
 	}
 
+	/**
+	 * Get all richUser administrators for complementary object and role.
+	 *
```
Vojtech-Sassmann, line 1307: I would also mention here that the returned users are without any attributes.

Vojtech-Sassmann (changes requested):

balcirakpeter (approved):

PR #3302, opened 2021-07-27
feat(core): store cause of group-resource assignment failure

Johaney-s:
* The error message is now saved to DB when assignment fails
* When the assignment state is switched to ACTIVE or INACTIVE, previous cause of failure is removed
* The cause can only be retrieved for FAILED assignments

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
@@ -47,10 +47,12 @@ import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedFromResourceException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotDefinedOnResourceException;
+import cz.metacentrum.perun.core.api.exceptions.GroupNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.GroupResourceStatusException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.MemberResourceMismatchException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceNotExistsException;
@@ -1113,6 +1115,7 @@ public class ResourcesManagerBlImpl implements ResourcesManagerBl {
 			log.error("Activation of group-resource assignment failed.", e);
 			try {
 				getResourcesManagerImpl().setGroupResourceStatus(sess, group, resource, GroupResourceStatus.FAILED);
+				getResourcesManagerImpl().setFailedGroupResourceAssignmentCause(sess, group, resource, e.getMessage());
 			} catch (GroupNotDefinedOnResourceException groupNotDefinedOnResourceException) {
 				log.error("Group-resource assignment doesn't exist.", groupNotDefinedOnResourceException);
 			}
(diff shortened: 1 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -1175,6 +1181,19 @@ public void deactivateGroupResourceAssignment(PerunSession sess, Group group, Re
...
+		// reset assignment failure cause
+		getResourcesManagerImpl().setFailedGroupResourceAssignmentCause(sess, group, resource, null);
+	}
+
+	@Override
+	public String getFailedGroupResourceAssignmentCause(PerunSession sess, Group group, Resource resource) throws GroupNotDefinedOnResourceException, GroupResourceStatusException {
+		GroupResourceStatus status = getResourcesManagerImpl().getGroupResourceStatus(sess, group, resource);
+		if (status != GroupResourceStatus.FAILED) {
+			throw new GroupResourceStatusException("Cannot read assignment failure cause for not FAILED assignment.");
+		}
```
Vojtech-Sassmann, line 1194: In such scenarios, it is better to just return `null` value (or `Optional<String>` if you want to do it the best way). If you throw an exception, you limit the usage of such a method (you force the method users to handle the exception even if they first check that the status is failing for example).

Vojtech-Sassmann (changes requested):
I'm not sure if we don't want to put the cause directly to objects like `AssignedGroup` and `AssignedResource`? I don't think it is necessary, but it could make life easier for GUI development. What do you think?

Johaney-s:
> I'm not sure if we don't want to put the cause directly to objects like `AssignedGroup` and `AssignedResource`? I don't think it is necessary, but it could make life easier for GUI development. What do you think?

Well, what I was thinking is the cause would be retrieved only in case user clicks the failed status icon, so not very often. But I'll ask GUI people what they prefer and change it accordingly.

Vojtech-Sassmann:
> > I'm not sure if we don't want to put the cause directly to objects like `AssignedGroup` and `AssignedResource`? I don't think it is necessary, but it could make life easier for GUI development. What do you think?
> 
> Well, what I was thinking is the cause would be retrieved only in case user clicks the failed status icon, so not very often. But I'll ask GUI people what they prefer and change it accordingly.

Yeah, I also think this way it is sufficient. If they will agree with you, you can keep it this way.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
Johaney-s: Added `failureCause` to `AssignedResource`, `AssignedGroup` and `GroupResourceAssignment`, replaced throwing `GroupResourceStatusException` for returning `null`.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -3916,6 +3916,17 @@ perun_policies:
...
+    policy_roles:
+      - RESOURCEADMIN: Resource
+      - VOADMIN: Vo
+      - RESOURCESELFSERVICE: Resource
+        GROUPADMIN: Group
+      - TRUSTEDFACILITYADMIN: Vo
+        FACILITYADMIN: Facility
+    include_policies:
+      - default_policy
+
```
Vojtech-Sassmann, line 3929: This can be probably removed.

Vojtech-Sassmann (changes requested):

perun-base/src/main/resources/perun-roles.yml
Johaney-s: Removed

perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,10 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.84
+ALTER TABLE groups_resources ADD COLUMN failure_cause default null;
```
cuadradek, line 10:
```suggestion
ALTER TABLE groups_resources ADD COLUMN failure_cause varchar default null;
```

cuadradek (changes requested):

perun-core/src/main/resources/postgresChangelog.txt
Johaney-s: thanks, added

cuadradek (approved):

PR #3303, opened 2021-07-28
fix(registrar): Constant-time MAC comparison

cuadradek:
- While validating user email, message authentication code (MAC) used to
be compared as String. This could lead to a timing attack, because
String.equals() doesn't implement constant-time algorithm.
- Now, MACs are compared using constant-time MessageDigest.isEqual().

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -15,6 +15,8 @@ import cz.metacentrum.perun.audit.events.RegistrarManagerEvents.MembershipExtend
 import cz.metacentrum.perun.core.api.*;
 import cz.metacentrum.perun.core.api.exceptions.*;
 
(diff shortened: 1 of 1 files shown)


Vojtech-Sassmann (commented):
Hi @cuadradek, did you check all places in our application where do we use the `String.equals()` method similarly? I don't think this is the only place.

Vojtech-Sassmann:
You could also create some general method in our Utils `secureEquals(String s1, String s2)` so the logic can be reused.

cuadradek:
> Hi @cuadradek, did you check all places in our application where do we use the `String.equals()` method similarly? I don't think this is the only place.

Yes, I think this is the only place. We replaced all similar methods for UUID tokens which are compared in Postgres.

Vojtech-Sassmann:
> > Hi @cuadradek, did you check all places in our application where do we use the `String.equals()` method similarly? I don't think this is the only place.
> 
> Yes, I think this is the only place. We replaced all similar methods for UUID tokens which are compared in Postgres.

Cool, I didn't know it was done in Postgres.

Vojtech-Sassmann (approved):

PR #3305, opened 2021-07-29
feat(core): allow empty login in module login-namespace:mu-adm

Johaney-s:
* Login in this login-namespace can be null now

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_mu_adm.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_mu_adm.java
@@ -1,0 +1,23 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace.java
```diff
@@ -73,7 +75,7 @@ public void checkAttributeSyntax(PerunSessionImpl sess, User user, Attribute att
 	@Override
 	public void checkAttributeSemantics(PerunSessionImpl sess, User user, Attribute attribute) throws WrongReferenceAttributeValueException, WrongAttributeAssignmentException {
 		String userLogin = attribute.valueAsString();
-		if (userLogin == null) throw new WrongReferenceAttributeValueException(attribute, null, user, null, "Value can't be null");
+		if (userLogin == null && !allowNoLogin) throw new WrongReferenceAttributeValueException(attribute, null, user, null, "Value can't be null");
 
 		// Get all users who have set attribute urn:perun:member:attribute-def:def:login-namespace:[login-namespace], with the value.
 		List<User> usersWithSameLogin = sess.getPerunBl().getUsersManagerBl().getUsersByAttribute(sess, attribute);
```
metodej, line 81: I am not sure how the `getUsersByAttribute` and `checkReservedLogins` behave for null login but I think the latter would throw `InternalErrorException`. I think we could skip all these checks in this case, what do you think?

Johaney-s: I think the checks could be skipped for null value, then there is no need for setting `allowNoLogin ` and rather check it before calling the superclass method.

metodej (approved):

Vojtech-Sassmann (approved):

PR #3325, opened 2021-08-18
feat(core): allow auto-assignment of subgroups to resource and inacti…

Kuliak:
…ve group-resource assignment

RPC, Entry, BL, Impl levels adjusted to support two flags:
- assignInactive: group-resource assignment with "INACTIVE" status
- autoAssignSubgroups: all subgroups are recursively assigned to given resource as well
BlImpl level changed to provide needed functionality for both flags.
Impl level changed to store information about subgroup auto-assignment.

BREAKING CHANGE: new db version, assignGroupToResource (and its variants) has two new flags for inactive assignment and all subgroups auto-assignment.


cuadradek (changes requested):
Nice work! Please, update also AssignedGroup and AssignedResource objects in file `perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl`.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -394,34 +398,84 @@ public boolean isGroupAssigned(PerunSession sess, Group group, Resource resource
...
 	@Override
-	public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> groups, Resource resource, boolean async) throws WrongAttributeValueException, WrongReferenceAttributeValueException, GroupResourceMismatchException {
-
+	public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> groups, Resource resource, boolean async, boolean assignInactive, boolean autoAssignSubgroups) throws WrongAttributeValueException, WrongReferenceAttributeValueException, GroupResourceMismatchException {
 		for(Group g: groups) {
 			getPerunBl().getAttributesManagerBl().checkGroupIsFromTheSameVoLikeResource(perunSession, g, resource);
 
+			// if group is already assigned with active status, we don't want to change its state or other properties
+			if (isGroupAssignedActive(perunSession, g, resource)) {
+				continue;
```
Johaney-s, line 412: There might be use case when user would manually assign already automatically assigned subgroup to make sure it stays assigned in case the automatically assigned groups are unassigned. I think this special case could be handled here.

perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,11 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.87
+ALTER TABLE groups_resources ADD COLUMN auto_assign_subgroups boolean default false not null;
+ALTER TABLE groups_resources_automatic ADD COLUMN auto_assign_subgroups boolean default false not null;
```
Johaney-s, line 11: default value for `groups_resources_automatic` could be true as it is set to it every time

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -494,39 +495,117 @@ public void assginGroupToResource() throws Exception {
...
+		resourcesManager.assignGroupToResource(sess, group, resource, false, false, true);
+
+		List<AssignedGroup> assignedGroups = resourcesManager.getGroupAssignments(sess, resource, null);
+		assertEquals("two groups (group with subgroup) should be assigned to our Resource", 2, assignedGroups.size());
 
+		assertTrue("Our group should be assigned to resource.", assignedGroups.contains(expectedGroup) || assignedGroups.contains(expectedGroupProcessing));
+		assertTrue("Our subgroup should be assigned to resource.", assignedGroups.contains(expectedSubGroup) || assignedGroups.contains(expectedSubGroupProcessing));
+
+		for (AssignedGroup group : assignedGroups) {
+			assertTrue("All groups should have 'autoAssignSubgroups' flag set to true", group.isAutoAssignSubgroups());
```
cuadradek, line 555: We should probably check this only for the main group. This flag is irrelevant for automatic subgroup assignments.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
```diff
@@ -219,7 +219,17 @@
...
 	/**
-	 * Returns true if the group is assigned to the current resource, false otherwise.
+	 * Returns true if the group is assigned to the current resource with ACTIVE status, false otherwise.
+	 * @param sess
+	 * @param group
+	 * @param resource
+	 * @return true if the group is assigned to the current resource with active status.
+	 * @throws InternalErrorException
+	 */
+	boolean isGroupAssignedActive(PerunSession sess, Group group, Resource resource);
```
cuadradek, line 229: I like the way you created this new method. But I would prefer `isGroupAssigned` to remain as it was. The reason is that we have many other methods like `isUserAssigned`, etc. where _assigned_ means that there is an ACTIVE assignment and it could become a little confusing if we changed it only here. Instead, a method checking if there is a group-resource assignment in any state could be named `hasGroupAssignment(Resource, Group)` or `groupResourceAssignmentExists(Group, Resource)`. What do you think @Kuliak @Johaney-s ?

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
Kuliak: sure, I will change it, so it will not be confusing for anyone.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -394,34 +398,84 @@ public boolean isGroupAssigned(PerunSession sess, Group group, Resource resource
...
+						if (isGroupAssigned(perunSession, subgroup, resource)) {
+							getResourcesManagerImpl().setGroupResourceStatus(perunSession, subgroup, resource, status);
+						} else {
+							getResourcesManagerImpl().assignGroupToResourceState(perunSession, subgroup, resource, status);
+						}
+
+						// if status is PROCESSING, activate group-resource assignment
+						if (!assignInactive) {
+							// automatic subgroups-resource assignment should be async
+							activateGroupResourceAssignment(perunSession, subgroup, resource, true);
```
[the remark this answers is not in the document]
Johaney-s: Now the status even for existing ACTIVE subgroup changes to PROCESSING, saves this status to database with `setGroupResourceStatus` and would pass this condition as `isGroupAssigned` returns` false` for PROCESSING assignments, so the group would be activated again. You should probably not set PROCESSING status in case there is already ACTIVE assignment.

perun-base/src/main/java/cz/metacentrum/perun/core/api/AssignedGroup.java
```diff
@@ -12,12 +12,14 @@
...
 	private String failureCause;
+	private boolean autoAssignSubgroups;
 
-	public AssignedGroup(EnrichedGroup enrichedGroup, GroupResourceStatus status, Integer sourceGroupId, String failureCause) {
+	public AssignedGroup(EnrichedGroup enrichedGroup, GroupResourceStatus status, Integer sourceGroupId, String failureCause, boolean autoAssignSubgroups) {
 		this.enrichedGroup = enrichedGroup;
 		this.status = status;
 		this.sourceGroupId = sourceGroupId;
 		this.failureCause = failureCause;
+		this.autoAssignSubgroups = autoAssignSubgroups;
```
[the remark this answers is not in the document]
cuadradek: Please, add it to equals() and hashCode() in AssignedResource as well.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -356,8 +355,18 @@ public boolean isUserAllowed(PerunSession sess, User user, Resource resource) {
 	}
 
 	@Override
-	public boolean isGroupAssigned(PerunSession sess, Group group, Resource resource) {
-		return getResourcesManagerImpl().isGroupAssigned(sess, group, resource);
+	public boolean isGroupAssigned(Group group, Resource resource) {
```
cuadradek, line 358: Please, add PerunSession argument to these methods. We usually include it even if it is not used.

cuadradek (changes requested):

cuadradek (approved):

Johaney-s (approved):

PR #3328, opened 2021-08-24
feat: added method for rerieving members assigned to resource with status

Johaney-s:
* GetMembersAssignedToResource returns members of assigned groups to given resource and evaluates
the assignment statuses. Returning members are wrapped in AssignedMember object together with the
status.


perun-openapi/openapi.yml
```diff
@@ -149,6 +149,16 @@ components:
       discriminator:
         propertyName: beanName
 
+    AssignedMember:
+      allOf:
+        - $ref: '#/components/schemas/Member'
```
cuadradek, line 154: I think "_allOf_" should be used in case of inheritance, see how _AssignedGroup_ or _AssignedResource_ objects are defined in this file.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -938,6 +940,50 @@ public void getAssignedMembers() throws Exception {
...
+		perun.getGroupsManager().addMember(sess, group1, member1);
+		perun.getGroupsManager().addMember(sess, group2, member1);
+		perun.getGroupsManager().addMember(sess, group3, member1);
+		perun.getGroupsManager().addMember(sess, group2, member2);
+		perun.getGroupsManager().addMember(sess, group3, notAssignedMember);
+
+		Facility facility = new Facility(0, "testFacility", "testFacility");
+		facility = perun.getFacilitiesManagerBl().createFacility(sess, facility);
+		Resource resource = new Resource(0, "testResource", "testResource", facility.getId(), vo.getId());
+		resource = perun.getResourcesManagerBl().createResource(sess, resource, vo, facility);
```
cuadradek, line 966: You can probably use _setUpFacility()_ and _setUpResource()_ methods here.

cuadradek (changes requested):

perun-openapi/openapi.yml
Johaney-s: fixed

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
Johaney-s: good point

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -538,6 +538,22 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * Returns members of groups assigned to resource with status of group-resource assignment.
+	 * @param sess perunSession
+	 * @param resource resource
+	 * @return list of members of groups assigned to given resource
+	 *
+	 * @throw PrivilegeException insufficient permissions
+	 */
+	getAssignedMembersWithStatus {
+		@Override
+		public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
```
metodej, line 551: I think the return type should be `List<AssignedMember>`.

metodej (changes requested):
The code looks great, just two comments.

PR #3329, opened 2021-08-24
fix(core): correctly close connection to extSources

Vojtech-Sassmann:
* In some scenarios we didn't close connectinos to extSources. This might cause a problem on the
extsource system.
 * Missing closements were detected and added.


balcirakpeter (approved):

stavamichal (changes requested):
Please add comment to the Entry part that this operation is not standard but it is necessary.

stavamichal (approved):

PR #3331, opened 2021-08-25
feat(core): New method getAssignedResourcesWithStatus for member

metodej:
* Added new method getAssignedResourcesWithStatus which returns all
resources assigned to any member's group with GroupResourceStatus
value.


perun-openapi/openapi.yml
```diff
@@ -12032,6 +12032,20 @@ paths:
...
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/resourcesManager/getAssignedResourcesWithStatus:
+    get:
+      tags:
+        - ResourcesManager
+      operationId: getAssignedResourcesWithStatus
+      summary: List all resources with statuses associated with a member's group.
+      parameters:
+        - $ref: '#/components/parameters/groupId'
```
Johaney-s, line 12042: `memberId `should be used as the parameter. For the summary I think the description in javadocs is more clear than the one here, I would prefer that instead. The same in rpc javadoc.

Johaney-s (changes requested):
Really nifty solution, I like it. Only small comment on the openApi.

perun-base/src/main/java/cz/metacentrum/perun/core/api/GroupResourceStatus.java
```diff
@@ -6,8 +6,19 @@
...
 public enum GroupResourceStatus {
-	ACTIVE,
-	INACTIVE,
-	FAILED,
-	PROCESSING
+
+	ACTIVE(3),
+	INACTIVE(2),
+	FAILED(1),
+	PROCESSING(0);
```
cuadradek, line 13: Maybe I would give PROCESSING level 2, because it is "closer" to being ACTIVE. It might be more interesting for users to know that one of the group-resource assignments is being activated. What do you think, @metodej @Johaney-s?

Johaney-s: @cuadradek I think the PROCESSING state should be the least important one because you can't rely on it for it is only temporary, plus it could end up as FAILED anyway. I believe the final statuses are more important.

perun-openapi/openapi.yml
metodej: Thank you, updated.

perun-base/src/main/java/cz/metacentrum/perun/core/api/GroupResourceStatus.java
metodej: Now that I think about it, even though it could end up FAILED, it could end up ACTIVE, making the present INACTIVE/FAILED state temporal. It could be beneficial for users to know they should wait a moment and check later. It makes sense for me to increase its _importance_.

Johaney-s: Alright, then please change the values accordingly.

PR #3337, opened 2021-08-30
fix(core): convert Member in AssignedMember to RichMember

Johaney-s:
* we need RichMember to show AssignedMember objects in GUI

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/AssignedMember.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/AssignedMember.java
@@ -8,21 +8,21 @@ import java.util.Objects;
  * @author Johana Supikova <xsupikov@fi.muni.cz>
  */
 public class AssignedMember {
-	private Member member;
+	private RichMember richMember;
 	private GroupResourceStatus status;
 
-	public AssignedMember(Member member, GroupResourceStatus status) {
-		this.member = member;
+	public AssignedMember(RichMember richMember, GroupResourceStatus status) {
+		this.richMember = richMember;
 		this.status = status;
 	}
 
 
-	public Member getMember() {
-		return member;
+	public RichMember getRichMember() {
+		return richMember;
 	}
 
-	public void setMember(Member member) {
-		this.member = member;
+	public void setRichMember(RichMember richMember) {
+		this.richMember = richMember;
 	}
 
 	public GroupResourceStatus getStatus() {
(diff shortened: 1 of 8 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/AssignedMember.java
```diff
@@ -8,20 +8,20 @@
  * @author Johana Supikova <xsupikov@fi.muni.cz>
  */
 public class AssignedMember {
-	private Member member;
+	private RichMember member;
```
cuadradek, line 11: I would maybe rename this field to richMember and its getter and setter as well, so it is clearer that we are getting / setting RichMember object instead of Member. What do you think?

cuadradek (commented):
Looks good. I just realized that AssignedMember object is missing in `perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl`, so add it there as well, please.

HejdaJakub (changes requested):
I agree with @cuadradek, otherwise it looks good.

perun-base/src/main/java/cz/metacentrum/perun/core/api/AssignedMember.java
Johaney-s: Thanks, renamed. Example added to `parseRpcMethods`.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -197,13 +198,24 @@
...
+
+				GroupResourceStatus assignmentStatus = GroupResourceStatus.valueOf(resultSet.getString("group_resource_status"));
+
+				// if member is repeated, only update assignment status and richmember's group statuses
 				if (members.containsKey(member.getId())) {
-					members.get(member.getId()).getMember().putGroupStatuses(member.getGroupStatuses());
+					AssignedMember storedMember = members.get(member.getId());
+					storedMember.getRichMember().putGroupStatuses(member.getGroupStatuses());
+					if (assignmentStatus.isMoreImportantThan(storedMember.getStatus())) {
+						storedMember.setStatus(assignmentStatus);
```
Johaney-s, line 211: guys, I think I found the problem failing the tests randomly. I didn't update the assignment status of our member if he was already in the map structure and because memberId is used as key, he couldn't appear more times with different statuses which I relied on later, so the assignment status was always copied from his first entry in the database. I fixed it here and removed the iteration comparing the statuses that was used later.

cuadradek (approved):

HejdaJakub (approved):

PR #3342, opened 2021-09-02
fix(core): removed attribute authorization in method sendPasswordResetLinkEmail

balcirakpeter:
- We want to perform an authorization on attributes only when we return
  those attributes to users. We do not want to do that when some method
  internaly uses them. Therefore method sendPasswordResetLinkEmail now
  uses BL version of getAttribute methods.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -1169,10 +1169,10 @@ public class MembersManagerEntry implements MembersManager {
 		try {
 			if (ad.getEntity().equals("user")) {
 				User user = perunBl.getUsersManagerBl().getUserByMember(sess, member);
-				mailAttribute = getPerunBl().getAttributesManager().getAttribute(sess, user, mailAttributeUrn);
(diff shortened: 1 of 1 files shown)


metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1169,10 +1169,10 @@ public void sendPasswordResetLinkEmail(PerunSession sess, Member member, String
 		try {
 			if (ad.getEntity().equals("user")) {
 				User user = perunBl.getUsersManagerBl().getUserByMember(sess, member);
-				mailAttribute = getPerunBl().getAttributesManager().getAttribute(sess, user, mailAttributeUrn);
+				mailAttribute = getPerunBl().getAttributesManagerBl().getAttribute(sess, user, mailAttributeUrn);
 			}
 			if (ad.getEntity().equals("member")) {
-				mailAttribute = getPerunBl().getAttributesManager().getAttribute(sess, member, mailAttributeUrn);
+				mailAttribute = getPerunBl().getAttributesManagerBl().getAttribute(sess, member, mailAttributeUrn);
```
zlamalp, line 1175: Please fix the same bug also in `sendAccountActivationLinkEmail()`, thank you.

zlamalp (changes requested):

zlamalp (approved):

PR #3344, opened 2021-09-02
feat(core): Handle automatic group-resource assignments in existing m…

Johaney-s:
…ethods

* When creating subgroup or removing group from resource we need to handle automatic subgroups assignment
* Subgroup will be assigned to resources where parent group is assigned with autoAssignSubgroups flag set to true
* Subgroups will be removed from resource if source group of the assignment is removed
* Fixed auditing of group-resource removal

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
@@ -24,6 +24,7 @@ import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
 import cz.metacentrum.perun.core.api.exceptions.BanAlreadyExistsException;
 import cz.metacentrum.perun.core.api.exceptions.BanNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedFromResourceException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotDefinedOnResourceException;
@@ -359,6 +360,22 @@ public interface ResourcesManagerBl {
 	 * @throws GroupResourceMismatchException
 	 */
 	void assignGroupsToResource(PerunSession perunSession, Iterable<Group> groups, Resource resource, boolean async, boolean assignInactive, boolean autoAssignSubgroups) throws WrongAttributeValueException, WrongReferenceAttributeValueException, GroupResourceMismatchException;
+
+	/**
+	 * Asynchronously assigns single subgroup to resource as automatically assigned source group's subgroup.
+	 * Source group must have existing assignment on resource with autoAssignSubgroups set to true.
+	 * @param perunSession sess
+	 * @param sourceGroup source group (containing groupToAssign in hierarchy as a subgroup)
+	 * @param groupToAssign source group's subgroup to be assigned on resource as by automatic assignment
+	 * @param resource resource
+	 * @param assignInactive assign subgroup to inactive state
+	 *
+	 * @throws GroupResourceMismatchException
+	 * @throws GroupAlreadyAssignedException
+	 * @throws WrongReferenceAttributeValueException
+	 * @throws WrongAttributeValueException
+	 */
+	void assignAutomaticGroupToResource(PerunSession perunSession, Group sourceGroup, Group groupToAssign, Resource resource, boolean assignInactive) throws GroupResourceMismatchException, GroupAlreadyAssignedException, WrongReferenceAttributeValueException, WrongAttributeValueException;
 
 	/**
 	 * Assign group to the resources. Check if attributes for each member from group are valid.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -33,6 +33,7 @@ import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.ExtSourcesManager;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.GroupResourceStatus;
 import cz.metacentrum.perun.core.api.GroupsManager;
 import cz.metacentrum.perun.core.api.Host;
 import cz.metacentrum.perun.core.api.Member;
@@ -66,6 +67,7 @@ import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
 import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
+import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyAssignedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.GroupAlreadyRemovedFromResourceException;
 import cz.metacentrum.perun.core.api.exceptions.GroupExistsException;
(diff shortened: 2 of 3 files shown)


Kuliak (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -251,6 +253,24 @@ public Group createGroup(PerunSession sess, Group parentGroup, Group group) thro
...
 
+		// check if new subgroup should be automatically assigned to resources
+		List<AssignedResource> assignedResources = getPerunBl().getResourcesManagerBl().getResourceAssignments(sess, parentGroup, null)
+			.stream()
+			.filter(AssignedResource::isAutoAssignSubgroups)
+			.collect(toList());
+
+		for (AssignedResource assignedResource : assignedResources) {
+			try {
+				Group sourceGroup = assignedResource.getSourceGroupId() == null ? parentGroup : getPerunBl().getGroupsManagerBl().getGroupById(sess, assignedResource.getSourceGroupId());
```
cuadradek, line 264: I think `getSourceGroupId()` would always be null because you filtered only assignments with `autoAssignSubgroups` set to true - but only direct assignments can have this set to true. So we probably want to to filter assignments with `autoAssignSubgroups == true` or `getSourceGroupId() != null`. It could be more readable to filter this in the for loop, so we don't need the stream.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -458,24 +444,7 @@ public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> gr
...
-						} else {
-							getResourcesManagerImpl().assignGroupToResourceState(perunSession, subgroup, resource, status);
-						}
-
-						// if status is not ACTIVE and should be, activate group-resource assignment
-						if (!assignInactive) {
-							// automatic subgroups-resource assignment should be async
-							activateGroupResourceAssignment(perunSession, subgroup, resource, true);
-						}
+						setAssignedGroupStatusAndActivate(perunSession, resource, true, assignInactive, subgroup);
```
cuadradek, line 447: I think we can call here the new method `assignAutomaticGroupToResource`. It should be more error-resilient if we decide to change the automatic group behavior - we don't have to change it at 2 places.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
```diff
@@ -360,6 +361,17 @@
...
 	void assignGroupsToResource(PerunSession perunSession, Iterable<Group> groups, Resource resource, boolean async, boolean assignInactive, boolean autoAssignSubgroups) throws WrongAttributeValueException, WrongReferenceAttributeValueException, GroupResourceMismatchException;
 
+	/**
+	 * Asynchronously assigns single subgroup to resource as automatically assigned source group's subgroup.
+	 * Source group must have existing assignment on resource with autoAssignSubgroups set to true.
+	 * @param perunSession sess
+	 * @param sourceGroup source group (containing groupToAssign in hierarchy as a subgroup)
+	 * @param groupToAssign source group's subgroup to be assigned on resource as by automatic assignment
+	 * @param resource resource
+	 * @param assignInactive assign subgroup to inactive state
```
cuadradek, line 371: Please, add here also the exceptions that can be thrown.

cuadradek (changes requested):
Nice work, I liked the detailed documentation. We should also handle use-case when we move a group structure under another group (method moveGroup in GroupsManagerBl).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
Johaney-s: I believe automatic assignments have `autoAssignSubgroups` set to true too.

cuadradek: Yes, you are right.

cuadradek (commented):
Could you please add some tests as well? E.g. checking a subgroup is removed from resource when its parent is removed, subgroup is automatically assigned to resource when it is created in such hierarchy, etc.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ResourcesManagerBl.java
Johaney-s: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -486,6 +455,54 @@ public void assignGroupsToResource(PerunSession perunSession, Iterable<Group> gr
...
+			getResourcesManagerImpl().setGroupResourceStatus(perunSession, g, resource, status);
+		} else {
+			getResourcesManagerImpl().assignGroupToResourceState(perunSession, g, resource, status);
+		}
+
+		// if status is PROCESSING, activate group-resource assignment
+		if (!assignInactive) {
+			activateGroupResourceAssignment(perunSession, g, resource, async);
+		}
+	}
```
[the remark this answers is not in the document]
Johaney-s: done

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
Johaney-s: right, done

Johaney-s:
> We should also handle use-case when we move a group structure under another group
Thanks, good point, I was thinking only of copying them instead of moving them. Added the logic to `moveGroup`.


> Could you please add some tests as well? E.g. checking a subgroup is removed from resource when its parent is removed, subgroup is automatically assigned to resource when it is created in such hierarchy, etc.
As for the tests, we'd first need to ensure subgroups are assigned synchronously, then we can test if moving/removing them works...

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -789,6 +809,26 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
...
 
+		//remove movingGroup-resource autoassignments where destination group is not subgroup of assignment's source group
+		List<AssignedResource> autoAssignedResources = perunBl.getResourcesManagerBl().getResourceAssignments(sess, movingGroup, List.of()).stream()
+			.filter(g -> g.getSourceGroupId() != null)
+			.collect(toList());
+
+		for (AssignedResource autoAssignedResource : autoAssignedResources) {
+			int sourceGroupId = autoAssignedResource.getSourceGroupId();
+			try {
+				Group sourceGroup = this.getGroupById(sess, autoAssignedResource.getSourceGroupId());
```
cuadradek, line 820: You can use here `sourceGroupId` variable.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -789,6 +809,26 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
...
+				if (destinationGroup == null || !this.getAllSubGroups(sess, sourceGroup).contains(destinationGroup)) {
+					// remove automatic group and subgroups' autoassignments
+					perunBl.getResourcesManagerBl().removeAutomaticGroupFromResource(sess, movingGroup, autoAssignedResource.getEnrichedResource().getResource(), sourceGroupId);
```
cuadradek, line 823: I think we should also remove assignments for subgroups of the movingGroup.

cuadradek (changes requested):
Also, if you want to, you could extract the new code from `moveGroup` into separate private methods since the method is way too long.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -803,6 +826,81 @@ public void moveGroup(PerunSession sess, Group destinationGroup, Group movingGro
...
+	private void fixMovedTreeAutoassignments(PerunSession sess, Group destinationGroup, Group movingGroup) {
+		List<AssignedResource> autoAssignedResources = perunBl.getResourcesManagerBl().getResourceAssignments(sess, movingGroup, List.of()).stream()
+			.filter(g -> g.getSourceGroupId() != null)
+			.collect(toList());
+
+		for (AssignedResource autoAssignedResource : autoAssignedResources) {
+			int sourceGroupId = autoAssignedResource.getSourceGroupId();
+			try {
+				Group sourceGroup = this.getGroupById(sess, sourceGroupId);
+				List<Group> groupsToRemove = this.getAllSubGroups(sess, sourceGroup);
```
cuadradek, line 889:
```suggestion
				List<Group> groupsToRemove = this.getAllSubGroups(sess, movingGroup);
```
I think we should remove autoassignments of moving group's subgroups. The following if condition should be changed as well.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
Johaney-s: Right, it was wrong tree to remove. Now the check is the same - if destination group is subgroup of source group, then it creates new tree starting from the moving group.

cuadradek (approved):

metodej (approved):
Looks good to me!

PR #3349, opened 2021-09-10
fix(core): use u:d:titleBeforeDc2 instead of u:d:titleBeforeVema at VŠUP

zlamalp:
We no longer use VEMA as personel system, hence attribute modules must be switched to use new
attribute.


Vojtech-Sassmann (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_titleBeforeKos.java
```diff
@@ -36,9 +36,9 @@ public void changedAttributeHook(PerunSessionImpl session, User user, Attribute
 		if (attribute.getValue() != null) {
 			Attribute titleBeforeVema;
```
Johaney-s, line 37: maybe the name of this variable should be changed to `titleBeforeDc2`?

zlamalp: You are right, thank you. Fixed.

PR #3351, opened 2021-09-13
feat(core): added attribute modules for o365 Exchange mails at VŠUP

zlamalp:
- Added attribute modules for u:d:vsupExchangeMail and u:d:vsupExchangeMailAliases.
- Old modules are now checking consistency also across new attributes.
- Fixed unchecked cast in old modules.
- Added comments and commented out parts which should be active only after migration.
- Added deprecation notices and comments.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_vsupExchangeMail.java
```diff
@@ -0,0 +1,293 @@
...
+				// Remove mail, which is not in attributes anymore
+				reservedMailsAttributeValue.remove(mail);
+				// since this attribute holds single value, we can break the cycle here
+				break;
+			}
+		}
+
+		// Put in which is in attribute but not in a map
+		if (attribute.getValue() != null) {
+			reservedMailsAttributeValue.putIfAbsent((String)attribute.getValue(), String.valueOf(user.getId()));
```
metodej, line 252:
```suggestion
			reservedMailsAttributeValue.putIfAbsent(attribute.valueAsString(), String.valueOf(user.getId()));
```

metodej (approved):
I think everything is ok, just a few missed `.valueAsString` opportunities.

zlamalp:
> I think everything is ok, just a few missed `.valueAsString` opportunities.

Thank you, it`s fixed now.

PR #3352, opened 2021-09-14
feat: Added pagination of users

metodej:
* Added to new method getUsersPage which returns paginated users
based on UsersPageQuery.
* UsersPageQuery consist of page size, offset, order, sort column and
possibly search string and withoutVo option. Search string filters
users to return only those with given string as name or other
aspect we search by. If withoutVo is set true, the method returns
only users that are not in any vo.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
@@ -25,6 +25,7 @@ import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Host;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.Paginated;
 import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.PerunPrincipal;
 import cz.metacentrum.perun.core.api.PerunSession;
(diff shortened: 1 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1406,4 +1456,37 @@ public void deleteSponsorLinks(PerunSession sess, User sponsor) {
...
+				" ,count(*) OVER() AS total_count" +
+				" FROM users";
+
+		String selectWithMembers =
+			"SELECT " + userMappingSelectQuery +
+				" ,count(*) OVER() AS total_count" +
+				" FROM users LEFT JOIN members on members.user_id = users.id";
+
+
+		return !query.isWithoutVo() && Objects.equals(query.getSearchString(), "") ? select : selectWithMembers;
```
cuadradek, line 1471: We can use `isEmpty(query.getSearchString()))` here as well.

Johaney-s (approved):
Nice work, everything look good to me.

cuadradek (commented):
Looking good, I have only couple of nit-picks.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -902,6 +904,26 @@ public UserExtSource getUserExtSourceByExtLogin(PerunSession sess, ExtSource sou
...
+		List<AttributeDefinition> attrDefs = new ArrayList<>();
+		for (String attrsName : attrNames) {
+			try {
+				attrDefs.add(getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, attrsName));
+			} catch (AttributeNotExistsException e) {
+				//pass
+			}
+		}
+
+		richUsers = convertUsersToRichUsersWithAttributes(sess, richUsers, attrDefs);
```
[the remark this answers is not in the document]
metodej: Nice, that looks better, thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
metodej: Fixed.

cuadradek (approved):

PR #3355, opened 2021-09-16
feat(core): add periodic automatic resource assignments checker

Johaney-s:
* this component checks every hour if automatic assignments are consistent
* removes redundant subgroups from resource when source group is no longer assigned
* assigns missing subgroups for manually assigned groups with autoAssignSubgroups set to true


perun-core/src/test/java/cz/metacentrum/perun/core/impl/ResourceAssignmentCheckerTest.java
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	@Test
+	public void leftoutSubgroupGetsAssigned() throws Exception {
+		System.out.println(CLASS_NAME + "leftoutSubgroupGetsAssigned");
+
+		setUpVo();
+		setUpGroups();
+		setUpFacility();
+		setUpResource();
```
metodej, line 61: I think we could move these to setup, and not repeat them in every test.

metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentChecker.java
```diff
@@ -0,0 +1,126 @@
...
+			List<AssignedGroup> automaticallyAssignedSubgroups = assignedGroups.stream()
+				.filter(group -> group.getSourceGroupId() != null)
+				.collect(Collectors.toList());
+
+			List<AssignedGroup> sourceGroups = assignedGroups.stream()
+				.filter(a -> a.isAutoAssignSubgroups() && a.getSourceGroupId() == null)
+				.collect(Collectors.toList());
+
+			for (AssignedGroup assignedSubgroup : automaticallyAssignedSubgroups) {
+				boolean sourceIsAssigned;
```
cuadradek, line 83: We should probably extract content of the loop into a new method that runs in a separate transaction.

cuadradek (changes requested):

perun-core/src/test/java/cz/metacentrum/perun/core/impl/ResourceAssignmentCheckerTest.java
```diff
@@ -0,0 +1,177 @@
...
+	private Facility facility;
+	private Resource resource;
+
+
+	@Before
+	public void setUp() throws Exception {
+
+		resourcesManagerImpl = (ResourcesManagerImplApi) ReflectionTestUtils.getField(perun.getResourcesManagerBl(), "resourcesManagerImpl");
+		if (resourcesManagerImpl == null) {
+			throw new RuntimeException("Failed to get vosManagerImpl");
```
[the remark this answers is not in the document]
Johaney-s: Thanks. fixed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentChecker.java
Johaney-s: Extracted both assigning and removing loop.

Kuliak (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ResourceAssignmentChecker.java
```diff
@@ -0,0 +1,148 @@
...
+import cz.metacentrum.perun.core.api.exceptions.GroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.GroupResourceMismatchException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.bl.PerunBl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.transaction.annotation.Propagation;
+import org.springframework.transaction.annotation.Transactional;
```
cuadradek, line 23: Unused imports.

cuadradek (changes requested):

PR #3357, opened 2021-09-17
fix(core): Fix of users pagination

metodej:
* The pagination returned user for every vo he was in, that's not an
issue now.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
@@ -1462,7 +1462,7 @@ public class UsersManagerImpl implements UsersManagerImplApi {
 				" FROM users";
 
 		String selectWithMembers =
-			"SELECT " + userMappingSelectQuery +
+			"SELECT distinct " + userMappingSelectQuery +
 				" ,count(*) OVER() AS total_count" +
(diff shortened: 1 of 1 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1462,7 +1462,7 @@ private String getSQLSelectForUsersPage(UsersPageQuery query) {
 				" FROM users";
 
 		String selectWithMembers =
-			"SELECT " + userMappingSelectQuery +
+			"SELECT distinct " + userMappingSelectQuery +
```
zlamalp, line 1465: Wasn't removing `distinct` our great speed optimization when searching users/members? It was probably in a different place, but maybe we should think about the way we select data.

metodej: I believe it was in member-group search, I will think more about it and revise it if I find a better way...

metodej: I didn't find a better way. I think we need each of the user's member, so we can find it by searchString, and we want to group them up at the end of the query...

cuadradek (approved):

PR #3358, opened 2021-09-17
fix(core): fix database dump problem caused by unqualified table name…

martin-kuba:
… in SQL function

The SQL function relation_group_resource_exist contained table name without schema which caused
problem with dumps when loaded.

BREAKING CHANGE: New database version

### DIFF
--- a/perun-db/postgres.sql
+++ b/perun-db/postgres.sql
@@ -1,4 +1,4 @@
--- database version 3.1.87 (don't forget to update insert statement at the end of file)
+-- database version 3.1.88 (don't forget to update insert statement at the end of file)
 
 -- VOS - virtual organizations
 create table vos (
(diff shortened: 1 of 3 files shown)


perun-db/postgres.sql
```diff
@@ -782,7 +782,7 @@ create table groups_resources_automatic (
 );
 
 create function relation_group_resource_exist(integer, integer) returns integer as
-    'select count(1)::integer from (SELECT group_id, resource_id FROM groups_resources UNION SELECT group_id, resource_id FROM groups_resources_automatic) gr_res where group_id=$1 and resource_id=$2;' language sql;
+    'select count(1)::integer from (SELECT group_id, resource_id FROM perun.groups_resources UNION SELECT group_id, resource_id FROM perun.groups_resources_automatic) gr_res where group_id=$1 and resource_id=$2;' language sql;
```
metodej, line 785: Isn't the same problem with function `delete_group_resource_status`, which references `groups_resources_state`?

martin-kuba: While `delete_group_resource_status` also uses unqualified table name, it is not a problem. The difference is that the function `relation_group_resource_exist` is used as a constraint on a table, thus it is invoked during initial data import into that table, which is performed with permissions of the user `postgres` with empty `search_path`, so the unqualified name is not found. In comparison, the `delete_group_resource_status` function is used as a trigger, thus it is invoked always with permissions of  the user `perun` with `search_path` that includes the `perun` schema, so the unqualified name is always correctly resolved.

metodej: Ok, thanks for explanation!

metodej (approved):

zlamalp (approved):

PR #3369, opened 2021-10-04
updated to Java 17

martin-kuba:


pom.xml
```diff
@@ -60,9 +60,6 @@
 		<!-- by default we run perun in memory -->
 		<spring.profiles.default>default</spring.profiles.default>
 
-		<!-- redefine versions of some libraries defined by Spring Boot Starter Parent -->
-		<tomcat.version>8.5.57</tomcat.version>
```
zlamalp, line 64: Can you please move this change outside of this pull-request? I will prepare change for this including update of cargo plugin. Its not really related to java update.

zlamalp (changes requested):
I think this pull-request is mixing two changes (build vs. CI) and I think it should be separate commits (base on conventional commits guidelines).

pom.xml
martin-kuba: removed

zlamalp (approved):

zlamalp:
We will merge this later in the sprint (after release from previous sprint is done).

PR #3394, opened 2021-10-26
feat(core): perform case-ignore checks in login-namespace:einfra

xkostka2:
* added support for case-ignore checking of logins

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_einfraTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_einfraTest.java
@@ -1,0 +1,158 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.*;
+import cz.metacentrum.perun.core.api.exceptions.*;
+import cz.metacentrum.perun.core.bl.ModulesUtilsBl;
+import cz.metacentrum.perun.core.bl.PerunBl;
+import cz.metacentrum.perun.core.bl.UsersManagerBl;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.impl.modules.pwdmgr.EinfraPasswordManagerModule;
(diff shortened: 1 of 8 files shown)


cuadradek (changes requested):
Looks good, can you please add one more test - checkAttributeSemantics when the login is already reserved in specific namespace? Mock `sess.getPerunBl().getUsersManagerBl().checkReservedLogins` to throw `AlreadyReservedLoginException` and the test should then expect `WrongReferenceAttributeValueException`.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_einfraTest.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	@Test(expected = WrongReferenceAttributeValueException.class)
+	public void testAlreadyReservedLogin() throws Exception {
+		System.out.println("testAlreadyReservedLogin()");
+		UsersManagerBl usersManager = mock(UsersManagerBl.class);
+		when(session.getPerunBl().getUsersManagerBl()).thenReturn(usersManager);
+		List<User> tmp = new ArrayList<>();
+		tmp.add(user);
+		tmp.add(user2);
```
cuadradek, line 152:
```suggestion
```
Otherwise, it would be the same test case as the previous one.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_einfraTest.java
```diff
@@ -0,0 +1,159 @@
...
+		when(usersManager.getUsersByAttribute(session, attributeToCheck, true)).thenReturn(tmp);
+		doThrow(new AlreadyReservedLoginException(""))
+			.when(usersManager).checkReservedLogins(session, "einfra", attributeToCheck.valueAsString(), true);
+		attributeToCheck.setValue(attributeToCheck.getValue().toString().toUpperCase());
```
cuadradek, line 156: In order to _doThrow()_ correctly work, you will have to move this line before _doThrow()_ or remove it since it is irrelevant.

cuadradek (changes requested):

cuadradek (approved):

PR #3404, opened 2021-11-08
feat(core): ssh public key format validation

xkostka2:
* added validation for ssh public keys

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
@@ -1,5 +1,6 @@
 package cz.metacentrum.perun.core.impl.modules.attributes;
 
+import com.google.api.client.util.Base64;
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -17,18 +25,124 @@
  */
 public class urn_perun_user_attribute_def_def_sshPublicKey extends UserAttributesModuleAbstract implements UserAttributesModuleImplApi {
 
+	private byte[] bytes = null;
+	private int pos = 0;
```
cuadradek, line 29: I think these fields should be moved to methods as local variables because an instance of this class is actually shared between threads/requests and it wouldn't be thread-safe with these fields.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -17,18 +25,124 @@
...
+				try {
+					validateSSH(sshKey);
+				} catch (Exception e) {
+					throw new WrongAttributeValueException("Invalid SSH key format");
+				}
+			}
+		}
+	}
+
+	public void validateSSH(String keyLine) throws Exception {
```
cuadradek, line 51: This method can be private.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -17,18 +25,124 @@
...
+	 * Provides a means to get from a parsed Q value to the X and Y point values.
+	 * that can be used to create and ECPoint compatible with ECPublicKeySpec.
+	 *
+	 * @param q          According to RFC 5656:
+	 *                   "Q is the public key encoded from an elliptic curve point into an octet string"
+	 * @param identifier According to RFC 5656:
+	 *                   "The string [identifier] is the identifier of the elliptic curve domain parameters."
+	 * @return An ECPoint suitable for creating a JCE ECPublicKeySpec.
+	 */
+	ECPoint getECPoint(BigInteger q, String identifier) {
```
cuadradek, line 100: This method can be private.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -17,18 +25,124 @@
...
+	}
+
+	/**
+	 * Gets the curve parameters for the given key type identifier.
+	 *
+	 * @param identifier According to RFC 5656:
+	 *                   "The string [identifier] is the identifier of the elliptic curve domain parameters."
+	 * @return An ECParameterSpec suitable for creating a JCE ECPublicKeySpec.
+	 */
+	ECParameterSpec getECParameterSpec(String identifier) {
```
cuadradek, line 116: Same here.

cuadradek (changes requested):

cuadradek (approved):

balcirakpeter (changes requested):
It looks good but there are two things missing.
1. we will need a better error explanation so the user knows what is wrong with the format.
2. You do not check whether the algorithm in the base64 string matches the given algorithm in the ssh key which should be put before the base64.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +28,148 @@
...
+		// look for the Base64 encoded part of the line to decode
+		boolean typeFound = false;
+		String typeCheck = "";
+		for (String part : keyLine.split(" ")) {
+			if (part.matches("^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$")) {
+				bytes = Base64.decodeBase64(part);
+				if (typeFound) {
+					break;
+				} else {
+					throw new IllegalArgumentException("no type found before key part");
```
balcirakpeter, line 61: Please review texts for all exceptions so they are grammatically correct.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +28,148 @@
...
+				try {
+					validateSSH(sshKey);
+				} catch (Exception e) {
+					throw new WrongAttributeValueException("Invalid SSH key format: " + e.getMessage());
+				}
+			}
+		}
+	}
+
+	private void validateSSH(String keyLine) throws Exception {
```
balcirakpeter, line 48: Please add a JavaDoc for this method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +28,148 @@
...
+		byte[] bytes = null;
+		int[] pos = {0};
+
+		// look for the Base64 encoded part of the line to decode
+		boolean typeFound = false;
+		String typeCheck = "";
+		for (String part : keyLine.split(" ")) {
```
balcirakpeter, line 55: If you will check the ssh key like this you can insert random text between the ssh algorithm type and the ssh base64 key. The base64 key has to be put directly after the algorithm type.

balcirakpeter (changes requested):
Please update tests as well.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +36,180 @@
...
+	 * @param sshKey SSH key to be checked
+	 */
+	private void validateSSH(String sshKey) throws Exception {
+		byte[] bytes = null;
+		int[] pos = {0};
+
+		// look for the Base64 encoded part of the line to decode
+		boolean typeFound = false;
+		String sshKeyTypeValue = "";
+		final String BASE64_FORMAT = "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$";
```
balcirakpeter, line 68: See how to use the pattern in `urn_perun_group_resource_attribute_def_def_vomsRoles`.
Also, all the final variables should be created outside of the method.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +36,180 @@
...
+		for (String part : sshKey.split(" ")) {
+			if (part.matches(String.valueOf(pattern))) {
+				if (typeFound) {
+					bytes = Base64.decodeBase64(part);
+					break;
+				} else {
+					throw new IllegalArgumentException("type hasn't been found before the key part");
+				}
 			}
+			if (part.equals(RSA)
```
balcirakpeter, line 89: please use stream.anyMatch() if possible.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +36,180 @@
...
+		try {
+			if (sshKeyType.equals(RSA)) {
+				decodeRSA(bytes, pos);
+			} else if (sshKeyType.equals(DSA)) {
+				decodeDSA(bytes, pos);
+			} else if (sshKeyType.equals(ECDSA_SHA2_NISTP256)
+				|| sshKeyType.equals(ECDSA_SHA2_NISTP384)
+				|| sshKeyType.equals(ECDSA_SHA2_NISTP521)) {
+				decodeEcdsa(bytes, pos);
+			} else if (!(sshKeyType.startsWith("sk-ecdsa") || sshKeyType.startsWith("ssh-ed25519") || sshKeyType.startsWith("sk-ed25519"))) {
```
balcirakpeter, line 121: Please use the defined constants also in this condition.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +36,180 @@
...
+					validateSSH(sshKey);
+				} catch (Exception e) {
+					throw new WrongAttributeValueException("Invalid SSH key format: " + e.getMessage());
+				}
+			}
+		}
+	}
+
+	/**
+	 * Checks whether is the SSH key in the correct format. Throws exception if validation fails for any reason.
```
balcirakpeter, line 57: Please use an appropriate tag to describe the exception which can be thrown.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -20,17 +36,180 @@
...
+			String name = identifier.replace("nist", "sec") + "r1";
+			AlgorithmParameters parameters = AlgorithmParameters.getInstance("EC");
+			parameters.init(new ECGenParameterSpec(name));
+			return parameters.getParameterSpec(ECParameterSpec.class);
+		} catch (InvalidParameterSpecException | NoSuchAlgorithmException e) {
+			throw new IllegalArgumentException("Unable to parse curve parameters: ", e);
 		}
 	}
 
+	private String decodeType(byte[] bytes, int[] pos) {
```
balcirakpeter, line 193: All private methods should have their own JavaDoc.

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -8,29 +9,248 @@
...
+				}
 			}
 		}
 	}
 
+	/**
+	 * Checks whether is the SSH key in the correct format. Throws exception if validation fails for any reason.
+	 *
+	 * @param sshKey SSH key to be checked
+	 * @exception Exception that is thrown whenever SSH key is not in correct format
```
balcirakpeter, line 72: Correct tag is `@throws`

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -8,29 +9,248 @@
...
+					validateSSH(sshKey);
+				} catch (Exception e) {
+					throw new WrongAttributeValueException("Invalid SSH key format: " + e.getMessage());
+				}
 			}
 		}
 	}
 
+	/**
+	 * Checks whether is the SSH key in the correct format. Throws exception if validation fails for any reason.
```
balcirakpeter, line 69: You can remove the second sentence about the exception.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -8,29 +9,248 @@
...
+
+		// look for the Base64 encoded part of the line to decode
+		boolean typeFound = false;
+		String sshKeyTypeValue = "";
+
+		List<String> sshTypes = Arrays.asList(RSA, DSA, ECDSA_SHA2_NISTP256, ECDSA_SHA2_NISTP384, ECDSA_SHA2_NISTP521, ED25519);
+
+		for (String part : sshKey.split(" ")) {
+			Matcher matcher = pattern.matcher(part);
+			if (matcher.matches()) {
```
balcirakpeter, line 86: This way you do not check whether the ssh type is put directly before the bsae64 part.

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
balcirakpeter:
```
if (mathcer.matches()) {
...
} else if (typeFound) {
error
}
```

balcirakpeter (changes requested):

Kuliak (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKey.java
```diff
@@ -8,29 +9,250 @@
...
+				if (typeFound) {
+					bytes = Base64.decodeBase64(part);
+					break;
+				} else {
+					throw new IllegalArgumentException("type hasn't been found before the key part");
+				}
+			} else if(typeFound){
```
balcirakpeter, line 93:
```suggestion
			} else if (typeFound) {
```

balcirakpeter (changes requested):

balcirakpeter (approved):

PR #3412, opened 2021-11-18
feat(core): Paginating supports filtering groups by member

Kuliak:
getGroupsPage method now checks presence of memberId parameter and if it is present, groups for given member are returned together with requested member-group attributes.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichGroup.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichGroup.java
@@ -26,12 +26,12 @@ public class RichGroup extends Group {
 		this.groupAttributes = attrs;
 	}
 
-	public List<Attribute> getAttributes() {
+	public List<Attribute> getGroupAttributes() {
 		return this.groupAttributes;
 	}
 
-	public void setAttributes(List<Attribute> attributes) {
-		this.groupAttributes = attributes;
+	public void setGroupAttributes(List<Attribute> groupAttributes) {
+		this.groupAttributes = groupAttributes;
 	}
 
 	@Override
@@ -54,7 +54,7 @@ public class RichGroup extends Group {
 		ret.append("', voId='");
 		ret.append(this.getVoId());
 		ret.append("', groupAttributes='");
-		ret.append(this.getAttributes());
+		ret.append(this.getGroupAttributes());
 		ret.append("']");
 		return ret.toString();
 	}
@@ -65,7 +65,9 @@ public class RichGroup extends Group {
 
 		String sGroupAttrs;
 		List<String> nAttrs = new ArrayList<>();
-		List<Attribute> oAttrs = this.getAttributes();
+
+		// serialize group attributes
+		List<Attribute> oAttrs = this.getGroupAttributes();
 		if (oAttrs == null) {
 			sGroupAttrs = "\\0";
 		} else {
@@ -110,10 +112,10 @@ public class RichGroup extends Group {
 			return false;
 		}
 		if (groupAttributes == null) {
-			if (other.getAttributes() != null) {
+			if (other.getGroupAttributes() != null) {
 				return false;
 			}
-		} else if (!this.getAttributes().equals(other.getAttributes())) {
+		} else if (!this.getGroupAttributes().equals(other.getGroupAttributes())) {
 			return false;
 		}
 		return true;
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -1674,10 +1674,15 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 	}
 
 	@Override
-	public Paginated<RichGroup> getGroupsPage(PerunSession sess, Vo vo, GroupsPageQuery query, List<String> attrNames) {
+	public Paginated<RichGroup> getGroupsPage(PerunSession sess, Vo vo, GroupsPageQuery query, List<String> attrNames) throws GroupNotExistsException, MemberNotExistsException, MemberGroupMismatchException {
 		Paginated<Group> paginatedGroups = groupsManagerImpl.getGroupsPage(sess, vo, query);
 
-		List<RichGroup> richGroups = convertGroupsToRichGroupsWithAttributes(sess, paginatedGroups.getData(), attrNames);
+		List<RichGroup> richGroups;
+		if (query.getMemberId() == null) {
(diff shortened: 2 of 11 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupsManagerEntryIntegrationTest.java
```diff
@@ -5902,6 +5901,108 @@ public void getGroupsPage_ParentGroup_ReturnsAttributes () throws Exception {
...
+		perun.getGroupsManager().createGroup(sess, vo, group3);
+
+		Member member = setUpMember(vo);
+
+		groupsManagerBl.addMember(sess, group, member);
+		groupsManagerBl.addMember(sess, group2, member);
+
+		Map<UUID, List<Attribute>> attributes = new HashMap<>();
+
+		AttributeDefinition attr = new AttributeDefinition();
```
Johaney-s, line 5973: It would be nice to have the repetitive code extracted to private method. There's also `setUpMemberGroupAttributes` you can use.

Johaney-s (changes requested):
Nice work, I have a few mainly cosmetic suggestions left.

perun-base/src/main/java/cz/metacentrum/perun/core/api/RichGroup.java
```diff
@@ -53,7 +53,7 @@ public String toString() {
 		ret.append(this.getDescription());
 		ret.append("', voId='");
 		ret.append(this.getVoId());
-		ret.append("', groupAttributes='");
+		ret.append("', attributes='");
```
Johaney-s, line 56: AuditParser.java uses method `createRichGroup()` where it expects 'groupAttributes', that's where the tests fail.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -2825,6 +2845,28 @@ public RichGroup convertGroupToRichGroupWithAttributesByName(PerunSession sess,
 		return richGroups;
 	}
 
+	@Override
+	public List<RichGroup> convertGroupsToRichGroupsWithAttributes(PerunSession sess, Member member, List<Group> groups, List<String> attrNames) throws MemberGroupMismatchException, MemberNotExistsException, GroupNotExistsException {
+		if (attrNames == null) return convertGroupsToRichGroupsWithAttributes(sess, member, groups);
+		List<RichGroup> richGroups = new ArrayList<>();
+
+		// separate group and member-group attribute names
```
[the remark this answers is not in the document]
Johaney-s: my bad, this was correct

cuadradek (commented):
I think we should exclude _members_ group, at least that's how _getMemberGroups_ works.

cuadradek (commented):
Method _getGroupsPage_ in Entry layer should have different authorization if memberId is present in GroupsPageQuery. In this case, the authorization should look like in method _getMemberGroups_.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -856,17 +858,30 @@ public void removeAdmin(PerunSession sess, Group group, Group authorizedGroup) t
...
+		if (query.getMemberId() != null) {
+			if (!AuthzResolver.authorizedInternal(sess, "member-getGroupsPage_Vo_GroupsPageQuery_List<String>_policy",  perunBl.getMembersManagerBl().getMemberById(sess, query.getMemberId()))) {
+				throw new PrivilegeException(sess, "getGroupsPage");
+			}
+		} else if (!AuthzResolver.authorizedInternal(sess, "getGroupsPage_Vo_GroupsPageQuery_List<String>_policy", vo)) {
 			throw new PrivilegeException(sess, "getGroupsPage");
 		}
 
 		Paginated<RichGroup> result = groupsManagerBl.getGroupsPage(sess, vo, query, attrNames);
 		result.setData(getGroupsManagerBl().filterOnlyAllowedAttributes(sess, result.getData(), null, true));
```
cuadradek, line 879: I suggest calling this overloaded method:
```suggestion
		result.setData(getGroupsManagerBl().filterOnlyAllowedAttributes(sess, result.getData(), member, null, true));
```

Then we won't need the next if statement.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -575,20 +600,28 @@ public int getSubGroupsCount(PerunSession sess, Group parentGroup) {
...
+				searchQuery +
+				" ORDER BY " + query.getSortColumn().getSqlOrderBy(query) +
+				" OFFSET (:offset)" +
+				" LIMIT (:limit);",
 			namedParams,
 			getPaginatedGroupsExtractor(query));
+
+		// remove "members" group
+		if(groups != null && !groups.getData().isEmpty() && query.getMemberId() != null) {
+			groups.getData().removeIf(g -> g.getName().equals(VosManager.MEMBERS_GROUP));
```
Johaney-s, line 621: Filtering members group here is too late, because the page size is given by the LIMIT parameter in the sql query and removing one of the groups after performing the query would return result not fitting the page size. You should put this  condition into the query.

Johaney-s (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -856,16 +858,31 @@ public void removeAdmin(PerunSession sess, Group group, Group authorizedGroup) t
...
+		} else if (!AuthzResolver.authorizedInternal(sess, "getGroupsPage_Vo_GroupsPageQuery_List<String>_policy", vo)) {
 			throw new PrivilegeException(sess, "getGroupsPage");
 		}
 
 		Paginated<RichGroup> result = groupsManagerBl.getGroupsPage(sess, vo, query, attrNames);
-		result.setData(getGroupsManagerBl().filterOnlyAllowedAttributes(sess, result.getData(), null, true));
+
+		if (member != null) {
+			result.setData(getGroupsManagerBl().filterOnlyAllowedAttributes(sess, result.getData(), member, null, true));
+		} else {
```
cuadradek, line 883: This can be even more simplified, there is no need for IF statement - the IF block should be sufficient for both cases, because member parameter can be null in method `filterOnlyAllowedAttributes`.

cuadradek (changes requested):

cuadradek (approved):

PR #3420, opened 2021-11-25
fix(core): allow duplicates in user identities for the same user

zlamalp:
- We need to support two identities with same login (eppn) from IdP
  (identity for proxy). As we don't want to allow duplicates globally,
  only 2 duplicates and for the same user are allowed.
- Updated module for einfraid persistent shadow login, we will generate identity
  for both proxies using the same attribute (login) value.
- Ensure uniqueness when returning value of user:virt:eduPersonPrincipalNames.
- Fixed variable naming in module user:virt:login-namespace:einfraif_persistent
  from elixir to einfra.
- Added tests for updated IdP identities logic.


zlamalp:
Resolved by #3429 closing.

PR #3430, opened 2021-12-07
fix(core): groups and subgroups assignment to resource fixed

Kuliak:
Subgroups are assigned to resource always with ACTIVE status, instead of copying its source group status.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
@@ -2678,8 +2678,8 @@ public class ResourcesManagerEntryIntegrationTest extends AbstractPerunIntegrati
 	}
 
 	@Test
-	public void autoAssignSubgroupInactive() throws Exception {
-		System.out.println(CLASS_NAME + "autoAssignSubgroupInactive");
+	public void autoAssignSubgroupWithInactiveSourceGroup() throws Exception {
+		System.out.println(CLASS_NAME + "autoAssignSubgroupWithInactiveSourceGroup");
 
 		vo = setUpVo();
 		member = setUpMember(vo);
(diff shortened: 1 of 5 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -2678,7 +2678,7 @@ public void createGroupAssignsSubgroupTree() throws Exception {
 	}
 
 	@Test
-	public void autoAssignSubgroupInactive() throws Exception {
+	public void autoAssignSubgroupWithInactiveSourceGroup() throws Exception {
 		System.out.println(CLASS_NAME + "autoAssignSubgroupInactive");
```
Johaney-s, line 2682: Please rename also the printed text.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -2696,7 +2696,33 @@ public void autoAssignSubgroupInactive() throws Exception {
...
+
+		// automatically assigned subgroups should always be ACTIVE
+		AssignedGroup assignedSubgroup = new AssignedGroup(new EnrichedGroup(subGroup, List.of()), GroupResourceStatus.ACTIVE, group.getId(), null, true);
+
+		assertThat(assignedGroups).containsExactlyInAnyOrder(assignedGroup, assignedSubgroup);
+	}
+
+	@Test
+	public void assignGroupsToResourceWithInactiveSourceGroupWithSubgroup() throws Exception {
+		System.out.println(CLASS_NAME + "autoAssignSubgroupsWithInactiveSourceGroup");
```
Johaney-s, line 2708: Rename the print to match test name here too

cuadradek: While we are at it, I would like if we named tests according cases they are testing. The code is more readable, if we know what it should do thanks to the test name. I suggest something like _assignInactiveGroupToResourceActivatesItsSubgroups_.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -2696,7 +2696,33 @@ public void autoAssignSubgroupInactive() throws Exception {
...
+		group = setUpGroup(vo, member);
+		subGroup = setUpSubGroup(group);
+
+		sess.getPerun().getResourcesManager().assignGroupToResource(sess, group, inactiveResource, false, true, true);
+
+		List<AssignedGroup> assignedGroups = sess.getPerun().getResourcesManager().getGroupAssignments(sess, inactiveResource, List.of());
+
+		AssignedGroup assignedGroup = new AssignedGroup(new EnrichedGroup(group, List.of()), GroupResourceStatus.INACTIVE, null, null, true);
+
+		// automatically assigned subgroups should always be ACTIVE
```
cuadradek, line 2724: Now, we won't need this comment if we know what the code should do thanks to the test name. In general, it is good to avoid inline comments if possible, as code should be as self-explanatory as possible and inline comments are then often useless.

cuadradek (changes requested):

Johaney-s (approved):

cuadradek (approved):

PR #3437, opened 2021-12-13
feat: add method for getting attribute policy collections 

cuadradek:
- Added new objects AttributePolicy and AttributePolicyCollection. They
will be used for evaluating users' rights on attributes. User has
rights to perform an action (READ/WRITE) on the attribute, if he
satisfies all policies in at least one of the attribute policy
collections with given action.
- Added new method getAttributePolicyCollections, it gets attribute
policy collections for given attribute.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
@@ -3,6 +3,7 @@ package cz.metacentrum.perun.core.bl;
 import cz.metacentrum.perun.core.api.ActionType;
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributePolicyCollection;
 import cz.metacentrum.perun.core.api.AttributeRights;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Group;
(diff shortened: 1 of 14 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -4620,6 +4621,16 @@ void mergeAttributesValues(PerunSession sess, Member member, List<Attribute> att
 	 */
 	void setAttributeRights(PerunSession sess, List<AttributeRights> rights);
 
+	/**
+	 * Gets attribute policy collections for an attribute definition with given id.
+	 *
+	 * @param sess perun session
+	 * @param attributeId id of the attribute definition
+	 * @return all policy collections of the attribute definition
+	 * @throws InternalErrorException
```
metodej, line 4630: I think there is no need for that in a javadoc, can we remove it?

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AttributesManagerImplApi.java
```diff
@@ -2706,4 +2707,14 @@
...
 	 */
 	String escapeQueryParameters(String value);
+
+	/**
+	 * Gets attribute policy collections for an attribute definition with given id.
+	 *
+	 * @param sess perun session
+	 * @param attributeId id of the attribute definition
+	 * @return all policy collections of the attribute definition
+	 * @throws InternalErrorException
```
metodej, line 2717: Same here.

metodej (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4923,6 +4968,20 @@ public void setAttributeRight(PerunSession sess, AttributeRights rights) {
 		}
 	}
 
+	@Override
+	public List<AttributePolicyCollection> getAttributePolicyCollections(PerunSession sess, final int attributeId) {
+		List<AttributePolicyCollection> attributePolicyCollections;
+		try {
+			attributePolicyCollections = jdbc.query("SELECT " + attributeRightSelectQuery + " FROM attributes_policies " +
```
Johaney-s, line 4975: I think `attributeRightSelectQuery` expects different tables than you provide

Johaney-s (changes requested):
Looks good, just a few comments. Should we also add the new objects to `ParseRpcMethods`?

PR #3442, opened 2022-01-05
fix(core): change regex to forbid using reserved characters

Kuliak:
* Regex used for validation forbids using space at the beginning or at the end of attribute value


perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_group_attribute_def_def_adGroupNameTest.java
```diff
@@ -77,8 +93,26 @@ public void testCorrectSyntax() {
...
+		System.out.println("testCorrectSyntaxWithUnderscore()");
+		attributeToCheck.setValue("_correct_Value_with_Dash_");
+
+		assertThatNoException().isThrownBy(
+			() -> classInstance.checkAttributeSyntax(sess, group, attributeToCheck));
+	}
+
+	@Test
+	public void testCorrectSyntaxWithSpace() {
 		System.out.println("testCorrectSyntax()");
```
Johaney-s, line 114: Looks good, just please match the printed test name to the actual test name.

Johaney-s (approved):

cuadradek (approved):

PR #3445, opened 2022-01-11
feat(core): add set method for attribute rights (policies)

Kuliak:
* new methods for setting attribute rights added to all corresponding layers

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/audit/events/AttributesManagerEvents/AttributePolicyCollectionSet.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/audit/events/AttributesManagerEvents/AttributePolicyCollectionSet.java
@@ -1,0 +1,34 @@
+package cz.metacentrum.perun.audit.events.AttributesManagerEvents;
+
+import cz.metacentrum.perun.audit.events.AuditEvent;
+import cz.metacentrum.perun.audit.events.EngineIgnoreEvent;
+import cz.metacentrum.perun.core.api.AttributePolicyCollection;
+
+public class AttributePolicyCollectionSet extends AuditEvent implements EngineIgnoreEvent {
+
+	private AttributePolicyCollection policyCollection;
+	private String message;
+
+	@SuppressWarnings("unused") // used by jackson mapper
+	public AttributePolicyCollectionSet() {
+	}
+
+	public AttributePolicyCollectionSet(AttributePolicyCollection apc) {
+		this.policyCollection = apc;
+		this.message = formatMessage("Attribute policy collection set: %s for %s", policyCollection, apc.getAttributeId());
+	}
+
+	public AttributePolicyCollection getPolicyCollection() {
+		return policyCollection;
+	}
+
+	@Override
+	public String getMessage() {
+		return message;
+	}
+
+	@Override
+	public String toString() {
+		return message;
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -18,6 +18,7 @@ import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AllUserFacility
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeAuthzDeleted;
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeCreated;
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeDeleted;
+import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributePolicyCollectionSet;
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeRemovedForFacility;
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeRemovedForFacilityAndUser;
 import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeRemovedForGroup;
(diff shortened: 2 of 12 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -9,6 +9,7 @@
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotMarkedUniqueException;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
+import cz.metacentrum.perun.rpc.Api;
```
cuadradek, line 12: Unused import.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -8256,6 +8257,16 @@ public void setAttributeRights(PerunSession sess, List<AttributeRights> rights)
 		}
 	}
 
+	@Override
+	public void setAttributePolicyCollections(PerunSession sess, List<AttributePolicyCollection> policies, int attributeId) {
+		for (AttributePolicyCollection apc : policies) {
+			getAttributesManagerImpl().setAttributePolicyCollections(sess, policies, attributeId);
```
cuadradek, line 8263: This line should be probably before the cycle.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
```diff
@@ -4512,6 +4512,19 @@ public void setAttributeRights(PerunSession sess, List<AttributeRights> rights)
 		getAttributesManagerBl().setAttributeRights(sess, rights);
 	}
 
+	@Override
+	public void setAttributePolicyCollections(PerunSession sess, List<AttributePolicyCollection> policies, int attributeId) throws PrivilegeException, AttributeNotExistsException {
+		Utils.checkPerunSession(sess);
+		getAttributeDefinitionById(sess, attributeId);
+
```
cuadradek, line 4519: I think we should check here whether roles of given attribute policy collections exist. Similarly like in `setAttributeRights`.

perun-base/src/main/java/cz/metacentrum/perun/audit/events/AttributesManagerEvents/AttributePolicyCollectionSet.java
```diff
@@ -0,0 +1,38 @@
...
+
+import cz.metacentrum.perun.audit.events.AuditEvent;
+import cz.metacentrum.perun.audit.events.EngineIgnoreEvent;
+import cz.metacentrum.perun.core.api.AttributePolicyCollection;
+import cz.metacentrum.perun.core.api.AttributeRights;
+
+public class AttributePolicyCollectionSet extends AuditEvent implements EngineIgnoreEvent {
+
+	private AttributePolicyCollection policyCollection;
+	private int attributeId;
```
cuadradek, line 11: This might be unnecessary field because `attributeId` is already part of `policyCollection`. What do you think?

cuadradek (commented):
Nice work. Please, add the new method to openAPI as well.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTest.java
```diff
@@ -9835,6 +9839,44 @@ public void setAttributeRights() throws Exception {
...
+		List<AttributePolicyCollection> policyCollections = new ArrayList<>();
+		List<AttributePolicy> policies = new ArrayList<>();
+		int attrId = 42;
+
+		policies.add(new AttributePolicy(42, Role.VOADMIN, RoleObject.User, 1));
+		policies.add(new AttributePolicy(43, Role.PERUNOBSERVER, RoleObject.User, 1));
+		policyCollections.add(new AttributePolicyCollection(100, userAttribute1.getId(), AttributeAction.READ, new ArrayList<>(policies)));
+
+		perun.getAttributesManager().setAttributePolicyCollections(sess, policyCollections, userAttribute1.getId());
+		policies.clear();
```
cuadradek, line 9857: I think clearing this list is not needed since it isn't used later.

cuadradek (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AttributesManagerMethod.java
```diff
@@ -3957,6 +3957,25 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 		}
 	},
 
+	/*#
+	 * Sets all attribute policies in the list for attribute, both given as parameters.
+	 *
+	 * @param policies List<AttributePolicyCollection> list of policies to set
```
cuadradek, line 3963:
```suggestion
	 * @param policyCollections List<AttributePolicyCollection> list of AttributePolicyCollection to set
```

perun-openapi/openapi.yml
```diff
@@ -6875,6 +6875,33 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/attributesManager/setAttributePolicyCollections:
+    post:
+      tags:
+        - AttributesManager
+      operationId: setAttributePolicyCollections
+      summary: "Sets attribute policy collections for an attribute definition provided in given policy collections."
```
cuadradek, line 6883: When I think about it, it might be reasonable to mention that original policy collections are deleted. Same for all the layers. What do you think?

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/AttributesManagerEntry.java
```diff
@@ -4512,6 +4514,37 @@ public void setAttributeRights(PerunSession sess, List<AttributeRights> rights)
...
+			getAttributeDefinitionById(sess, attributeId);
+		}
+
+		// check validity of roles, existence of attributes
+		for (AttributePolicyCollection apc : policyCollections) {
+			for (AttributePolicy ap : apc.getPolicies()) {
+				if (!AuthzResolver.roleExists(ap.getRole())) {
+					throw new RoleNotSupportedException("Role: " + ap.getRole() + " does not exists.", ap.getRole());
+				}
+				getAttributeDefinitionById(sess, apc.getAttributeId());
```
Johaney-s, line 4536: Isn't this the same check already done with `attributeIds` list?

Johaney-s (approved):

cuadradek (commented):
One last thing - I think it will be more correct if the method runs in a nested transaction in IMPL layer. It's not necessary since we don't catch exceptions in BL layer right now, but that could change in the future. Plus, we execute multiple write commands in the IMPL layer method, hence it make sense to make a nested transaction to group them. What do you think?

cuadradek (approved):

PR #3449, opened 2022-01-18
fix(core): fix transactions when copying facilityadmins

Johaney-s:
* when copying facilityadmin between facilities, AlreadyAdminException is thrown
* because this exception is thrown in nested transaction that continues with other operations, postgres raises error
* using transaction on setRole method should correctly rollback copying already present admin and continue


bodnara (approved):
Seems fine to me, but then again, what do I know.

zlamalp (approved):

PR #3452, opened 2022-01-20
feat(core): allow ssh keys with command prefixes

xflord:
* added new method removeSSHKeyCommandPrefix which checks for and removes the command prefix, then the validation proceeds as before.


Johaney-s (approved):

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_sshPublicKeyTest.java
```diff
@@ -98,6 +98,29 @@ public void testCheckCorrectSshDssAttributeSyntax() throws Exception {
 		classInstance.checkAttributeSyntax(session, user, attributeToCheck);
 	}
 
+	@Test
+	public void testCheckCorrectSshRSAWithPrefixAttributeSyntax() throws Exception {
+		System.out.println("testCheckCorrectSshPrefix()");
```
Johaney-s, line 103: Looks good, just please match the test name to what's printed.

Johaney-s (changes requested):

Johaney-s (approved):

PR #3463, opened 2022-01-31
feat(core): create a new role that interacts with methods in AuditMes…

xflord:
…sagesManagerEntry

* created a new role AUDITCONSUMERADMIN and set corresponding policy rules for AuditMessagesManagerEntry methods
* edited PERUNOBSERVER policy rules to allow access to get methods in AuditMessagesManagerEntry


perun-core/src/test/java/cz/metacentrum/perun/core/entry/AuditMessagesManagerEntryIntegrationTest.java
```diff
@@ -169,6 +178,75 @@ public void getMessagesPage_ascendingOrder() throws Exception {
 		assertThat(messages.getData().get(0).getId()).isLessThan(messages.getData().get(1).getId());
 	}
 
+	@Test
```
Johaney-s, line 181: Authorization tests are used mainly in `AuthzResolverIntegrationTest`, maybe it would be better to move these to the same file?

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AuditMessagesManagerEntryIntegrationTest.java
```diff
@@ -169,6 +178,75 @@ public void getMessagesPage_ascendingOrder() throws Exception {
...
+
+		PerunSession sess1 = perun.getPerunSession(pp1, new PerunClient());
+
+		assertFalse(AuthzResolver.authorizedInternal(sess1, "setLastProcessedId_String_int_policy"));
+		assertFalse(AuthzResolver.authorizedInternal(sess1, "createAuditerConsumer_String_policy"));
+	}
+
+	@Test
+	public void isUnauthorizedOther() throws Exception {
+		System.out.println(CLASS_NAME + "isUnauthorizedPerunObserver");
```
Johaney-s, line 233: Please fix this line to match the testcase name

cuadradek (commented):
Good job.

Please, add a brief description of the role to the beginning of file `perun-roles.yml`. Also, I think the role with a simple policy should be added to file `test-roles.yml` with a simple test in `AuthzResolverIntegrationTest.java` like other roles.

Then we should add the role to CLI scripts `setRole` and `unsetRole`, so we can actually (un)set this role to / from a user at least from CLI. It should be very similar to (un)setting CABINETADMIN in these scripts.

Lastly, we can also update documentation of methods setRole and unsetRole in RPC API so it contains the new role as well.

cuadradek (approved):

Johaney-s (approved):

PR #3464, opened 2022-02-02
feat(base): allow paging for Applications

xflord:
* added getApplicationsPage method to RegistrarManager to enable paged SQL queries (similar to members and groups)


perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/RegistrarBaseIntegrationTest.java
```diff
@@ -794,6 +799,291 @@ private Application prepareApplicationToGroup(User user, Group group) {
...
+
+		List<User> returnedUsers = result.getData().stream()
+			.map(Application::getUser).toList();
+
+		assertThat(returnedUsers).containsOnly(user2);
+	}
+
+	@Test
+	public void getApplicationsPageFindByUsername() throws Exception {
+		System.out.println("getApplicationsPageFindByGroupName");
```
Johaney-s, line 901:
```suggestion
		System.out.println("getApplicationsPageFindByUsername");
```

Anyway, this test seems to do the same thing as the previous one, I'd say we can remove it.

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/RegistrarBaseIntegrationTest.java
```diff
@@ -794,6 +799,291 @@ private Application prepareApplicationToGroup(User user, Group group) {
...
+		Paginated<Application> result = registrarManager.getApplicationsPage(session, vo, query);
+
+		List<Group> returnedGroups = result.getData().stream()
+			.map(Application::getGroup).filter(Predicate.not(Objects::isNull)).toList();
+
+		assertThat(returnedGroups).containsOnly(group2);
+	}
+
+	@Test
+	public void getApplicationsPageFindByGroupId() throws Exception {
```
Johaney-s, line 975: Could you also add tests where groupId is not used as search string, but as groupId variable in the query and the same for the memberId?

Johaney-s (changes requested):
Really nice job and excellent test coverage, I have mainly cosmetic suggestions in the comments.

I think there are now not included string matches with app data though, which include additional information. Shouldn't it be part of this PR?

And please, add the new method also to `openapi `so we can use it in GUI. We also add new objects to `parseRpcMethods.pl`, add it there too.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2054,6 +2066,55 @@ public Application getApplicationById(PerunSession sess, int appId) throws Regis
...
+	@Override
+	public Paginated<Application> getApplicationsPage(PerunSession userSession, Vo vo, ApplicationsPageQuery query) throws PerunException {
+		vosManager.checkVoExists(userSession, vo);
+		MapSqlParameterSource namedParams = new MapSqlParameterSource();
+
+		//Authorization
+		if (!AuthzResolver.authorizedInternal(userSession, "getApplicationsForVo_Vo_List<String>_Boolean_policy", Collections.singletonList(vo))) {
+			throw new PrivilegeException(userSession, "getApplicationsForVo");
+		}
+		if (query.getGroupId() != null && !AuthzResolver.authorizedInternal(userSession, "getApplicationsForGroup_Group_List<String>_policy", Collections.singletonList(groupsManager.getGroupById(userSession, query.getGroupId())))) {
```
Johaney-s: This check will fail already on the first policy if groupId is used, because groupAdmin is not allowed to call getApplications for Vo, only for Group.

xflord:
> I think there are now not included string matches with app data though, which include additional information. Shouldn't it be part of this PR?

Pretty sure this is the last part of the 3-part task, no?

Hopefully I added everything that's necessary into openapi, also when it comes to constructors I'm not sure whether to have more of them or force people to use null values in fewer - let me know.

perun-openapi/openapi.yml
```diff
@@ -965,6 +965,47 @@ components:
...
+        pageSize: { type: integer }
+        offset: { type: integer }
+        order: { $ref: '#/components/schemas/SortingOrder' }
+        sortColumn: { $ref: '#/components/schemas/ApplicationsOrderColumn' }
+        searchString: { type: string }
+        states: { type: array, items: { $ref: '#/components/schemas/AppState'  } }
+        dateFrom: { type: string, format: date }
+        dateTo: { type: string, format: date }
+        memberId: { type: integer }
+        groupId: { type: integer }
```
Johaney-s, line 991: `includeGroupApplications` should be here too

perun-openapi/openapi.yml
```diff
@@ -14242,6 +14290,32 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/membersManager/getApplicationsPage:
```
Johaney-s, line 14293:
```suggestion
  /json/registrarManager/getApplicationsPage:
```

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2054,6 +2056,54 @@ public Application getApplicationById(PerunSession sess, int appId) throws Regis
...
+		namedParams.addValue("memberId", query.getMemberId());
+		namedParams.addValue("groupId", query.getGroupId());
+		String searchQuery = getSQLWhereForApplicationsPage(query, namedParams);
+
+		Paginated<Application> applications = namedJdbc.query(
+			APP_SELECT_PAGE +
+				" WHERE a.vo_id=(:voId)" +
+				(query.getStates() == null || query.getStates().isEmpty() ? "" : " AND a.state IN (:states) ") +
+				(query.getIncludeGroupApplications() ? "" : " AND a.group_id is null") +
+				(query.getMemberId() == null ? "" : "  AND a.user_id=(:memberId)") +
```
Johaney-s, line 2087: MemberId shouldn't be matched to userId, userId should be obtained by calling `getMemberById`and retrieving its userId, which can then be used in the WHERE clause. It will need fixing in the tests then too.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
```diff
@@ -439,6 +439,32 @@ public ApplicationFormItem call(ApiCaller ac, Deserializer parms) throws PerunEx
...
+	 * @throw PrivilegeException if user doesnt have sufficient privileges
+	 * @throw PerunException
+	 */
+	getApplicationsPage {
+
+		@Override
+		public Object call(ApiCaller ac, Deserializer parms) throws PerunException {
+			return ac.getRegistrarManager().getApplicationsPage(ac.getSession(),
+				ac.getVoById(parms.readInt("vo")),
+				parms.read("query", ApplicationsPageQuery.class));
```
Johaney-s, line 464: I get CANNOT_DESERIALIZE_VALUE error every time I include dateFrom or dateTo in the query, it seems it can't deserialize it correcly if it is part of the object.

cuadradek (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -4480,6 +4530,27 @@ private void processNotificationsAfterCreation(PerunSession session, Application
...
 	}
+
+	/**
+	 * Returns ResultSetExtractor that can be used to extract returned paginated applications
+	 * from db.
+	 *
+	 * @param query query data
+	 * @return extractor, that can be used to extract returned paginated applications from db
+	 */
+	private static ResultSetExtractor<Paginated<Application>> getPaginatedApplicationsExtractor(ApplicationsPageQuery query) {
```
cuadradek, line 4541: Nit: you can move this method under `---MAPPERS AND SELECTS---` below because it is related logic.

cuadradek (changes requested):
We don't use scope `base` iirc. Scopes of this PR should actually be `registrar` and `openapi`, so there should be 2 commits.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2054,6 +2056,54 @@ public Application getApplicationById(PerunSession sess, int appId) throws Regis
...
+		namedParams.addValue("limit", query.getPageSize());
+		namedParams.addValue("memberId", query.getMemberId());
+		namedParams.addValue("groupId", query.getGroupId());
+		String searchQuery = getSQLWhereForApplicationsPage(query, namedParams);
+
+		Paginated<Application> applications = namedJdbc.query(
+			APP_SELECT_PAGE +
+				" WHERE a.vo_id=(:voId)" +
+				(query.getStates() == null || query.getStates().isEmpty() ? "" : " AND a.state IN (:states) ") +
+				(query.getIncludeGroupApplications() ? "" : " AND a.group_id is null") +
```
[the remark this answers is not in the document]
xflord: includeGroupApplications has a default value (true) in ApplicationsPageQuery, so null shouldn't be possible

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
xflord: Changed the name of the attribute in ApplicationPageQuery to userId, as this was what I wanted it to represent in the first place. When searching for applications a member of a group/vo has in that group/vo, it doesn't matter whether we use his memberId or userId, does it? And as his userId is stored in the database with the applications, I believe this to be the simpler solution.

perun-base/pom.xml
```diff
@@ -139,6 +139,13 @@
 			<version>${testcontainers.version}</version>
 			<scope>compile</scope>
 		</dependency>
+		<!-- https://mvnrepository.com/artifact/com.fasterxml.jackson.datatype/jackson-datatype-jsr310 -->
+		<dependency>
+    		<groupId>com.fasterxml.jackson.datatype</groupId>
+    		<artifactId>jackson-datatype-jsr310</artifactId>
+    		<version>2.13.1</version>
```
cuadradek, line 146: I think we can omit the version tag, because this project has `spring-boot-starter-parent` as a parent project. It should automatically provide versions for some dependencies and update them in the future.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2054,6 +2056,60 @@ public Application getApplicationById(PerunSession sess, int appId) throws Regis
...
+		if (query.getGroupId() != null && !AuthzResolver.authorizedInternal(userSession, "getApplicationsForGroup_Group_List<String>_policy", Collections.singletonList(groupsManager.getGroupById(userSession, query.getGroupId())))) {
+			throw new PrivilegeException(userSession, "getApplicationsPage");
+		} else if (!AuthzResolver.authorizedInternal(userSession, "getApplicationsForVo_Vo_List<String>_Boolean_policy", Collections.singletonList(vo))) {
+			throw new PrivilegeException(userSession, "getApplicationsPage");
+		}
+
+
+		namedParams.addValue("voId", vo.getId());
+		if (query.getStates() != null) {
+			namedParams.addValue("states", query.getStates().stream()
```
cuadradek, line 2074: This would cause NullPointerException if `getStates()` is null.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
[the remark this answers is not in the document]
cuadradek: I see, but null can still be passed to constructor or setter. I don't think we have use case for that, so you don't have to change it, I just thought it would be a little more correct.

perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl
```diff
@@ -93,6 +93,17 @@
...
+$objectExamples{"List&lt;Application&gt;"} = $listPrepend . $objectExamples{"Application"} . $listAppend;
+$objectExamples{"List<Application>"} = $objectExamples{"List&lt;Application&gt;"};
+
+$objectExamples{"ApplicationsPageQuery"} = "{ \"pageSize\" : 3 , \"offset\" : 0 , \"order\" : \"ASCENDING\" , \"sortColumn\" : \"ID\" , \"includeGroupApplications\" : true , \"searchString\" : \"Doe\" , \"states\" : [\"VERIFIED\" , \"NEW\"] , \"dateFrom\" : \"2011-05-17\", \"dateTo\" : \"2011-05-17\", \"memberId\" : 10 , \"groupId\" : 10 }";
+$objectExamples{"List&lt;ApplicationsPageQuery&gt;"} = $listPrepend . $objectExamples{"ApplicationsPageQuery"} . $listAppend;
+$objectExamples{"List<ApplicationsPageQuery>"} = $objectExamples{"List&lt;ApplicationsPageQuery&gt;"};
+
+$objectExamples{"Paginated&lt;Application&gt;"} = "{ \"offset\" : 0 , \"pageSize\" : 3 , \"totalCount\" : 1 , \"data\" : " . $objectExamples{"List&lt;Application&gt;"} . "  }";
+$objectExamples{"Paginated<Application>"} = @objectExamples{"Paginated&lt;Application&gt;"};
```
cuadradek, line 105:
```suggestion
$objectExamples{"Paginated<Application>"} = $objectExamples{"Paginated&lt;Application&gt;"};
```
This looks like a typo, it is only in Paginated objects - can you please fix it in all objects?

cuadradek (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
cuadradek: Sorry, I prepared this comment yesterday, I can see it is fixed now.

Johaney-s (approved):

cuadradek (approved):

PR #3467, opened 2022-02-04
feat(registrar): make searchString in applications page search in application_data table as well

xflord:


perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/RegistrarBaseIntegrationTest.java
```diff
@@ -36,6 +36,7 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Bean;
```
Johaney-s, line 39: unused import

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -2123,6 +2123,15 @@ public static String prepareGroupNameDscAppsSearchQuerySimilarMatch() {
 		return "(strpos(lower("+Compatibility.convertToAscii("COALESCE(g.name,'') || COALESCE(g.dsc,'')")+"), lower(" + Compatibility.convertToAscii(":searchString") + ")) > 0)";
 	}
 
+	/**
+	 * Returns search query to search by group name or description (similar match) based on databased in use.
+	 *
+	 * @return search query
+	 */
+	public static String prepareApplicationDataSearchQuerySimilarMatch() {
+		return "(strpos(lower("+Compatibility.convertToAscii("COALESCE(d.value,'')")+"), lower(" + Compatibility.convertToAscii(":searchString") + ")) > 0)";
```
cuadradek, line 2132: Compatibility class was used to resolve differences between supported RDBMS. Since we support only Postgres now, you don't need to call methods of this class and you can use directly Postgres-specific expressions in SQL queries. You don't need to change this now, just so you know in the future.

Johaney-s (approved):

cuadradek (approved):

PR #3471, opened 2022-02-09
fix(core): modules marked as deprecated

sarkapalkovicova:
Marked researcher_access_persistent and researcher_access_persistent_shadow
modules as deprecated, because they are not used.


Johaney-s (commented):
I think these changes don't require two separate commits. Also, we usually write commit descriptions about what changes were made and why, please add some too.

Johaney-s (approved):

PR #3474, opened 2022-02-15
fix: mfaTokens marked as obsolete

sarkapalkovicova:
LDAP attribute mfaTokens was marked as deleted.


xflord (commented):
You should change the commit message to include the scope of the PR like all the other commits in Perun (e.g. fix(ldap): mfaTokens.....), other than that looks good to me.

xflord (approved):

PR #3475, opened 2022-02-15
feat(core): solve default and associated roles for attribute rights

Johaney-s:
* These changes are related only to new attribute authorization
* By default PERUNADMIN needs read and write privilege for attributes, default read roles are also PERUNOBSERVER, RPC, ENGINE
* Default roles are automatically resolved before fetching related objects to speed up the check
* Observer roles are defined in perun-roles.yml as associated read roles to admin roles
* Associated roles can read attribute if the main role can (groupobserver can read it if groupadmin can)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1290,6 +1330,11 @@ public static boolean isAuthorizedForAttribute(PerunSession sess, AttributeActio
 		log.trace("Entering isAuthorizedForAttribute: sess='{}', actionType='{}', attrDef='{}', primaryHolder='{}', " +
 			"secondaryHolder='{}'", sess, actionType, attrDef, resource, null);
 
+		// Check roles which are authorized by default
+		if (sess.getPerunPrincipal().getRoles() != null && !sess.getPerunPrincipal().getRoles().isEmpty() && hasAccessByDefault(sess, actionType)) {
```
cuadradek, line 1334: Just a nit: maybe we could move the first two conditions to hasAccessByDefault so it is more readable.

cuadradek (commented):
Nice work, could you please add one more test to check that associated read roles only affect READ operations and not WRITE operations?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
Johaney-s: Good point, moved it. Also added test for the write case.

cuadradek (approved):

Kuliak (approved):

PR #3482, opened 2022-02-23
feat: add methods for enabling/disabling member VOs for a VO

cuadradek:
feat(core): add methods for enabling/disabling member VOs for a VO
- New field isTenant was added to Vo class along with getter and
setter method.
- New column is_tenant was added to vos table.
- New method enableMemberVos() was added to all layers and RPC. It
enables the given VO to contain member organizations.
- New method disableMemberVos() was added to all layers and RPC. It
disables the given VO to contain member organizations.

feat(openapi): add methods for enabling/disabling member VOs for a VO
- Added memberVosEnabled property to Vo object schema.
- New method vosManager/enableMemberVos was added. It enables the given
VO to contain member organizations.
- New method vosManager/disableMemberVos was added. It disables the
given VO to contain member organizations.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/Vo.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/Vo.java
@@ -8,6 +8,7 @@ import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 public class Vo extends Auditable implements Comparable<PerunBean> {
 	private String name;
 	private String shortName;
+	private boolean memberVosEnabled = false; // can other VOs be member organizations of this VO?
 
 	public Vo() {
 	}
@@ -19,6 +20,11 @@ public class Vo extends Auditable implements Comparable<PerunBean> {
 		this.name = name;
 		this.shortName = shortName;
 
+	}
+
+	public Vo(int id, String name, String shortName, boolean memberVosEnabled) {
+		this(id, name, shortName);
+		this.memberVosEnabled = memberVosEnabled;
 	}
 
 	@Deprecated
@@ -36,6 +42,11 @@ public class Vo extends Auditable implements Comparable<PerunBean> {
 		if (shortName == null)  throw new InternalErrorException(new NullPointerException("shortName is null"));
 		this.name = name;
 		this.shortName = shortName;
+	}
+
+	public Vo(int id, String name, String shortName, boolean memberVosEnabled, String createdAt, String createdBy, String modifiedAt, String modifiedBy, Integer createdByUid, Integer modifiedByUid) {
+		this(id, name, shortName, createdAt, createdBy, modifiedAt, modifiedBy, createdByUid, modifiedByUid);
+		this.memberVosEnabled = memberVosEnabled;
 	}
 
 	public String getName() {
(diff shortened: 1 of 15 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/Vo.java
```diff
@@ -8,6 +8,7 @@
 public class Vo extends Auditable implements Comparable<PerunBean> {
 	private String name;
 	private String shortName;
+	private boolean isTenant = false;
```
Johaney-s, line 11: Shouldn't it be also part of `equals`, `hashCode` and `compareTo` methods?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5667,6 +5667,18 @@ perun_policies:
     include_policies:
       - default_policy
 
+  setTenant_Vo_policy:
+    policy_roles:
+      - VOADMIN: Vo
```
Johaney-s, line 5672: only perunadmin should be authorized

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -74,6 +75,17 @@ public void createVo() throws Exception {
 		assertTrue("id must be greater than zero", newVo.getId() > 0);
 	}
 
+	@Test
+	public void createVoIsNotTenant() throws Exception {
+		System.out.println(CLASS_NAME + "createVoIsNotTenant");
```
Johaney-s, line 80: This behaviour is quite tricky, because it doesn't even raise an exception and just ignores the isTenant==true parameter. I'd say we should either allow this variant and set it to be tenant directly when creating the vo, or at least raise an exception.

Johaney-s (changes requested):
A few more comments:
- We should also update vo object in `parserpcmethods`.
- It would be nice to have a method returning all tenants in Perun, what do you think?
- Please reduce the number of commits

HejdaJakub (commented):
I agree with @Johaney-s. Otherwise, I found only missing exceptions in one Javadoc.

perun-base/src/main/java/cz/metacentrum/perun/core/api/Vo.java
```diff
@@ -56,6 +67,14 @@ public void setShortName(String shortName) {
 		this.shortName = shortName;
 	}
 
+	public boolean isTenant() {
+		return isTenant;
+	}
+
+	public void setTenant(boolean tenant) {
```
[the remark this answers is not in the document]
cuadradek: This was actually generated by IDE and it is a naming convention for boolean "is" variables.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
cuadradek: This actually follows the method's documentation in the RPC API that only VO's name and shortname are used and other parameters are ignored.  Therefore, I think an exception doesn't need to be thrown, but at least Javadoc of the method in other layers should be updated according to RPC API. And we discussed that we don't need the possibility to create tenants directly because it is a very rare use case and it would be probably an unnecessary breaking change. What do you think?

Johaney-s: We should remove the option from the constructor then.

cuadradek:
Since we decided to use the term _member organizations_ instead of _tenants_, I renamed everything in the PR accordingly.

cuadradek:
Don't merge this until the DB is updated on devel.

Kuliak (approved):

cuadradek:
I am closing this PR because we decided not to enable/disable memberVos - instead, any VO can have memberVos.

PR #3488, opened 2022-03-07
feat: add method for retrieving RichMembers for VO without user attributes

xflord:


HejdaJakub (changes requested):
I don't think we want to change all these existing methods, especially to implementation with the new required parameter. I thought that we just want one or a few new method(s) (more than one only if we want to allow the definition of attrNames, Vo statuses, etc. - but I am not sure if we want it) with no impact on existing methods. 

According to the task definition, we want this use case only for Vos, so I think you don't have to deal with groups, too.

perun-openapi/openapi.yml
```diff
@@ -9101,6 +9101,22 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/membersManager/getRichMembersWithoutUserAttributes:
```
Johaney-s, line 9104:
```suggestion
  /json/membersManager/getRichMembersNoUserAttributes:
```
This has to match the method name in MembersManager.

perun-openapi/openapi.yml
```diff
@@ -9101,6 +9101,22 @@ paths:
...
+    get:
+      tags:
+        - MembersManager
+      operationId: getRichMembersWithoutUserAttributes
```
Johaney-s, line 9108:
```suggestion
      operationId: getRichMembersNoUserAttributes
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -846,6 +846,20 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
...
+		Utils.checkPerunSession(sess);
+
+		getPerunBl().getVosManagerBl().checkVoExists(sess, vo);
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "getRichMembers_Vo_policy", vo)) {
+			throw new PrivilegeException(sess, "getRichMembersNoUserAttributes");
+		}
+
+		return getPerunBl().getMembersManagerBl().filterOnlyAllowedAttributes(sess, getMembersManagerBl().getRichMembers(sess, vo), null, true);
```
Johaney-s, line 860:
```suggestion
		return getPerunBl().getMembersManagerBl().filterOnlyAllowedAttributes(sess, getMembersManagerBl().getRichMembersNoUserAttributes(sess, vo), null, true);
```
we need to call the right method

Johaney-s (changes requested):
it would be nice to have a test for this.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -846,6 +846,20 @@ public RichMember getRichMemberWithAttributes(PerunSession sess, Member member)
...
 	}
 
+	@Override
+	public List<RichMember> getRichMembersNoUserAttributes(PerunSession sess, Vo vo) throws PrivilegeException, VoNotExistsException {
+		Utils.checkPerunSession(sess);
+
+		getPerunBl().getVosManagerBl().checkVoExists(sess, vo);
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "getRichMembers_Vo_policy", vo)) {
```
HejdaJakub, line 856: You should also add the new method to the `perun-roles.yml`. The yml definition will be the same as `getRichMembers_Vo_policy`.
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "getRichMembersNoUserAttributes_Vo_policy", vo)) {
```

HejdaJakub (changes requested):

Johaney-s (approved):

HejdaJakub (approved):

PR #3492, opened 2022-03-10
feat: allow to lookup users by IUID and ext source login

HejdaJakub:
* In this PR is created new flag which allows to lookup users by IUID (additionalIdentifiers) and if
no user is found, then start lookup by user ext source login and name.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
@@ -57,7 +57,6 @@ import cz.metacentrum.perun.core.bl.VosManagerBl;
 import cz.metacentrum.perun.core.impl.Auditer;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.impl.ResourceAssignmentChecker;
-import cz.metacentrum.perun.core.impl.ResourceAssignmentActivator;
 import cz.metacentrum.perun.core.implApi.AttributesManagerImplApi;
 import cz.metacentrum.perun.core.implApi.ResourceAssignmentActivatorApi;
 import org.slf4j.Logger;
@@ -121,6 +120,7 @@ public class PerunBlImpl implements PerunBl {
 
 	private final static Set<String> dontLookupUsersForLogins = BeansUtils.getCoreConfig().getDontLookupUsers();
 	private final static Set<String> extSourcesWithMultipleIdentifiers = BeansUtils.getCoreConfig().getExtSourcesMultipleIdentifiers();
+	private final static boolean lookupUserByIdentifiersAndExtSourceLogin = BeansUtils.getCoreConfig().getLookupUserByIdentifiersAndExtSourceLogin();
 
 	public PerunBlImpl() {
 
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -783,11 +785,20 @@ public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess,
...
-			UserExtSource ues = getUserExtSourceFromMultipleIdentifiers(sess, principal);
-			return getUserByUserExtSource(sess, ues);
-		} else {
-			return getUserByExtSourceNameAndExtLogin(sess, principal.getExtSourceName(), principal.getActor());
+			try {
+				UserExtSource ues = getUserExtSourceFromMultipleIdentifiers(sess, principal);
+				return getUserByUserExtSource(sess, ues);
+			} catch (UserExtSourceNotExistsException ex) {
+				if (!lookupUserByIdentifiersAndExtSourceLogin) {
+					throw new UserExtSourceNotExistsException(ex);
```
cuadradek, line 793: Nit: we can simply rethrow the existing exception instead of a new one. Creating a new exception can be an expensive operation (if happening too often) but I don't think it would make a difference in this case.
```suggestion
					throw ex;
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -783,11 +785,20 @@ public UserExtSource getUserExtSourceFromMultipleIdentifiers(PerunSession sess,
...
+				}
+			} catch (UserNotExistsException ex) {
+				if (!lookupUserByIdentifiersAndExtSourceLogin) {
+					throw new UserNotExistsException(ex);
```
cuadradek, line 797: Same here. If we simply rethrow the existing exception, I think we could even merge the two catch blocks like:
`catch (UserExtSourceNotExistsException | UserNotExistsException ex) {`

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -143,11 +143,14 @@ public PerunSession getPerunSession(PerunPrincipal principal, PerunClient client
 
 				if (client.getType() != PerunClient.Type.OAUTH) {
 					// Try to update LoA for userExtSource
-					UserExtSource ues;
+					UserExtSource ues = null;
 					String shibIdentityProvider = principal.getAdditionalInformations().get(UsersManagerBl.ORIGIN_IDENTITY_PROVIDER_KEY);
+					boolean findByExtSourceNameAndLogin = true;
 					if(shibIdentityProvider != null && extSourcesWithMultipleIdentifiers.contains(shibIdentityProvider)) {
 						ues = usersManagerBl.getUserExtSourceFromMultipleIdentifiers(internalSession, principal);
```
cuadradek, line 150: We should probably catch `UserExtSourceNotExistsException` here so we can try to find ues by name and login in such a case (similarly to `getUserByExtSourceInformation` method). What do you think?

cuadradek (commented):
Looks good; I only have two small suggestions.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
HejdaJakub: You are right. I updated this block of code.

Johaney-s (approved):

cuadradek (approved):

PR #3494, opened 2022-03-10
feat(core): create missing members in hierarchical vos and update attribute

Johaney-s:
* When adding new member vo, missing users are added as members to parent vo
* When member vo is added or removed, parent vo members' memberOrganizations attribute is updated accordingly

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
@@ -5,6 +5,9 @@ import cz.metacentrum.perun.audit.events.MembersManagerEvents.MemberUnsuspended;
 import cz.metacentrum.perun.audit.events.VoManagerEvents.VoCreated;
 import cz.metacentrum.perun.audit.events.VoManagerEvents.VoDeleted;
 import cz.metacentrum.perun.audit.events.VoManagerEvents.VoUpdated;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.BanOnVo;
 import cz.metacentrum.perun.core.api.Candidate;
 import cz.metacentrum.perun.core.api.EnrichedVo;
@@ -27,12 +30,14 @@ import cz.metacentrum.perun.core.api.UserExtSource;
 import cz.metacentrum.perun.core.api.Vo;
 import cz.metacentrum.perun.core.api.VosManager;
 import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
+import cz.metacentrum.perun.core.api.exceptions.AlreadyMemberException;
 import cz.metacentrum.perun.core.api.exceptions.AlreadySponsorException;
 import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.BanNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.CandidateNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
 import cz.metacentrum.perun.core.api.exceptions.ExtSourceUnsupportedOperationException;
+import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
 import cz.metacentrum.perun.core.api.exceptions.GroupExistsException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
@@ -47,6 +52,9 @@ import cz.metacentrum.perun.core.api.exceptions.UserNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.VoExistsException;
 import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
 import cz.metacentrum.perun.core.bl.MembersManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.bl.UsersManagerBl;
@@ -82,6 +90,7 @@ public class VosManagerBlImpl implements VosManagerBl {
 
 	private final static Logger log = LoggerFactory.getLogger(VosManagerBlImpl.class);
 	private final static Date FAR_FUTURE = new GregorianCalendar(2999, Calendar.JANUARY, 1).getTime();
+	public final static String A_MEMBER_DEF_MEMBER_ORGANIZATIONS = AttributesManager.NS_MEMBER_ATTR_DEF + ":memberOrganizations";
 
 	private final VosManagerImplApi vosManagerImpl;
 	private PerunBl perunBl;
@@ -171,6 +180,24 @@ public class VosManagerBlImpl implements VosManagerBl {
 				} else {
 					throw new RelationExistsException("Vo vo=" + vo + " contains groups");
 				}
+			}
+
+			// Remove hierarchical relations
+			log.debug("Removing {} hierarchical relations with parent organizations", vo);
+			List<Vo> parentVos = getParentVos(sess, vo.getId());
+			if (parentVos.size() != 0 && !forceDelete) {
+				throw new RelationExistsException("Vo vo=" + vo + " is member of hierarchical organization");
+			}
+			for (Vo parentVo : parentVos) {
+				removeMemberVo(sess, parentVo, vo);
+			}
+			log.debug("Removing {} hierarchical relations with member organizations", vo);
+			List<Vo> memberVos = getMemberVos(sess, vo.getId());
+			if (memberVos.size() != 0 && !forceDelete) {
+				throw new RelationExistsException("Vo vo=" + vo + " has member organizations");
+			}
+			for (Vo memberVo : memberVos) {
+				removeMemberVo(sess, vo, memberVo);
 			}
 
 			// Finally delete binding between Vo and external source
@@ -918,6 +945,61 @@ public class VosManagerBlImpl implements VosManagerBl {
 	public void addMemberVo(PerunSession sess, Vo vo, Vo memberVo) throws RelationExistsException {
 		checkParentVos(sess, vo, memberVo);
 		vosManagerImpl.addMemberVo(sess, vo, memberVo);
+
+		List<Member> memberVoMembers = perunBl.getMembersManagerBl().getMembers(sess, memberVo);
+		AttributeDefinition memberExpirationAttrDef;
+
+		try {
+			memberExpirationAttrDef = perunBl.getAttributesManagerBl().getAttributeDefinition(sess, AttributesManager.NS_MEMBER_ATTR_DEF + ":membershipExpiration");
+		} catch (AttributeNotExistsException e) {
+			throw new InternalErrorException(e);
+		}
+
+		for (Member member : memberVoMembers) {
+			if (member.getStatus() == Status.DISABLED || member.getStatus() == Status.INVALID) {
+				continue;
+			}
+			try {
+
+				try {
+					Member existingMember = perunBl.getMembersManagerBl().getMemberByUserId(sess, vo, member.getUserId());
+					if (member.getStatus() == Status.EXPIRED) {
+						// member is expired in memberVo but exists in parentVo, don't update
+						continue;
+					}
+
+					// member is valid in memberVo, so reset expiration in parentVo
+					perunBl.getAttributesManagerBl().removeAttribute(sess, existingMember, memberExpirationAttrDef);
+					if (existingMember.getStatus() != Status.VALID) {
+						perunBl.getMembersManagerBl().validateMemberAsync(sess, existingMember);
+					}
+					//update memberOrganizations
+					Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, existingMember, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+					ArrayList<String> currentValue = attribute.valueAsList();
+					currentValue.add(memberVo.getShortName());
+					attribute.setValue(currentValue);
+					perunBl.getAttributesManagerBl().setAttribute(sess, existingMember, attribute);
+
+				} catch (MemberNotExistsException e) {
+					// if user is member only in member vo, create member in parent vo
+					Member newMember = perunBl.getMembersManagerBl().createMember(sess, vo, perunBl.getUsersManagerBl().getUserByMember(sess, member));
+					if (member.getStatus() == Status.VALID) {
+						// remove expiration set according to parentVo and update memberOrganizations
+						perunBl.getAttributesManagerBl().removeAttribute(sess, newMember, memberExpirationAttrDef);
+						Attribute attribute = new Attribute(perunBl.getAttributesManagerBl().getAttributeDefinition(sess, A_MEMBER_DEF_MEMBER_ORGANIZATIONS));
+						ArrayList<String> newValue = new ArrayList<>(List.of(memberVo.getShortName()));
+						attribute.setValue(newValue);
+						perunBl.getAttributesManagerBl().setAttribute(sess, newMember, attribute);
+					}
+					perunBl.getMembersManagerBl().validateMemberAsync(sess, newMember);
+				}
+
+			} catch (WrongAttributeAssignmentException | WrongReferenceAttributeValueException | AttributeNotExistsException | WrongAttributeValueException | AlreadyMemberException e) {
+				throw new InternalErrorException(e);
+			} catch (ExtendMembershipException e) {
+				log.warn("Could not set expiration for member " + member + " after adding vo " + memberVo + " as member of vo " + vo + " for reason: " + e.getReason());
+			}
+		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -775,6 +778,75 @@ public void addExistingMemberVo() throws Exception {
...
+
+		assertThat(perun.getMembersManagerBl().getMembers(sess, vo).size()).isEqualTo(1);
+
+		vosManagerEntry.removeMemberVo(sess, vo, memberVo);
+		assertThat((ArrayList<String>) perun.getAttributesManagerBl().getAttribute(sess, member, A_MEMBER_DEF_MEMBER_ORGANIZATIONS).getValue())
+			.isNullOrEmpty();
+	}
+
+	@Test
+	public void deleteVoInHierarchy() throws Exception {
```
cuadradek, line 837: We should also probably test deleting the parent VO and it would be nice to have `RelationExistsException` use cases in tests as well.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +929,59 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+		for (Member member : memberVoMembers) {
+			try {
+
+				try {
+					// if user is member in both vos, update memberOrganizations list attribute
+					Member existingMember = perunBl.getMembersManagerBl().getMemberByUserId(sess, vo, member.getUserId());
+					Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, existingMember, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+					ArrayList<String> currentValue = (ArrayList<String>) attribute.getValue();
+					if (currentValue == null || currentValue.isEmpty()) {
+						throw new ConsistencyErrorException("Member " + existingMember + " of hierarchical vo " + vo + " has no origin vo set in attribute.");
```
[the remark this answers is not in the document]
Johaney-s: Right, haven't thought about that.  Removed the check.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
Johaney-s: DeleteVo method re-throws all exceptions of existing relationships as InternalErrorException, so I added check that one is thrown if called with forceDelete == false.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +928,54 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+					Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, existingMember, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+					ArrayList<String> currentValue = attribute.valueAsList();
+					currentValue.add(memberVo.getShortName());
+					attribute.setValue(currentValue);
+					perunBl.getAttributesManagerBl().setAttribute(sess, existingMember, attribute);
+
+				} catch (MemberNotExistsException e) {
+					// if user is member only in member vo, create member in parent vo and set up memberOrganizations list attribute
+					Member newMember = perunBl.getMembersManagerBl().createMember(sess, vo, perunBl.getUsersManagerBl().getUserByMember(sess, member));
+					perunBl.getMembersManagerBl().validateMemberAsync(sess, member);
```
cuadradek, line 950: I am not sure if we can validate the members asynchronously. The async method waits at first for 5 sec and then starts validating. It waits 5 seconds because the transaction, which inserts the member to DB, might not have been committed yet. In our case, I am not sure if the transaction will be committed within these 5 seconds if there are thousands of members to be created.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +928,54 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
 		vosManagerImpl.addMemberVo(sess, vo, memberVo);
+
+		List<Member> memberVoMembers = perunBl.getMembersManagerBl().getMembers(sess, memberVo);
+
+		for (Member member : memberVoMembers) {
+			try {
+
+				try {
+					// if user is member in both vos, update memberOrganizations list attribute
+					Member existingMember = perunBl.getMembersManagerBl().getMemberByUserId(sess, vo, member.getUserId());
```
cuadradek, line 940: Maybe we could check his current status and try to validate him if he is not valid yet. And perhaps also remove his expiration time if it is set for him because the member should follow the parent VOs lifecycle only if he is not valid in any member organization.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +929,65 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+					}
+					Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, existingMember, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+					ArrayList<String> currentValue = attribute.valueAsList();
+					currentValue.add(memberVo.getShortName());
+					attribute.setValue(currentValue);
+					perunBl.getAttributesManagerBl().setAttribute(sess, existingMember, attribute);
+
+				} catch (MemberNotExistsException e) {
+					// if user is member only in member vo, create member in parent vo and set up memberOrganizations list attribute
+					Member newMember = perunBl.getMembersManagerBl().createMember(sess, vo, perunBl.getUsersManagerBl().getUserByMember(sess, member));
```
cuadradek, line 961: I just realized that creating members also sets expiration times for them according to the parent VO rules. Therefore, we should most likely unset these expiration times for them because they should follow expiration rules only when they are not valid in any member organization.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +929,65 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+		}
+
+		for (Member member : memberVoMembers) {
+			try {
+
+				try {
+					// if user is member in both vos, update memberOrganizations list attribute, validate and reset expiration
+					Member existingMember = perunBl.getMembersManagerBl().getMemberByUserId(sess, vo, member.getUserId());
+					perunBl.getAttributesManagerBl().removeAttribute(sess, existingMember, memberExpirationAttrDef);
+					if (existingMember.getStatus().equals(Status.INVALID) || existingMember.getStatus().equals(Status.DISABLED)) {
```
cuadradek, line 950: We should probably validate existing members if they are expired as well, shouldn't we?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
cuadradek: We will test this and adjust the code in separate PR.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +929,65 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
-		// todo - add necessary logic for removing member vo
 		vosManagerImpl.removeMemberVo(sess, vo, memberVo);
+
+		List<Member> parentVoMembers = perunBl.getMembersManagerBl().getMembers(sess, vo);
+
+		for (Member member : parentVoMembers) {
+			try {
+				Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, member, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+				ArrayList<String> currentValue = attribute.valueAsList();
+				attribute.setValue(currentValue.remove(vo.getShortName()));
```
xflord, line 985: This all seems confusing to me. If we're removing memberVo from the hierarchy, don't we want to remove memberVo from the memberOrganizations attribute? If that's the case, I believe we only do this with members that are from memberVo, not all parentVo members.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -789,6 +795,120 @@ public void addExistingMemberVo() throws Exception {
...
+		Vo vo = perun.getVosManagerBl().createVo(sess, myVo);
+		Vo memberVo = perun.getVosManagerBl().createVo(sess, new Vo(-1, "Vo2", "vo2"));
+
+		Member member = createMemberFromExtSource(memberVo);
+		vosManagerEntry.addMemberVo(sess, vo, memberVo);
+
+		assertThat(perun.getMembersManagerBl().getMembers(sess, vo).size()).isEqualTo(1);
+
+		vosManagerEntry.removeMemberVo(sess, vo, memberVo);
+		assertThat(perun.getAttributesManagerBl().getAttribute(sess, member, A_MEMBER_DEF_MEMBER_ORGANIZATIONS).valueAsList())
```
xflord, line 849: In my understanding memberOrganizations is an attribute relevant only for members in parent VOs, so checking it for member of the member VO seems pointless. I would use getMemberByUser and check it for the member of the parent VO. Also this test might cause problems in the future when we delete members from parent VOs who are no longer members of any member VOs. Not sure whether this is relevant though.

xflord (commented):
Still not 100% familiar with everything related to tenant VOs so correct me if I'm wrong on anything

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
Johaney-s: Good catches in both comments, hopefully all fixed now. I think we can keep the test for now and adjust/remove/ignore it later as needed.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Johaney-s: I'm not sure, expired members should already have validated attributes, so we'd set the status from EXPIRED to VALID and run the attributes validation check again, but wouldn't the status get changed to VALID after adding the new member vo to memberOrganizations anyway?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -902,14 +929,86 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+		for (Member member : parentVoMembers) {
+			try {
+				Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, member, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+				ArrayList<String> currentValue = attribute.valueAsList();
+				if (currentValue != null && currentValue.contains(memberVo.getShortName())) {
+					currentValue.remove(memberVo.getShortName());
+					attribute.setValue(currentValue);
+					perunBl.getAttributesManagerBl().setAttribute(sess, member, attribute);
+				}
+				if (currentValue != null && currentValue.equals(new ArrayList<>(List.of(memberVo.getShortName())))) {
```
xflord, line 1003:
```suggestion
				if (currentValue != null && currentValue.equals(new ArrayList<>(List.of(vo.getShortName())))) {
```
I think you meant to do this.

cuadradek (commented):
PR #3498 was just merged; hence resolve possible merge conflicts, please.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Johaney-s: What I wanted to achieve here was to check, if before removing the memberVo from memberOrganizations it was the only one in the list. If so, parentVo would overtake to set expiration. If there was another vo in the list, then it should stay intact i think.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -904,6 +931,59 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+						perunBl.getAttributesManagerBl().removeAttribute(sess, newMember, memberExpirationAttrDef);
+						Attribute attribute = new Attribute(perunBl.getAttributesManagerBl().getAttributeDefinition(sess, A_MEMBER_DEF_MEMBER_ORGANIZATIONS));
+						ArrayList<String> newValue = new ArrayList<>(List.of(memberVo.getShortName()));
+						attribute.setValue(newValue);
+						perunBl.getAttributesManagerBl().setAttribute(sess, newMember, attribute);
+					}
+					perunBl.getMembersManagerBl().validateMemberAsync(sess, newMember);
+				}
+
+			} catch (WrongAttributeAssignmentException | WrongReferenceAttributeValueException | AttributeNotExistsException | WrongAttributeValueException | AlreadyMemberException | ExtendMembershipException e) {
```
cuadradek, line 983: I am thinking about `ExtendMembershipException` - maybe we could silently skip these members (e.g. this exception is thrown when they don't have required loa) instead of the whole operation of adding a member organization to fail. Or perhaps creating such members in a different way because they shouldn't follow the parent VO's expiration lifecycle (and this exception is thrown because of the parent VO's rules). Just a thought.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Johaney-s: I get it now, it indeed should check only parent vo stayed in the memberOrganizations. I added empty list as a possibility too.

xflord (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -918,6 +945,60 @@ public boolean usesEmbeddedGroupRegistrations(PerunSession sess, Vo vo) {
...
+						ArrayList<String> newValue = new ArrayList<>(List.of(memberVo.getShortName()));
+						attribute.setValue(newValue);
+						perunBl.getAttributesManagerBl().setAttribute(sess, newMember, attribute);
+					}
+					perunBl.getMembersManagerBl().validateMemberAsync(sess, newMember);
+				}
+
+			} catch (WrongAttributeAssignmentException | WrongReferenceAttributeValueException | AttributeNotExistsException | WrongAttributeValueException | AlreadyMemberException e) {
+				throw new InternalErrorException(e);
+			} catch (ExtendMembershipException ignored) {
```
cuadradek, line 999: We might log this exception, just like in `removeMemberVo`.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -854,6 +859,122 @@ public void getEnrichedVoById() throws Exception {
 		assertThat(enrichedVo.getParentVos()).containsOnly(parent);
 	}
 
+	public void addMemberVoNoExistingMember() throws Exception {
```
cuadradek, line 862: `@Test ` is missing.

cuadradek (commented):
It would probably be a good idea to have a test for a case when we remove a memberVo and the members in a parentVo should start being managed by its expiration lifecycle.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Johaney-s: added

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
Johaney-s: added and also added test `removeMemberVoStartsParentVoLifecycle` to check that overtaking lifecycle by parent vo works

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -854,6 +861,148 @@ public void getEnrichedVoById() throws Exception {
...
+
+		// set expiration rules for parent vo
+		Attribute membershipExpirationRulesAttribute = perun.getAttributesManagerBl().getAttribute(sess, vo, MembersManager.membershipExpirationRulesAttributeName);
+		LinkedHashMap<String, String> expirationRules = new LinkedHashMap<>();
+		expirationRules.put(AbstractMembershipExpirationRulesModule.membershipPeriodKeyName, "01.01.");
+		membershipExpirationRulesAttribute.setValue(expirationRules);
+		perun.getAttributesManagerBl().setAttribute(sess, vo, membershipExpirationRulesAttribute);
+
+		vosManagerEntry.addMemberVo(sess, vo, memberVo);
+		Member memberInParentVo = perun.getMembersManagerBl().getMemberByUserId(sess, vo, member.getUserId());
```
cuadradek, line 999: Just a quick suggestion - we could check here that his expiration attribute is empty (that parentVo's lifecycle starts only after removing memberVo).

Johaney-s: added

cuadradek (approved):

PR #3497, opened 2022-03-15
Registrar list and map widgets

bodnara:
Added support for new application form items - list / map.
Widgets used in old gui are just temporary solution, better visualization can be added later.

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationFormItem.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationFormItem.java
@@ -251,7 +251,24 @@ public class ApplicationFormItem {
 		 * Special type for specifying if will be allowed to register to group(s) through VO application form. This type
 		 * is represented by standard HTML checkbox.
 		 */
-		EMBEDDED_GROUP_APPLICATION
+		EMBEDDED_GROUP_APPLICATION,
+		/**
+		 * Widgets for list like input.
+		 * Represented by string of comma (',') separated values.
+		 * Should also have a comma after last value.
+		 * Special characters as part of value should be escaped.
+		 * Example list: item,item\,withComma,lastItem,
+		 */
+		LIST_INPUT_BOX,
+		/**
+		 * Widgets for map like input (key - value pairs).
+		 * Represented by string of comma (',') separated values.
+		 * Should also have a comma after last value.
+		 * Keys and values are separated by a colon (':').
+		 * Special characters as part of key or value should be escaped.
+		 * Example map: key:value,key\,withComma:value\:withColon,last:item,
+		 */
+		MAP_INPUT_BOX
 	}
 
 	public static class ItemTexts {
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -3825,7 +3825,13 @@ public class RegistrarManagerImpl implements RegistrarManager {
 				// if attribute exists
 				if (a != null) {
 					if (a.getType().equalsIgnoreCase(LinkedHashMap.class.getName())) {
-						// FIXME do not set hash map attributes - not supported in GUI and registrar
+
+						// we expect that map contains string keys and values
+						LinkedHashMap<String, String> value = a.valueAsMap();
+						value = handleMapValue(value, newValue);
+
+						a.setValue(value);
+						attributes.add(a);
 					} else if (a.getType().equalsIgnoreCase(Boolean.class.getName())) {
 						a.setValue(BeansUtils.stringToAttributeValue(newValue, Boolean.class.getName()));
 						attributes.add(a);
@@ -3833,22 +3839,8 @@ public class RegistrarManagerImpl implements RegistrarManager {
 
 						// we expect that list contains strings
 						ArrayList<String> value = a.valueAsList();
+						value = handleArrayValue(value, newValue);
 
-						if (Objects.equals(AttributesManager.NS_USER_ATTR_DEF+":sshPublicKey", a.getName())) {
-							// normalize value for SSH keys
-							value = handleSSHKeysValue(value, newValue);
-						} else {
-
-							if (value == null) {
-								// set as new value
-								value = new ArrayList<>();
-								value.add(newValue);
-							} else if (!value.contains(newValue)) {
-								// add value between old values
-								value.add(newValue);
-							}
-
-						}
 						a.setValue(value);
 						attributes.add(a);
 					} else {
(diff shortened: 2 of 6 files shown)


perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationFormItem.java
```diff
@@ -251,7 +251,15 @@ public void setUpdatable(boolean updatable) {
...
-		EMBEDDED_GROUP_APPLICATION
+		EMBEDDED_GROUP_APPLICATION,
+		/**
+		 * Widgets for array like values.
+		 */
+		LIST_INPUT_BOX,
+		/**
+		 * Widgets for map like values (key - value pairs).
+		 */
+		MAP_INPUT_BOX
```
cuadradek, line 262: Maybe I would specify how values of such items should look like - separated by spaces and colons. And should special characters (`,` or `:`) be escaped so we could actually process such values - e.g., an array with 3 items like this: `item,item\,withComma,anotherItem`?

cuadradek: Sorry, I meant separated by **commas** and colons.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3820,28 +3820,20 @@ private void storeApplicationAttributes(Application app) throws UserNotExistsExc
...
 					if (a.getType().equalsIgnoreCase(LinkedHashMap.class.getName())) {
-						// FIXME do not set hash map attributes - not supported in GUI and registrar
+
+						// we expect that map contains string keys and values
+						LinkedHashMap<String, String> value = a.valueAsMap();
+						value = handleMapValue(value, newValue);
+
+						a.setValue(value);
+						attributes.add(a);
 						continue;
```
zlamalp, line 3830: Not your change, but I think `continue` is not necessary here and in second branch as there is no more code after this whole `if` section.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3891,6 +3883,37 @@ private void storeApplicationAttributes(Application app) throws UserNotExistsExc
...
+	 * Normalize input value from registration form for map values and correctly merge them with the originalValue from User.
+	 *
+	 * @param originalValue Value from Users attribute currently stored in Perun
+	 * @param newValue Value provided by registration form
+	 * @return Map of attribute keys, values after merge
+	 */
+	private LinkedHashMap<String, String> handleMapValue(LinkedHashMap<String, String> originalValue, String newValue) {
+
+		// Normalize value representing map
+		String preparedVal = newValue.replaceAll("(\n)+", ",");
```
zlamalp, line 3896: I would be careful with this first normalization of input value. It was originally done to handle list values from multi-line input fields. After all these changes, we probably shouldn't use such widgets to handle list attributes, then it shouldn't be necessary to normalize this and would actually malform multi-line input.
I'm not saying it shouldn't be there, just we should know all possible use-cases and constraint for list/map attribute values and make changes and safeguards accordingly. It should be all tested thoroughly.

zlamalp (commented):
I would suggest to start implementing WUI part of this so this whole logic can be tested in real-time or implement tests for array/map values processing. I really can't tell, if it will just work.

perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationFormItem.java
bodnara: I have added a better specification based on the `stringToAttributeValue` method in `BaseUtils`.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
bodnara: Redundant continue was removed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
bodnara: I agree that this normalization of data will probably not be needed with the new widgets and so decided to remove it. However as You said, some sort of processing may be needed as a outcome of the implementation and testing of WUI part of this task.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3820,31 +3820,21 @@ private void storeApplicationAttributes(Application app) throws UserNotExistsExc
...
+						LinkedHashMap<String, String> value = a.valueAsMap();
+						value = handleMapValue(value, newValue);
+
+						a.setValue(value);
+						attributes.add(a);
 					} else if (a.getType().equalsIgnoreCase(ArrayList.class.getName())) {
 
 						// we expect that list contains strings
 						ArrayList<String> value = a.valueAsList();
+						value = handleArrayValue(value, newValue);
```
cuadradek, line 3834: I think this method would correctly handle only values ending with `,`.  We should still support current behavior, e.g. `TEXTFIELD` with ArrayList destination attribute - its value doesn't end with `,`. So if it is an ArrayList attribute, we should handle differently `LIST_INPUT_BOX` type (method `handleArrayValue`) and other types (current behavior - don't split value, simply merge it to current value).

bodnara: I have decided to just add a `,`  to values from different widgets and that way they can be processed the same way.

cuadradek (approved):

zlamalp (commented):
I recently did some fixes to ssh keys handling in registrar and I think it should be also reflected in this pull-request, please do check it.

zlamalp (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3866,27 +3858,17 @@ private void storeApplicationAttributes(Application app) throws UserNotExistsExc
...
 	 *
 	 * @param originalValue Value from Users attribute currently stored in Perun
 	 * @param newValue Value provided by registration form
-	 * @return List of SSH keys after merge
+	 * @return List of array attribute values after merge
 	 */
-	private ArrayList<String> handleSSHKeysValue(ArrayList<String> originalValue, String newValue) {
-
-		// blank input means no change to original attribute
-		if (StringUtils.isBlank(newValue)) {
```
zlamalp, line 3878: You must keep this blank check and also use it in handling map values, since GUI send empty items as empty strings and without this later processing will fail, as it will try to store list with empty string as single/first value.

zlamalp (changes requested):

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/CreateFormItemTabItem.java
```diff
@@ -221,7 +221,11 @@ public void onChange(ChangeEvent event) {
 					layout.setHTML(3,0,"Selection box with pre-defined values of UTC timezones.");
 				} else if (type.equals("EMBEDDED_GROUP_APPLICATION")) {
 					layout.setHTML(3,0,"Checkbox of real organization groups which the user can check and thus submit an application to these groups");
-				} else{
+				} else if (type.equals("LIST_INPUT_BOX")) {
```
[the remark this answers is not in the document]
bodnara: New widgets were added to initialization.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
bodnara: Added this check to both methods.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/CreateFormItemTabItem.java
```diff
@@ -79,6 +79,8 @@ public class CreateFormItemTabItem implements TabItem {
 		aMap.put("HEADING", "Header");
 		aMap.put("TIMEZONE", "Selection of timezone");
 		aMap.put("EMBEDDED_GROUP_APPLICATION", "Groups for embedded applications");
+		aMap.put("LIST_INPUT_BOX", "Field used to gather attribute list values");
+		aMap.put("MAP_INPUT_BOX", "Field used to gather attribute value in form of key - value pairs");
```
zlamalp, line 83: I probably wouldn't use the same text for the "name" (here) and "description". So maybe something similar to texts above:
```java
		aMap.put("LIST_INPUT_BOX", "Multiple input text fields");
		aMap.put("MAP_INPUT_BOX", "Multiple key-value input text fields");
```

bodnara: More appropriate name was used.

zlamalp (commented):
I think it is now mostly OK, just don't want to merge it, before we finalize GUI part in registrar, since it might affect how we process the values. Also currently deployed config would stop working, if it gets to the release, so we need to plan it before we are sure.

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/EditFormItemTabItem.java
```diff
@@ -531,6 +531,15 @@ public void onFinished(JavaScriptObject jso) {
 					// sort
 					list = new TableSorter<AttributeDefinition>().sortByFriendlyName(list);
 					for (AttributeDefinition def : list) {
+
+						// allow ArrayList and LinkedHashMap attributes only to designated widgets, unless it was already selected
+						boolean isMapOrListWidget = item.getType().equalsIgnoreCase("MAP_INPUT_WIDGET") || item.getType().equalsIgnoreCase("LIST_INPUT_WIDGET");
```
HejdaJakub, line 536:
```suggestion
						boolean isMapOrListWidget = item.getType().equalsIgnoreCase("MAP_INPUT_BOX") || item.getType().equalsIgnoreCase("LIST_INPUT_BOX");
```

HejdaJakub (changes requested):

perun-web-gui/src/main/java/cz/metacentrum/perun/webgui/tabs/registrartabs/EditFormItemTabItem.java
```diff
@@ -531,6 +531,15 @@ public void onFinished(JavaScriptObject jso) {
...
+						boolean isMapOrListAttribute = def.getType().contains("ArrayList") || def.getType().contains("LinkedHashMap");
+						boolean alreadySelected = item.getPerunSourceAttribute().contains(def.getFriendlyName()) || item.getPerunDestinationAttribute().contains(def.getFriendlyName());
```
zlamalp, line 538: To prevent null pointer when source/dest attribute is not yet set on the form item. Also we should compare whole attr name, not just friendly name.
```suggestion
						boolean alreadySelected = def.getName().equalsIgnoreCase(item.getPerunSourceAttribute()) || def.getName().equalsIgnoreCase(item.getPerunDestinationAttribute());
```

zlamalp (changes requested):

zlamalp (approved):

PR #3498, opened 2022-03-15
feat(core): avoid cyclic memberships

sarkapalkovicova:
Check if VO is not an ancestor before adding to members.


cuadradek (changes requested):

HejdaJakub (commented):
It looks good. I am just not sure if the `ancestor` is the best name for it. Maybe something like `checkParentVos` (similarly to groups) or `checkParenthood` or `checkCyclicMembership` could be better? But it's just my individual opinion so I would be ok also with the current name.

cuadradek (approved):

PR #3500, opened 2022-03-16
feat(core): when adding/removing members to/from member vos, accordingly edit the membersOrganizations attribute and add/remove to/from parent vos

xflord:
merge after https://github.com/CESNET/perun/pull/3494

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -159,6 +159,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 	private static final String NAME = "name";
 	private static final String GROUP_ADDING_ERRORS = "group_adding_errors";
 	private static final String MEMBER = "member";
+	private final static String A_MEMBER_DEF_MEMBER_ORGANIZATIONS = AttributesManager.NS_MEMBER_ATTR_DEF + ":memberOrganizations";
 
 	public static final List<String> SPONSORED_MEMBER_REQUIRED_FIELDS = Arrays.asList(
 			"firstname",
@@ -236,6 +237,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 		membersManagerImpl.rejectAllMemberOpenApplications(sess, member);
 
 		// Remove member from the DB
+		removeMemberFromParentVos(sess, member);
 		getMembersManagerImpl().deleteMember(sess, member);
 		getPerunBl().getAuditer().log(sess, new MemberDeleted(member));
 	}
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1510,6 +1513,82 @@ public Member setStatus(PerunSession sess, Member member, Status status) throws
 		}
 	}
 
+	private void removeMemberFromParentVos(PerunSession sess, Member member ) {
+		User user = getPerunBl().getUsersManagerBl().getUserByMember(sess, member);
+		try {
+			Vo memberVo = getPerunBl().getVosManager().getVoById(sess, member.getVoId());
+			List<Vo> parentVos = getPerunBl().getVosManager().getParentVos(sess, member.getVoId());
```
cuadradek, line 1520: Here in the BL layer, we should usually call BL methods instead of entry layer methods because it is unnecessary to check privileges and other things the entry layer checks. Then, you can remove PrivilegeException from imports.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1579,6 +1660,7 @@ public Member expireMember(PerunSession sess, Member member) throws WrongReferen
 		Status oldStatus = member.getStatus();
 		getMembersManagerImpl().setStatus(sess, member, Status.EXPIRED);
 		member.setStatus(Status.EXPIRED);
+		removeMemberFromParentVos(sess, member);
```
cuadradek, line 1663: Requirements for expired members changed a bit - if a member is expired in a memberVo, he should still be in its parentVos but managed by their lifecycle rules (so we should only remove the memberVo from memberOrganizations attribute and call `extendMembership` method to start managing him by a parentVos expiration lifecycle).

cuadradek (changes requested):
Looks good so far.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
cuadradek: Sorry, I just realized this is how you actually implemented :D

cuadradek: When I think about it again, if his status changed from INVALID/DISABLED to EXPIRED, he might not exist in parentVos, so we should probably create him there (use a little changed `addMemberToParentVos` instead of `removeMemberFromParentVos`) - what do you think?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1510,6 +1512,96 @@ public Member setStatus(PerunSession sess, Member member, Status status) throws
...
+						Member existingMember = perunBl.getMembersManagerBl().getMemberByUser(sess, vo, user);
+						// validate and reset expiration
+						perunBl.getAttributesManagerBl().removeAttribute(sess, existingMember, memberExpirationAttrDef);
+						if (existingMember.getStatus() != Status.VALID) {
+							perunBl.getMembersManagerBl().validateMemberAsync(sess, existingMember);
+						}
+						Attribute attribute = perunBl.getAttributesManagerBl().getAttribute(sess, existingMember, A_MEMBER_DEF_MEMBER_ORGANIZATIONS);
+						ArrayList<String> currentValue = (ArrayList<String>) attribute.getValue();
+						if (currentValue == null) {
+							throw new ConsistencyErrorException("Member " + existingMember + " of hierarchical vo " + vo + " has no memberOrganizations attribute set.");
```
cuadradek, line 1590: Same here.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1510,6 +1513,82 @@ public Member setStatus(PerunSession sess, Member member, Status status) throws
...
+					currentValue.remove(memberVo.getShortName());
+					// if user no longer in any memberVos, follow parentVo lifecycle
+					if (currentValue.isEmpty() || (currentValue.equals(new ArrayList<>(List.of(vo.getShortName()))))) {
+						this.getPerunBl().getMembersManagerBl().extendMembership(sess, existingMember);
+					}
+					attribute.setValue(currentValue);
+					perunBl.getAttributesManagerBl().setAttribute(sess, existingMember, attribute);
+				} catch (WrongAttributeValueException | WrongAttributeAssignmentException | AttributeNotExistsException | MemberNotExistsException | WrongReferenceAttributeValueException e) {
+					throw new InternalErrorException(e);
+				} catch (ExtendMembershipException ignored) {
```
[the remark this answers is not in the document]
xflord: It seems to me that there are tiny differences amongst the common parts and even then it's still just a couple of lines of code + across different classes. Kinda think its not necessary to write new private classes to be used in existing private classes, however if you really think it's going to make things easier in the future, I'll see what I can do.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
xflord: Yes, I see what you mean. Wrote a small if statement.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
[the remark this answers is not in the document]
cuadradek: I didn't realize the methods are in different classes and with tiny differences, so we don't have to extract them.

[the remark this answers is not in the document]
cuadradek: I didn't realize the methods are in different classes and with tiny differences, so we don't have to extract them.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -3037,6 +3037,92 @@ public void getAllMembers_twoMembersFromOneVo() throws Exception {
...
+
+		Member member = setUpMember(memberVo);
+
+		perun.getVosManagerBl().addMemberVo(sess, parentVo, memberVo);
+
+		assertThat(perun.getMembersManagerBl().getMembers(sess, parentVo).size()).isEqualTo(1);
+
+		perun.getMembersManagerBl().disableMember(sess, member);
+
+		assertThat(perun.getMembersManagerBl().getMembers(sess, parentVo).size()).isEqualTo(0);
```
cuadradek, line 3100: I think this should be 1 because the member in the parentVo shouldn't be completely deleted, but he should start its lifecycle instead. That's why this test fails. You should probably disable the member in the memberVo and then create a relationship between memberVo and parentVo, then the size should be 0, then validate the member, then the size should be 1.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
```diff
@@ -3037,6 +3037,92 @@ public void getAllMembers_twoMembersFromOneVo() throws Exception {
...
+
+		perun.getMembersManagerBl().validateMemberAsync(sess, member);
+
+		assertThat(perun.getMembersManagerBl().getMembers(sess, parentVo).size()).isEqualTo(1);
+
+	}
+
+	@Test
+	public void removeMemberFromMemberVoAndParentVo() throws Exception {
```
cuadradek, line 3109: Same problem here.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -1579,6 +1679,25 @@ public Member expireMember(PerunSession sess, Member member) throws WrongReferen
...
 		member.setStatus(Status.EXPIRED);
+		// if member went from invalid/disabled to expired, they might not exist in some parentVos - create them, but don't
+		// update MemberOrganizations, etc...
+		if (oldStatus == Status.INVALID || oldStatus == Status.DISABLED) {
+			try {
+				User user = getPerunBl().getUsersManagerBl().getUserByMember(sess, member);
+				List<Vo> parentVos = getPerunBl().getVosManagerBl().getParentVos(sess, member.getVoId());
+
+				for (Vo vo : parentVos) {
+					Member newMember = perunBl.getMembersManagerBl().createMember(sess, vo, user);
```
Johaney-s, line 1690: I think the try-catch should go inside the forcycle, if the member exists in parentVo already, we still want to check the other vos.

xflord: Big oversight on my part, thank you.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/MembersManagerEntryIntegrationTest.java
xflord: Yeah, my bad. Wrote the tests waay before we cleared up how the lifecycle should work and then forgot to change them.

Johaney-s (approved):

cuadradek (approved):

PR #3503, opened 2022-03-17
feat(openapi): added method

sarkapalkovicova:
Added getUserExtSourceByExtLoginAndExtSourceName to OpenApi specification.

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -8480,6 +8480,21 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/usersManager/getUserExtSourceByExtLoginAndExtSourceName:
+    get:
+      tags:
+        - UsersManager
+      operationId: getUserExtSourceByExtLoginAndExtSourceName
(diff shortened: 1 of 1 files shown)


perun-openapi/openapi.yml
```diff
@@ -8480,6 +8480,21 @@ paths:
...
 
+  /json/usersManager/getUserExtSourceByExtLoginAndExtSourceName:
+    get:
+      tags:
+        - UsersManager
+      operationId: getUserExtSourceByExtLoginAndExtSourceName
+      summary: Returns user's external source by the user's external login and external source name.
+      parameters:
+        - $ref: '#/components/parameters/extSourceName'
+        - $ref: '#/components/parameters/extLogin'
```
Johaney-s, line 8491: The definition of extLogin in openApi uses name `extLogin`, but the method in RPC requires parameter called `extSourceLogin`, which would not match. I think you can write inline definition for the parameter instead of the used reference.

Johaney-s (changes requested):

Johaney-s (approved):

PR #3504, opened 2022-03-17
docs(rpc): Documentation for CSRF protection added

Kuliak:
* documentation for CSRF protection added to RPC API documentation at https://perun-aai.org/documentation/technical-documentation/rpc-api/index.html
* CSRF protection error refers to CSRF protection section in RPC API documentation

### DIFF
--- a/perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl
+++ b/perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl
@@ -633,6 +633,16 @@ Response: call1(response);
 
 		<p>When using JSONP, returned objects are stripped of non relevant properties like <em>createdAt</em>, <em>createdBy</em>, <em>modifiedAt</em>, <em>modifiedBy</em> etc. You can get them when using standard JSON.</p>
 
+		<h3>CSRF Protection</h3>
+
+		<p>Domains with both GUI and API deployed are protected against CSRF attacks as a whole. Even only-API clients must implement this protection or use different domain reserved for the API calls (example below). Perun will generate a CSRF token during the first API call. The token value is returned to the client in a cookie.The client must read the value from the cookie and return it to the API with each call in the appropriate HTTP header along with the cookie itself. As a result, the token value must be the same in the cookie, HTTP header and server-side session. CSRF protection is enabled only for state-changing requests (POST, PUT). Also, CSRF is not used in OIDC, since it has its own CSRF protection using own access tokens.</p>
+
+<pre>GUI: perun.&ltdomain&gt.cz
+API: perun-api.&ltdomain&gt.cz
+</pre>
+
+		<p>Presence of CSRF token is checked at first by looking at value of Cookie ('XSRF-TOKEN') and HTTP Session. If either of them is missing, new token is generated and stored into both original Session and Cookie that is stored in the response. Token value must be same for all Cookie, Header ('X-XSRF-TOKEN') and server side Session, in case it is not, corresponding error message is sent in response.</p>
+
 	</div>
 </div>
 </div>
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/csrf/CsrfFilter.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/csrf/CsrfFilter.java
@@ -118,7 +118,8 @@ public final class CsrfFilter implements Filter {
 				// token is either invalid or missing
 				response.sendError(HttpStatus.SC_FORBIDDEN, "This domain is under CSRF protection; however, " +
 				    "a valid CSRF token was not provided. Configure the client to provide the token in header " +
-					"or use the domain reserved only for API calls - without CSRF protection");
+					"or use the domain reserved only for API calls - without CSRF protection. Information about " +
+					"CSRF protection can also be found at https://perun-aai.org/documentation/technical-documentation/rpc-api/index.html");
 				return;
 			}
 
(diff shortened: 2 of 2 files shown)


zlamalp (changes requested):
You should start this whole section with more generic message to the reader and not detailed specification, what types or request are protected. That sentence is currently at the end of the paragraph.

I would modify that sentence:

`Generally, domain with both GUI and API should be protected against CSRF as a whole, thus API clients`

like this 

`Domains with both GUI and API deployed are protected against CSRF attacks as a whole. Even only-API clients must implement this protection or use different domain reserved for the API calls.`.

In you example of the domain, you can't use `<domain>` as it would be interpreted as HTML tag. If you wish to use it, please do HTML escaping. Also you might want to use some formatting (&lt;b&gt;, &lt;i&gt; or &lt;pre&gt;) for "examples" as it is done in the rest of document.

I wouldn't go into the depths of implementation when describing the checking mechanism as it is not beneficial to the reader, that check is done by servlet api filter implementation.

From:
`Perun uses specialized tokens generated in the first request that have to be read from Cookie, put into HTTP Request Header and checked by implemented filter.`

To:
`Perun will generate a CSRF token during the first API call. The token value is returned to the client in a cookie.The client must read the value from the cookie and return it to the API with each call in the appropriate HTTP header along with the cookie itself. As a result, the token value must be the same in the cookie, HTTP header and server-side session.`

I guess the above sentence sort of describes what is in more details in the paragraph below. So I will let it up to you if you wish to rephrase it or add details (like cookie/header name) in it.

Key information for the reader is:

* When is the token generated (at first call, can change anytime, so the value must be read from cookie before each subsequent call).
* Where do I get it and where do I put it (cookie, cookie/header names)
* Protection principle (values on all places must match)
* What kind of requests are protected (data changing operations - POST/PUT)
* What kind of requests are not protected (GET or any type of request using OIDC/oauth)
* Pointers to API only domains without protection, if client update would be to difficult.

perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl
```diff
@@ -633,6 +633,16 @@ sub buildVersion {
 
 		<p>When using JSONP, returned objects are stripped of non relevant properties like <em>createdAt</em>, <em>createdBy</em>, <em>modifiedAt</em>, <em>modifiedBy</em> etc. You can get them when using standard JSON.</p>
 
+		<h3>CSRF Protection</h3>
+
+		<p>CSRF protection is enabled only for state-changing requests (POST, PUT). Also, CSRF is not used in OIDC, since it has its own CSRF protection using own access tokens. Perun will generate a CSRF token during the first API call. The token value is returned to the client in a cookie.The client must read the value from the cookie and return it to the API with each call in the appropriate HTTP header along with the cookie itself. As a result, the token value must be the same in the cookie, HTTP header and server-side session. Domains with both GUI and API deployed are protected against CSRF attacks as a whole. Even only-API clients must implement this protection or use different domain reserved for the API calls, e.g. for:.</p>
+
+		<pre> GUI: perun.&ltdomain&gt.cz
```
zlamalp, line 640: Alignment in the used &lt;pre&gt;. Please always check the script output.
obrazek

perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl
```diff
@@ -633,6 +633,16 @@ sub buildVersion {
...
+		API: perun-api.&ltdomain&gt.cz
+        </pre>
+
+		<p>More precisely, presence of CSRF token is checked at first by looking at value of Cookie ('XSRF-TOKEN') and HTTP Session. If either of them is missing, new token is generated and stored into both original Session and Cookie that is stored in the response. Token value must be same for all Cookie, Header ('X-XSRF-TOKEN') and server side Session, in case it is not, corresponding error message is sent in response. After CSRF token check is completed, following filter (from filter chain) is applied.</p>
```
zlamalp, line 644: I think you can remove `More precisely,` and just start with the rest of the sentence.

Also drop `After CSRF token check is completed, following filter (from filter chain) is applied.` at the end of paragraph. This information is not relevant to the API client at all.

perun-utils/rpc-methods-javadoc-generator/parseRpcMethods.pl
```diff
@@ -633,6 +633,16 @@ sub buildVersion {
 
 		<p>When using JSONP, returned objects are stripped of non relevant properties like <em>createdAt</em>, <em>createdBy</em>, <em>modifiedAt</em>, <em>modifiedBy</em> etc. You can get them when using standard JSON.</p>
 
+		<h3>CSRF Protection</h3>
+
+		<p>CSRF protection is enabled only for state-changing requests (POST, PUT). Also, CSRF is not used in OIDC, since it has its own CSRF protection using own access tokens. Perun will generate a CSRF token during the first API call. The token value is returned to the client in a cookie.The client must read the value from the cookie and return it to the API with each call in the appropriate HTTP header along with the cookie itself. As a result, the token value must be the same in the cookie, HTTP header and server-side session. Domains with both GUI and API deployed are protected against CSRF attacks as a whole. Even only-API clients must implement this protection or use different domain reserved for the API calls, e.g. for:.</p>
```
zlamalp, line 638: As i said, I would start with the most generic info (about protecting domains), then describe its principle and then specified, that only state-changing requests are protected and protection is not used with OIDC. Basically take this paragraph, split it into three parts and revers the order. I think it will make more sense like this.

zlamalp (changes requested):

zlamalp (approved):

PR #3507, opened 2022-03-18
fix(core): make parameter attrNames optional

mvocu:
make parameter attrNames of method getAllRichGroupsWithAttributesByName() optional, as is in BlImpl layer.

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
@@ -1481,6 +1481,18 @@ public enum GroupsManagerMethod implements ManagerMethod {
 	 * @return List<RichGroup> RichGroups containing selected attributes
 	 */
 	/*#
+	 * Returns sub-list of all RichGroups, each containing all attributes, starting at fromIndex (included)
+	 * and ending at the size of the original list.
+	 *
+	 * Example: [1,2,3,4], fromIndex=1 => [2,3,4]
+	 *
+	 * @throw VoNotExistsException When Vo doesn't exist
+	 *
+	 * @param vo int <code>id</code> of vo
+	 * @param fromIndex int begin index of returned subList, included
+	 * @return List<RichGroup> RichGroups containing all attributes
+	 */
+	/*#
 	 * Returns sub-list of all RichGroups, each containing selected attributes, starting at first index of the original
 	 * list (included) and ending at the toIndex (included).
 	 *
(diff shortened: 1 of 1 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1523,7 +1523,7 @@ public Vo call(ApiCaller ac, Deserializer parms) throws PerunException {
 
 			List<RichGroup> listOfRichGroups = ac.getGroupsManager().getAllRichGroupsWithAttributesByNames(ac.getSession(),
 				ac.getVoById(parms.readInt("vo")),
-				parms.readList("attrNames", String.class));
+				parms.contains("attrNames") ? parms.readList("attrNames", String.class) : null);
```
zlamalp, line 1526: By the method docs above, parameter `attrNames` is required and I think that is consistent with the other relevant methods. This commit changes this requirement without updating the docs.
Until now it worked the way, that it distinguish value of parameter - null, empty list, non-empty list. Now you distinguish the presence of it.
I'm not sure what is reasoning behind this change, is there a problem with corresponding OpenAPI definition for the clients? I believe this method is heavily used by de.NBI at Elixir instance, so I'm just cautious about the changes to it.

mvocu: When this method is called via openapi with attrNames == null, the result is 
`org.springframework.web.client.HttpClientErrorException: 400 Missing the required parameter 'attrNames' when calling getAllRichGroupsWithAttributesByNames
`
I guess the serializer omits the null parameter altogether. 

The change does not alter the behaviour for current clients, it just makes the attrNames parameter optional. I do not understand how the current implementation could have worked when called with "null" attrNames list? Please note that UrlDeserializer does not return "null" from readList method, ever (either possibly empty list or exception).

zlamalp: You are right, that OpenAPI client probably works like that. I think @martin-kuba wanted to fix this (or something similar) with #3447. Optional presence of query param is handled differently when `JsonDeserializer` is used (especially when it is list).
So generally I'm not against the change, just make sure its backwards compatible and update javadoc as "optional" param presence means doubling each method version docs (with and without the param).

mvocu: I think that making parameter optional does not have the potential to break compatibility with existing code. Moreover it is not my intention to enable clients to omit the parameter willingly (I did not mark the parameter as optional in openapi either), just to handle the situation when null value passed as an argument causes the parameter to be skipped and server throwing exception subsequently. 

I have updated the javadoc as requested.

zlamalp: You are probably right. Just please fix `selected` -> `all` in all `return` parts of the new docs, otherwise I think its ok.

mvocu:
Done.

zlamalp (changes requested):
I think `FIX(core)` in the commit message should be lowercase `fix(core)`, as suggested by @cuadradek. Otherwise its OK.

mvocu:
Hmmmm.... done.

zlamalp (approved):

PR #3508, opened 2022-03-18
feat: add support for consent

HejdaJakub:
* New object ConsentDefinition was created and added to parseRpcMethod.
* Five new tables were created to support consent.
* New object ConsentDefinition was added to openapi.yml.

### DIFF
--- a/perun-base/src/test/resources/test-schema.sql
+++ b/perun-base/src/test/resources/test-schema.sql
@@ -1,4 +1,4 @@
--- database version 3.1.90 (don't forget to update insert statement at the end of file)
+-- database version 3.1.91 (don't forget to update insert statement at the end of file)
 CREATE EXTENSION IF NOT EXISTS "unaccent";
 CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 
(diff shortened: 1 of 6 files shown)


perun-db/postgres.sql
```diff
@@ -1550,6 +1550,58 @@ create table groups_to_register (
     constraint grpreg_group_fk foreign key (group_id) references groups(id) on delete cascade
 );
 
+-- CONSENTS_DEFINITIONS - consents definitions for facilities, services and attributes
+create table consents_definitions (
+	id integer not null,
+	name varchar not null,
+	decription varchar,
```
Johaney-s, line 1557:
```suggestion
	description varchar,
```

perun-db/postgres.sql
```diff
@@ -1550,6 +1550,58 @@ create table groups_to_register (
...
+	created_at timestamp default statement_timestamp() not null,
+	created_by varchar default user not null,
+	modified_at timestamp default statement_timestamp() not null,
+	modified_by varchar default user not null,
+	created_by_uid integer,
+	modified_by_uid integer,
+	constraint con_def_pk primary key(id)
```
Johaney-s, line 1564: I think we should also add constraint for the name to be unique, what do you think?

perun-db/postgres.sql
```diff
@@ -1550,6 +1550,58 @@ create table groups_to_register (
...
+  	constraint facility_consent_cons_def_fk foreign key (consent_def_id) references consents_definitions(id)
+);
+
+-- SERVICE_CONSENT - relation between services and consent definitions
+create table service_consent (
+	service_id integer not null,       --identifier of service
+	consent_def_id integer not null,   --identifier of consent definition
+	constraint service_consent_pk primary key (service_id,consent_def_id),
+  	constraint service_consent_service_fk foreignkey(service_id) references services(id),
+  	constraint service_consent_cons_def_fk foreign ke y (consent_def_id) references consents_definitions(id)
```
Johaney-s, line 1594:
```suggestion
  	constraint service_consent_cons_def_fk foreign key (consent_def_id) references consents_definitions(id)
```

perun-db/postgres.sql
```diff
@@ -1550,6 +1550,58 @@ create table groups_to_register (
...
+  	constraint facility_consent_fac_fk foreign key (facility_id) references facilities(id),
+  	constraint facility_consent_cons_def_fk foreign key (consent_def_id) references consents_definitions(id)
+);
+
+-- SERVICE_CONSENT - relation between services and consent definitions
+create table service_consent (
+	service_id integer not null,       --identifier of service
+	consent_def_id integer not null,   --identifier of consent definition
+	constraint service_consent_pk primary key (service_id,consent_def_id),
+  	constraint service_consent_service_fk foreignkey(service_id) references services(id),
```
Johaney-s, line 1593:
```suggestion
  	constraint service_consent_service_fk foreign key(service_id) references services(id),
```

perun-openapi/openapi.yml
```diff
@@ -1385,6 +1385,19 @@ components:
         email: { type: string }
         identities: { type: array, items: { $ref: '#/components/schemas/ExtSource' } }
 
+    ConsentDefinition:
+      properties:
+        name: { type: string }
+        desciption: { type: string }
```
Johaney-s, line 1391:
```suggestion
        description: { type: string }
```

Johaney-s (changes requested):
Looks good, just a few suggestions. Also, I think the database changes should be also reflected in `test-schema.sql` and `postgresChangelog.txt`.

perun-db/postgres.sql
```diff
@@ -1550,6 +1550,58 @@ create table groups_to_register (
...
+	status varchar not null,
+	createdDate timestamp not null
+	constraint consents_pk primary key (user_id,consent_def_id),
+  	constraint consents_user_fk foreign key (user_id) references users(id),
+  	constraint consent_cons_def_fk foreign key (consent_def_id) references consents_definitions(id),
+  	constraint consents_stat_chk check (status in ('GRANTED','REVOKED','EXPIRED'))
+);
+
+-- FACILITY_CONSENT - relation between facilities and consent definitions
+create table facility_consent (
```
cuadradek, line 1580: I think this would be a little more accurate, what do you think? Same for other tables.
```suggestion
create table facilities_consent_definitions (
```

cuadradek (changes requested):

perun-db/postgres.sql
HejdaJakub: You're right, I edited the names of these tables.

cuadradek (commented):
I think we could create DB indices on each foreign key column which should speedup joining tables in queries. Also, I am not sure, but maybe we could add auditing columns (created_at, created_by, ...) to all tables. What do you think, @HejdaJakub @Johaney-s ?

HejdaJakub:
Latest changes:
 - added auditing columns to `posgres.sql`,` test-schema.sql`, and `postgresChangelog.txt`
 - added indexes on all foreign keys
 - hash code changed
 - changed the `description` to `text` and made it required

perun-base/src/test/resources/test-schema.sql
```diff
@@ -1552,6 +1552,84 @@ create table groups_to_register (
...
+	modified_at timestamp default statement_timestamp() not null,
+	modified_by varchar default user not null,
+	created_by_uid integer,
+	modified_by_uid integer,
+	constraint con_def_pk primary key(id),
+	constraint con_def_u unique(name)
+);
+
+-- CONSENTS - relation between users and consent definitions
+create table consents (
```
cuadradek, line 1571: One last thing - we might want to have a column for ID in this table too. I am not sure if it is necessary, but I assume the Consent object will extend the PerunBean object, which has an ID attribute. What do you think?

perun-base/src/test/resources/test-schema.sql
```diff
@@ -1552,6 +1552,84 @@ create table groups_to_register (
...
+	user_id integer not null,		   --identifier of user
+	consent_def_id integer not null,   --identifier of consent definition
+	status varchar not null,
+	createdDate timestamp not null,
```
cuadradek, line 1575: Now that we have auditing columns, this column seems redundant at this moment.

HejdaJakub:
The architecture of consents has been changed, so the new class and basic objects are implemented in https://github.com/CESNET/perun/pull/3517.

PR #3510, opened 2022-03-21
feat(openapi): added methods to OpenApi

mattjoke:
* added methods to VosManagerMethod
* added methods to OpenApi specification

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -2692,6 +2692,14 @@ components:
       in: query
       required: true
 
+    memberVoId:
+      name: memberVo
+      description: "id of member Vo"
+      schema:
+        type: integer
+      in: query
+      required: true
+
     banId:
       name: banId
       description: "id of a ban"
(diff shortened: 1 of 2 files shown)


perun-openapi/openapi.yml
```diff
@@ -14231,6 +14231,63 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /urlinjsonout/vosManager/addMemberVo:
+    post:
+      tags:
+        - VosManager
+      operationId: addMemberVo
+      summary: Add member to vo.
```
bodnara, line 14239: What this method does is that it adds a `Vo` as a member of another `Vo`, rather than adding a `Member` as a "person".

perun-openapi/openapi.yml
```diff
@@ -14231,6 +14231,63 @@ paths:
...
+      parameters:
+        - $ref: '#/components/parameters/voId'
+        - $ref: '#/components/parameters/memberId'
```
bodnara, line 14242: `'#/components/parameters/voId'`

perun-openapi/openapi.yml
```diff
@@ -14231,6 +14231,63 @@ paths:
...
+          $ref: '#/components/responses/VoidResponse'
+        default:
+          $ref: '#/components/responses/ExceptionResponse'
+
+  /urlinjsonout/vosManager/removeMemberVo:
+    post:
+      tags:
+        - VosManager
+      operationId: removeMemberVo
+      summary: Remove member from vo.
```
bodnara, line 14254: Same here, we are removing `member Vo` from `Vo`

perun-openapi/openapi.yml
```diff
@@ -14231,6 +14231,63 @@ paths:
...
+      parameters:
+        - $ref: '#/components/parameters/voId'
+        - $ref: '#/components/parameters/memberId'
```
bodnara, line 14257: `'#/components/parameters/voId'`

perun-openapi/openapi.yml
```diff
@@ -14231,6 +14231,63 @@ paths:
...
+          $ref: '#/components/responses/ExceptionResponse'
+
+  /json/vosManager/getParentVos:
+    get:
+      tags:
+        - VosManager
+      operationId: getParentVos
+      summary: Get list of all parent organizations of a vo.
+      parameters:
+        - $ref: '#/components/parameters/memberId'
```
bodnara, line 14285: `'#/components/parameters/voId'`

bodnara (changes requested):
All of these methods are connected to the concept of hierarchical organizations and their member organizations. So in this context adding "member" is refering to object `Vo` and not `Member`. Which I must admit, can be misleading.

bodnara:
One suggestion regarding commit messages. Try to be more specific. It should be somewhat clear what was actually added.

bodnara (approved):
Seems ok like this.
I would suggest if you need some changes that are already merged to master you can just rebase your branch.

xflord (commented):
I think we ideally want 1 commit per scope that's being changed (I believe you had it right on the first push, then it got a bit messy). Other than that the code seems fine.

HejdaJakub (approved):

PR #3511, opened 2022-03-22
feat(openapi): added methods from Searcher

sarkapalkovicova:
Added methods from Searcher to OpenApi specification.

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -3220,6 +3220,10 @@ tags:
     description: IntegrationManager RPC API in Perun
     externalDocs:
       url: https://perun-aai.org/documentation/technical-documentation/rpc-api/rpc-javadoc-IntegrationManager.html
+  - name: Searcher
+    description: Searcher RPC API in Perun
+    externalDocs:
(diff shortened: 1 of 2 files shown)


bodnara (commented):
Inside the method `getMembersByUserAttribues` in `SearcherMethod.java` there is a call to read param. `vo`. Which I assume is correct, however it is missing from the method doc. Please look into this and if it is indeed just missing add info about this param. to method doc and update openapi specification accordingly.

bodnara (approved):

perun-openapi/openapi.yml
```diff
@@ -16652,3 +16656,159 @@ paths:
...
+                  additionalProperties:
+                    type: string
+
+  /json/Searcher/getResources/attributes-match:
+    post:
+      tags:
+        - Searcher
+      operationId: getAttributesMatchResources
+      summary: Get list of resources that have attributes with partially matched values
+      if allowPartialMatchForString is set to true, else with exactly matched values.
```
Johaney-s, line 16790: So that it is parsed correctly, the summary should be on one line.

Johaney-s (approved):

PR #3515, opened 2022-03-28
feat: add manager classes handling consents logic

cuadradek:
feat(core): add manager classes handling consents logic

- New manager classes handling consents logic were added to every layer.
They contain no methods so far.

feat(openapi): add new manager handling consents logic

- New manager handling consents logic were added, it contains no
methods so far.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/PerunBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/PerunBl.java
@@ -2,6 +2,8 @@ package cz.metacentrum.perun.core.bl;
 
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.AuditMessagesManager;
+import cz.metacentrum.perun.core.api.ConfigManager;
+import cz.metacentrum.perun.core.api.ConsentsManager;
 import cz.metacentrum.perun.core.api.DatabaseManager;
 import cz.metacentrum.perun.core.api.ExtSourcesManager;
 import cz.metacentrum.perun.core.api.FacilitiesManager;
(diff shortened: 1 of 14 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/PerunBl.java
```diff
@@ -256,4 +256,10 @@
 	 * @return ConfigManagerBl
 	 */
 	ConfigManagerBl getConfigManagerBl();
+
+	/**
+	 * Gets a ConsentsManagerBl
+	 * @return ConsentsManagerBl
+	 */
+	ConsentsManagerBl getConsentsManagerBl();
```
Johaney-s, line 264: I think getConsentsManager() method should be here too, same as other managers have it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/PerunBlImpl.java
```diff
@@ -115,7 +118,7 @@
 	private SearcherBl searcherBl = null;
 	private TasksManagerBl tasksManagerBl = null;
 	private ConfigManagerBl configManagerBl = null;
-
+	private ConsentsManagerBl consentsManagerBl = null;
```
Johaney-s, line 121: should be probably added to toString() method also

cuadradek: Done, I added missing ConfigManager as well.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/PerunBl.java
cuadradek: Done, I added missing getConfigManager() as well.

xflord (approved):

PR #3517, opened 2022-03-28
feat: add new class ConsentHub

cuadradek:
feat(core): add new class ConsentHub

- New class ConsentHub was created, it represents group of facilities
with joint consent management.
- New DB tables for consent hubs were added as well.

BREAKING CHANGE: update DB

feat(openapi): add new schema ConsentHub

- New schema ConsentHub was added, it represents group of facilities
with joint consent management.

### DIFF
--- a/perun-base/src/test/resources/test-schema.sql
+++ b/perun-base/src/test/resources/test-schema.sql
@@ -1,4 +1,4 @@
--- database version 3.1.90 (don't forget to update insert statement at the end of file)
+-- database version 3.1.91 (don't forget to update insert statement at the end of file)
 CREATE EXTENSION IF NOT EXISTS "unaccent";
 CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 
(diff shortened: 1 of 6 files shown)


perun-openapi/openapi.yml
```diff
@@ -520,6 +520,16 @@ components:
...
+    ConsentHub:
+      allOf:
+        - $ref: '#/components/schemas/Auditable'
+        - properties:
+            name: { type: string }
+            enforceConsents: { type: boolean }
+            facilities: { type: array, items: { $ref: '#/components/schemas/Facility' } }
+      discriminator:
+        propertyName: beanName
+
```
HejdaJakub, line 532: Maybe there should be some of these properties required? I think that `name` and `facilities` should be required, `enforceConsents` has a default value, so it can be optional.

HejdaJakub (changes requested):

xflord (commented):
The task states that we need a new 'Hub' object and 'ConsentHub' should be an extension of that. Did we drop that idea? Otherwise looks good.

perun-base/src/test/resources/test-schema.sql
```diff
@@ -1,4 +1,4 @@
--- database version 3.1.90 (don't forget to update insert statement at the end of file)
+-- database version 3.1.91 (don't forget to update insert statement at the end of file)
```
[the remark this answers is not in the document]
cuadradek: Yeah, I thought about that too, but we actually don't deal with permissions in the test schema.

cuadradek:
> The task states that we need a new 'Hub' object and 'ConsentHub' should be an extension of that. Did we drop that idea? Otherwise looks good.

Yes, I decided to drop that idea for now because it would be a little confusing and unnecessary, in my opinion. If we need it in the future, we might change it then.

perun-openapi/openapi.yml
cuadradek: Good point; I made them all required.

HejdaJakub (approved):

PR #3524, opened 2022-03-29
feat(core): methods to create/delete/find consentHubs

Johaney-s:
* Added get methods to all layers
* Added new test file for ConsentsManager, added tests
* Added new get methods to openapi.yml
* Consent hub is created automatically when new facility is created, default name is same as for facility
* Removing consentHub with facility if it is the only facility of the hub

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
@@ -1,10 +1,48 @@
 package cz.metacentrum.perun.core.api;
 
+import cz.metacentrum.perun.core.api.exceptions.ConsentHubNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
+
+import java.util.List;
+
 /**
  * Consents entry logic.
  *
  * @author Radoslav Čerhák <r.cerhak@gmail.com>
  */
 public interface ConsentsManager {
+
+	/**
+	 * Get list of all Consent Hubs
+	 *
+	 * @param sess perun session
+	 * @return list of Consent Hubs
+	 * @throws PrivilegeException
+	 * @throws InternalErrorException
+	 */
+	List<ConsentHub> getAllConsentHubs(PerunSession sess) throws PrivilegeException;
+
+	/**
+	 * Finds existing Consent Hub by id.
+	 *
+	 * @param sess perun session
+	 * @param id id of the Consent Hub you are looking for
+	 * @return found Consent Hub
+	 * @throws ConsentHubNotExistsException
+	 * @throws InternalErrorException
+	 */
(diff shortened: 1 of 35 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
```diff
@@ -1,10 +1,62 @@
...
+	 *
+	 * @param sess perun session
+	 * @param facilityId id of facility for which consent hub is searched
+	 * @return found Consent Hub
+	 * @throws ConsentHubNotExistsException
+	 * @throws InternalErrorException
+	 * @throws PrivilegeException
+	 * @throws FacilityNotExistsException
+	 */
+	ConsentHub getConsentHubByFacility(PerunSession sess, int facilityId) throws ConsentHubNotExistsException, PrivilegeException, FacilityNotExistsException;
```
xflord, line 60: The task states that in the **future** 1 facility will be a part of max 1 consent hub, are we implementing it now? If not, then this could potentially return multiple CHs, no? I also agree that this should be implemented now, but just make sure that everybody is aware. (e.g. https://github.com/CESNET/perun/pull/3528 doesn't deal with this), maybe update associated tasks?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -359,6 +365,20 @@ public void deleteFacility(PerunSession sess, Facility facility, Boolean force)
 			}
 		}
 
+		//remove consent hub with consents
+		try {
+			ConsentHub hub = getPerunBl().getConsentsManagerBl().getConsentHubByFacility(sess, facility.getId());
```
xflord, line 370: This would need to change if multiple CHs per facility are a possibility

perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
Johaney-s: Should be solved in the PR you mentioned. We want 1 hub/facility now.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -13,10 +25,156 @@
...
+
+	@Override
+	public void deleteConsentHub(PerunSession perunSession, ConsentHub consentHub) throws ConsentHubAlreadyRemovedException {
+		//TODO: remove all user consents first
+
+		try {
+			jdbc.update("delete from consent_hubs_facilities where consent_hub_id=?", consentHub.getId());
+
+			int numAffected = jdbc.update("delete from consent_hubs where id=?", consentHub.getId());
+			if (numAffected == 0) throw new ConsentHubAlreadyRemovedException("ConsentHub: " + consentHub);
```
cuadradek, line 144: We could add log info just like in the create method.

cuadradek (commented):
Looks good; only minor suggestions.

cuadradek (approved):

PR #3525, opened 2022-03-29
feat(dispatcher): block propagation of services with no destination

HejdaJakub:
* Due to the new consent logic, we don't want to allow the propagation of services without any destination.

### DIFF
--- a/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
+++ b/perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
@@ -371,6 +371,10 @@ public class TaskScheduler extends AbstractRunner {
 				// when service is blocked globally in Perun or on facility as a whole.
 				return DENIED;
 			}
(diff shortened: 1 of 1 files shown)


perun-dispatcher/src/main/java/cz/metacentrum/perun/dispatcher/scheduling/TaskScheduler.java
```diff
@@ -371,17 +371,19 @@ protected TaskScheduled sendToEngine(Task task) {
 				// when service is blocked globally in Perun or on facility as a whole.
 				return DENIED;
 			}
+		} else {
+			log.error("No destination found for given service: {}", service.getId());
```
zlamalp, line 375: From the perspective of logs I would consider this to be warning or just debug. Also destinations are bound to Task, since its combination of facility/service, where the data is provisioned.
Also see generic template for log messages in Dispatcher/Engine, eg. `[{}] Removed blocked destination: {}`. They are like this in order to easily "grep" the whole Task processing.

zlamalp (approved):

PR #3529, opened 2022-03-30
feat(core): Added methods for work with Consent

mattjoke:
* Added createConsent method to Impl and BlImpl
* Added deleteConsent method to Impl and BlImpl
* Created exceptions ConsentExistsException and ConsentNotExistsException

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ConsentNotExistsException.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ConsentNotExistsException.java
@@ -1,0 +1,56 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+import cz.metacentrum.perun.core.api.Consent;
+
+/**
+ * Thrown when trying to retrieve a consent that doesn't exist from the database
+ *
+ * @author David Flor <493294@mail.muni.cz>
+ */
+public class ConsentNotExistsException extends EntityNotExistsException {
+	static final long serialVersionUID = 0;
+
+	private Consent consent;
+
+	/**
+	 * Simple constructor with a message
+	 * @param message message with details about the cause
+	 */
+	public ConsentNotExistsException(String message) {
+		super(message);
+	}
+
+	/**
+	 * Constructor with a message and Throwable object
+	 * @param message message with details about the cause
+	 * @param cause Throwable that caused throwing of this exception
+	 */
+	public ConsentNotExistsException(String message, Throwable cause) {
+		super(message, cause);
+	}
+
+	/**
+	 * Constructor with a Throwable object
+	 * @param cause Throwable that caused throwing of this exception
+	 */
+	public ConsentNotExistsException(Throwable cause) {
+		super(cause);
+	}
+
+	/**
+	 * Constructor with the consent that does not exist
+	 * @param consent the consent
+	 */
+	public ConsentNotExistsException(Consent consent) {
+		super(consent.toString());
+		this.consent = consent;
+	}
+
+	/**
+	 * Getter for the consent that doesn't exist
+	 * @return the consent that doesn't exist
+	 */
+	public Consent getConsent() {
+		return this.consent;
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
@@ -13,8 +13,6 @@ import cz.metacentrum.perun.core.api.exceptions.ConsentHubNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubAlreadyRemovedException;
-import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
-import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 
 import java.util.List;
 
@@ -29,7 +27,7 @@ public interface ConsentsManagerBl {
 	/**
 	 * Gel all consents
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @return all existing consents in the database
 	 */
 	List<Consent> getAllConsents(PerunSession sess);
@@ -37,7 +35,7 @@ public interface ConsentsManagerBl {
 	/**
 	 * Get all consents for chosen ConsentHub with the specified status
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param id     id of the ConsentHub
 	 * @param status status of the consent
 	 * @return consents for chosen ConsentHub with the specified status
@@ -47,7 +45,7 @@ public interface ConsentsManagerBl {
 	/**
 	 * Get all consents for chosen ConsentHub
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param id   id of the ConsentHub
 	 * @return consents for chosen ConsentHub
 	 */
@@ -56,7 +54,7 @@ public interface ConsentsManagerBl {
 	/**
 	 * Get all consents for chosen User with the specified status
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param id     id of the User
 	 * @param status status of the consent
 	 * @return consents for chosen User with the specified status
@@ -66,7 +64,7 @@ public interface ConsentsManagerBl {
 	/**
 	 * Get all consents for chosen User
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param id   id of the user
 	 * @return consents for chosen User
 	 */
@@ -75,7 +73,7 @@ public interface ConsentsManagerBl {
 	/**
 	 * Get all consents for chosen user in specified consent hub
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param userId       id of the user
 	 * @param consentHubId id of the consent hub
 	 * @return consents
@@ -85,19 +83,19 @@ public interface ConsentsManagerBl {
 	/**
 	 * Get consent for chosen user in specified consent hub with specified status
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param userId       if of the user
 	 * @param consentHubId id of the consent hub
 	 * @param status       specified status
 	 * @return consent
-	 * @throws ConsentNotExistsException
+	 * @throws ConsentNotExistsException thrown if consent with the id doesn't exist
 	 */
 	Consent getConsentForUserAndConsentHub(PerunSession sess, int userId, int consentHubId, ConsentStatus status) throws ConsentNotExistsException;
 
 	/**
 	 * Get consent object with specified id
 	 *
-	 * @param sess
+	 * @param sess perun session
 	 * @param id   id of desired consent object
 	 * @return consent object with specified id
 	 * @throws ConsentNotExistsException thrown if consent with the id doesn't exist
@@ -119,6 +117,9 @@ public interface ConsentsManagerBl {
 	 * the attributes of the user and the consent hub.
 	 * Attributes in the consent are ignored!
 	 *
+	 * If the user has already 'UNSIGNED' consent for the same consent hub,
+	 * that consent is deleted and replaced with the new one.
+	 *
 	 * @param perunSession perun session
 	 * @param consent Consent to create
 	 *
@@ -126,9 +127,8 @@ public interface ConsentsManagerBl {
 	 *
 	 * @throws ConsentExistsException if consent already exists
 	 * @throws ConsentHubNotExistsException if consent hub doesn't exist
-	 * @throws ConsentNotExistsException if consent doesn't exist
 	 */
-	Consent createConsent(PerunSession perunSession, Consent consent) throws ConsentExistsException, ConsentHubNotExistsException, ConsentNotExistsException;
+	Consent createConsent(PerunSession perunSession, Consent consent) throws ConsentExistsException, ConsentHubNotExistsException;
 
 	/**
 	 * Deletes consent
@@ -145,7 +145,7 @@ public interface ConsentsManagerBl {
 	 *
 	 * @param sess perun session
 	 * @return list of Consent Hubs
-	 * @throws InternalErrorException
+	 * @throws InternalErrorException if an exception occurred
 	 */
 	List<ConsentHub> getAllConsentHubs(PerunSession sess);
 
@@ -155,8 +155,8 @@ public interface ConsentsManagerBl {
 	 * @param sess perun session
 	 * @param id id of the Consent Hub you are looking for
 	 * @return found Consent Hub
-	 * @throws ConsentHubNotExistsException
-	 * @throws InternalErrorException
+	 * @throws ConsentHubNotExistsException if Consent Hub doesn't exist
+	 * @throws InternalErrorException if an exception occurred
 	 */
 	ConsentHub getConsentHubById(PerunSession sess, int id) throws ConsentHubNotExistsException;
 
@@ -166,8 +166,8 @@ public interface ConsentsManagerBl {
 	 * @param sess perun session
 	 * @param name name of the Consent Hub you are looking for
 	 * @return found Consent Hub
-	 * @throws ConsentHubNotExistsException
-	 * @throws InternalErrorException
+	 * @throws ConsentHubNotExistsException if Consent Hub doesn't exist
+	 * @throws InternalErrorException if an exception occurred
 	 */
 	ConsentHub getConsentHubByName(PerunSession sess, String name) throws ConsentHubNotExistsException;
 
@@ -177,7 +177,7 @@ public interface ConsentsManagerBl {
 	 * @param sess perun session
 	 * @param facilityId facility for which consent hub is searched
 	 * @return found Consent Hub
-	 * @throws ConsentHubNotExistsException
+	 * @throws ConsentHubNotExistsException if Consent Hub doesn't exist
 	 */
 	ConsentHub getConsentHubByFacility(PerunSession sess, int facilityId) throws ConsentHubNotExistsException;
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
@@ -1,14 +1,20 @@
 package cz.metacentrum.perun.core.bl;
 
+import cz.metacentrum.perun.core.api.Consent;
 import cz.metacentrum.perun.core.api.ConsentHub;
 import cz.metacentrum.perun.core.api.Facility;
+import cz.metacentrum.perun.core.api.ConsentStatus;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.exceptions.FacilityAlreadyAssigned;
 import cz.metacentrum.perun.core.api.exceptions.RelationNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentHubAlreadyRemovedException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.ConsentHubAlreadyRemovedException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
+import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 
 import java.util.List;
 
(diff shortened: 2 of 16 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -13,10 +19,42 @@
...
+
 	private static JdbcPerunTemplate jdbc;
 
 	public ConsentsManagerImpl(DataSource perunPool) {
 		jdbc = new JdbcPerunTemplate(perunPool);
 		jdbc.setQueryTimeout(BeansUtils.getCoreConfig().getQueryTimeout());
 	}
+
+	@Override
+	public Consent createConsent(PerunSession perunSession, Consent consent) {
```
xflord, line 32: Missing check whether consent already exists. Also, when creating a new consent, we want to check and remove any existing unsigned consents for that user within that consent hub

xflord (commented):
Another part of this task was making sure that when a user or a facility is deleted, associated Consents are also removed. Please include this.

xflord (changes requested):
.

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/ConsentNotExistsException.java
```diff
@@ -0,0 +1,45 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+import cz.metacentrum.perun.core.api.Consent;
+
+/**
+ * This exception is thrown when trying to create a consent
```
Johaney-s, line 6: Wrong description. Also, we usually add @ author tags to new classes/exceptions, please add yours.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -13,10 +19,42 @@
...
+		try {
+			int newId = Utils.getNewId(jdbc, "consents_id_seq");
+
+			jdbc.update("insert into consents(id, user_id, consent_hub_id, consent_status, created_at, created_by, modified_at, modified_by, created_by_uid, modified_by_uid) " +
+					"values (?,?,?,?,"+Compatibility.getSysdate()+",?,"+Compatibility.getSysdate()+",?,?,?)", newId, consent.getUserId(),
+				consent.getConsentHub().getId(), "UNSIGNED", perunSession.getPerunPrincipal().getActor(), perunSession.getPerunPrincipal().getActor(),
```
Johaney-s, line 38: i find it more convenient to use `ConsentStatus.UNSIGNED`, because if we choose to rename it at some point, we won't have to rewrite it at all places where we didn't use the enum value.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/ConsentsManagerImplApi.java
```diff
@@ -1,10 +1,33 @@
...
  */
 public interface ConsentsManagerImplApi {
 
+	/**
+	 * Save consent to database.
+	 *
+	 * @param perunSession PerunSession
+	 * @param consent Consent
+	 * @return created consent
+	 * @throws cz.metacentrum.perun.core.api.exceptions.IllegalArgumentException if consent is null
```
Johaney-s, line 21:
```suggestion
	 * @throws ConsentExistsException if consent exists
```
isn't InternalErrorException thrown when consent is null? please fix the javadoc to match the actual exceptions that are thrown. you won't need to use the whole package name, just the exception name is okay.

Johaney-s (changes requested):
We should also have tests for this.

As @xflord mentioned about handling the deletion of facility and user - just note, that for facility this is already being handled in PR #3524 and we'd need to add the logic for removing consents to the one used there.

Johaney-s:
As other methods creating objects, it should be also added to `perun-core.xml` to run in transaction.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
xflord, line 32: We also need to create a new entry in the _consent_attr_defs_ table for each attribute the Consent object has.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -13,10 +19,42 @@
...
+			consent.setId(newId);
+
+			return consent;
+		} catch (Exception e){
+			throw new InternalErrorException(e);
+		}
+	}
+
+	@Override
+	public void deleteConsent(PerunSession perunSession, Consent consent) {
```
xflord, line 51: Same goes for deleting.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -38,6 +45,28 @@ public void setPerunBl(PerunBl perunBl) {
 		this.perunBl = perunBl;
 	}
 
+	public Consent createConsent(PerunSession sess, Consent consent) throws ConsentExistsException, UserNotExistsException, PrivilegeException, ConsentHubNotExistsException, ConsentNotExistsException {
```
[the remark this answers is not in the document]
cuadradek: When creating a consent, it should always contain all attribute definitions required in the consent hub (all user-related attributes required on services assigned to resources of the consent hub's facilities). Therefore, we should always calculate these attributes and set them to the consent object before it is passed to the impl layer, where the attributes are inserted into the `consent_attr_defs` table.

Please, mention this (ignoring attributes in the consent object and calculating them instead) in the method's Javadoc because it is not a typical "create" behavior.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -38,6 +45,28 @@ public void setPerunBl(PerunBl perunBl) {
...
+		Utils.notNull(consent, "consent");
+
+		consent = getConsentsManagerImpl().createConsent(sess, consent);
+		getPerunBl().getAuditer().log(sess, new ConsentCreated(consent));
+
+		return consent;
+	}
+
+	public void deleteConsent(PerunSession sess, Consent consent) throws ConsentNotExistsException {
```
cuadradek, line 57: Missing `@Override` annotation.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -135,7 +220,14 @@ public boolean consentHubExists(PerunSession sess, ConsentHub consentHub) {
 
 	@Override
 	public void deleteConsentHub(PerunSession perunSession, ConsentHub consentHub) throws ConsentHubAlreadyRemovedException {
-		//TODO: remove all user consents first
+
+		try {
+			for (Consent consent : getConsentsForConsentHub(perunSession, consentHub.getId())) {
```
cuadradek, line 225: I think deleting all the consent hub's consents should be moved to the BL layer because the `deleteConsent` method in the BL layer should also audit the deletion message. If it remained here, we would lose the audit information.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -38,6 +51,112 @@ public void setPerunBl(PerunBl perunBl) {
...
+			attributeNamespace.startsWith(AttributesManager.NS_UES_ATTR);
+	}
+
+
+	public void deleteConsent(PerunSession sess, Consent consent) throws ConsentNotExistsException {
+		Utils.notNull(consent, "consent");
+		Utils.notNull(consent.getId(), "consent.id");
+
+		// Remove UNSIGNED consents from consent hub
+		try {
```
xflord, line 101: I think this check should be in createConsent()

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -51,11 +55,303 @@
...
 	}
 
+	@Override
+	public Consent createConsent(PerunSession perunSession, Consent consent) throws ConsentNotExistsException, ConsentExistsException, ConsentHubNotExistsException {
+		// Check if consent already exists
+		if(consentExists(perunSession, consent)){
+			throw new ConsentExistsException("Consent already exists.");
+		}
+
+		// Check if UNSIGNED consent exists and delete it
```
xflord, line 113: If you move the check in BL from deleteConsent to createConsent, this is no longer needed (On second thought, the check also probably **should** be in BL, not here.)

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -51,11 +55,303 @@
...
+			return consent;
+		} catch (Exception e){
+			throw new InternalErrorException(e);
+		}
+	}
+
+	@Override
+	public void deleteConsent(PerunSession perunSession, Consent consent) throws ConsentNotExistsException {
+		// Check if consent exists
+		checkConsentExists(perunSession, consent);
```
xflord, line 142: Don't think this check is necessary.

xflord (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -38,6 +51,112 @@ public void setPerunBl(PerunBl perunBl) {
...
+			attributeNamespace.startsWith(AttributesManager.NS_MEMBER_ATTR_CORE) ||
+			attributeNamespace.startsWith(AttributesManager.NS_USER_FACILITY_ATTR) ||
+			attributeNamespace.startsWith(AttributesManager.NS_USER_ATTR) ||
+			attributeNamespace.startsWith(AttributesManager.NS_UES_ATTR);
+	}
+
+
+	public void deleteConsent(PerunSession sess, Consent consent) throws ConsentNotExistsException {
+		Utils.notNull(consent, "consent");
+		Utils.notNull(consent.getId(), "consent.id");
```
cuadradek, line 112: We can omit this check because id is a primitive integer and thus never a null :)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -72,6 +193,23 @@ public ConsentHub createConsentHub(PerunSession sess, ConsentHub consentHub) thr
 
 	@Override
 	public void deleteConsentHub(PerunSession sess, ConsentHub consentHub) throws ConsentHubAlreadyRemovedException {
+
+		// Check if ConsentHub exists
+		try {
+			getConsentHubById(sess, consentHub.getId());
```
cuadradek, line 199: I think we can omit this check because this method in the impl layer would throw this exception anyway. What do you think?

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -41,6 +56,109 @@ public void setPerunBl(PerunBl perunBl) {
...
+			if (oldConsent != null) {
+				deleteConsent(sess, oldConsent);
+			}
+		} catch (ConsentNotExistsException e) {
+			// OK
+		}
+
+		// Add attributes from ConsentHub
+		ConsentHub consentHub = getConsentHubById(sess, consent.getConsentHub().getId());
+		Set<AttributeDefinition> filter = new HashSet<>(consent.getAttributes());
```
cuadradek, line 76:
```suggestion
		Set<AttributeDefinition> filter = new HashSet<>();
```
Just a detail, so the original consent attributes are ignored, as the Javadoc says.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -53,11 +63,285 @@
...
+	}
+
+	@Override
+	public void deleteConsent(PerunSession perunSession, Consent consent) throws ConsentNotExistsException {
+		try {
+			int numberOfRows = jdbc.update("delete from consents where id=?", consent.getId());
+			if (numberOfRows != 1) throw new ConsentNotExistsException(consent);
+			log.info("Consent {} deleted.", consent);
+			consent.setAttributes(null);
+			removeAttrsForConsent(perunSession, consent.getId());
```
Johaney-s, line 139: This should be done before removing the consent, because the table storing attribute-consent relationship uses consent id as foreign key. Would be nice to have that behavior tested also.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/ConsentsManagerImplApi.java
```diff
@@ -19,12 +23,127 @@
  */
 public interface ConsentsManagerImplApi {
 
+	/**
+	 * Save consent to database.
+	 *
+	 * @param perunSession PerunSession
+	 * @param consent Consent
+	 * @return created consent
+	 * @throws IllegalArgumentException if consent is null
```
Johaney-s, line 32: update exceptions in javadoc

Johaney-s (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -53,11 +63,289 @@
...
+			return consent;
+		} catch (Exception e){
+			throw new InternalErrorException(e);
+		}
+	}
+
+	@Override
+	public void deleteConsent(PerunSession perunSession, Consent consent) throws ConsentNotExistsException {
+		try {
+			if (!consentExists(perunSession, consent)) {
```
Johaney-s, line 135: can be simplified by calling `checkConsentExists` directly

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -53,11 +63,289 @@
...
+				throw new ConsentNotExistsException("Consent: " + consent + " doesn't exists.");
+			}
+			removeAttrsForConsent(perunSession, consent.getId());
+			consent.setAttributes(null);
+
+			int numberOfRows = jdbc.update("delete from consents where id=?", consent.getId());
+			if (numberOfRows != 1) throw new ConsentNotExistsException(consent);
+			log.info("Consent {} deleted.", consent);
+		} catch (Exception e){
```
Johaney-s, line 144: i think we don't want to return any thrown exception in the block as ConsentNotExistsException

cuadradek: We should also catch here only runtime exceptions and rethrow them as `InternalErrorException`. Otherwise, `ConsentNotExistsException` from two lines above would also be caught here and rethrown again.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -32,6 +45,185 @@ public void setUp() throws Exception {
...
+
+		Facility facility = setUpFacility();
+
+		Consent consentOld = new Consent(-1, user.getId(), perun.getConsentsManager().getConsentHubByName(sess, facility.getName()), new ArrayList<>());
+		consentOld = perun.getConsentsManagerBl().createConsent(sess, consentOld);
+
+		Consent consentNew = new Consent(-11, user.getId(), perun.getConsentsManager().getConsentHubByName(sess, facility.getName()), new ArrayList<>());
+		consentNew = perun.getConsentsManagerBl().createConsent(sess, consentNew);
+
+		perun.getConsentsManagerBl().checkConsentExists(sess, consentOld);
```
cuadradek, line 77: In longer tests, it is better to avoid asserting expected exceptions using `@Test (expected = ConsentNotExistsException.class)`, because this exception can be thrown at multiple lines in the test and we want to be sure it was thrown in this very line. You can use _assertThatExceptionOfType_ method:

```
assertThatExceptionOfType(ConsentNotExistsException.class).isThrownBy(
			() -> perun.getConsentsManagerBl().checkConsentExists(sess, consentOld));
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/ConsentsManagerImpl.java
```diff
@@ -53,11 +63,289 @@
...
+			return jdbc.query("select " + AttributesManagerImpl.attributeDefinitionMappingSelectQuery + " from attr_names join consent_attr_defs on attr_names.id = consent_attr_defs.attr_id" +
+				" where consent_attr_defs.consent_id=?", AttributesManagerImpl.ATTRIBUTE_DEFINITION_MAPPER, consentId);
+		} catch (EmptyResultDataAccessException ex) {
+			return new ArrayList<>();
+		} catch (RuntimeException err) {
+			throw new InternalErrorException(err);
+		}
+	}
+
+	private void createAttrDefsForConsent(PerunSession sess, Consent consent) throws ConsentNotExistsException {
```
cuadradek, line 290: `ConsentNotExistsException` is not thrown here, I think. Please, remove it from the signature and all methods calling this one, too.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -32,6 +45,185 @@ public void setUp() throws Exception {
...
+	@Test (expected = ConsentNotExistsException.class)
+	public void deleteConsent() throws Exception {
+		System.out.println(CLASS_NAME + "deleteConsent");
+
+		User user = setUpUser("John", "Doe");
+
+		Facility facility = setUpFacility();
+
+		Consent consent = new Consent(-1, user.getId(), perun.getConsentsManager().getConsentHubByName(sess, facility.getName()), new ArrayList<>());
+		perun.getConsentsManagerBl().deleteConsent(sess, consent);
```
cuadradek, line 90: I think this test is the case I was talking about above. I believe the exception `ConsentNotExistsException` would be thrown already here. We should also probably have two separate tests - one for deleting existing consent and one for deleting not-existing one.

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
cuadradek: Okay, you can delete this then.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
```diff
@@ -19,6 +25,121 @@
...
+	 * Attributes in the consent are ignored!
+	 *
+	 * @param perunSession perun session
+	 * @param consent Consent to create
+	 *
+	 * @return created consent
+	 *
+	 * @throws ConsentExistsException if consent already exists
+	 * @throws ConsentHubNotExistsException if consent hub doesn't exist
+	 * @throws ConsentNotExistsException if consent doesn't exist
```
cuadradek, line 129: I believe this exception won't be thrown, and we can omit it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -475,13 +477,13 @@ public void deleteUser(PerunSession sess, User user) throws RelationExistsExcept
...
 			this.deleteUser(sess, user, forceDelete, false);
-		} catch (AnonymizationNotSupportedException ex) {
+		} catch (AnonymizationNotSupportedException | ConsentNotExistsException | UserNotExistsException ex) {
 			//this shouldn't happen with 'anonymizedInstead' set to false
 			throw new InternalErrorException(ex);
 		}
 	}
 
-	private void deleteUser(PerunSession sess, User user, boolean forceDelete, boolean anonymizeInstead) throws RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException, AnonymizationNotSupportedException {
+	private void deleteUser(PerunSession sess, User user, boolean forceDelete, boolean anonymizeInstead) throws RelationExistsException, MemberAlreadyRemovedException, UserAlreadyRemovedException, SpecificUserAlreadyRemovedException, AnonymizationNotSupportedException, ConsentNotExistsException, UserNotExistsException {
```
cuadradek, line 486: I believe UserNotExistsException won't be thrown and should be omitted from the signature.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -614,6 +616,12 @@ private void deleteUser(PerunSession sess, User user, boolean forceDelete, boole
 			}
 		}
 
+		//Remove all user's consents
+		for (Consent consent : getPerunBl().getConsentsManagerBl().getConsentsForUser(sess, user.getId())) {
+			getPerunBl().getConsentsManagerBl().deleteConsent(sess, consent);
```
cuadradek, line 621: I think it would be better to catch the ConsentNotExists exception here, ignore it (we want to delete the consent anyway), and omit it from the signature. Because right now, it is caught in the other method and rethrown as InternalErrorException, I think it would be better to catch it here and skip it, so the whole method doesn't fail.

cuadradek (commented):
I have the last couple of minor comments.

cuadradek (approved):

Johaney-s (approved):

PR #3530, opened 2022-03-31
fix(core): fix consentHubs

Johaney-s:
* added test for removing facility
* logging deletion of consentHub
* changed filtering for future use

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
@@ -523,6 +523,8 @@ public enum FacilitiesManagerMethod implements ManagerMethod {
 	 * Parameter description is optional.
 	 * Other parameters are ignored.
 	 * @param facility Facility JSON object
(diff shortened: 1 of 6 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/FacilitiesManagerMethod.java
```diff
@@ -523,13 +523,19 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 	 * Parameter description is optional.
 	 * Other parameters are ignored.
 	 * @param facility Facility JSON object
+	 * @throw ConsentHubExistsException if consent hub with facility name exists
+	 * @throw FacilityExistsException if facility already exists
+	 * @throw PrivilegeException for insufficient rights
```
cuadradek, line 528: PrivilegeExceptions are added automatically, I think.

Johaney-s: good to know, fixed

cuadradek (approved):

PR #3541, opened 2022-04-06
feat(core): Added method changeConsentStatus 

sarkapalkovicova:
Added method for changing consent status.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
@@ -1,13 +1,13 @@
 package cz.metacentrum.perun.core.api;
 
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.FacilityNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InvalidConsentStatusException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubExistsException;
-
 import java.util.List;
 
 /**
@@ -167,4 +167,18 @@ public interface ConsentsManager {
 	 * @throws PrivilegeException           insufficient rights
 	 */
 	ConsentHub updateConsentHub(PerunSession perunSession, ConsentHub consentHub) throws ConsentHubNotExistsException, PrivilegeException, ConsentHubExistsException;
+
+	/**
+	 * Set consent status
+	 *
+	 * @param sess perun session
+	 * @param consent consent
+	 * @param status status that should be set
+	 * @return consent
+	 * @throws ConsentNotExistsException if consent hub does not exist
+	 * @throws PrivilegeException if insufficient rights
+	 * @throws InvalidConsentStatusException if passed status value can not be set
+	 * @throws UserNotExistsException if user does not exist
+	 */
+	Consent changeConsentStatus(PerunSession sess, Consent consent, ConsentStatus status) throws ConsentNotExistsException, PrivilegeException, InvalidConsentStatusException, UserNotExistsException;
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
@@ -2,15 +2,16 @@ package cz.metacentrum.perun.core.bl;
 
 import cz.metacentrum.perun.core.api.Consent;
 import cz.metacentrum.perun.core.api.ConsentHub;
-import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.ConsentStatus;
+import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.PerunSession;
-import cz.metacentrum.perun.core.api.exceptions.FacilityAlreadyAssigned;
-import cz.metacentrum.perun.core.api.exceptions.RelationNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.FacilityAlreadyAssigned;
+import cz.metacentrum.perun.core.api.exceptions.InvalidConsentStatusException;
+import cz.metacentrum.perun.core.api.exceptions.RelationNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentHubAlreadyRemovedException;
 
@@ -244,4 +245,13 @@ public interface ConsentsManagerBl {
 	 */
 	void removeFacility(PerunSession sess, ConsentHub consentHub, Facility facility) throws RelationNotExistsException, ConsentHubAlreadyRemovedException;
 
+	/**
+	 * Set consent status
+	 *
+	 * @param sess perun session
+	 * @param consent consent
+	 * @return consent
+	 * @throws InvalidConsentStatusException if passed status value can not be set
+	 */
+	Consent changeConsentStatus(PerunSession sess, Consent consent, ConsentStatus status) throws InvalidConsentStatusException;
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
@@ -1,22 +1,25 @@
 package cz.metacentrum.perun.core.blImpl;
 
+import cz.metacentrum.perun.audit.events.ConsentManager.ChangedConsentStatus;
 import cz.metacentrum.perun.audit.events.ConsentManager.ConsentCreated;
 import cz.metacentrum.perun.audit.events.ConsentManager.ConsentDeleted;
+import cz.metacentrum.perun.core.api.ConsentHub;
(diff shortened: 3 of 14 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntry.java
```diff
@@ -139,4 +145,27 @@ public ConsentHub updateConsentHub(PerunSession sess, ConsentHub consentHub) thr
...
+
+		if (status != ConsentStatus.GRANTED && status != ConsentStatus.REVOKED) {
+			throw new InvalidConsentStatus("Invalid consent status value.");
+		}
+		if (status == consent.getStatus()) {
+			throw new InvalidConsentStatus("Tried to set consent status on current value.");
+		}
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "changeConsentStatus_int_ConsentStatus_int_policy", user)) {
```
Johaney-s, line 163: we should check this for the consent's user, not for the user who was sent with this request. otherwise any user would be authorized to change any consent's status. speaking of which, do we need the user argument in this method?

perun-base/src/main/java/cz/metacentrum/perun/core/api/exceptions/InvalidConsentStatus.java
```diff
@@ -0,0 +1,36 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+/**
+ * This exception is thrown when trying to set consent status on an invalid value.
+ *
+ * @author Sarka Palkovicova
+ */
+public class InvalidConsentStatus extends PerunException {
```
Johaney-s, line 8:
```suggestion
public class InvalidConsentStatusException extends PerunException {
```
to match convention of other custom created exceptions.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
```diff
@@ -72,4 +75,18 @@
...
+	 * @param consentId id of consent
+	 * @param status status that should be set
+	 * @param userId id of user
+	 * @return consent
+	 * @throws ConsentNotExistsException
+	 * @throws PrivilegeException
+	 * @throws UserNotExistsException
+	 * @throws InvalidConsentStatus
+	 */
+	Consent changeConsentStatus(PerunSession sess, int consentId, ConsentStatus status, int userId) throws ConsentNotExistsException, PrivilegeException, UserNotExistsException, InvalidConsentStatus;
```
xflord, line 91: Do we need to pass userId? Can't we always get the relevant User object from the Consent object?

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
```diff
@@ -123,4 +180,13 @@
...
 
+	/**
+	 * Set consent status
+	 *
+	 * @param sess perun session
+	 * @param consent consent
+	 * @param user user
+	 * @return consent
+	 */
+	Consent changeConsentStatus(PerunSession sess, Consent consent, User user) throws ConsentNotExistsException;
```
xflord, line 191: Same here.

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntry.java
```diff
@@ -139,4 +145,27 @@ public ConsentHub updateConsentHub(PerunSession sess, ConsentHub consentHub) thr
...
 		return getConsentsManagerBl().updateConsentHub(sess, consentHub);
 	}
+	@Override
+	public Consent changeConsentStatus(PerunSession sess, int consentId, ConsentStatus status, int userId) throws ConsentNotExistsException, PrivilegeException, UserNotExistsException, InvalidConsentStatus {
+		Utils.notNull(sess, "sess");
+
+		Consent consent = getPerunBl().getConsentsManagerBl().getConsentById(sess, consentId);
+		User user = getPerunBl().getUsersManager().getUserById(sess, userId);
+
+		if (status != ConsentStatus.GRANTED && status != ConsentStatus.REVOKED) {
```
xflord, line 155: These checks and the logic of setting the status should probably be in BL layer. (not sure, check with @Johaney-s  or @cuadradek )

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntry.java
```diff
@@ -139,4 +145,27 @@ public ConsentHub updateConsentHub(PerunSession sess, ConsentHub consentHub) thr
 
 		return getConsentsManagerBl().updateConsentHub(sess, consentHub);
 	}
+	@Override
+	public Consent changeConsentStatus(PerunSession sess, int consentId, ConsentStatus status, int userId) throws ConsentNotExistsException, PrivilegeException, UserNotExistsException, InvalidConsentStatus {
+		Utils.notNull(sess, "sess");
+
+		Consent consent = getPerunBl().getConsentsManagerBl().getConsentById(sess, consentId);
+		User user = getPerunBl().getUsersManager().getUserById(sess, userId);
```
xflord, line 153: Here you can easily get the user from the consent, no need to pass it.

xflord (changes requested):
Good job, only minor tweaks.

Johaney-s (commented):
looks good to me, just a few suggestions

perun-core/src/main/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntry.java
cuadradek: Yeah, I agree

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ConsentsManagerMethod.java
```diff
@@ -81,5 +83,27 @@ public ConsentHub call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * Returns consent with changed status.
+	 *
+	 * @param consent int Consent <code>id</code>
+	 * @param status String UNSIGNED | GRANTED | REVOKED
+	 * @param user int User <code>id</code>
+	 *
+	 * @throws ConsentNotExistsException
+	 * @throws PrivilegeException
+	 * @throws UserNotExistsException
+	 * @throws InvalidConsentStatus
```
cuadradek, line 98:
```suggestion
	 * @throw InvalidConsentStatus
```
I think it should be like this here in RPC (I think it's because of parsing, I am not sure if _throws_ would be incorrect, tho). Also, `PrivilegeException` is added automatically to every method in RPC, so we should omit it. 

Please, add a brief description of when these exceptions happen.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ConsentsManagerMethod.java
```diff
@@ -81,5 +83,27 @@ public ConsentHub call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 *
+	 * @return Consent
+	 */
+	changeConsentStatus {
+		@Override
+		public Consent call(ApiCaller ac, Deserializer params) throws PerunException {
+			return ac.getConsentsManager().changeConsentStatus(ac.getSession(), params.readInt("consent"), ConsentStatus.valueOf(params.readString("status")), params.readInt("user"));
```
cuadradek, line 105:
```suggestion
			return ac.getConsentsManager().changeConsentStatus(ac.getSession(), ac.getConsentById(parms.readInt("consent")), ConsentStatus.valueOf(params.readString("status")));
```
In RPC, we usually get the object by the ID and the object is then passed to the Entry layer instead of the ID. Then we won't have to get the object by ID in the entry layer. You will have to create a simple method `getConsentById` in ApiCaller.

cuadradek (commented):
Looks good. Please rebase this branch because another PR was just merged.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -232,4 +234,28 @@ public void removeFacility(PerunSession sess, ConsentHub consentHub, Facility fa
...
+		checkExistingConsents(sess, consent);
+		getConsentsManagerImpl().changeConsentStatus(sess, consent);
+		getPerunBl().getAuditer().log(sess, new ChangedConsentStatus(consent));
+		return getConsentById(sess, consent.getId());
+	}
+
+	private void checkExistingConsents(PerunSession sess, Consent consent) throws ConsentNotExistsException {
+		for (Consent currentConsent : getConsentsForUserAndConsentHub(sess, consent.getUserId(), consent.getConsentHub().getId())) {
+			if (currentConsent.getId() != consent.getId() && currentConsent.getStatus() != ConsentStatus.UNSIGNED) {
+				deleteConsent(sess, currentConsent);
```
xflord, line 257: You could catch ConsentNotExistsException here and throw InternalErrorException/ConsitencyError, since theoretically the Consent should always exist. Then you could remove the exception from the method signature (which would make more sense, because you checked whether the consent exists in the entry layer).

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -404,6 +404,40 @@ public void addFacilityToConsentHubAgain() throws Exception {
...
+
+		ConsentStatus status = ConsentStatus.GRANTED;
+		Facility facility = setUpFacility();
+		User user = setUpUser("John", "Doe");
+
+		Consent consent = new Consent(-1, user.getId(), perun.getConsentsManager().getConsentHubByName(sess, facility.getName()), new ArrayList<>());
+		perun.getConsentsManagerBl().createConsent(sess, consent);
+
+		Consent updatedConsent = consentsManagerEntry.changeConsentStatus(sess, consent, status);
+		assertEquals("Updated consent has different status than sent consent.", status, updatedConsent.getStatus());
```
cuadradek, line 419: Just a detail, I would probably assert that `getConsentById` returns the consent with GRANTED status as well. Just to be sure that the status was really set in DB, not just to the object returned by `changeConsentStatus`.

xflord (changes requested):
One tiny thing

Johaney-s (approved):

cuadradek (commented):
Can you please add two simple tests for `InvalidConsentStatusException`? Tests for setting UNSIGNED status and setting the same status.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
sarkapalkovicova: No problem, I can add one more assert, but changeStatus already returns consent from the database by using getConsentById. Should I return consent in bl layer with set status passed like argument instead of one from the database?

cuadradek: I think both ways are alright, you don't have to change it.

cuadradek (approved):

PR #3548, opened 2022-04-10
perf(core): speedup getHashedDataWithGroups

cuadradek:
- Method getHashedDataWithGroups used to evaluate consents for every
group's members (in method getDataForGroup). That was not necessary
because all members on the resource were evaluated (in method
getDataForResource) before and all members with required GRANTED
consents are saved in the set allMembers.


Johaney-s (approved):

cuadradek:
This is solved by #3549, too.

balcirakpeter (approved):

balcirakpeter:
:D

PR #3549, opened 2022-04-10
fix(core): fix possible deadlock while evaluating consents

cuadradek:
- A deadlock could happen while concurrently evaluating consents for the
same consent hub. If a user didn't have an existing consent, he would be
locked and a consent would be created. Evaluation1 could lock user1 and
wait for user2 locked by evaluation2. If evaluation2 wanted to lock
user1, a deadlock would happen.
- Another problem in concurrent evaluating was creating a new consent.
All concurrent evaluations could possibly create a new one - that would
create multiple ConsentCreated audit messages and therefore send
multiple emails to user.
- Now, users are not locked individually but the whole consent hub is
locked while evaluating which should solve both problems.
- Evaluating now happens only once for one propagation instead of for
every resource and every group.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
@@ -33,7 +33,7 @@ public class PerunLocksUtils {
 	//Maps for saving and working with specific locks
 	private static final ConcurrentHashMap<Group, ReadWriteLock> groupsLocks = new ConcurrentHashMap<>();
 	private static final ConcurrentHashMap<Group, ConcurrentHashMap<Member, Lock>> groupsMembersLocks = new ConcurrentHashMap<>();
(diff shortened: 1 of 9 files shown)


balcirakpeter (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunLocksUtils.java
```diff
@@ -34,6 +34,7 @@
 	private static final ConcurrentHashMap<Group, ReadWriteLock> groupsLocks = new ConcurrentHashMap<>();
 	private static final ConcurrentHashMap<Group, ConcurrentHashMap<Member, Lock>> groupsMembersLocks = new ConcurrentHashMap<>();
 	private static final ConcurrentHashMap<ConsentHub, ConcurrentHashMap<Integer, Lock>> consentHubsUsersLocks = new ConcurrentHashMap<>();
```
Johaney-s, line 36: nitpick - now this can be removed

Johaney-s (approved):

balcirakpeter (approved):

PR #3554, opened 2022-04-13
Manual consent evaluation

Kuliak:
* consents evaluation for consent hub - for all its facilities with corresponding resources and services
* consents evaluation for service - full evaluation for all consent hubs that have given service assigned
* flag for usage of expired members added to Service class (value is copied from gen script directly)
* database changed to support new flag
* openapi updated - 2 new methods added
* policies updated - 2 new methods added

breaking change: new database version (Service flag for expired members usage added)

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
@@ -187,6 +187,16 @@ public interface ConsentsManagerBl {
 	ConsentHub getConsentHubByFacility(PerunSession sess, int facilityId) throws ConsentHubNotExistsException;
 
 	/**
+	 * Finds all existing Consent Hubs by service id (consent hubs that have given service assigned through facilities).
+	 *
+	 * @param sess perun session
+	 * @param serviceId service for which consent hubs is searched
+	 * @return found Consent Hubs
+	 * @throws ConsentHubNotExistsException if Consent Hub doesn't exist
+	 */
+	List<ConsentHub> getConsentHubsByService(PerunSession sess, int serviceId) throws ConsentHubNotExistsException;
+
+	/**
 	 * Creates new consent hub.
 	 * @param perunSession session
 	 * @param consentHub consent hub
@@ -282,4 +292,30 @@ public interface ConsentsManagerBl {
 	 * @param members the given members
 	 */
 	List<Member> evaluateConsents(PerunSession sess, Service service, Facility facility, List<Member> members);
+
+	/**
+	 * Evaluates consents for given consent hub with enforced consents enabled.
+	 *
+	 * For given ConsentHub, collects all attributes from services assigned to it
+	 * and compares it against attributes from users' Consent objects that are
+	 * assigned to resources of given ConsentHub object.
+	 *
+	 * Service defines whether only active users will be evaluated or expired ones as well.
+	 *
+	 *
+	 * @param sess session
+	 * @param consentHub consent hub
+	 */
+	void evaluateConsents(PerunSession sess, ConsentHub consentHub);
+
+	/**
+	 * Consolidate consents using given service for all consent hubs the service is assigned to.
+	 *
+	 * Method finds all consent hubs that contain given service in any of its facilities and start
+	 * full consent consolidation for each consent hub (all services in consent hubs will be checked).
+	 *
+	 * @param sess session
+	 * @param service service
+	 */
+	void evaluateConsents(PerunSession sess, Service service);
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
@@ -3,27 +3,28 @@ package cz.metacentrum.perun.core.blImpl;
 import cz.metacentrum.perun.audit.events.ConsentManager.ChangedConsentStatus;
 import cz.metacentrum.perun.audit.events.ConsentManager.ConsentCreated;
 import cz.metacentrum.perun.audit.events.ConsentManager.ConsentDeleted;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.BeansUtils;
-import cz.metacentrum.perun.core.api.ConsentHub;
 import cz.metacentrum.perun.core.api.Consent;
+import cz.metacentrum.perun.core.api.ConsentHub;
 import cz.metacentrum.perun.core.api.ConsentStatus;
 import cz.metacentrum.perun.core.api.Facility;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.PerunBean;
 import cz.metacentrum.perun.core.api.PerunSession;
-import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.Resource;
+import cz.metacentrum.perun.core.api.Service;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.exceptions.ConsentExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentHubAlreadyRemovedException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentHubExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsentHubNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ConsentNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ConsistencyErrorException;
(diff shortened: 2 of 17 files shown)


cuadradek (changes requested):
Looks good so far; I have just a couple of suggestions.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -346,4 +358,137 @@ private boolean hasValidConsent(PerunSession sess, Member member, ConsentHub con
...
+				.filter((member -> assignedUsers.contains(member.getUserId())))
+				.collect(Collectors.toList());
+
+			// all services assigned to given resource
+			List<Service> servicesOnResource = getPerunBl().getResourcesManagerBl().getAssignedServices(sess, resource);
+
+			// get only services assigned to given resource as well as to facility
+			List<Service> filteredServicesOnResource = facilityAssignedServices.stream()
+				.filter((servicesOnResource::contains))
+				.collect(Collectors.toList());
```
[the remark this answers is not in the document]
Kuliak: Yes, I was not sure whether it is the case, so I added this to be sure (removed now).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -346,4 +358,137 @@ private boolean hasValidConsent(PerunSession sess, Member member, ConsentHub con
...
+
+			// For each resource, check consents service by service
+			for (Service service : filteredServicesOnResource) {
+				List<Member> filteredMembersOnResource;
+
+				// depending on service, use active (possible expired too) members assigned to given resource
+				if (service.isUseExpiredMembers()) {
+					filteredMembersOnResource = membersOnResource.stream()
```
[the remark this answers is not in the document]
Kuliak: Okay, I expected it to work the other way, but this makes sense as well. I fixed it - now group status is used (in the first case, status is compared against both VALID and EXPIRED value despite the fact, that only these two values are possible for group status, I wanted to make sure that in case some new status will be added in the future, it will not affect this)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -346,4 +358,137 @@ private boolean hasValidConsent(PerunSession sess, Member member, ConsentHub con
...
+
+		PerunLocksUtils.lockConsentHub(consentHub);
+
+		// map of users ids to map of their consents (consent status to the user's consent with such status)
+		Map<Integer, Map<ConsentStatus, Consent>> userIdToConsents = this.getConsentsForConsentHub(sess, consentHub.getId())
+			.stream()
+			.collect(groupingBy(Consent::getUserId, toMap(Consent::getStatus, Function.identity())));
+
+		for (Facility facility : facilities) {
+			facilityAssignedServices = getPerunBl().getServicesManagerBl().getAssignedServices(sess, facility);
```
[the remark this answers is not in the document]
Kuliak: Yes, it should work like this. Very nice suggestion, thank you! I have missed those two methods, now it is significantly simplified. 
I have created set to contain all the processed users, is it OK?

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -620,6 +639,282 @@ public void evaluateConsentsWithInsufficientUnsignedConsent() throws Exception {
...
+		perun.getGroupsManagerBl().addMember(sess, group, member1);
+		perun.getGroupsManagerBl().addMember(sess, group, member2);
+		perun.getGroupsManager().setMemberGroupStatus(sess, member1, group, MemberGroupStatus.EXPIRED);
+		perun.getGroupsManager().setMemberGroupStatus(sess, member2, group, MemberGroupStatus.VALID);
+
+		// validate both members in Vo, otherwise they will be skipped
+		perun.getMembersManagerBl().validateMember(sess, member1);
+		perun.getMembersManagerBl().validateMember(sess, member2);
+
+		System.out.println("members:\n" + perun.getGroupsManagerBl().getGroupMembers(sess, group));
```
cuadradek, line 659: You probably forgot to delete this line.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -620,6 +639,282 @@ public void evaluateConsentsWithInsufficientUnsignedConsent() throws Exception {
...
+		assertEquals(0, shouldBeEmptyConsents.size());
+
+		// evaluate again, still should have only the one UNSIGNED consent
+		try {
+			BeansUtils.getCoreConfig().setForceConsents(true);
+			perun.getConsentsManagerBl().evaluateConsents(sess, consentHub);
+		} finally {
+			BeansUtils.getCoreConfig().setForceConsents(originalForce);
+		}
+		assertThat(consents).containsExactly(expectedConsent);
```
cuadradek, line 687: We should probably retrieve the consents list again after the second evaluation.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -620,6 +639,282 @@ public void evaluateConsentsWithInsufficientUnsignedConsent() throws Exception {
...
+
+			perun.getConsentsManagerBl().evaluateConsents(sess, consentHub);
+		} finally {
+			BeansUtils.getCoreConfig().setForceConsents(originalForce);
+			service.setUseExpiredMembers(originalUseExpiredMembers);
+			perun.getServicesManagerBl().updateService(sess, service);
+
+		}
+
+		assertThat(consents1).containsExactly(expectedConsent1);
```
cuadradek, line 747: Same here.

cuadradek (changes requested):

xflord (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
```diff
@@ -282,4 +292,30 @@ public interface ConsentsManagerBl {
...
+	 * and compares it against attributes from users' Consent objects that are
+	 * assigned to resources of given ConsentHub object.
+	 *
+	 * Service defines whether only active users will be evaluated or expired ones as well.
+	 *
+	 *
+	 * @param sess session
+	 * @param consentHub consent hub
+	 */
+	ConsentHub evaluateConsents(PerunSession sess, ConsentHub consentHub);
```
cuadradek, line 309:
```suggestion
	void evaluateConsents(PerunSession sess, ConsentHub consentHub);
```

perun-core/src/main/java/cz/metacentrum/perun/core/bl/ConsentsManagerBl.java
```diff
@@ -282,4 +292,30 @@ public interface ConsentsManagerBl {
...
+	/**
+	 * Consolidate consents using given service for all consent hubs the service is assigned to.
+	 *
+	 * Method finds all consent hubs that contain given service in any of its facilities and start
+	 * full consent consolidation for each consent hub (all services in consent hubs will be checked).
+	 *
+	 * @param sess session
+	 * @param service service
+	 */
+	Service evaluateConsents(PerunSession sess, Service service);
```
cuadradek, line 320:
```suggestion
	void evaluateConsents(PerunSession sess, Service service);
```

cuadradek (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ServicesManagerBlImpl.java
```diff
@@ -379,6 +377,9 @@ private ServiceAttributes getData(PerunSession sess, Service service, Resource r
 	}
 
 	private ServiceAttributes getData(PerunSession sess, Service service, Facility facility, Resource resource, boolean filterExpiredMembers) {
+		service.setUseExpiredMembers(!filterExpiredMembers);
+		servicesManagerImpl.updateService(sess, service);
```
cuadradek, line 381: I would maybe compare `service.isUseExpiredMembers()` and `filterExpiredMembers` and update the service only if they differ (so it should be updated at most once). We should probably call updateService in the BL layer instead of impl layer because an audit message is created there. What do you think?

Kuliak: Thank you for your suggestion, BL version of update will be much better. Also, service will be updated only if corresponding value is changed, however, I was thinking whether:

```**!filterExpiredMembers != service.isUseExpiredMembers()**```

would not be more clear, but now, there is simplified version of it.

cuadradek (approved):

PR #3558, opened 2022-04-28
feat(core): Manually run consents in gen scripts

mattjoke:
* Fixed getHashedHierarchicalData to work with new parameter
* Fixed getHashedDataWithGroups to work with new parameter
* Added openpai optional parameter for the gen scripts

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
@@ -296,6 +296,11 @@ public class ConsentsManagerBlImpl implements ConsentsManagerBl {
 
 	@Override
 	public List<Member> evaluateConsents(PerunSession sess, Service service, Facility facility, List<Member> members) {
+		return evaluateConsents(sess, service, facility, members, true);
+	}
+
+	@Override
+	public List<Member> evaluateConsents(PerunSession sess, Service service, Facility facility, List<Member> members, boolean consentEval) {
 		if (!BeansUtils.getCoreConfig().getForceConsents()) {
 			return members;
 		}
@@ -325,7 +330,7 @@ public class ConsentsManagerBlImpl implements ConsentsManagerBl {
 			.collect(groupingBy(Consent::getUserId, toMap(Consent::getStatus, Function.identity())));
 
 		return members.stream()
-			.filter(member -> hasValidConsent(sess, member, consentHub, requiredAttributes, userIdToConsents.get(member.getUserId())))
+			.filter(member -> hasValidConsent(sess, member, consentHub, requiredAttributes, userIdToConsents.get(member.getUserId()), consentEval))
 			.toList();
 	}
 
@@ -341,13 +346,16 @@ public class ConsentsManagerBlImpl implements ConsentsManagerBl {
 	 * @param consentHub the consent hub
 	 * @param requiredAttributes a list of attributes a consent should contain
 	 * @param usersConsents map of user's consents (consent status to the user's consent witch such status)
+	 * @param consentEval true if consent evaluation is enabled
 	 * @return true if the given member has a valid consent
 	 */
 	private boolean hasValidConsent(PerunSession sess, Member member, ConsentHub consentHub, List<AttributeDefinition> requiredAttributes,
-									Map<ConsentStatus, Consent> usersConsents) {
+									Map<ConsentStatus, Consent> usersConsents, boolean consentEval) {
 		if (usersConsents == null || usersConsents.isEmpty()) {
 			try {
-				this.createConsent(sess, new Consent(-1, member.getUserId(), consentHub, null));
+				if (consentEval) {
+					this.createConsent(sess, new Consent(-1, member.getUserId(), consentHub, null));
+				}
 			} catch (ConsentExistsException | ConsentHubNotExistsException | UserNotExistsException e) {
 				throw new InternalErrorException(e);
 			}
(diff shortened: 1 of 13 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/provisioning/HierarchicalHashedDataGenerator.java
```diff
@@ -106,12 +110,14 @@ private GenResourceDataNode getDataForResource(Resource resource) {
 		} else {
 			members = sess.getPerunBl().getResourcesManagerBl().getAllowedMembers(sess, resource);
 		}
-		if (BeansUtils.getCoreConfig().getForceConsents()) {
-			// remove the members without granted consents on required attributes
-			members.removeIf(member -> !membersWithConsent.contains(member));
-		} else {
-			// we skipped this part if consents were required, so add them now
-			membersWithConsent.addAll(members);
+		if (consentEval) {
```
xflord, line 113: Join into one IF statement. We still want to run the code in this else block if consentEval is false.

perun-core/src/main/java/cz/metacentrum/perun/core/provisioning/GroupsHashedDataGenerator.java
```diff
@@ -87,7 +89,9 @@ public HashedGenData generateData() {
 			} else {
 				membersToEvaluate = sess.getPerunBl().getFacilitiesManagerBl().getAllowedMembers(sess, facility, service);
 			}
-			membersWithConsent.addAll(sess.getPerunBl().getConsentsManagerBl().evaluateConsents(sess, service, facility, membersToEvaluate));
+			if (consentEval) {
```
xflord, line 92: Join into one if statement on line 85. We want to check if we force consents globally AND for that specific service

xflord (changes requested):
Also update the comments/documentation in ServicesAgent.pm to include the new parameter. Some of the tests are failing, because you forgot to update them with the new consentEval parameter -  try to run relevant tests yourself before pushing.

perun-cli/Perun/Agent.pm
```diff
@@ -56,6 +57,12 @@ use constant {
 	WRONG_AGENT_VERSION   => "Version of the command tools and Perun server mismatch",
 };
 
+$ENV{"CONSENT_EVAL"} = 0;
+GetOptions("consent-eval|c" => sub{
+	$ENV{"CONSENT_EVAL"} = 1;
+});
```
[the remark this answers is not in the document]
mattjoke: You are right, this approach enables the consent eval in every script using Agent.pm.

xflord (changes requested):
Also as @Johaney-s probably already told you, we need to make sure that automatic calls of services set consentEval to true (probably?). Ask around how/where this is done.

perun-core/src/main/java/cz/metacentrum/perun/core/provisioning/HierarchicalHashedDataGenerator.java
```diff
@@ -72,7 +73,7 @@ public HashedGenData generateData() {
 		List<Resource> resources =
 				sess.getPerunBl().getFacilitiesManagerBl().getAssignedResources(sess, facility, null, service);
 
-		if (BeansUtils.getCoreConfig().getForceConsents()) {
+		if (consentEval && BeansUtils.getCoreConfig().getForceConsents()) {
```
Johaney-s, line 76: we\ll want to filter out members that do not have consent even if calling it without the consentEval flag, we just want to skip creation of new consents in that case

perun-core/src/main/java/cz/metacentrum/perun/core/provisioning/HierarchicalHashedDataGenerator.java
```diff
@@ -106,7 +107,7 @@ private GenResourceDataNode getDataForResource(Resource resource) {
 		} else {
 			members = sess.getPerunBl().getResourcesManagerBl().getAllowedMembers(sess, resource);
 		}
-		if (BeansUtils.getCoreConfig().getForceConsents()) {
+		if (consentEval && BeansUtils.getCoreConfig().getForceConsents()) {
```
Johaney-s, line 110: we'll want to filter out members that do not have consent even if calling it without the consentEval flag, we just want to skip creation of new consents in that case

Johaney-s (approved):

balcirakpeter (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -343,7 +351,7 @@ private boolean hasValidConsent(PerunSession sess, Member member, ConsentHub con
 			return true;
 		} else if (revokedConsent != null && revokedConsent.getAttributes().containsAll(requiredAttributes)) {
 			return false;
-		} else if (unsignedConsent == null || !unsignedConsent.getAttributes().containsAll(requiredAttributes)) {
+		} else if (unsignedConsent == null || !unsignedConsent.getAttributes().containsAll(requiredAttributes) && consentEval) {
```
balcirakpeter, line 354: Please add extra parentheses so the condition is more understandable.
```suggestion
		} else if ((unsignedConsent == null || !unsignedConsent.getAttributes().containsAll(requiredAttributes)) && consentEval) {
```

balcirakpeter (changes requested):
Please update also Javadoc in `ServicesAgent.pm`  for `getHashedData` subroutines.

PR #3559, opened 2022-04-29
feat(cli): authentication and token revocation

sarkapalkovicova:
Added scripts for authentication and token revocation.

### DIFF
--- a/perun-cli/oidcAuthentication
+++ b/perun-cli/oidcAuthentication
@@ -1,0 +1,34 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+use Perun::auth::OidcAuth;
+use Perun::Exception;
+use Getopt::Long qw(:config no_ignore_case);
+
+sub help {
+	return qq{
+	Performs OIDC authentication which stores new access and refresh tokens. Can be used to renew tokens before performing batch operations.
+	------------------------------------
+	Available options:
+	--help                          | -h prints this help
+
+};
+}
+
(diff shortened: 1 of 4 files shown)


Johaney-s (commented):
this commit should be marked as breaking change, because it requires adding new property to config

cuadradek (commented):
Looks good, and I tested it - it works. I have only one suggestion - please, add _help_ method and batch mode to both scripts - just like in script _removeOidcTokens_.

cuadradek:
> Looks good, and I tested it - it works. I have only one suggestion - please, add _help_ method and batch mode to both scripts - just like in script _removeOidcTokens_.

We discussed this with @Johaney-s and batch mode for _oidcAuthentication_ script is not needed, because the authentication link would be printed anyway (when batch mode is on, nothing is printed but we have to print the authentication link anyway).

perun-cli/oidcAuthentication
```diff
@@ -0,0 +1,34 @@
...
+use strict;
+use warnings;
+
+use Perun::auth::OidcAuth;
+use Perun::Exception;
+use Getopt::Long qw(:config no_ignore_case);
+
+sub help {
+	return qq{
+	OIDC authentication
```
Johaney-s, line 12:
```suggestion
	Performs OIDC authentication which stores new access and refresh tokens. Can be used to renew tokens before performing batch operations.
```

Johaney-s (approved):

cuadradek (approved):

PR #3569, opened 2022-05-04
fix(core): renamed attribute module namespace_UIDRanges

sarkapalkovicova:
Fixed typo in name.


sarkapalkovicova:
A deployment note was created, but please check if SQL is correct. My VPN is not working, so I can not do it myself.

cuadradek (approved):

PR #3571, opened 2022-05-05
fix(core): add new Service parameter to AuditParser and parseRpcMethods

cuadradek:
- New parameter useExpiredMembers in Service object was missing in
AuditParser and in file parseRpcMethods.pl. Now, it is added there.


Johaney-s (approved):

zlamalp (commented):
Not sure if it's useful, but we have a test `AuditParserTest`, we should probably add check that (de)serailized property matches. Around line 328 like this:
```
assertEquals(service.isUseExpiredMembers(), ((Service) serviceInList.get(0)).isUseExpiredMembers());```

cuadradek:
> Not sure if it's useful, but we have a test `AuditParserTest`, we should probably add check that (de)serailized property matches. Around line 328 like this:
> 
> ```
> assertEquals(service.isUseExpiredMembers(), ((Service) serviceInList.get(0)).isUseExpiredMembers());```
> ```

Added, thanks.

zlamalp (approved):

PR #3572, opened 2022-05-05
fix(core): remove updating service during getData methods

cuadradek:
- Updating services during getData methods was added recently to set
useExpiredMembers parameter to DB. This raised exceptions because
getData methods run in read-only transactions.
- Updating was removed from getData methods and useExpiredMembers will
be set during deployment.


Kuliak (approved):
All update occurences removed, good job!

PR #3573, opened 2022-05-05
refactor(core): change default value of useExpiredMembers

cuadradek:
- Changed default value of useExpiredMembers in Service object to true
because most of the services use expired members.


zlamalp (changes requested):
Please fix also expected results in the tests:
```
ConsentsManagerEntryIntegrationTest.evaluateConsentsForConsentHubCreatesConsents:678 expected:<0> but was:<1>
ConsentsManagerEntryIntegrationTest.evaluateConsentsForServiceCreatesConsents:806 expected:<0> but was:<1>
```

PR #3575, opened 2022-05-09
fix(core): don't include attributes from services that don't process expired members if user is assigned as expired member in that resource

xflord:
* check in createConsent() whether direct MemberGroupStatus of member in all the assigned groups to the resource is expired - if so and the service skips expired members, skip here as well
* wrote test to cover changes made

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
@@ -9,12 +9,15 @@ import cz.metacentrum.perun.core.api.Consent;
 import cz.metacentrum.perun.core.api.ConsentHub;
 import cz.metacentrum.perun.core.api.ConsentStatus;
 import cz.metacentrum.perun.core.api.Facility;
+import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.MemberGroupStatus;
 import cz.metacentrum.perun.core.api.PerunBean;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.Service;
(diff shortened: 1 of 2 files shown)


Johaney-s (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
```diff
@@ -90,7 +95,20 @@ public Consent createConsent(PerunSession sess, Consent consent) throws ConsentE
...
+				Member member;
+				try {
+					Vo vo = getPerunBl().getVosManagerBl().getVoById(sess, resource.getVoId());
+					member = getPerunBl().getMembersManagerBl().getMemberByUser(sess, vo, user);
+				} catch (VoNotExistsException | MemberNotExistsException e) {
+					throw new InternalErrorException(e);
+				}
+				List<Group> groups = getPerunBl().getGroupsManagerBl().getAssignedGroupsToResource(sess, resource, member);
+				boolean isExpiredInResource = groups.stream()
+							.allMatch(group -> getPerunBl().getGroupsManagerBl().getDirectMemberGroupStatus(sess, member, group).equals(MemberGroupStatus.EXPIRED));
```
cuadradek, line 107:
```suggestion
							.allMatch(group -> getPerunBl().getGroupsManagerBl().getTotalMemberGroupStatus(sess, member, group).equals(MemberGroupStatus.EXPIRED));
```
The difference is that a member could have EXPIRED direct status and VALID indirect status through a subgroup. But totally, he is VALID. What do you think?

cuadradek (commented):
Good job, just one small suggestion.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ConsentsManagerBlImpl.java
xflord: You're right, I wasn't sure which one to use.

cuadradek (approved):

mattjoke (approved):

PR #3581, opened 2022-05-11
feat(core): add force property to anonymizeUsers method

xkureck:
* add possibility to call method anonymizeUsers with force switch to remove the user from groups and vos

### DIFF
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
@@ -507,12 +507,29 @@ public enum UsersManagerMethod implements ManagerMethod {
 	 * @throw RelationExistsException When the User has some members assigned.
 	 * @throw AnonymizationNotSupportedException When an attribute should be anonymized but its module doesn't specify the anonymization process.
 	 */
+	/*#
+	 * Anonymizes user (force) - according to configuration, each of user's attributes is either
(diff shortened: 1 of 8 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -507,12 +507,29 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 
 		@Override
 		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
 			parms.stateChangingCheck();
-			ac.getUsersManager().anonymizeUser(ac.getSession(), ac.getUserById(parms.readInt("user")));
+
+			if (parms.contains("force") && parms.readBoolean("force")) {
+				ac.getUsersManager().anonymizeUser(ac.getSession(), ac.getUserById(parms.readInt("user")), true);
+			} else {
+				ac.getUsersManager().anonymizeUser(ac.getSession(), ac.getUserById(parms.readInt("user")));
```
xflord, line 531:
```suggestion
				ac.getUsersManager().anonymizeUser(ac.getSession(), ac.getUserById(parms.readInt("user")), false);
```
Maybe we can do this and keep only the version of anonymizeUser with the force parameter for cleaner code, what do you think?

xflord (changes requested):

cuadradek (approved):

xflord (approved):

PR #3582, opened 2022-05-11
feat(cli): Added useExpiredMembers parameter

mattjoke:
* added useExpiredMembers to Service.pm
* added parameter to createService script
* fixed undefined description bug

### DIFF
--- a/perun-cli/createService
+++ b/perun-cli/createService
@@ -15,13 +15,14 @@ sub help {
 	--description  | -d service description
 	--script       | -s service script path (default './service_name')
 	--enabled      | -e enabled (default 'true')
+	--expired      | -x use expired members (default 'true')
 	--batch        | -b batch
 	--help         | -h prints this help
 
 	};
 }
 
-my ($name, $description, $script, $enabled, $batch);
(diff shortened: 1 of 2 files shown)


perun-cli/createService
```diff
@@ -15,13 +15,14 @@ sub help {
 	--description  | -d service description
 	--script       | -s service script path (default './service_name')
 	--enabled      | -e enabled (default 'true')
+	--expired      | -x use expired members (default 'false')
```
cuadradek, line 18: We recently changed the `useExpiredMembers` parameter to have `true` as a default value in `perun-core` because most of the services use expired members. So I think it would be better to have it as `true` here as well.

cuadradek (changes requested):
Looks good, I noticed a little bug - useExpiredMembers is not printed e.g. in script `listOfServices`. It is most likely because it is uninitialized - renaming `$self->{_expired}` to `$self->{_useExpiredMembers}` should fix it.

cuadradek (approved):

xflord (approved):

PR #3583, opened 2022-05-11
feat(cli): support multiple OIDC configurations

xflord:
* user can now use multiple OIDC configurations through CLI
* configurations are stored in oidc_config.yml as dictionary entries
* to swap between configurations, change environmental variable PERUN_OIDC_CONFIG to desired configuration id
* edited revokeOidcToken to support revoking specific/all configurations' tokens, same with removeOidcTokens
* added listOidcTokens, which lists all the configurations and their stored tokens (if there are any)

BREAKING CHANGE: config file oidc_config.yml changed, see sample_oidc_config.yml

### DIFF
--- a/perun-cli/Perun/auth/OidcAuth.pm
+++ b/perun-cli/Perun/auth/OidcAuth.pm
@@ -9,21 +9,34 @@ use LWP::UserAgent;
 use File::Basename;
 use Perun::Exception;
 
-my $PERUN_OIDC = "perun_oidc";
+my $chosenConfigName = $ENV{PERUN_OIDC_CONFIG};
+if (!defined $chosenConfigName) {
+	die "ERROR: environmental variable PERUN_OIDC_CONFIG not set!\n"
+}
+my $PERUN_OIDC = "perun_oidc" . "_" . $chosenConfigName;
 my $PYTHON = "python3";
 my $dirname = dirname(__FILE__);
+my $configFilename = $dirname . '/oidc_config.yml';
 my $contentType = "application/x-www-form-urlencoded";
 my $authenticationFailed = "Authentication failed";
 my $tokenRevocationFailed = "Token revocation failed";
 
 sub getAccessToken
 {
+	if (defined $_[0]) {
+		$PERUN_OIDC = "perun_oidc" . "_" . $_[0];
+	}
+
 	my $ret = `$PYTHON -c "import keyring; print('token:' + str(keyring.get_password('$PERUN_OIDC', 'access_token')), end='')"`;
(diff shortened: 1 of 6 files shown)


sarkapalkovicova:
I tested it, and it seems to be working right. I have no problem with the code. GJ

Johaney-s (commented):
really nice job, i think the format of listing the tokens is okay also.

please mark this commit with breaking change as it requires changing the oidc config file

Johaney-s (approved):

perun-cli/Perun/auth/OidcAuth.pm
```diff
@@ -268,22 +289,78 @@ sub loadAccessToken
...
 	`$PYTHON -c "import keyring; keyring.delete_password('$PERUN_OIDC', 'access_token_validity')"`;
 }
 
+sub removeAllAccessTokens
+{
+	my %config = % { loadAllConfigurations() };
+	foreach my $key (keys %config) {
+		if (getAccessToken($key)) {
+			removeAccessToken($key);
+			print("Access token from configuration $key successfully removed from keyring.\n");
```
cuadradek, line 306: I think we can omit printing in this file because we also print it in separate scripts where we also consider whether batch mode is on (if it is on, nothing should be printed). What do you think @Johaney-s @xflord ?

perun-cli/revokeOidcToken
```diff
@@ -14,33 +14,45 @@ sub help {
...
+	Perun::auth::OidcAuth::revokeAllRefreshTokens();
+} else {
+	Perun::auth::OidcAuth::checkConfigExists($configId);
+	if (Perun::auth::OidcAuth::getRefreshToken($configId)) {
 
-if (Perun::auth::OidcAuth::getRefreshToken()) {
-
-	Perun::auth::OidcAuth::revokeToken(Perun::auth::OidcAuth::getRefreshToken(), "refresh_token");
-	Perun::auth::OidcAuth::removeRefreshToken();
+		Perun::auth::OidcAuth::revokeToken($configId, "refresh_token");
```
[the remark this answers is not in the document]
xflord: revokeToken() removes the token as well (see end of the method)

perun-cli/Perun/auth/OidcAuth.pm
xflord: I’m not too familiar with how batch mode but I will remove the prints in OidcAuth. Question is, do we want to move the prints into the scripts or is the amount of information provided by them enough?

cuadradek (approved):

PR #3588, opened 2022-05-16
feat(core): virt member attribute module mails

Johaney-s:
* virtual module collects member's mail addresses
* by default returns list of preferredMail and memberMail attributes
* attributes can be changed in module configuration file
* supports user and member attributes with email syntax
* works with string and list attributes


Johaney-s:
The custom configuration for manual testing can be put in `/etc/perun/module/urn_perun_member_attribute_def_virt_mails.yaml`, sample:

```
attributeNames:
  - urn:perun:user:attribute-def:def:preferredMail
  - urn:perun:member:attribute-def:def:mail
```

xflord (approved):
Tested manually as well and seems to be working as intended.

cuadradek (approved):

PR #3595, opened 2022-05-18
fix(cli): don't require PERUN_OIDC_CONFIG variable if not used

Johaney-s:
* env variable PERUN_OIDC_CONFIG was required to be defined even if not used, which is now fixed


xflord (approved):
Oh yeah it's environment not environmental haha, good job.

cuadradek (approved):

PR #3597, opened 2022-05-23
feat(core): allow observer roles on operand group to create union

Johaney-s:
* we want to extend the rights on creating and removing group unions such that observer role is sufficient on the operand group to create/remove a union

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -2008,17 +2008,37 @@ perun_policies:
     include_policies:
       - default_policy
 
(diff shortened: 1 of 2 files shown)


xkureck (commented):
When you are changing some policy names, you need to check if those policies are used in GUI
this policies are used in GUI, so you need to change them also there

Johaney-s:
> When you are changing some policy names, you need to check if those policies are used in GUI this policies are used in GUI, so you need to change them also there

Thanks, created https://github.com/CESNET/perun-web-apps/pull/1044 for that

cuadradek (approved):

PR #3612, opened 2022-06-01
feat(cli): use new method of attribute rights in cli scripts

xflord:


Johaney-s (commented):
Good job, looks good and works as expected. Please make the scripts executable and squash the commits. The printed table of policy collections is not as nicely formatted as the previous version's one, but that's just a cosmetic nitpick. Logic-wise I think we should add in a separate commit check on backend that the combination role+object makes sense before we save it to db.

perun-cli/setAttributePolicyCollections
```diff
@@ -0,0 +1,102 @@
...
+	Available options:
+	--attributeId       | -a attribute identifier
+	--attributeName     | -A attribute name (including namespace)
+	--policyCollections | -p list of policy collections in the following format:
+
+	 	[ACTION]=[ROLE]:[OBJECT]&&[ROLE]:[OBJECT]||[ACTION]=[ROLE]:[OBJECT]&&[ROLE]:[OBJECT]
+
+	 	ACTION = R (Read), W (Write), R+W (Read + Write)
+	 	ROLE   = one of the available Perun roles.
+	 	OBJECT = None, Group, Vo, Facility, Resource, User, Member
```
Johaney-s, line 24:
```suggestion
	 	OBJECT = None, Group, Vo, Facility, Resource, User, Member, SecurityTeam
```

Johaney-s: Please update also openapi's RoleObject definition as we forgot to update it before. Then it looks all good to me, I like the new formatting of the get script output.

Johaney-s (approved):

cuadradek (approved):

PR #3620, opened 2022-06-08
feat(core): check Role + RoleObject combinations when setting attribute policies

xflord:
* added a check whether the combination of Role + RoleObject in policies makes sense
* added a test to cover changes made
* edited other tests to work with this change

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -6872,6 +6872,7 @@ perun_roles_management:
       - PERUNOBSERVER:
     associated_read_roles:
       - PERUNOBSERVER
+    assignable_to_attributes: false
 
   PERUNOBSERVER:
     primary_object:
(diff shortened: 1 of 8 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -7159,6 +7179,7 @@ perun_roles_management:
       - VOADMIN: Vo
       - VOOBSERVER: Vo
     associated_read_roles: []
+    assignable_to_attributes: false
```
cuadradek, line 7182: I think this should be true, we should be already able to evaluate attribute rights for SPONSOR role.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6957,14 +6962,18 @@ perun_roles_management:
...
       - GROUPOBSERVER: Group
     associated_read_roles: []
+    assignable_to_attributes: true
 
   SELF:
     primary_object:
-    assign_to_objects: {}
+    assign_to_objects:
+      User: user_id
+      Member: member_id
```
cuadradek, line 6971: We decided that we will not use SELF-Member combination, you can delete it.

cuadradek (approved):

Johaney-s (commented):
Looks good, you can add the new exception also to AttributesManagerMethod.

PR #3630, opened 2022-06-15
feat(registrar): prefill login app items with reserved logins

cuadradek:
- Reserved logins used to be tied to applications with 1 to 1
relationship.
- Now, reserved logins are tied to users with many to one relationship.
- Reserved logins can now be used to prefill application USERNAME items
with same namespace.
- Reserved logins can now be deleted from DB and KDC only if they are
not used in multiple applications (previously they were always used by
only one application, hence we didn't have to check it).
- Deleting reserved logins when deleting groups / vos was refactored -
the common logic was extracted to UsersManager.


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1259,8 +1257,8 @@ public Application createApplicationInternal(PerunSession session, Application a
 					if (loginAvailable) {
 						try {
 							// Reserve login
-							jdbc.update("insert into application_reserved_logins(login,namespace,app_id,created_by,created_at) values(?,?,?,?,?)",
-									login, loginNamespace, appId, application.getCreatedBy(), new Date());
+							jdbc.update("insert into application_reserved_logins(login,namespace,user_id,extsourcename,created_by,created_at) values(?,?,?,?,?)",
+									login, loginNamespace, userId, session.getPerunPrincipal().getExtSourceName(), session.getPerunPrincipal().getActor(), new Date());
```
zlamalp, line 1261: Related to this:
```
application.getCreatedBy() -> session.getPerunPrincipal().getActor()
```
we should just check how is original value constructed in WUI/registrar and if it can be changed.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1366,23 +1364,14 @@ public void deleteApplication(PerunSession sess, Application app) throws PerunEx
...
-				// Try to get reservedLogin and reservedNamespace before deletion
-				List<Pair<String, String>> logins;
-				try {
-					logins = jdbc.query("select namespace,login from application_reserved_logins where app_id=?", (resultSet, arg1) -> new Pair<>(resultSet.getString("namespace"), resultSet.getString("login")), app.getId());
-				} catch (EmptyResultDataAccessException e) {
-					// set empty logins
-					logins = new ArrayList<>();
-				}
-				// delete passwords in KDC
+				List<Pair<String, String>> logins = usersManager.getReservedLoginsOnlyByGivenApp(sess, app.getId());
```
zlamalp, line 1367: I wonder, if we shouldn't have some kind of synchronization, since we could be deleting various applications for the same user with same reserved login at once and it could be resolved like there is still some other application holding the reserved login and it would be kept.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1476,20 +1465,17 @@ public Application rejectApplication(PerunSession sess, int appId, String reason
 			log.info("Application {} marked as REJECTED.", appId);
 
 			// get all reserved logins
-			List<Pair<String, String>> logins = jdbc.query("select namespace,login from application_reserved_logins where app_id=?",
-					(resultSet, arg1) -> new Pair<>(resultSet.getString("namespace"), resultSet.getString("login")), appId);
-
-			// delete passwords for reserved logins
+			List<Pair<String, String>> logins = usersManager.getReservedLoginsOnlyByGivenApp(sess, app.getId());
```
zlamalp, line 1468: I think we are repeating this part of code in deleteApplication() and rejectApplication() and it could be extracted to own private method.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2677,6 +2665,20 @@ public List<ApplicationFormItemWithPrefilledValue> getFormItemsWithPrefilledValu
 
 			}
 
+			// prefill USERNAME items with reserved logins
+			if (itemW.getFormItem().getType() == USERNAME && (itemW.getPrefilledValue() == null || itemW.getPrefilledValue().isEmpty())) {
```
zlamalp, line 2669: You should iterate over form items and if this login is prefilled, you should remove password form item for this namespace if its present. We do the same when login is prefilled from attribute or federation.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
zlamalp: When I think about it more, we could probably do something like select for update when getting specific namespace/login combination from reserved logins table so we can be sure, that other processes are waiting for the first one to complete its thing.

perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,19 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.95
+ALTER TABLE application_reserved_logins ADD COLUMN user_id integer;
+ALTER TABLE application_reserved_logins ADD COLUMN extsourcename varchar;
+UPDATE application_reserved_logins arl SET user_id = a.user_id FROM application a WHERE arl.app_id = a.id;
```
zlamalp, line 12: `user_id` will be probably null in most cases when migrating data, but I guess its ok, because of the way you select reserved logins.

zlamalp (changes requested):
Overall good work, thank you. Just a few small things and probably we should think about proper locking when working with reserved logins.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
cuadradek: The principal's actor is used as createdBy in WUI, I just thought it would be safer to use the session's actor rather than what is sent from frontend to RPC (they should be the same, though).

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
cuadradek: Good point, I added select for update to every select from the reserved logins table.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
zlamalp: Ok, then its fine.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -2677,6 +2673,24 @@ public List<ApplicationFormItemWithPrefilledValue> getFormItemsWithPrefilledValu
 
 			}
 
+			// prefill USERNAME items with reserved logins
+			if (itemW.getFormItem().getType() == USERNAME || itemW.getFormItem().getType() == PASSWORD) {
+				String destAttribute = itemW.getFormItem().getPerunDestinationAttribute();
```
zlamalp, line 2678: I'm undecided if we should use source or destination attribute here. Generally when pre-filling form items we check source attribute. We could theoretically want to pre-fill login from another namespace (some kind of migration or requirement for the same login), but in the end we store reserved logins by the destination attribute namespace, so I think its fine like this.

zlamalp (approved):
I think its ready for manual testing.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
zlamalp: Just a note, since we select login,namespace for the same user one by one, two processes could lock each other, but we don't generally use more than one login on registration form and probability of this is very low. We shall see in our manual tests.

xflord (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
cuadradek: I think this is prevented by locking them in alphabetical order.

PR #3632, opened 2022-06-21
feat: method to retrieve rich resources assigned to facility with selected service assigned

xflord:
* overloaded getAssignedRichResources to also accept service as an argument


cuadradek (commented):
Good job, I would just add a simple test again.

cuadradek (approved):

PR #3633, opened 2022-06-21
feat: method to retrieve services assigned to resources with selected vo on selected facility

xflord:
* overloaded getAssignedServices to also accept vo as an argument


cuadradek (commented):
Good job, I would just add a simple test for the new method.

Johaney-s (approved):

PR #3636, opened 2022-06-23
fix(core): consents optimization

Kuliak:
Before, every service under every consent hub that contained selected service, was evaluated. Now, only selected service on each consent hub (and each facility) is evaluated instead. 

In other words, when we have valid consent (with all required attributes) for given service, we do not care for other services. On the other hand, when consent does not have all the required attributes, we create a new one. While creating a new consent, all the facilities with all the services under given consent hub are checked, thus all required attributes from all services will be present in the new consent.

Note: Evaluation time for service 921 went from around 100s to 16s.

* new method for checking whether service is assigned to facility
* new private method created that uses selected service to evaluate consents only for its members
* if there was no selected service (service is null), all services assigned to facilities under processed consent hub are used instead
* tests that used old logic (evaluate everything for each consent hub that contains selected service) were removed, since evaluation of other services is not done anymore.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
@@ -52,8 +52,8 @@ public class ConsentsManagerEntryIntegrationTest extends AbstractPerunIntegratio
 	private Facility facility;
 	private Resource resource;
 	private Service service;
-	private Vo vo, otherVo;
-	private Group group, otherGroup;
+	private Vo vo;
+	private Group group;
 	private Member member;
 	private AttributeDefinition attrDef;
 	private AttributeDefinition facAttrDef;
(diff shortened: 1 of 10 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -967,21 +822,53 @@ public void createConsentNotIncludeAttrFromExpiredServices() throws Exception {
 	private ConsentHub setUpAnotherConsentHub() throws Exception {
```
xflord, line 822: Not quite sure what these changes are for. Couldn't we delete this method, since it was only used in the tests that you deleted?

xflord (changes requested):
I like your solution, but please edit the javadocs of evaluateConsents(PerunSession sess, Service service) to reflect the changes you made.

Johaney-s (commented):
I have no more comments to add to @xflord's.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
Kuliak: Thank you for your suggestions! I edited the javadoc in several layers, openAPI including.
As for the test part, I wanted to leave it there for future needs (if needed), however, it makes more sense to remove it now, since it is not used at all.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ConsentsManagerMethod.java
```diff
@@ -234,8 +234,8 @@ public Void call(ApiCaller ac, Deserializer params) throws PerunException {
 	/*#
 	 * Returns Service.
```
Johaney-s, line 235: You can remove this part and @ return parameter, also in the evaluateConsentsForConsentHub, since it returns nothing.

perun-openapi/openapi.yml
```diff
@@ -17387,7 +17387,7 @@ paths:
       tags:
         - ConsentsManager
       operationId: evaluateConsentsForService
-      summary: Evaluates consents for all consent hubs with given service.
+      summary: Evaluates consents ONLY for given service on consent hubs containing it.
```
Johaney-s, line 17390: This summary sounds to me like we're ignoring other hub's services' attributes when creating the consents, which we already discussed. Maybe we could explain it a bit more in the docs and the summary, what do you think?

xflord:
I agree with @Johaney-s , also the changes to openapi.yml could theoretically be in a seperate commit.

Johaney-s (approved):

xflord (approved):

PR #3639, opened 2022-06-24
feat(mvn): upgrade to Spring Boot 2.7.1 / Spring 5.3.21

martin-kuba:
upgraded Spring Boot and all its dependencies


Johaney-s (commented):
More suitable commit type would be `build`.

martin-kuba:
> More suitable commit type would be `build`.

Then it would not be included in the generated list of changes in a release. It is important to have it there for comparing versions against known vulnerabilities in dependencies.

PR #3642, opened 2022-06-28
feat: method for reserving login for user

xflord:
* created method reserveLogin() in all layers
* it sets the user attribute loginNamespace in the selected namespace for the selected user, effectively reserving the login
* added the method to openapi

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
@@ -38,6 +38,7 @@ import cz.metacentrum.perun.core.api.exceptions.PasswordResetLinkExpiredExceptio
 import cz.metacentrum.perun.core.api.exceptions.PasswordResetLinkNotValidException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthFailedException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.RelationExistsException;
 import cz.metacentrum.perun.core.api.exceptions.RelationNotExistsException;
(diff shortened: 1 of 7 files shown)


bodnara (commented):
Looks good to me. Only question would be, if we want to write tests for this ?

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/UsersManagerMethod.java
```diff
@@ -1774,6 +1774,29 @@ public User call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @throw InvalidLoginException When login to reserve has invalid syntax or is not allowed.
+	 * @throw AlreadyReservedLoginException When login is already reserved
+	 */
+	reserveLogin {
+		@Override
+		public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
+			parms.stateChangingCheck();
+			ac.getUsersManager().reserveLogin(ac.getSession(),
+				ac.getUserById(parms.readInt("user")),
+				parms.readString("userLogin"),
```
Johaney-s, line 1795: openapi's login is called just "login", so i think it wouldn't match to "userLogin" here, the same goes for "namespace" vs "loginNamespace"...

perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
```diff
@@ -38,6 +38,7 @@
 import cz.metacentrum.perun.core.api.exceptions.PasswordResetLinkNotValidException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthFailedException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
```
Johaney-s, line 41: unused import

zlamalp (commented):
We already have similar API logic - method `setLogin()` which is used to set login for service accounts by their owners. I'm not sure about this method and its relation to VOADMIN role, we should probably discuss it.

PR #3646, opened 2022-06-29
feat(core): Added consolidator and linker to config

mattjoke:
BREAKING CHANGE: properties "consolidator" and "linker" added to the perun-apps-config.yml


xflord (commented):
Looks good, just mark the commit as 'breaking change' since we're editing config files.

xflord:
also we don't use scope 'base', use 'core' instead.

xflord (approved):

PR #3648, opened 2022-06-30
feat(cli): default OIDC configuration

xflord:
* when the environment variable PERUN_OIDC_CONFIG isn't defined, the alphabetically first config from the config file is used by default.

### DIFF
--- a/perun-cli/Perun/auth/OidcAuth.pm
+++ b/perun-cli/Perun/auth/OidcAuth.pm
@@ -8,14 +8,13 @@ use JSON::XS;
 use LWP::UserAgent;
 use File::Basename;
 use Perun::Exception;
+use Try::Tiny;
 
-my $UNDEFINED_CONFIGURATION = "UNDEFINED_OIDC_CONFIGURATION";
-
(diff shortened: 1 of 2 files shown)


perun-cli/Perun/auth/OidcAuth.pm
```diff
@@ -386,13 +384,20 @@ sub revokeToken
...
 	if (!exists $configs{$configName}) {
 		die "ERROR: configuration \"$configName\" does not exist!\n";
 	}
 }
 
+sub getDefaultConfig
+{
+	my %configs = % { loadAllConfigurations() };
+	if (length keys %configs eq 0) {
+		die "ERROR: configuration file is empty!\n";
```
Johaney-s, line 398: This for some reason does not get thrown, but compilation error arises.

Johaney-s (changes requested):

perun-cli/Perun/auth/OidcAuth.pm
xflord: A compilation error would occur when converting an empty file to hash, so used try block instead.

Johaney-s (approved):

cuadradek (approved):

PR #3654, opened 2022-07-13
feat(core): added consolidate method and call to userInfoEndpoint whe…

xkureck:
…n creating Perun Principal

* added new method for consolidate with access token
* added new class
UserInfoEndpoint which is making calls to proxy's userInfoEndpoint and
from that is retrieving information about sub/issuer/email/.. of logged
user
* added new class EnrichedExtSource and EnrichedIdentity

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/oidc/UserInfoEndpointCall.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/oidc/UserInfoEndpointCall.java
@@ -1,0 +1,150 @@
+package cz.metacentrum.perun.oidc;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.exceptions.ExpiredTokenException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpEntity;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * Class for executing call to User info endpoint.
+ *
+ * @author Lucie Kureckova <luckureckova@gmail.com>
+ */
+public class UserInfoEndpointCall {
+
+	private final static Logger log = LoggerFactory.getLogger(UserInfoEndpointCall.class);
+
+	public UserInfoEndpointResponse getUserInfoEndpointData(String accessToken, String issuer, Map<String, String> additionalInformation) throws ExpiredTokenException {
+		RestTemplate restTemplate = new RestTemplate();
+		JsonNode config = restTemplate.getForObject(issuer + "/.well-known/openid-configuration", JsonNode.class);
+		HttpHeaders headers = new HttpHeaders();
+		headers.setBearerAuth(accessToken);
+		HttpEntity<Void> entity = new HttpEntity<>(headers);
+		ResponseEntity<JsonNode> userInfoResponse;
+		try {
+			userInfoResponse = restTemplate.exchange(config.path("userinfo_endpoint").textValue(), HttpMethod.GET, entity, JsonNode.class);
+		} catch (HttpClientErrorException ex) {
+			if (ex.getStatusCode() == HttpStatus.FORBIDDEN) {
+				log.error("Token {} is no longer valid for issuer: {}", accessToken, issuer);
+				throw new ExpiredTokenException("Your token is no longer valid. Please retry from the start.");
+			} else {
+				log.error("Failed to get userInfoResponse for access token: {}. The response code was: {}", accessToken, ex.getStatusCode());
+				throw new InternalErrorException("Failed to get userInfoResponse for access token: " + accessToken + ". The response code was: " + ex.getStatusCode());
+			}
+		}
+		JsonNode userInfo = userInfoResponse.getBody();
+		if(StringUtils.isNotEmpty(userInfo.path("error").asText())) {
+			log.error("Call to user info endpoint failed, the error is: {}", userInfo);
+			throw new InternalErrorException("Call to user info endpoint failed, the error is" + userInfo);
+		}
+		log.debug("user info retrieved: {}", userInfo);
+
+		fillAdditionalInformationWithDataFromUserInfo(userInfo, additionalInformation);
+
+		String extSourceName = getExtSourceName(userInfo);
+
+		String extSourceLogin = getExtSourceLogin(userInfo);
+
+		return new UserInfoEndpointResponse(extSourceName, extSourceLogin);
+	}
+
+	/**
+	 * Parsing extsource login (sub) from user info.
+	 * We are looking for the first nonempty property value that can be login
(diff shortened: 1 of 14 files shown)


perun-base/src/main/java/cz/metacentrum/perun/oidc/UserInfoEndpointCall.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Map;
+
+import static org.testcontainers.shaded.org.apache.commons.lang3.StringUtils.isEmpty;
+import static org.testcontainers.shaded.org.apache.commons.lang3.StringUtils.isNotEmpty;
+
+public class UserInfoEndpointCall {
+
+	private final static Logger log = LoggerFactory.getLogger(UserInfoEndpointCall.class);
+
+	public UserInfoEndpointResponse getUserInfoEndpointData(String accessToken, String issuer, Map<String, String> additionalInformation) throws ExpiredTokenException {
```
Kuliak, line 22: This method seems to be quite long. I would recommend to extract some of its parts to separate methods for better readability. For example, you might put retrieving of each part to separate method. What do you think?

perun-base/src/main/java/cz/metacentrum/perun/oidc/UserInfoEndpointResponse.java
```diff
@@ -0,0 +1,27 @@
+package cz.metacentrum.perun.oidc;
+
```
Kuliak, line 2: javadoc

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ConsolidatorManagerImpl.java
```diff
@@ -3,17 +3,14 @@
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import cz.metacentrum.perun.core.api.*;
-import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.ExtSourceNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.PerunException;
-import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
-import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.*;
```
Kuliak, line 6: normal imports instead of wildcard one

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/ConsolidatorManager.java
```diff
@@ -103,4 +103,11 @@ public interface ConsolidatorManager {
 	 */
 	List<UserExtSource> consolidateIdentityUsingToken(PerunSession sess, String token) throws PerunException;
 
+	/**
+	 * Join user identities by access tokens.
```
Kuliak, line 107: I would separate method description from list of its parameters.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ConsolidatorManagerImpl.java
```diff
@@ -460,8 +458,102 @@ public List<UserExtSource> consolidateIdentityUsingToken(PerunSession sess, Stri
 
 	}
 
+    @Override
+    public Void consolidate(PerunSession sess, String accessToken) throws PerunException {
+		Map<String, String> additionalInformationNonCaller = new HashMap<>();
+		UserInfoEndpointResponse userInfoNonCaller = userInfoEndpointCall.getUserInfoEndpointData(accessToken, BeansUtils.getCoreConfig().getOidcIssuersExtsourceNames().keySet().iterator().next(), additionalInformationNonCaller);
 
-	/**
+		//trying to find in the perun user with identity which we obtained throught access token
```
Kuliak, line 466: typo (throught -> through)

Kuliak (changes requested):
I think you did a great job, thank you! I have some small esthetic changes. 
There are some more like leaving no space between IF and its CONDITION, but you do not have to adjust it.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ConsolidatorManagerImpl.java
```diff
@@ -600,7 +692,20 @@ else if (ues.getExtSource().getName().equals("https://login.elixir-czech.org/idp
...
 							// don't care
 						}
-						es.add(ues.getExtSource());
+						Attribute nameAttr = null;
+						try {
+							nameAttr = perun.getAttributesManagerBl().getAttribute(registrarSession, ues, AttributesManager.NS_UES_ATTR_DEF + ":sourceIdPName");
+						} catch (Exception e) {
+							log.debug("Failed to get sourceIdpName attribute" + e);
+						}
+						var richExtSource = new RichExtSource(ues.getExtSource());
```
Johaney-s, line 701: so the proposed solution to the problem with the enriched ext sources is to return List of Identities from this method that doesn't contain any additional attributes, create new object EnrichedIdentity which would contain EnrichedExtSource and a new method that would "enrich" the Identity object to create enriched one from it. this way a new method could be added checking similar users that would be declared to return List of EnrichedIdentity and would allow us to know that it contains the attributes inside the EnrichedExtSource property of the EnrichedIdentity. the new method would first call the old one and then enrich the identities.

Johaney-s (commented):
a few suggestions besides what @Kuliak already mentioned

perun-base/src/main/java/cz/metacentrum/perun/oidc/UserInfoEndpointCall.java
```diff
@@ -0,0 +1,110 @@
...
+		if(isEmpty(extSourceName)) {
+			log.error("issuer was empty or null: {}", extSourceName);
+			throw new InternalErrorException("issuer was null: " + extSourceName);
+		}
+
+		//retrieve ext source login
+		List<String> pathToExtSourceLogin = BeansUtils.getCoreConfig().getUserInfoEndpointExtSourceLogin();
+		String login = "";
+		for(String path: pathToExtSourceLogin) {
+			JsonNode loginNode = userInfo.path(path);
```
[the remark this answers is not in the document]
xkureck: the search strategy for login is different - we are looking for first nonempty property, which can only be 'one level'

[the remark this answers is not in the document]
xkureck: I have added better comment and also I hope better named the variables

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ConsolidatorManagerImpl.java
```diff
@@ -617,4 +720,43 @@ else if (ues.getExtSource().getName().equals("https://login.elixir-czech.org/idp
...
+
+		if (list != null && !list.isEmpty()) {
+
+			for (Identity identity : list) {
+				Set<EnrichedExtSource> enrichedExtSources = new HashSet<>();
+
+				for (ExtSource extSource : identity.getIdentities()) {
+					Attribute nameAttr = null;
+					UserExtSource ues = new UserExtSource();
+					ues.setId(extSource.getId());
```
Johaney-s, line 740: this seems odd - we probably don't want to use the extSource id as ues id

Johaney-s (approved):

xflord (approved):

cuadradek (approved):

PR #3656, opened 2022-07-17
Mailing services resources with member

Kuliak:
We were supposed to optimize workflow where we need to filter only those resources that have mailing service(s) assigned (mailman...) where given member is assigned. 

One possible approach could have been a change of policies for current methods, however, I have decided to create new method instead with its own policy to keep current policies consistent.


cuadradek (commented):
Just a thought: `getMailingServiceRichResources` could return resources which have an assigned service with the required attribute `optOutMailingList` instead of an service with name `mailman%`. Or we could do it without a new method: use `getServicesByAttributeDefinition` and then for every returned service `getAssignedRichResources(member, service)`. What do you think?

Kuliak:
> Just a thought: `getMailingServiceRichResources` could return resources which have an assigned service with the required attribute `optOutMailingList` instead of an service with name `mailman%`. Or we could do it without a new method: use `getServicesByAttributeDefinition` and then for every returned service `getAssignedRichResources(member, service)`. What do you think?

Yes, I was thinking about the same thing when creating the new method. Your suggestion using AttributeDefinition looks nice, I think it is worth trying. Thank you!

Kuliak:
> > > Just a thought: `getMailingServiceRichResources` could return resources which have an assigned service with the required attribute `optOutMailingList` instead of an service with name `mailman%`. Or we could do it without a new method: use `getServicesByAttributeDefinition` and then for every returned service `getAssignedRichResources(member, service)`. What do you think?
> > 
> > 
> > Yes, I was thinking about the same thing when creating the new method. Your suggestion using AttributeDefinition looks nice, I think it is worth trying. Thank you!
> 
> On the other hand, `getServicesByAttributeDefinition` can be called only by Perun admin and we probably want to keep it this way; so this approach wouldn't work.

Yes, that's correct, however, I was thinking this might be done by using getServicesByAttributeDefinition in business layer while creating new method for it as well. What do you think?

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -783,6 +784,17 @@ public List<RichResource> getRichResources(PerunSession sess, Vo vo) {
 		return getResourcesManagerImpl().getRichResources(sess, vo);
 	}
 
+	@Override
+	public List<RichResource> getMailingServiceRichResourcesWithMember(PerunSession sess, Member member) {
+		AttributeDefinition ad;
+		try {
+			ad = getPerunBl().getAttributesManagerBl().getAttributeDefinition(sess, AttributesManager.NS_ENTITYLESS_ATTR_DEF + ":optOutMailingList");
```
cuadradek, line 791: I believe the attribute is a member-resource attribute, not entityless.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/ResourcesManagerBlImpl.java
```diff
@@ -783,6 +784,17 @@ public List<RichResource> getRichResources(PerunSession sess, Vo vo) {
...
+		} catch (AttributeNotExistsException e) {
+			throw new InternalErrorException(e);
```
cuadradek, line 793: Maybe we can return an empty list instead of this exception, because the attribute definition is not created during Perun startup, thus it doesn't have to be used on every Perun instance. In such cases, we probably don't want the operation to fail.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ResourcesManagerEntryIntegrationTest.java
```diff
@@ -3076,6 +3117,20 @@ private Attribute setUpAttribute3() throws Exception {
 		return attribute;
 	}
 
+	private Attribute setUpMailingListAttribute() throws Exception {
+		AttributeDefinition attrDef = new AttributeDefinition();
+		attrDef.setNamespace(AttributesManager.NS_ENTITYLESS_ATTR_DEF);
```
cuadradek, line 3122: Again, member-resource instead of entityless.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
```diff
@@ -565,6 +565,15 @@ public interface ResourcesManager {
 	 */
 	List<RichResource> getRichResources(PerunSession perunSession, Vo vo) throws PrivilegeException, VoNotExistsException;
 
+	/**
+	 * Return all rich resources with mailing service(s) where given member is assigned.
+	 *
+	 * @param perunSession session
+	 * @param member member
+	 * @return list of corresponding rich resources
+	 */
```
xflord, line 574:
```suggestion
	 * @throws MemberNotExistsException
	 */
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ResourcesManagerMethod.java
```diff
@@ -961,6 +976,19 @@ public List<RichResource> call(ApiCaller ac, Deserializer parms) throws PerunExc
 		}
 	},
 
+	/*#
+	 * Return all rich resources with mailing service(s) where given member is assigned.
+	 *
+	 * @param member int VO <code>id</code>
+	 * @return List<RichResource> member resources with mailing service
+	 */
```
xflord, line 984:
```suggestion
	 * @throw MemberNotExistsException
	 */
```

xflord: nitpick

perun-core/src/main/java/cz/metacentrum/perun/core/api/ResourcesManager.java
xflord: nitpick

xflord (commented):
Seems good with the changes Radek suggested, just maybe squash the commits.

xflord (approved):

cuadradek (approved):

PR #3660, opened 2022-07-25
fix(openapi): group in Application and ApplictaionForm can be null

xpavlic:
On cesnet-devel there are Applications and ApplicationForms with group set as null.
In openapi generated by python generator, there is problem when group is null (None). None is NoneType object and it cannot be interpreted as Group. It causes exception on place where it probably shoud not be.

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -728,6 +728,8 @@ components:
     Group:
       allOf:
         - $ref: '#/components/schemas/Auditable'
+      type: object
+      nullable: true
       properties:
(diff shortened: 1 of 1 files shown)


martin-kuba (changes requested):
I do not like the complex type using oneOf. I recommend changing the definition of Group to be nullable:
```yaml
    Group:
      allOf:
        - $ref: '#/components/schemas/Auditable'
      type: object
      nullable: true
      properties:
```

martin-kuba (approved):

PR #3669, opened 2022-08-01
feat(core): system role and mfa tags added to perun-roles.yml and cod…

xflord:
…e to load then into objects

* new tags in perun-roles.yml
* updated objects to include the tags
* code to load to tags


perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunRolesLoader.java
```diff
@@ -247,7 +248,18 @@ private Set<PerunPolicy> loadPoliciesFromJsonNode(JsonNode rootNode) {
 			List<String> includePolicies = new ArrayList<>(objectMapper.convertValue(policyNode.get("include_policies"), new TypeReference<List<String>>() {
 			}));
 
-			policies.add(new PerunPolicy(policyName, perunRoles, includePolicies));
+			List<Map<String,String>> mfaRules = new ArrayList<>();
+			JsonNode perunMFARulesNode = policyNode.get("mfa_rules");
+			// remove if after mfa rules added
```
Johaney-s, line 253: just a thought - maybe we could skip the whole `mfa_rules` section in case it is not critical operation? it would save us some time setting it empty for every non-critical method. and then the `if` here would be valid check

mattjoke (approved):

cuadradek (approved):

Johaney-s (approved):

PR #3671, opened 2022-08-02
feat(core): mark critical operations in FacilitiesManager

Johaney-s:
* added mfa_rules section to critical methods in perun-roles.yml

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -684,24 +684,32 @@ perun_policies:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   addOwner_Facility_Owner_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   removeOwner_Facility_Owner_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -920,42 +934,25 @@ perun_policies:
...
       - default_policy
-
-  addAdmin_Facility_User_policy:
-    policy_roles:
-      - FACILITYADMIN: Facility
-    include_policies:
-      - default_policy
-
-  addAdmin_Facility_Group_policy:
-    policy_roles:
```
xflord, line 943: Why delete all of this? I see that these policies aren't used in the Entry layers of the methods, however shouldn't they still require MFA? Are we somehow going to implement MFA into setRole() or am I missing something?

cuadradek (approved):

perun-base/src/main/resources/perun-roles.yml
Johaney-s: good point, we'll want to have setRole() methods covered. that method is using `authorizedToManageRole` check, which throws PrivilegeException, so I imagine we could create public method in Utils like isMfaCritical and send the complementary object to it, then decide if MFA requirements are met etc. - so not using perun-roles for this use case at all, but handle it separately. if you'd agree with this solution, I would comment it to the original task.

xflord (approved):

perun-base/src/main/resources/perun-roles.yml
xflord: Sounds good to me

PR #3673, opened 2022-08-03
feat(core): Members Manager

mattjoke:
* added mfa_rules for Members Manager in perun-roles.yml


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2795,58 +2851,77 @@ perun_policies:
...
+    mfa_rules:
+      - MFA: Member
+      - MFA: User
 
   unsetSponsorshipForMember_Member_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
Johaney-s, line 2908:
```suggestion
      - MFA: Member
      - MFA: User
      - MFA: Vo
```
would make sense to me if `setSponsorshipForMember_Member_User_LocalDate_policy`, `unsetSponsorshipForMember_Member_policy`, `sponsored-sponsorMember_Member_User_LocalDate_policy` and `sponsor-sponsorMember_Member_User_LocalDate_policy` would have the same rules

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2198,55 +2198,76 @@ perun_policies:
...
+      - MFA: Group
+      - MFA: Member
 
   createSpecificMember_Vo_Candidate_List<User>_SpecificUserType_List<Group>_policy:
     policy_roles:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
Johaney-s, line 2229: discussed with @balcirakpeter and checking if VO is critical is enough, no need for global MFA check. please check also other methods where you check mfa globaly if it makes sense to check just vo. also, besides checking if member or user is critical in the other methods, vo should be checked too, so please add it where it makes sense.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2965,18 +3040,25 @@ perun_policies:
...
+      - MFA: Member
+      - MFA: User
 
   sponsored-removeSponsor_Member_User_policy:
     policy_roles:
       - SPONSORSHIP: Member
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
Johaney-s, line 3053: again, I'd go with Member, User and Vo for `extendExpirationForSponsoredMember_Member_User_policy`, `sponsored-removeSponsor_Member_User_policy`, `sponsor-removeSponsor_Member_User_policy`

Johaney-s (commented):
i have a few suggestions, but if you find some of them wrong, it would be a good idea to discuss it also with others on daily, because i wasn't quite sure in most cases... basically, as i understand it, member is a relationship between VO and User, so mostly all three objects should be checked.

please rename the commit message to something more specific, because it will appear in the release notes and then it's easier to understand what changes were made.

perun-base/src/main/resources/perun-roles.yml
mattjoke: Yes, this makes more sense than a global check.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2696,31 +2737,41 @@ perun_policies:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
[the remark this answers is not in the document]
Johaney-s: status changing doesn't need to be globally checked

perun-base/src/main/resources/perun-roles.yml
Johaney-s: when i think about it again, i'd add also Group in the createSpecificMember and createMember rules

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2795,58 +2842,88 @@ perun_policies:
...
+      - MFA: Vo
 
   setSponsoredMember_Vo_User_String_String_User_LocalDate_policy:
     policy_roles:
       - SPONSOR: Vo
         SELF: User
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Member
```
xflord, line 2855:
```suggestion
      - MFA: User
```

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2795,58 +2842,88 @@ perun_policies:
...
+      - MFA: Vo
 
   createSponsoredMembersFromCSV_Vo_String_List<String>_User_policy:
     policy_roles:
       - SPONSOR: Vo
         SELF: User
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Member
```
xflord, line 2865:
```suggestion
```

xflord: Not sure what Member object would be checked - there is none before the method is called

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2795,58 +2842,88 @@ perun_policies:
...
 
   group-createSponsoredMembersFromCSV_Vo_String_List<String>_User_policy:
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
     include_policies:
       - default_policy
-
+    mfa_rules:
+      - MFA: Member
```
xflord, line 2875:
```suggestion
```

xflord: same here

xflord (changes requested):
These reviews are a bit tricky, so as @Johaney-s said, if you disagree with anything, speak up and we will discuss it with everyone.

perun-base/src/main/resources/perun-roles.yml
mattjoke: Yes, the User makes more sense

perun-base/src/main/resources/perun-roles.yml
mattjoke: Should the Group be added to the create Sponsor Member methods?

Johaney-s: yes, if they are assigned to groups

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2815,38 +2846,58 @@ perun_policies:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
xflord, line 2850:
```suggestion
      - MFA: User
      - MFA: Member
      - MFA: Vo
```
Forgot this

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2815,38 +2846,58 @@ perun_policies:
...
 
   setStatus_Member_Status_String_policy:
     policy_roles:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
xflord, line 2858:
```suggestion
      - MFA: User
      - MFA: Member
      - MFA: Vo
```
And this

xflord (changes requested):

xflord (approved):
gj

perun-base/src/main/resources/perun-roles.yml
Johaney-s: I think we should remove `- MFA: Group` from `createSponsoredMember_Vo_String_Map<String_String>_String_User_LocalDate_policy`, `createSponsoredMembersFromCSV_Vo_String_List<String>_User_policy`, `createSponsoredMembers_Vo_String_List<String>_User_policy` because it shouldn't affect groups anyhow. Besides that the rest looks all good to me

Johaney-s (approved):

PR #3674, opened 2022-08-03
feat(core): critical operations in GroupsManager

sarkapalkovicova:
Critical operations in perun-roles.yml were marked.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -1152,6 +1152,8 @@ perun_policies:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Vo
 
   createGroup_Group_Group_policy:
     policy_roles:
@@ -1159,6 +1161,9 @@ perun_policies:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Group
+      - MFA: Vo
 
   deleteGroup_Group_boolean_policy:
     policy_roles:
@@ -1166,12 +1171,18 @@ perun_policies:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Group
+      - MFA: Vo
 
   deleteAllGroups_Vo_policy:
     policy_roles:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Vo
+      - MFA: Group
 
   deleteGroups_List<Group>_boolean_policy:
     policy_roles:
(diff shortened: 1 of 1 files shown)


Johaney-s (commented):
sometimes we'll mark the whole VO as critical and not set every group as a critical object, and because groups influence the VO itself, we should add checks also for - MFA: Vo where possible. this goes for deleting groups, extending memberships etc.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -1784,20 +1800,29 @@ perun_policies:
...
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Group
+      - MFA: Vo
 
   synchronizeGroups_policy:
```
Johaney-s, line 1827: please add global MFA check here and to `synchronizeGroupsStructures_policy`

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2034,13 +2062,20 @@ perun_policies:
...
 
   result-removeGroupUnion_Group_Group_policy:
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Group
+      - MFA: Member
```
Johaney-s, line 2077: we don't need to check members in the two policies removing group unions

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2068,6 +2107,8 @@ perun_policies:
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Member
```
Johaney-s, line 2111: please add also check for Group and Vo here, changing the member's status can gain him access that is managed by group or VO...

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2086,6 +2127,9 @@ perun_policies:
...
+      - MFA: Vo
```
Johaney-s, line 2132: i'd add  `- MFA: Group` because it affects also the group.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -2122,13 +2166,17 @@ perun_policies:
...
+      - MFA: Group
 
   addGroupsToAutoRegistration_List<Group>_policy:
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Group
```
xflord, line 2179: in both of the AutoRegistration methods, we want to check Vo as well, since it affects the registration process for the Vos of the groups.

xflord (changes requested):

Johaney-s:
please change the base of the PR from CESNET:master to CESNET:mfa

xflord (approved):

Johaney-s (approved):

PR #3677, opened 2022-08-05
feat(core): critical operations in UsersManager

sarkapalkovicova:
Critical operations in perun-roles.yml were marked.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -4957,11 +4957,15 @@ perun_policies:
       - SELF: User
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
 
   owner-removeSpecificUserOwner_User_User_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
 
   addSpecificUserOwner_User_User_policy:
     policy_roles:
@@ -4969,11 +4973,15 @@ perun_policies:
       - SELF: User
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
 
   owner-addSpecificUserOwner_User_User_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
 
   getSpecificUsers_policy:
     policy_roles:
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5095,50 +5103,71 @@ perun_policies:
...
+    mfa_rules:
+      - MFA: User
 
   deleteUser_User_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
+      - MFA: Member
```
xflord, line 5129: I think Member is affected only in the forceDelete version of this method (so the policy after this one). Also, user external sources are deleted in this method, so MFA: UserExtSource should be here imo. What do you think @Johaney-s @cuadradek ?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5095,50 +5103,71 @@ perun_policies:
...
 
   anonymizeUser_User_boolean_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
+      - MFA: Member
```
xflord, line 5145: Same deal with the UserExtSources here

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5204,22 +5233,33 @@ perun_policies:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
```
xflord, line 5237: I feel like whenever working with UES, we should check both the UES and User associated (so in `updateUserExtSource_UserExtSource_policy` as well) @Johaney-s  @cuadradek thoughts?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -5412,81 +5452,107 @@ perun_policies:
...
+    mfa_rules:
+      -MFA: User
 
   deletePassword_String_String_policy:
     policy_roles:
       - REGISTRAR:
     include_policies:
       - default_policy
+    mfa_rules:
+      -MFA
```
xflord, line 5538:
```suggestion
      -MFA:
```

xflord (commented):
Again, these reviews can be a bit tricky, if you disagree with anything, comment here or mention it at the daily meeting

perun-base/src/main/resources/perun-roles.yml
Johaney-s: here I agree; when deleting user I'm not so sure, but better make it stricter now and if needed we'll remove it later... so okay by me

xflord (approved):

Johaney-s (approved):

PR #3681, opened 2022-08-08
feat(core): MFA rules added for Security Teams, Services a Tasks Mana…

Kuliak:
…gers

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -4220,18 +4220,24 @@ perun_policies:
       - SECURITYADMIN:
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   updateSecurityTeam_SecurityTeam_policy:
     policy_roles:
       - SECURITYADMIN: SecurityTeam
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   deleteSecurityTeam_SecurityTeam__boolean_policy:
     policy_roles:
       - SECURITYADMIN: SecurityTeam
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   getSecurityTeamById_int_policy:
     policy_roles:
@@ -4267,41 +4273,21 @@ perun_policies:
     include_policies:
       - default_policy
 
-  addAdmin_SecurityTeam_User_policy:
-    policy_roles:
-      - SECURITYADMIN: SecurityTeam
-    include_policies:
-      - default_policy
-
-  addAdmin_SecurityTeam_Group_policy:
-    policy_roles:
-      - SECURITYADMIN: SecurityTeam
-    include_policies:
-      - default_policy
-
-  removeAdmin_SecurityTeam_User_policy:
-    policy_roles:
-      - SECURITYADMIN: SecurityTeam
-    include_policies:
-      - default_policy
-
-  removeAdmin_SecurityTeam_Group_policy:
-    policy_roles:
-      - SECURITYADMIN: SecurityTeam
-    include_policies:
-      - default_policy
-
   addUserToBlacklist_SecurityTeam_User_String_policy:
     policy_roles:
       - SECURITYADMIN: SecurityTeam
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   removeUserFromBlacklist_SecurityTeam_User_policy:
     policy_roles:
       - SECURITYADMIN: SecurityTeam
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   getBlacklist_SecurityTeam_policy:
     policy_roles:
@@ -4339,24 +4325,32 @@ perun_policies:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   blockServiceOnDestination_Service_int_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   blockAllServicesOnFacility_Facility_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   blockAllServicesOnDestination_int_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   getServicesBlockedOnFacility_Facility_policy:
     policy_roles:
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4395,52 +4427,70 @@ perun_policies:
...
+      - MFA: Facility
 
   forceServicePropagation_Facility_Service_policy:
     policy_roles:
       - FACILITYADMIN: Facility
       - ENGINE:
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Service
```
Johaney-s, line 4464: same problem - Service objects don't have attributes, so we can't mark them as critical, thus it wouldn't make sense to have this check here

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4395,52 +4427,70 @@ perun_policies:
...
   planServicePropagation_Service_policy:
     policy_roles:
       - FACILITYADMIN:
       - ENGINE:
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Service
 
   suspendTasksPropagation_policy:
```
Johaney-s, line 4495: missing policies here

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4772,24 +4856,35 @@ perun_policies:
...
+      - MFA: Facility
 
   addDestinationsDefinedByHostsOnFacility_Service_Facility_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
+      - MFA: Host
```
Johaney-s, line 4869: we're not affecting the Host object itself, just reading from it, so no need for this check here. same for the methods below

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4886,19 +4981,25 @@ perun_policies:
...
+      - MFA: Facility
+
   deleteTaskResults_Task_Destination_policy:
     policy_roles:
       - FACILITYADMIN: Facility
     include_policies:
       - default_policy
-  
+    mfa_rules:
+      - MFA:
```
Johaney-s, line 5001:
```suggestion
      - MFA: Facility
```

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4272,36 +4274,48 @@ perun_policies:
...
+    mfa_rules:
+      - MFA: User
 
   removeAdmin_SecurityTeam_Group_policy:
     policy_roles:
       - SECURITYADMIN: SecurityTeam
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Group
```
xflord, line 4302: All the remove/add admin policies can be deleted - we now use the `setRole` instead of the regular authorization method (check other PRs adding MFA rules)

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4272,36 +4274,48 @@ perun_policies:
...
 
   addUserToBlacklist_SecurityTeam_User_String_policy:
     policy_roles:
       - SECURITYADMIN: SecurityTeam
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: User
```
xflord, line 4310: Adding the user to a blacklist seems like a similar process to banning a member from resource/Vo, where we agreed that the member being banned isn't the primary object and doesn't need to be critical. (The SecurityTeams "rules" are being edited, not the user themselves).

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4395,31 +4417,41 @@ perun_policies:
...
   forceServicePropagation_Facility_Service_policy:
     policy_roles:
       - FACILITYADMIN: Facility
       - ENGINE:
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   forceServicePropagation_Service_policy:
```
xflord, line 4456: missing global MFA check for this method

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4434,6 +4466,8 @@ perun_policies:
       - ENGINE:
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Facility
 
   planServicePropagation_Service_policy:
```
xflord, line 4472: global MFA check here

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4464,6 +4500,8 @@ perun_policies:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   updateService_Service_policy:
```
xflord, line 4506: global MFA check for updating and creating

xflord (changes requested):
I see that you removed the MFA checks for creating/updating security teams, which I would keep as global MFA checks (you're creating/updating an object and hence changing the state of the perun instance) or alternatively I would completely omit MFA when it comes to security teams, since as I understand it's a feature that's not used. Probably something we should discuss further. @Johaney-s @cuadradek

perun-base/src/main/resources/perun-roles.yml
Kuliak: okay, I think we can remove mentioned methods. As for creating/updating, I did not use any MFA rule because I expected that no attributes were set during those operations. Perhaps, we should discuss this, too.

xflord: > Kritické operácie budú pre začiatok všetky politiky, ktoré sú používané v metódach meniacich stav v Perunovi.
Quote from the task description. 
This is why I believe that any method that changes the state of Perun (whether it be creating/deleting/updating any object) is supposed to be critical (and MFA covered). However, I do see where you're coming from, let's discuss it tomorrow at daily, hopefully with @balcirakpeter present.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -4395,57 +4389,75 @@ perun_policies:
...
+      - MFA:
 
   planServicePropagation_Facility_Service_policy:
     policy_roles:
       - FACILITYADMIN: Facility
       - ENGINE:
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
```
Johaney-s, line 4444:
```suggestion
      - MFA: Facility
```

perun-base/src/main/resources/perun-roles.yml
Johaney-s: still missing

Johaney-s (changes requested):

Johaney-s (approved):

xflord (approved):

PR #3682, opened 2022-08-08
fix(core): fix update facility

sarkapalkovicova:
When facility is renamed, the corresponding consent hub is also renamed.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
@@ -15,6 +15,7 @@ import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.BanOnFacility;
 import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.Consent;
 import cz.metacentrum.perun.core.api.ConsentHub;
 import cz.metacentrum.perun.core.api.Destination;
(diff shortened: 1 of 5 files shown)


cuadradek (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/FacilitiesManagerBlImpl.java
```diff
@@ -436,12 +437,23 @@ public void deleteFacility(PerunSession sess, Facility facility, Boolean force)
...
-	public Facility updateFacility(PerunSession sess, Facility facility) throws FacilityExistsException {
+	public Facility updateFacility(PerunSession sess, Facility facility) throws FacilityExistsException, ConsentHubExistsException {
 		//check facility name, it can contain only a-zA-Z.0-9_-
 		if (!facility.getName().matches("^[ a-zA-Z.0-9_-]+$")) {
 			throw new IllegalArgumentException("Wrong facility name, facility name can contain only a-Z0-9.-_ and space characters");
 		}
 
+		// if renaming the facility, then rename also the corresponding consent hub
+		try {
+			if (!facility.getName().equals(getFacilitiesManagerImpl().getFacilityById(sess, facility.getId()).getName())) {
```
Johaney-s, line 448: maybe we should first check that the consent hub name matches the original facility name, because once we start using the joint consent hubs for more facilities they could have some special consenthub name set and this would rewrite it. what do you think @sarkapalkovicova @cuadradek

cuadradek: yeah, that's a good point

cuadradek (approved):

Johaney-s (approved):

PR #3683, opened 2022-08-09
feat: critical operations in Registrar, Mail, Notif, Consent managers

cuadradek:
- mfa_rules in perun-roles.yml were added for Registrar, Mail,
Notification and Consent managers.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -6438,11 +6438,15 @@ perun_policies:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   removePerunNotifReceiverById_int_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   getPerunNotifRegexById_int_policy:
     policy_roles:
@@ -6460,16 +6464,22 @@ perun_policies:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   updatePerunNotifRegex_PerunNotifRegex_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   removePerunNotifRegexById_int_policy:
     policy_roles: []
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA:
 
   getPerunNotifTemplateById_int_policy:
     policy_roles:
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -7016,32 +7120,44 @@ perun_policies:
...
     policy_roles:
       - VOADMIN: Vo
       - GROUPADMIN: Group
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Vo
+      - MFA: Group
 
   source-copyFormFromGroupToGroup_Group_Group_policy:
```
sarkapalkovicova, line 7162: I'm not sure if you skipped it or if it's correct, but shouldn't there be mfa rules like for  `destination-copyFormFromGroupToGroup_Group_Group_policy`? The rest looks fine to me.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6653,99 +6675,137 @@ perun_policies:
...
 
   createApplicationFormInGroup_Group_policy:
     policy_roles:
       - GROUPADMIN: Group
       - VOADMIN: Vo
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Vo
+      - MFA: Group
```
Kuliak, line 6808: Are we really changing mentioned Vo or Group? Maybe general MFA would be better, since only empty application form is created and no attribute values are set, or?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6790,116 +6850,160 @@ perun_policies:
...
 
   group-addFormItem_ApplicationForm_ApplicationFormItem_policy:
     policy_roles:
       - VOADMIN: Vo
       - GROUPADMIN: Group
     include_policies:
       - default_policy
+    mfa_rules:
+      - MFA: Vo
+      - MFA: Group
```
Kuliak, line 6864: yes, it concerns Vo or group, however, do we change them / set something for them or for the given form only? Same applies for update/delete items or forms. Perhaps, we might discuss it to clarify it. What do you say?
@Johaney-s @xflord

Kuliak (commented):
Good job! I have some suggestions that we might discuss just to be sure.

perun-base/src/main/resources/perun-roles.yml
cuadradek: I thought the source group won't be changed, only its form would be copied so it is a read operation = non critical. What do you think?

perun-base/src/main/resources/perun-roles.yml
cuadradek: We discussed this, and global MFA would enforce MFA for all forms, this way only for the groups / Vos marked as critical.

perun-base/src/main/resources/perun-roles.yml
cuadradek: We discussed this, and we probably want to keep it this way. We might remove it in the future if we do not want it.

Kuliak (approved):

sarkapalkovicova (approved):

PR #3684, opened 2022-08-09
feat: new method getMessagesByIdAndCount

balcirakpeter:
Created a new method that returns a defined number of messages whose ids are less or equal to the
given id.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
@@ -204,6 +204,15 @@ public class AuditMessagesManagerImpl implements AuditMessagesManagerImplApi {
 	}
 
 	@Override
+	public List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int id, int count) {
+		try {
+			return jdbc.query("select " + auditMessageMappingSelectQuery + " from auditer_log where id<=? order by id desc limit ?", AUDIT_MESSAGE_MAPPER, id, count);
+		} catch (RuntimeException err) {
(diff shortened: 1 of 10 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
```diff
@@ -203,6 +203,16 @@ public List<AuditMessage> getMessagesByCount(PerunSession perunSession, int coun
 		}
 	}
 
+	@Override
+	public List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int id, int count) {
+		int lowestId = id - count + 1;
+		try {
+			return jdbc.query("select " + auditMessageMappingSelectQuery + " from auditer_log where id between ? and ? order by id desc", AUDIT_MESSAGE_MAPPER, lowestId, id);
```
Johaney-s, line 210: just a thought -  i see there's similar problem with `getMessagesByCount` method - by the name of the method, wouldn't it be better to try to return the exact count of the messages than subtract the ids? or rather change the name to `getMessagesByIdsRange`?
```suggestion
			return jdbc.query("select " + auditMessageMappingSelectQuery + " from auditer_log where id<=? order by id desc limit ?", AUDIT_MESSAGE_MAPPER, id, count);
```

cuadradek (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
balcirakpeter: Good point. Changed so it is using `limit` instead of `between`.

Johaney-s (approved):

PR #3685, opened 2022-08-10
feat(cli): scripts for critical attribute actions

Johaney-s:
* added script setAttributeCriticalActions, which allows to mark critical actions on attributes
* extended getAttributePolicyCollections script to return also critical actions


mattjoke (approved):
The code looks good; however, manual testing of the scripts would probably be needed.

xflord (approved):
Looks ok, good thinking sorting the collections.

PR #3691, opened 2022-08-23
fix: 🐛 Update BBMRI registration modules to new group structure

dBucik:
No longer the Network and Collection group structures use
representatives group.


dBucik:
Tested on devel, no issues found so far

balcirakpeter (approved):

PR #3693, opened 2022-08-23
feat(core): check MFA for attributes

Johaney-s:
* extended attribute authorization check with option for MFA
* if attribute action is marked as critical it requires MFA
* if user didn't perform MFA, uncatched MfaPrivilegeException is thrown
* when setting 'writable' property to attribute, MFA is not checked because if user is authorized to write to attribute and action requires MFA, exception will be thrown when calling setAttribute method

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/MfaPrivilegeException.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/MfaPrivilegeException.java
@@ -1,0 +1,32 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.exceptions.rt.PerunRuntimeException;
+
+/**
+ * This exception is thrown when principal is performing MFA-requiring action but is not authenticated with Multi-Factor
+ * @author Johana Supikova <xsupikov@fi.muni.cz>
+ */
+public class MfaPrivilegeException extends PerunRuntimeException {
+	static final long serialVersionUID = 0;
+
+	public MfaPrivilegeException(String message) {
+		super(message);
+	}
+
+	public MfaPrivilegeException(String message, Throwable cause) {
+		super(message, cause);
+	}
+
+	public MfaPrivilegeException(Throwable cause) {
+		super(cause);
+	}
+
+	public MfaPrivilegeException(PerunSession sess) {
+		super("Principal " + sess.getPerunPrincipal().getActor() + " is not authorized by MFA");
+	}
+
+	public MfaPrivilegeException(PerunSession sess, String action) {
+		super("Principal " + sess.getPerunPrincipal().getActor() + " is not authorized to perform action '" + action + "' without MFA");
+	}
+}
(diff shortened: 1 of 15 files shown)


Johaney-s:
blocked by #3690

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -1544,76 +1544,76 @@ public List<AttributeDefinition> getAttributesDefinitionWithRights(PerunSession
 			AttributeDefinition attrDef = iterator.next();
 
 			if (this.isFromNamespace(sess, attrDef, NS_USER_FACILITY_ATTR) && user != null && facility != null) {
-				if (!AuthzResolver.isAuthorizedForAttribute(sess, READ, attrDef, user, facility)) {
+				if (!AuthzResolver.isAuthorizedForAttribute(sess, READ, attrDef, user, facility, true)) {
 					iterator.remove();
 				} else {
-					attrDef.setWritable(AuthzResolver.isAuthorizedForAttribute(sess, WRITE, attrDef, user, facility));
+					attrDef.setWritable(AuthzResolver.isAuthorizedForAttribute(sess, WRITE, attrDef, user, facility, false));
```
HejdaJakub, line 1550: Wouldn't the true/false be in the opposite order (`false` for `READ` and `true` for `WRITE`)? We want to check MFA for `WRITE`, am I right?

HejdaJakub: NOTE: I agree that this order is correct after discussion

HejdaJakub (approved):

mattjoke (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/MfaPrivilegeException.java
```diff
@@ -0,0 +1,32 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.exceptions.rt.PerunRuntimeException;
+
+/**
+ * This exception is thrown when principal is performing MFA-requiring action but is not authenticated with Multi-Factor
+ * @author Johana Supikova <xsupikov@fi.muni.cz>
+ */
+public class MfaPrivilegeException extends PerunRuntimeException {
```
cuadradek, line 10: I was thinking about this, and we could make this exception a checked one. Unchecked exceptions usually reflect errors inside the program logic, but this exception represents a common situation. What do you think @Johaney-s @mattjoke @HejdaJakub ?

cuadradek: On the other side, we often use authorizing methods in lambda functions in streams, which can't reasonably handle checked exceptions. Hence, if we made this exception a checked one, we would also have to change a lot of code to make it work.

Kuliak (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/MfaPrivilegeException.java
mattjoke: From the design standpoint, it would be better, but we have to (as you mentioned) consider the amount of code needed to be updated and re-coded.

Johaney-s: we decided to keep it a RuntimeException for there are too many stream usages and the exception is not meant to be caught anyway.

PR #3695, opened 2022-08-25
feat(cli): resolve MFA exception for command line tools

Johaney-s:
* if Perun throws MfaPrivilegeException, CLI tools need to instruct user to authenticate with Multi-Factor
* this is done by setting enforce_mfa property in OIDC config
* then when authentication request is sent, it prepends MFA acr value so that MFA is done along with standard OIDC authentication
* refreshMfa() needs to be called afterwards so that mfa timestamp is retrieved in Perun from UserInfoEndpoint
* if user has no active tokens, unmet_authentication_requirements error is returned from token endpoint


mattjoke (approved):
I think the code looks good

sarkapalkovicova (approved):

xflord (approved):

PR #3697, opened 2022-08-30
feat(core): check MFA rules

cuadradek:
- Method AuthzResolverBlImpl.isAuthorized was extended to check MFA
rules. MFAPrivilegeException is thrown when a policy requires MFA on an
object marked as critical, but the principal isn't authenticated with
MFA.
- The MFA rules are checked only when config property enforceMfa is
true.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -11,6 +11,7 @@ import cz.metacentrum.perun.core.api.AttributeAction;
 import cz.metacentrum.perun.core.api.AttributeDefinition;
 import cz.metacentrum.perun.core.api.AttributePolicy;
 import cz.metacentrum.perun.core.api.AttributePolicyCollection;
+import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.AuthzResolver;
 import cz.metacentrum.perun.core.api.BanOnVo;
 import cz.metacentrum.perun.core.api.BeansUtils;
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -136,14 +143,121 @@ public static boolean authorized(PerunSession sess, String policyDefinition, Lis
...
+	 * @param id
+	 * @return
+	 */
+	private static boolean isCriticalObject(PerunSession sess, String objectType, Integer id) {
+		AttributesManagerBl attrManager = perunBl.getAttributesManagerBl();
+		Attribute attr;
+		try {
+			if ("Group".equals(objectType)) {
+				Group group = perunBl.getGroupsManagerBl().getGroupById(sess, id);
+				attr = attrManager.getAttribute(sess, group, AttributesManager.NS_GROUP_ATTR_DEF + ":" + MFA_CRITICAL_ATTR);
```
Johaney-s, line 227: would be nice to reuse `isAnyObjectMfaCritical` method to avoid code duplication

Johaney-s (commented):
potentially we could get into conflict if resolving policies together with MFA for we're fetching related objects which might get checked for MFA even if we don't want to do so - we cannot exclude them from the MFA check... but so far the cases that seem problematic to me can be resolved by duplicating policies (for example when copying facility attributes - sourceFacility would have separate policy from destinationFacility)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -136,14 +143,121 @@ public static boolean authorized(PerunSession sess, String policyDefinition, Lis
...
+
+	/**
+	 * Returns true if the object of given type with given ID is marked as critical.
+	 *
+	 * @param sess
+	 * @param objectType
+	 * @param id
+	 * @return
+	 */
+	private static boolean isCriticalObject(PerunSession sess, String objectType, Integer id) {
```
xflord, line 221: This pretty much does the same thing as `isAnyObjectMfaCritical` from AuthzResolver, have you considered using that instead? But I guess then we would need a method that retrieves objects based on bean name and ids, resulting in the same complexity, if not higher. Just a thought.

xflord: @Johaney-s  beat me to it

HejdaJakub:
> potentially we could get into conflict if resolving policies together with MFA for we're fetching related objects which might get checked for MFA even if we don't want to do so - we cannot exclude them from the MFA check... but so far the cases that seem problematic to me can be resolved by duplicating policies (for example when copying facility attributes - sourceFacility would have separate policy from destinationFacility)

Duplicating policies sounds like a good solution (at least it is not difficult and it should work), so I agree with this solution.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
cuadradek: Yes, I used `isAnyObjectMfaCritical` to save a little bit of code duplication, but as you said, we still need to retrieve the objects.

Johaney-s (approved):

xflord (approved):

PR #3698, opened 2022-09-01
feat(core): resolve MFA in FacilitiesManager

Johaney-s:
* added missing policies
* separated policies for destination-source objects (source object is only read from so MFA should be evaluated separately for it)

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -768,7 +768,13 @@ perun_policies:
     mfa_rules:
       - MFA: Facility
 
-  copyOwners_Facility_Facility_policy:
+  source-copyOwners_Facility_Facility_policy:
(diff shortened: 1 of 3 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -1059,7 +1071,13 @@ perun_policies:
       - MFA: Facility
       - MFA: User
```
xflord, line 1072:
```suggestion
```
I know this isn't part of your PR, but I think we should omit User here, since copyManagers uses the setRole() function (for which we will check MFA separately IIRC)

Johaney-s: good point, fixed

xflord (approved):

cuadradek (approved):

PR #3699, opened 2022-09-01
feat(core): resolve MFA in GroupsManager

xflord:
* edited objects evaluated for authorization in some methods to better work with MFA
* seperated policies for allowing/disallowing groups to hierarchical vos

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -1298,6 +1298,15 @@ perun_policies:
       - MFA: Group
       - MFA: Vo
 
+  parentGroup-moveGroup_Group_Group_policy:
+    policy_roles:
+      - SELF:
+    include_policies:
+      - default_policy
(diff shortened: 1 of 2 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -1298,6 +1298,14 @@ perun_policies:
       - MFA: Group
       - MFA: Vo
 
+  parentGroup-moveGroup_Group_Group_policy:
```
Johaney-s, line 1301: `policy_roles: []` with including default policies (PERUNADMIN) would mean that to pass privilege requirements one would need to be perunadmin - you probably want to add `policy_roles: Self:` to `parentGroup-moveGroup_Group_Group_policy` if you want to check just the MFA and always pass the other policies

Johaney-s (approved):

Kuliak (approved):

PR #3706, opened 2022-09-02
feat(core): resolve MFA for un/setting role

Johaney-s:
* when setting role to user or group, MFA might be required globally or for specific bounded object
* if the bounded object or the entity that gets the role are critical, MFA is required
* which bounded objects should be checked is specified in assignment_check property of the role

BREAKING CHANGE: extended perun-roles.yml with property assignment_check for roles

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
@@ -23,6 +23,10 @@ import java.util.Objects;
  *            Example entry: key: User; value: user_id
  * assignedObjects is a map of objects which can be assigned with the role. Key is a object name and value is mapping to the database.
  *            Example entry: key: Resource; value: resource_id
+ * assignmentCheck is a list of maps defining which of the assigned objects should be checked for being critical (requiring MFA) when setting the role.
+ *            Example entry: key: MFA; value: Resource <- If resource is critical, MFA is required
+ *            Example entry: {} <- No MFA is required to set this role
+ *            Example entry: Key: MFA; value: <- No value means MFA is always required to set this role
  * associatedReadRoles is a list of related roles which are authorized to read attribute value if the main role is authorized.
  *            Example list for groupadmin role - value: [GROUPOBSERVER]
  * assignableToAttributes is a flag that determines whether the role can appear in attribute policies.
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -4137,6 +4145,44 @@ private static boolean objectAndRoleManageableByEntity(String entityToManage, Pe
...
+		List<Map<String, String>> mfaPolicies = rules.getAssignmentCheck();
+		Map<String, Set<Integer>> mapOfBeans = new HashMap<>();
+
+		if (mfaPolicies != null && !mfaPolicies.isEmpty() && complementaryObject != null) {
+			mapOfBeans = fetchAllRelatedObjects(Arrays.asList(complementaryObject));
+		}
+
+		// check complementary objects for MFA requirements
+		if (!mfaAuthorized(sess, mfaPolicies, mapOfBeans)) {
+			throw new MfaPrivilegeException("Multi-Factor authentication is required - bounded object is critical.");
```
xflord, line 4172: Maybe providing the complementary object in question in this exception could prove useful (we can call `setRole()` with a list of multiple complementary objects and then we wouldn't know because of which object the exception was thrown).

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -1660,6 +1663,114 @@ public void mfaNotCriticalObject() throws Exception {
...
 
+	@Test
+	public void mfaGenericSetRole() throws Exception {
+		System.out.println(CLASS_NAME + "mfaGenericSetRole");
+
+		final Vo createdVo = perun.getVosManager().createVo(sess, new Vo(0,"test123test123","test123test123"));
+		final Member createdMember = createSomeMember(createdVo);
+		final User createdUser = perun.getUsersManagerBl().getUserByMember(sess, createdMember);
+
+		boolean originalMfaForce = BeansUtils.getCoreConfig().isEnforceMfa();
```
xflord, line 1674: Maybe set assignmentCheck for PERUNADMIN as you did for VOADMIN in the following tests? Just to be consistent and extra safe. What do you think?

perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
```diff
@@ -23,6 +23,10 @@
  *            Example entry: key: User; value: user_id
  * assignedObjects is a map of objects which can be assigned with the role. Key is a object name and value is mapping to the database.
  *            Example entry: key: Resource; value: resource_id
+ * assignmentCheck is a list of maps defining which of the assigned objects should be checked for being critical (requiring MFA) when setting the role.
```
xflord, line 26:
```suggestion
 * assignmentCheck is a list of maps defining which of the complementary objects should be checked for being critical (requiring MFA) when setting the role.
```
Wouldn't this be more consistent with the Javadocs?

xflord (commented):
Great job, just tiny cosmetic suggestions.

perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
Johaney-s: i think not - complementary object in javadocs is referred to as the object the role is assigned to (resourceadmin -> resource), while assigned objects refer to all the related objects (vo, resource, facility), which in perun-roles and the line above is called `assignedObjects`

xflord (approved):

cuadradek (approved):

PR #3707, opened 2022-09-05
feat(core): resolve MFA in RegistrarManager, MailManager

xflord:
* edited objects evaluated for authorization in some methods to better work with MFA
* added missing policies
* seperated some policies to better work with MFA

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -1167,10 +1167,23 @@ public class RegistrarManagerImpl implements RegistrarManager {
 	@Override
 	public void updateFormItemTexts(PerunSession sess, ApplicationFormItem item, Locale locale) throws PrivilegeException, FormNotExistsException {
 
+		ApplicationForm form;
+
 		try {
-			getFormByItemId(sess, item.getId());
+			form = getFormByItemId(sess, item.getId());
 		} catch (PrivilegeException ex) {
 			throw new PrivilegeException(sess, "updateFormItemTexts");
+		}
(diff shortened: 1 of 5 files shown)


cuadradek (commented):
Overall it looks good and I have couple of suggestions:

- Method `updateMailById` could have its own authorization policy (unlike the current `getFormById`, we probably want to add some MFA rules).
- Method `setSendingEnabled` also because it doesn't use any policy.
- In method `sendMessage`, we could maybe change the order of authz checks in IF conditions, so the `MFAPrivilegeException` isn't raised when the caller is self-authorized for the application.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3240,14 +3259,17 @@ public void updateApplicationUser(PerunSession sess, Application app) {
...
-		//Authorization
-		if (!AuthzResolver.authorizedInternal(sess, "updateFormItemData_int_ApplicationFormItemData_policy")) {
-			throw new PrivilegeException(sess, "updateFormItemData");
-		}
 
 		Application app = getApplicationById(sess, appId);
 		if (AppState.APPROVED.equals(app.getState()) || AppState.REJECTED.equals(app.getState())) throw new RegistrarException("Form items of once approved or rejected applications can't be modified.");
 
+		//Authorization
+		if (app.getGroup() == null && (!AuthzResolver.authorizedInternal(sess, "updateFormItemData_int_ApplicationFormItemData_policy", app.getVo()))) {
```
Johaney-s, line 3267: group==null and is authorized => will go to `else if ` branch and i guess would fail on the null group

Johaney-s (approved):

cuadradek (approved):

PR #3710, opened 2022-09-12
feat(core): check MFA for roles always requiring it

Johaney-s:
* we want to have option to configure roles which would always require MFA
* this would now apply for PERUNADMIN and PERUNOBSERVER
* user with this role will always need MFA to do ANY operation in Perun
* some internal components are given perunadmin role, which we need to skip for this check - these are defined in coreConfig

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -53,6 +53,7 @@ import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.MFAuthenticationException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MfaPrivilegeException;
+import cz.metacentrum.perun.core.api.exceptions.MfaRolePrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.PolicyNotExistsException;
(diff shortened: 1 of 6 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -4202,6 +4220,29 @@ private static void checkMfaForSettingRole(PerunSession sess, PerunBean assignin
 		}
 	}
 
+	/**
+	 * Checks if having a role requires MFA based on configuration in perun-roles, throws MfaPrivilegeException if requirements unmet.
+	 * @param roles roles to be set to principal
+	 * @throws MfaPrivilegeException if MFA requirements are not met
```
cuadradek, line 4226:
```suggestion
	 * @throws MfaRolePrivilegeException if MFA requirements are not met
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -4202,6 +4220,29 @@ private static void checkMfaForSettingRole(PerunSession sess, PerunBean assignin
 		}
 	}
 
+	/**
+	 * Checks if having a role requires MFA based on configuration in perun-roles, throws MfaPrivilegeException if requirements unmet.
```
cuadradek, line 4224:
```suggestion
	 * Checks if having a role requires MFA based on configuration in perun-roles, throws MfaRolePrivilegeException if requirements unmet.
```

cuadradek (changes requested):

cuadradek (approved):

PR #3711, opened 2022-09-13
feat(core): renamed system_role

mattjoke:
* renamed system_role to skip_mfa to indicate roles, for which mfa checks should be skipped
* updated RoleManagementRules and AuthzResolverBlImpl to use more generic names


perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
```diff
@@ -131,12 +131,12 @@ public void setAssignableToAttributes(boolean assignableToAttributes) {
 		this.assignableToAttributes = assignableToAttributes;
 	}
 
-	public boolean isSystemRole() {
-		return systemRole;
+	public boolean shouldSkipMFA() {
+		return skipMFA;
 	}
 
 	public void setSystemRole(boolean systemRole) {
```
Johaney-s, line 138:
```suggestion
	public void setSkipMFA(boolean skipMFA) {
```

Johaney-s (approved):

Johaney-s:
i think we don't use scope `base` for commits, please use `core` instead

cuadradek (approved):

PR #3718, opened 2022-09-20
feat(core): add new GROUPMEMBERSHIPMANAGER role

Johaney-s:
* new role in group which can resolve group applications, add, remove or invite members and extend their membership, but cannot change group settings
* extended management rules and attribute policies for new role
* added script to insert new attribute policy for existing user, member, user-facility and member-group attributes for WRITE action
* removed deprecated Privilege class

BREAKING CHANGE: running db script required for Groupmembershipmanager role to work


cuadradek (approved):

perun-db/addGroupMembershipManagerPolicies
```diff
@@ -0,0 +1,112 @@
...
+$sthCollections->execute();
+while (my @row = $sthCollections->fetchrow_array()) {
+	my ($attrId, $attrName) = @row;
+	my $attrType = (split ':', $attrName)[2];
+
+	# add READ policy for GROUPMEMBERSHIPMANAGER
+	my ($shouldRead) = grep {$attrName eq $_} @readAttributes;
+	my $object = ($attrType eq "member" || $attrType eq "vo" || $attrType eq "user") ? "Vo" : "Group";
+
+	if ($attrType eq "member" || $attrType eq "member_group" || $attrType eq "user_facility" || $attrType eq "user" || $shouldRead) {
```
cuadradek, line 91: If it is not too much work, I would add a check that GROUP ADMIN has a READ policy for the attribute.

Johaney-s: extended the select query for attribute definitions to ensure that

xflord (approved):

PR #3721, opened 2022-09-21
feat(core): added keepMemberStatus attribute

sarkapalkovicova:
Added attribute, which when is set to value true, statuses of existing members will not be updated during synchronization.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -87,7 +87,6 @@ import cz.metacentrum.perun.core.api.exceptions.GroupSynchronizationAlreadyRunni
 import cz.metacentrum.perun.core.api.exceptions.GroupSynchronizationNotEnabledException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.InvalidLoginException;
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3704,7 +3703,31 @@ public void updateExistingMemberWhileSynchronization(PerunSession sess, Group gr
...
+	 * @param sess session
+	 * @param group group to be synchronized
+	 *
+	 * @return false if change of member's status is not allowed, true otherwise
+	 *
+	 * @throws InternalErrorException if obtaining attribute failed
+	 * @throws WrongAttributeAssignmentException if incorrect assignment of attribute
+	 * @throws AttributeNotExistsException if attribute does not exist in database
+	 */
+	private boolean isMemberStatusChangeAllowed(PerunSession sess, Group group) throws WrongAttributeAssignmentException, AttributeNotExistsException {
```
xflord, line 3728: Should this method return the value of the new attribute? Seems to me that it only returns true when the attribute is not set, which doesn't make much sense. The javadocs are really confusing as well.

xflord (changes requested):

Johaney-s (approved):

Johaney-s:
We tested and discussed the intended behavior of the changes and concluded this is not the way it supposed to work, so in order to not include this in the release I'll revert the changes now and we'll reuse the attribute initialization and change the logic accordingly in new PR.

PR #3723, opened 2022-09-23
core(registrar): get user's open applications

Johaney-s:
* added method to get only open user's application


perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -10,6 +10,7 @@
 create table attribute_critical_actions (attr_id integer not null, action attribute_action not null, constraint attrcritops_pk primary key (attr_id, action),	constraint attrcritops_attr_fk foreign key (attr_id) references attr_names (id) on delete cascade);
 grant all on attribute_critical_actions to perun;
 create index idx_fk_attr_critops ON attribute_critical_actions(attr_id);
+create index app_state_idx ON application (state);
```
cuadradek, line 13: I think this change should be in 3.1.98 DB version.

Johaney-s: it should be both in the same release

cuadradek (approved):

xflord (approved):

PR #3725, opened 2022-09-26
feat(core): notification language check

bodnara:
* added syntax check for notification language attribute
* by default only values "en" and "cs" are allowed
* in future this could be configured


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_vo_attribute_def_def_notificationsDefLang.java
```diff
@@ -0,0 +1,36 @@
...
+		if (!defaultLanguages.contains(value)) {
+			throw new WrongAttributeValueException(attribute, "Language " + value + " is not supported for notifications.");
+		}
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
+		AttributeDefinition attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_VO_ATTR_DEF);
+		attr.setFriendlyName("notificationDefLang");
```
Johaney-s, line 30:
```suggestion
		attr.setFriendlyName("notificationsDefLang");
```
same for displayname

Johaney-s (approved):

PR #3730, opened 2022-10-04
fix(registrar): check form items on update

Johaney-s:
* rpc gets vo/group in parameters and retrieves its application form in updateFormItems and calls Impl layer with it
* items to be updated or removed need to belong to passed form, otherwise any groupadmin/voadmin could change any form item in the whole system

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -26,6 +26,7 @@ import java.util.stream.Collectors;
 
 import javax.sql.DataSource;
 
+import cz.metacentrum.perun.core.api.exceptions.IllegalArgumentException;
 import cz.metacentrum.perun.core.bl.AttributesManagerBl;
 import cz.metacentrum.perun.core.bl.GroupsManagerBl;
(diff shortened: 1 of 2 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -4795,6 +4802,19 @@ private void processNotificationsAfterCreation(PerunSession session, Application
...
+	/**
+	 * Checks if form item belongs to form with given id
+	 *
+	 * @param sess
+	 * @param formId id of application form
+	 * @param itemId id of form item
+	 * @return true if item belongs to the form, false otherwise
+	 */
+	private boolean checkItemBelongsToForm(PerunSession sess, int formId, int itemId) {
+		int storedFormId = jdbc.queryForInt("select form_id from application_form_items where id=?", itemId);
```
cuadradek, line 4814: Just a suggestion: runtime exception `IncorrectResultSizeDataAccessException` is possibly thrown by this query if the item is not in DB - maybe we could catch it and return false in such a case.

Johaney-s: good point, added

cuadradek (approved):

xflord (approved):

PR #3732, opened 2022-10-05
feat(core): exact match user attribute list/map values

Johaney-s:
* we don't want to allow substrings to be matched in map/list attributes when calling getUsersByAttributeValue method
* these changes require the value of list/map attributes to be surrounded by it's type delimiters so substrings are not matched anymore

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
@@ -37,6 +37,7 @@ import cz.metacentrum.perun.core.api.exceptions.UserExtSourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.bl.DatabaseManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
+import cz.metacentrum.perun.core.blImpl.AttributesManagerBlImpl;
 import cz.metacentrum.perun.core.implApi.UsersManagerImplApi;
 import cz.metacentrum.perun.core.implApi.modules.pwdmgr.PasswordManagerModule;
 import cz.metacentrum.perun.registrar.model.Application;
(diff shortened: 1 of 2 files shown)


xflord (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -849,27 +852,35 @@ public List<User> getUsersByAttributeValue(PerunSession sess, AttributeDefinitio
...
 			operator = "=";
 		}  else if (attributeDefinition.getType().equals(Boolean.class.getName())) {
-			value = attributeValue.trim();
+			value = "'" + attributeValue.trim() + "'";
 			operator = "=";
 		} else if (attributeDefinition.getType().equals(ArrayList.class.getName())) {
-			value = "%" + attributeValue.trim() + "%";
+			// escape backslashes from delimiters escaping, use delimiters to match value
+			String escapedValue = AttributesManagerBlImpl.escapeListAttributeValue(attributeValue.trim()).replace("\\", "\\\\");
+			String firstElementMatch = escapedValue + LIST_DELIMITER + "%";
```
cuadradek, line 866: We should always use parametrized queries (using JdbcTemplate's `named parameters` or `?`) instead of string concatenation to prevent SQL injections when passing input from users. 
I tested calling this method on my local instance with the parameters:
- attribute name `urn:perun:user:attribute-def:def:kerberosLogins` 
- attribute value `ok']);DROP TABLE user_attr_values CASCADE; commit;`

and the table was really dropped.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -846,31 +849,40 @@ public List<User> getUsersByAttribute(PerunSession sess, Attribute attribute, bo
...
+		} else if (attributeType.equals(Integer.class.getName())) {
+			namedParams.addValue("value", attributeValue.trim());
+		} else if (attributeType.equals(Boolean.class.getName())) {
+			namedParams.addValue("value", attributeValue.trim());
 		} else if (attributeDefinition.getType().equals(ArrayList.class.getName())) {
-			value = "%" + attributeValue.trim() + "%";
+			// escape backslashes from delimiters escaping, use delimiters to match value
 			operator = "like";
+			String escapedValue = AttributesManagerBlImpl.escapeListAttributeValue(attributeValue.trim()).replace("\\", "\\\\");
+			value_part = "any(array[(:regex1),(:regex2)])";
```
cuadradek, line 868: Just a nitpick: could you please rename the regex parameters or add a comment explaining what the regexes should represent (first element match / any other element)? It helped me understand the code quickly when I saw it for the first time.

Johaney-s:
Just realized values can contain backslashed delimiters that would allow partial match of the value, such as
'**cat**' in 'my\\,**cat**'
'**corgi**' in 'best_dogs:my\\,**corgi**,'
so opted to use slower postgres ~ operator instead of LIKE, which would not allow to check for preceding backslash before delimiters. even combination of `LIKE regex and NOT LIKE substring_regex` would not work, as we still need to match cases like 'alena' in 'alena,magdalena' where the NOT LIKE part would filter the correct results out.

Now only cases that are truly between not-backslashed delimiters should be matched.

cuadradek (approved):

PR #3737, opened 2022-10-07
fix(core): fix getMessagesByIdAndCount

mattjoke:
* getMessagesByIdAndCount now returns 'count' of newer messages by provided 'id'


xflord (changes requested):
Update javadocs for the method now that its changed. Also, all other getMessages methods return the messages in a descending order by id, which could be a bit confusing, but I guess it makes sense either way (since we want messages NEWER than 'id', just make sure to make that clear in the docs)

xflord (commented):
Don't forget openapi.yml :p

Johaney-s (approved):

xflord (approved):

PR #3741, opened 2022-10-12
feat: Add method to get OIDC config for Perun instance + CLi script to call this method

xflord:
* Added new method to ConfigManager, which returns new OidcConfig class, which holds the fields of the config.
* The configuration file perun-oidc-config.yml is expected to be in the same directory as other config files (eg. perun.properties, perun-roles.yml, etc.)
* Added loader for the perun-oidc-config.yml
* Added CLI script, which calls the new method and prints out the configuration in the desired format

### DIFF
--- a/perun-cli/getOidcConfig
+++ b/perun-cli/getOidcConfig
@@ -1,0 +1,49 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use JSON;
+
+use Perun::Agent;
+use Getopt::Long qw(:config no_ignore_case);
+
+
+sub help {
+	return qq{
+	Returns the OIDC configuration for this Perun instance.
+	The configuration can then be used in Perun/auth/oidc_config.yml to connect via OIDC.
+
+};
+}
+GetOptions("help|h"            => sub { print help(); exit 0; });
+
+my $agent = Perun::Agent->new();
+
+my $configAgent = $agent->getConfigAgent;
+
+my $oidcConfig = $configAgent->getPerunOidcConfig();
+
+my @result = ();
+
+foreach my $key (keys %$oidcConfig) {
+	my $fieldValue = $oidcConfig->{$key};
+	unless (defined $fieldValue) {
+		warn "WARNING: Field $_ is not defined in the OIDC config of this Perun instance!";
+		next;
+	}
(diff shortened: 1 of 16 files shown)


perun-cli/getOidcConfig
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+
+use Perun::Agent;
+use Getopt::Long qw(:config no_ignore_case);
+
+
+my @fields = ("client_id", "oidc_device_code_uri", "oidc_token_endpoint_uri", "oidc_token_revoke_endpoint_uri", "acr_values", "scopes", "perun_api_endpoint", "enforce_mfa");
```
Johaney-s, line 10: one thing that comes to my mind is if we have all the fields defined like this, then additional property (let's say we add some new url) would be skipped. theoretically users should always update their CLI tools if perun version changes, but it is not always the case. would you say this is better than iterating all the fields that perun sends? what do you think, @xflord , @cuadradek ? also, we probably wouldn't have to extend the OidcConfig.pm and this class if new property is added.

Johaney-s (commented):
really good job! just a thought about the CLI usage...

perun-cli/getOidcConfig
xflord: You're probably right. I did it this way mainly to keep the order of fields (which ultimately doesn't matter, I believe?) Is there a way to iterate all the fields sent in perl, or are you thinking of sending the list of fields from backend as well?

Johaney-s: in ConfigAgent it would go like
`return Perun::Common::callManagerMethod('getPerunOidcConfig', '', @_)`

and in the script something like
> foreach my $key (keys %$oidcConfig) {
> 	my $value = $oidcConfig->{$key};
> 	print $key . ": " . $value . "\n";
> }

but some smart handling of the json will probably be needed, because we should be able to use the string quotes on string and not use them for integer and boolean values (enforceMfa)...

xflord: Used json encoding instead, which adds quotes to strings. The returned fields now aren't always in the same order, however that shouldn't matter.

Kuliak (approved):
Looks good, nice job!

PR #3743, opened 2022-10-13
feat(registrar): added ApplicationNotNewException

xflord:
* ApplicationNotNewException is thrown when trying to send mail verification for an application that is no longer in the state "NEW"


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/exceptions/ApplicationNotNewException.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public class ApplicationNotNewException extends PerunException {
+
+
+
+	private static final long serialVersionUID = 1L;
+
+	private final String type;
+
+	public ApplicationNotNewException(String mess, String type) {
```
Johaney-s, line 19:
```suggestion
	public ApplicationNotNewException(String mess, String state) {
```

Johaney-s (approved):

PR #3748, opened 2022-10-17
feat(core): allow self to call get all entities

bodnara:
* methods of type getAll<Entity> are now callable with just role SELF
* if caller has no other role this would return an empty array

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -1671,7 +1671,7 @@ perun_policies:
 
   getAllGroups_policy:
     policy_roles:
-      - SELF: User
+      - SELF:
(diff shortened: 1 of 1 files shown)


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6477,6 +6477,7 @@ perun_policies:
 
   getAllVos_policy:
     policy_roles:
+      - SELF:
```
Johaney-s, line 6480: there's no filter applied to the result of the method and the method itself returns all vos, so if possible, do not add the SELF role here, or apply filter if needed

bodnara: SELF is gone.

PR #3749, opened 2022-10-17
fix(cli): default OIDC config message fix

Kuliak:
* Message about usage of default OIDC config is printed only when authenticating, not every time configuration is being loaded

### DIFF
--- a/perun-cli/Perun/auth/OidcAuth.pm
+++ b/perun-cli/Perun/auth/OidcAuth.pm
@@ -182,6 +182,10 @@ sub authenticationRequest
 {
 	my $config = loadConfiguration();
 
+	if (!$ENV{PERUN_OIDC_CONFIG} && $config->{"default"}) {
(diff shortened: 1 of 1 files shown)


perun-cli/Perun/auth/OidcAuth.pm
```diff
@@ -182,6 +182,10 @@ sub authenticationRequest
 {
 	my $config = loadConfiguration();
 
+	if ($config->{"default"}) {
+		print "Environment variable PERUN_OIDC_CONFIG not set, using default config $config.\n";
```
xflord, line 186: This would print the message even when PERUN_OIDC_CONFIG is set to the default config, which is contradictory.

xflord (changes requested):

xflord:
Screenshot from 2022-10-18 11-23-54
The name of the config doesn't print correctly

Johaney-s (approved):

xflord (approved):

PR #3753, opened 2022-10-19
fix(registrar): fix duplicate notifications

cuadradek:
- Two duplicate group notifications were sent to group managers when a user submitted a group application in the following scenario:

1. first the user submitted an application to the group's VO, which was auto-approved
2. then he submitted the group application in less than X seconds

- The reason is that after the VO application was approved, a new thread was created which slept for 5 seconds and then validated the member in VO and handled the user's group applications - sent notifications to the group's managers. If a group application was submitted in this time window (5 seconds + validation time) an APP_CREATED_VO_ADMIN notification was sent both in this method and in the second step - submitting the group application.
- #3694 was a quick-fix which removed the thread's sleep, but there was still a big enough time window when this bug could happen.
- Now, a new table created_app_notifications_sent is introduced to save the information whether APP_CREATED_VO_ADMIN notification was already sent.
- The notification is sent always in the second step (submitting group application) if the user has a required state.
- The notification is sent in the first step only if it wasn't sent yet and the user has a required state. There is no need for an additional synchronization (like checking if it is being sent in a concurrent thread), because the first step sends the notification only for committed applications, and they are committed together with information whether the notification was sent.

BREAKING CHANGE: update DB

### DIFF
--- a/perun-base/src/test/resources/test-schema.sql
+++ b/perun-base/src/test/resources/test-schema.sql
@@ -1,4 +1,4 @@
--- database version 3.1.97 (don't forget to update insert statement at the end of file)
+-- database version 3.1.98 (don't forget to update insert statement at the end of file)
 CREATE EXTENSION IF NOT EXISTS "unaccent";
 CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 
@@ -1638,6 +1638,27 @@ create table consent_attr_defs (
 	constraint consentatt_consent_fk foreign key (consent_id) references consents(id)
 );
 
+create type mail_type as enum (
+	'APP_CREATED_USER',
+	'APPROVABLE_GROUP_APP_USER',
+	'APP_CREATED_VO_ADMIN',
+	'MAIL_VALIDATION',
+	'APP_APPROVED_USER',
+	'APP_REJECTED_USER',
+	'APP_ERROR_VO_ADMIN',
+	'USER_INVITE'
+);
+
+-- APP_NOTIFICATIONS_SENT - sent applications notifications, used only for APP_CREATED_VO_ADMIN type for now
+create table app_notifications_sent (
+	app_id int not null,
+	notification_type mail_type not null,
+	created_at timestamp default statement_timestamp() not null,
+	created_by varchar default user not null,
+	constraint appnotifsent_pk primary key(app_id, notification_type),
+	constraint appnotifsent_app_fk foreign key (app_id) references application(id) on delete cascade
+);
+
 create sequence "attr_names_id_seq";
 create sequence "attribute_policies_id_seq";
 create sequence "attribute_policy_collections_id_seq";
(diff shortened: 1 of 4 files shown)


perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,11 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.1.98
+create table created_app_notifications_sent (app_id int not null, created_at timestamp default statement_timestamp() not null, created_by varchar default user not null, constraint appnotifsent_pk primary key(app_id), constraint appnotifsent_app_fk foreign key (app_id) references application(id) on delete cascade);
```
balcirakpeter, line 10: What about already existing applications? How they will be treated when we deploy this feature?

perun-base/src/test/resources/test-schema.sql
```diff
@@ -1638,6 +1638,15 @@ create table consent_attr_defs (
 	constraint consentatt_consent_fk foreign key (consent_id) references consents(id)
 );
 
+-- CREATED_APP_NOTIFICATIONS_SENT - applications for which notification APP_CREATED_VO_ADMIN was sent
+create table created_app_notifications_sent (
+	app_id int not null,
+	created_at timestamp default statement_timestamp() not null,
```
balcirakpeter, line 1644: What about making it more generic so it could be extended in the future by other types of notifications?

balcirakpeter (commented):
Couldn't there be a case when the VO application is in the approval process but at the same time the group application is in the submitting process, therefore, no email is sent?

cuadradek:
> Couldn't there be a case when the VO application is in the approval process but at the same time the group application is in the submitting process, therefore, no email is sent?

Good question. The email is not sent in the group application's submitting process only if the member doesn't exist or doesn't have a VALID or INVALID state. Then the email is not sent, and the group application is committed to the DB. Meanwhile, the approval process of the VO application is validating the member and then would send the email ONLY if the group application is already committed in DB.

Therefore, I believe there is a very short time window when no email would be sent:

- member validation in the VO process finished right after the Group process would decide that it wouldn't send a notification
- VO process then wouldn't see the group application because it is not committed in DB yet

So, the time window is very tiny, and it could happen already before this PR. Therefore, I would say we don't need to worry about that, even considering the new Registrar is on its way.

perun-core/src/main/resources/postgresChangelog.txt
cuadradek: Submitted group applications would be considered that no email was sent after their submission, but it might have been sent if the member already existed and was in a VALID or INVALID state (it was an EXTENSION group application).
Then we deploy these changes, and the email will be always sent when the VO application is approved. In the case of such EXTENSION applications, I believe the email would be indeed sent twice, once before deployment and once after.

Johaney-s (approved):

perun-base/src/test/resources/test-schema.sql
```diff
@@ -1638,6 +1638,16 @@ create table consent_attr_defs (
 	constraint consentatt_consent_fk foreign key (consent_id) references consents(id)
 );
 
+-- APP_NOTIFICATIONS_SENT - sent applications notifications, used only for APP_CREATED_VO_ADMIN type for now
+create table app_notifications_sent (
+	app_id int not null,
+	notification_type varchar not null,
```
balcirakpeter, line 1644: Wouldn't it be better to use an enum to represent the type?

cuadradek: Yes, thanks for the suggestion.

balcirakpeter (approved):

PR #3755, opened 2022-10-24
feat: make find_group_inconsistencies script use user IDS to work bet…

xflord:
…ter with hierarchical Vos


mattjoke (approved):
Looks good! I have also tested it and I haven't found any errors.

Johaney-s (commented):
Looks good, I would just alter the 'Inconsistency found' messages to clearly state the id is UID, not member's id.

Johaney-s (approved):

PR #3756, opened 2022-10-27
feat(core): Role display name

bodnara:
* Added property display name for roles
* Set up display names for roles in configuration
* Added display name to RoleManagementRules

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -7822,6 +7822,7 @@ perun_policies:
 #  assignable_to_attributes is a flag that determines whether the role can appear in attribute policies.
 #  mfa_critical_role is a flag marking roles always requiring MFA from users having that role
 #  skip_mfa is a flag that whether the role should skip MFA check.
+#  display_name is a more user-friendly name
 perun_roles_management:
 
   PERUNADMIN:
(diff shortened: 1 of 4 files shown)


perun-openapi/openapi.yml
```diff
@@ -730,6 +731,7 @@ components:
         - entitiesToManage
         - assignedObjects
         - assignableToAttributes
+        - displayName
```
Johaney-s, line 734: will this work if display name is null?

xflord: works for primaryObject so I suppose it should work for displayName as well?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -7822,6 +7822,7 @@ perun_policies:
 #  assignable_to_attributes is a flag that determines whether the role can appear in attribute policies.
 #  mfa_critical_role is a flag marking roles always requiring MFA from users having that role
 #  skip_mfa is a flag that whether the role should skip MFA check.
+#  display_name is a more user-friendly name
```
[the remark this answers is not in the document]
xflord: ++

perun-openapi/openapi.yml
Johaney-s: right, missed that

perun-base/src/main/resources/perun-roles.yml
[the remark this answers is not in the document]
bodnara: Javadoc was added to both `RoleManagementRules.java` and Ansible templates.

xflord (approved):

PR #3758, opened 2022-10-27
feat(core): support branding between new and old GUI

Johaney-s:
* Added vos field to branding config to distinguish brand to which vo belongs
* Added support for tags directing to new admin-gui in manager's notifications
* Invitations etc. coming from new GUI should be directed to correct branding (vos property in apps-config should ensure that)
* Banner in old gui should be configurable in apps-config to also follow branding rules

BREAKING CHANGE: New brand properties in perun-apps-config.yml, deprecated newAdminGuiAlert in perun-web-gui.properties moved to perun-apps-config and can be branded

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
@@ -35,6 +35,7 @@ import cz.metacentrum.perun.core.bl.GroupsManagerBl;
 import cz.metacentrum.perun.core.bl.MembersManagerBl;
 import cz.metacentrum.perun.core.bl.UsersManagerBl;
 import cz.metacentrum.perun.core.blImpl.AuthzResolverBlImpl;
+import cz.metacentrum.perun.core.impl.PerunAppsConfig;
 import cz.metacentrum.perun.registrar.exceptions.ApplicationMailAlreadyRemovedException;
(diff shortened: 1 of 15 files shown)


mattjoke (approved):
Looks good :+1:

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -1699,14 +1712,21 @@ private String replaceAppDetailUrl(String mailText, int appId, Vo vo, Group grou
...
-					if (newValue != null && !newValue.isEmpty()) {
-						if (!newValue.endsWith("/")) newValue += "/";
-						newValue += namespace + "/gui/";
-						newValue += "?vo/appdetail?id="+appId;
-						//newValue += getFedAuthz().contains(namespace) ? "?vo/appdetail?id="+appId : "#vo/appdetail?id="+appId;
+					if (namespace.equals("newGUI")) {
+						newValue = PerunAppsConfig.getBrandContainingVo(vo.getShortName()).getNewApps().getAdmin();
+						if (newValue != null && !newValue.isEmpty()) {
+							if (!newValue.endsWith("/")) newValue += "/";
+							newValue += "organizations/" + vo.getId() + "/applications/" + appId;
```
balcirakpeter, line 1719: There is a group with an id missing in the URL if it is a group application.

balcirakpeter (changes requested):

balcirakpeter (approved):

PR #3759, opened 2022-11-01
feat(core): optional login for namespace

mattjoke:
* urn allows login to be null or have a value if present
* any namespace can use optional namespace

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -110,6 +110,7 @@ import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_member_attrib
 import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_member_attribute_def_virt_isLifecycleAlterable;
 import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_member_group_attribute_def_virt_groupStatus;
 import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_member_group_attribute_def_virt_groupStatusIndirect;
+import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_user_attribute_def_virt_optional_login_namespace;
 import cz.metacentrum.perun.core.impl.modules.attributes.urn_perun_vo_attribute_def_def_applicationAutoRejectMessages;
 import cz.metacentrum.perun.core.implApi.AttributesManagerImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.AttributesModuleImplApi;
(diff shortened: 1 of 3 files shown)


sarkapalkovicova (commented):
According to the assignment, you should also create a new attribute. The added module looks OK, but I would avoid using the specific name of the namespace since it should be a general solution.

Johaney-s (commented):
What we need is a virtual attribute (with no attribute semantic check) that returns either login in specified namespace, or null if login is not set. That way, we can propagate users who do not have the login set (if the def:login-namespace attribute would be used on service, users without login couldn't be assigned to resource because the login value check would be run and fail on null value). We don't want to create any additional logins for this purpose, so no need to create einfa-login-not-enforced.

As @sarkapalkovicova mentioned, it should be a general solution allowing login in any namespace to be returned this way, so it could go  like this: when virt:optional-login-namespace:einfra is called, def:login-namespace:einfra value is returned.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_optional_login_namespace.java
```diff
@@ -0,0 +1,23 @@
...
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleAbstract;
+
+/**
+ * Attribute, that contains login, only if it is available in the namespace.
+ *
+ * @author Matej Hakoš
+ */
+public class urn_perun_user_attribute_def_virt_optional_login_namespace extends UserVirtualAttributesModuleAbstract {
```
Johaney-s, line 12: We'll need the getValue() implementation that returns user's `def:login-namespace:XY` attribute value - otherwise the implementation from parent class `UserVirtualAttributesModuleAbstract` would be used and returned value would be always null (no real value is stored in this attribute, it'll always be computed on demand).

Could you please add tests for this module? :rocket:

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -8149,6 +8150,21 @@ protected void initialize() {
 			policies.add(Triple.of(Role.FACILITYADMIN, READ, RoleObject.Facility));
 			attributes.put(attr, createInitialPolicyCollections(policies));
 
+			// optional-login-namespace
+			attr = new AttributeDefinition();
+			attr.setNamespace(AttributesManager.NS_USER_ATTR_VIRT);
+			attr.setType(String.class.getName());
+			attr.setFriendlyName("optional-login-namespace:"+namespace);
+			attr.setDisplayName("Optional login in namespace: "+namespace);
+
```
Johaney-s, line 8159:
```suggestion
attr.setDescription("Contains an optional login in namespace" + namespace + " if the user has it.");
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_optional_login_namespace.java
```diff
@@ -0,0 +1,43 @@
...
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl perunSession, User user, AttributeDefinition attribute) {
+		Attribute attr = new Attribute(attribute);
+		String namespace = AttributesManager.NS_USER_ATTR_DEF+":login-namespace:"+attribute.getFriendlyNameParameter();
+		try {
+			Attribute a = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, namespace);
+			attr.setValue(a.getValue());
+		} catch (AttributeNotExistsException | WrongAttributeAssignmentException | NullPointerException e) {
+			// It's OK, we just return attribute with value null
+			return attr;
```
Johaney-s, line 28: I think we should log error in this case - AttributeNotExistsException would be thrown if the login-namespace attribute definition does not exist in the system, which I consider important being notified about.

Returning attr here is not necessary as it is returned at the end of the method anyway.

Is catching the NullPointerException really used?

Johaney-s (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_optional_login_namespace.java
mattjoke: The NullPointerException was left by a mistake, when I was creating the tests.

Johaney-s (approved):

sarkapalkovicova (approved):

PR #3760, opened 2022-11-02
fix(core): fix createGroupUnion sql exception

sarkapalkovicova:
List of members for query was divided into parts.


Johaney-s (approved):

mattjoke (approved):
Look good :+1:

PR #3763, opened 2022-11-07
feat(core): New service account role for SP reg application created

Kuliak:
* Role added into Role.java file
* perun-roles file adjusted to contain newly created role (documentation, method call policies, role management settings)
* new script ensuring adding all necessary policies into DB created


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -8220,6 +8244,25 @@ perun_roles_management:
...
+  SPREGAPPLICATION:
+    primary_object:
+    assign_to_objects: {}
+    assignment_check:
+      - MFA:
+    entities_to_manage:
+      User: user_id
+    privileged_roles_to_manage:
+      - PERUNADMIN:
+      - VOADMIN: Vo
```
cuadradek, line 8256: I don't think we want to let VO admins to set / read this role.

perun-db/addSPRegServiceAccountPolicies
```diff
@@ -0,0 +1,182 @@
...
+
+	if ($sthGetAttributeId->rows() ne 1) {
+		# if no corresponding attribute for given name was found, skip
+		next;
+	}
+
+	my $attrId = $sthGetAttributeId->fetchrow_array();
+	my $attrType = (split ':', $attrName)[2];
+
+    if ($attrType eq "user") {
```
cuadradek, line 150: Do we need this check? Shouldn't we set all of the read attributes no matter their type?

cuadradek (commented):
Overall good work. I also think roles FACILITYADMIN and FACILITYOBSERVERS should have this new role in privileged_roles_to_manage / read so that it can set these roles to a group (addAdmin method from the docs).

perun-base/src/main/resources/perun-roles.yml
Johaney-s: Also please add `skip_mfa: true` as we want to avoid enforcing MFA for the role.

perun-db/addSPRegServiceAccountPolicies
```diff
@@ -0,0 +1,182 @@
...
+        --password  | -w Password for Perun DB (required)
+        --overwrite | -o removes all existing COLLECTIONS!
+        };
+}
+
+# Specific attributes besides READ for facility attributes
+my @readAttributes = (
+	"urn:perun:user:attribute-def:def:preferredMail"
+);
+my @writeAttributes = (
```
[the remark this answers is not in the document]
Johaney-s: there are similar attribute definitions in the original docs:
- OIDCIssueRefreshTokens, OIDCIssuedRefreshTokens
- singleLogoutService, singleLogoutServices
- assertionConsumerService,assertionConsumerServices

and in the docs there is rpPrivacyPolicyURL, but in the list here rpPrivacyPolicy - did you discuss the differences with the docs authors? Were they all a typo?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -8220,6 +8248,25 @@ perun_roles_management:
...
+      - MFA:
+    entities_to_manage:
+      User: user_id
+    privileged_roles_to_manage:
+      - PERUNADMIN:
+    privileged_roles_to_read:
+      - PERUNADMIN:
+      - PERUNOBSERVER:
+      - VOADMIN: Vo
+      - VOOBSERVER: Vo
```
cuadradek, line 8264: I would also remove VOADMIN and VOOBSERVER from here.

cuadradek (approved):

PR #3766, opened 2022-11-14
fix(openapi): Fix getBansForResource param name

bodnara:
* openapi now defines correct name for resource id: "resourceId"

### DIFF
--- a/perun-cli/listOfBans
+++ b/perun-cli/listOfBans
@@ -57,7 +57,7 @@ if (defined($resourceId) or defined($memberId)) {
 	my $banOnResourceAgent = $agent->getBanOnResourceAgent;
 
 	if (defined($resourceId)) {
-		@bans = $banOnResourceAgent->getBansForResource( resourceId => $resourceId );
+		@bans = $banOnResourceAgent->getBansForResource( resource => $resourceId );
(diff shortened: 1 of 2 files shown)


perun-openapi/openapi.yml
```diff
@@ -14228,7 +14228,7 @@ paths:
       operationId: getBansForResource
       summary: Get all bans for members on the resource.
       parameters:
-        - $ref: '#/components/parameters/resourceId'
+        - { name: resourceId, required: true, description: "Resource id", schema: { type: integer },  in: query }
```
xflord, line 14231: How about changing the param name in `ResourcesManagerMethod` instead of here? Then we can still use the component resourceId instead of just duplicating it with a different name.

bodnara: The reason for this was to avoid changing the param name in all other places (mostly CLI) that already send it as "resourceId".

xflord: Screenshot from 2022-11-15 10-45-46
This doesn't look like much. Would you mind fixing this so we are consistent across the board, now that we're going to be working with bans more often?

xflord (approved):

PR #3768, opened 2022-11-15
feat: extended the getPerunOidcConfig method to be able to return one of multiple stored OidConfigs + extended the CLI script

xflord:
* The configuration file perun-oidc-config.yml can now hold multiple named oidc configurations
* Added PerunOidcConfigContainer to store these configurations
* The CLI script getOidcConfig now takes 'configName' as parameter


perun-base/src/test/resources/perun-oidc-config.yml
```diff
@@ -1,8 +1,9 @@
-client_id: "363b656e-d139-4290-99cd-ee64eeb830d5" # Identifier of the CLI app on the OIDC server
-oidc_device_code_uri: "https://login.cesnet.cz/oidc/devicecode" # Here the CLI requests from the OIDC server to start the authentication
-oidc_token_endpoint_uri: "https://login.cesnet.cz/oidc/token" # Token endpoint URI
-oidc_token_revoke_endpoint_uri: "https://login.cesnet.cz/oidc/revoke" # Token revoke endpoint URI
-acr_values: "" # List of ACR values for authentication
-scopes: "openid perun_api perun_admin offline_access" # List of scopes the access_token should grant access to
-perun_api_endpoint: "https://perun-dev.cesnet.cz/oauth/rpc"
-enforce_mfa: false # Nastavenim true se vynuti MFA pri prihlaseni
+perun:
+  client_id: "363b656e-d139-4290-99cd-ee64eeb830d5" # Identifier of the CLI app on the OIDC server
```
Johaney-s, line 2: we got feedback that we shouldn't use real values for test configs

Johaney-s (commented):
i think we'll want to resolve the "name" parameter by domain of the request, but let's see

perun-core/src/main/java/cz/metacentrum/perun/core/impl/PerunOidcConfigLoader.java
```diff
@@ -20,10 +23,23 @@ public void setConfigurationPath(Resource configurationPath) {
...
+		while (configNames.hasNext()) {
+			String configName = configNames.next();
+			oidcConfigs.put(configName, getOidcConfigFromJsonNode(rootNode.get(configName)));
+		}
+		return oidcConfigs;
+	}
+
+	private static OidcConfig getOidcConfigFromJsonNode(JsonNode rootNode) {
+		OidcConfig oidcConfig = new OidcConfig();
+		System.out.println(rootNode.asText());
```
cuadradek, line 42: You probably forgot to delete this line.

cuadradek (changes requested):

perun-cli/Perun/Agent.pm
```diff
@@ -64,6 +64,16 @@ sub gotMfaPrivilegeException;
 sub new {
 	my $self = fields::new(shift);
 
+	$self->{_useNon} = shift;
+	if ($self->{_useNon}) {
+		if (defined($ENV{PERUN_OIDC}) && $ENV{PERUN_OIDC} eq "1") {
```
Johaney-s, line 69: can we use the oidc's perun url in this case (from the config file, same as would be used in other scripts)?

perun-base/src/test/resources/perun-oidc-config.yml
```diff
@@ -1,8 +1,9 @@
-client_id: "363b656e-d139-4290-99cd-ee64eeb830d5" # Identifier of the CLI app on the OIDC server
-oidc_device_code_uri: "https://login.cesnet.cz/oidc/devicecode" # Here the CLI requests from the OIDC server to start the authentication
-oidc_token_endpoint_uri: "https://login.cesnet.cz/oidc/token" # Token endpoint URI
-oidc_token_revoke_endpoint_uri: "https://login.cesnet.cz/oidc/revoke" # Token revoke endpoint URI
-acr_values: "" # List of ACR values for authentication
-scopes: "openid perun_api perun_admin offline_access" # List of scopes the access_token should grant access to
-perun_api_endpoint: "https://perun-dev.cesnet.cz/oauth/rpc"
-enforce_mfa: false # Nastavenim true se vynuti MFA pri prihlaseni
+dummy:
```
Johaney-s, line 1: please mark this commit with breaking change because the versions will be incompatible

Johaney-s (changes requested):

cuadradek (approved):

Johaney-s (commented):
One more thing - the file imho should not be required to exist on machine (some instances don't use oidc), now if the config file is missing, perun won't start cause 
`java.io.FileNotFoundException: /etc/perun/perun-oidc-config.yml (No such file or directory)`

xflord:
> One more thing - the file imho should not be required to exist on machine (some instances don't use oidc), now if the config file is missing, perun won't start cause `java.io.FileNotFoundException: /etc/perun/perun-oidc-config.yml (No such file or directory)`

Made it so that when the file does not exist, the information is logged and when the `getPerunOidcConfig` method is consequently called, an exception is thrown

Johaney-s (approved):

cuadradek (approved):

PR #3770, opened 2022-11-15
fix(core): remove duplicate combinations of facility-service-destinations, while assuring other duplicates won't happen

xflord:
* SQL command in postgresChangelog to delete any existing duplicates
* added primary key to the facility_service_destinations table
* throw exception when duplicate key detected

BREAKING CHANGE: db update


cuadradek (commented):
Looks good.

Johaney-s (approved):

PR #3773, opened 2022-11-21
feat(core): make roles receiving notifications configurable

Johaney-s:
* we're expanding roles that should get notifications when application is created/failed
* extended perun-roles configuration with receive_notifications property which can contain GROUP or VO objects
* when application triggers notification, all roles that contain such object are retrieved, not only voadmin/groupadmin

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
@@ -34,6 +34,8 @@ import java.util.Objects;
  * skipMFA is a flag that whether the role should skip MFA check.
  * mfaCriticalRole is a flag marking roles always requiring MFA from users having that role
  * displayName is a more user-friendly name
+ * receiveNotifications contains names of objects for which the role should get notifications
+ * 			   Example value: Vo ; meaning: will receive notifications when vo application is created/failed
  */
 public class RoleManagementRules {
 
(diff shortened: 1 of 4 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleManagementRules.java
```diff
@@ -34,6 +34,8 @@
  * skipMFA is a flag that whether the role should skip MFA check.
  * mfaCriticalRole is a flag marking roles always requiring MFA from users having that role
  * displayName is a more user-friendly name
+ * receiveNotifications contains names of objects for which the role should get notifications
+ * 			   Example value: VO ; meaning: will receive notifications when vo application is created/failed
```
bodnara, line 38: I think it would be nice to also add this to the perun-roles and https://github.com/CESNET/ansible-role-perun-docker-server

Johaney-s: thanks, added. also missed the docs in perun-roles.yml, added now.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -7949,6 +7953,8 @@ perun_roles_management:
       - GROUPOBSERVER
     assignable_to_attributes: true
     display_name: "Group admin"
+    receive_notifications:
+      - GROUP
```
xflord, line 7957: Why not use 'Group' and 'Vo' to be consistent with `primary_object` etc..? Is there any specific reason that I'm missing for it to be all caps?

Johaney-s: no specific reason, just personal preference for upper case list elements. changed here and in ansible template docs.

xflord (approved):

bodnara (approved):

PR #3775, opened 2022-11-22
feat(core): update ban on Vo

bodnara:
* updateBan for VoManager was added to missing layers
* created test for updateBan on Vo
* setBan no longer updates existing ban but throws exception to stay consistent with variants of this method for other objects

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
@@ -2,6 +2,7 @@ package cz.metacentrum.perun.core.blImpl;
 
 import cz.metacentrum.perun.audit.events.MembersManagerEvents.MemberSuspended;
 import cz.metacentrum.perun.audit.events.MembersManagerEvents.MemberUnsuspended;
+import cz.metacentrum.perun.audit.events.VoManagerEvents.BanUpdatedForVo;
 import cz.metacentrum.perun.audit.events.VoManagerEvents.VoCreated;
 import cz.metacentrum.perun.audit.events.VoManagerEvents.VoDeleted;
 import cz.metacentrum.perun.audit.events.VoManagerEvents.VoUpdated;
@@ -17,7 +18,6 @@ import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Host;
 import cz.metacentrum.perun.core.api.Member;
 import cz.metacentrum.perun.core.api.MemberCandidate;
-import cz.metacentrum.perun.core.api.Pair;
 import cz.metacentrum.perun.core.api.PerunSession;
 import cz.metacentrum.perun.core.api.Resource;
 import cz.metacentrum.perun.core.api.RichUser;
@@ -40,7 +40,6 @@ import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
 import cz.metacentrum.perun.core.api.exceptions.GroupExistsException;
 import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.LoginNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.NotGroupMemberException;
 import cz.metacentrum.perun.core.api.exceptions.PerunException;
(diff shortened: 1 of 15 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/VosManagerEntry.java
```diff
@@ -694,6 +696,23 @@ public BanOnVo setBan(PerunSession sess, BanOnVo ban) throws PrivilegeException,
...
 
+	@Override
+	public BanOnVo updateBan(PerunSession sess, BanOnVo banOnVo) throws PrivilegeException, MemberNotExistsException, BanNotExistsException, VoNotExistsException {
+		Utils.checkPerunSession(sess);
+		vosManagerBl.getBanById(sess, banOnVo.getId());
+		Member member = getPerunBl().getMembersManagerBl().getMemberById(sess, banOnVo.getMemberId());
+		Vo vo = getPerunBl().getVosManagerBl().getVoById(sess, banOnVo.getVoId());
+
+		// Authorization
+		if (!AuthzResolver.authorizedInternal(sess, "updateBan_BanOnVo_policy", Arrays.asList(banOnVo, member, vo))) {
```
xflord, line 707:
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "updateBan_BanOnVo_policy", Arrays.asList(banOnVo))) {
```
Vo and Member get fetched from BanOnVo, no need to include them here

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -872,15 +873,17 @@ public List<BanOnVo> getBansForVo(PerunSession sess, int voId) {
...
 
 		// if the validity is not specified, set a date from far future
 		if (banOnVo.getValidityTo() == null) {
 			banOnVo.setValidityTo(FAR_FUTURE);
 		}
-
-		return vosManagerImpl.updateBan(sess, banOnVo);
+		banOnVo = getVosManagerImpl().updateBan(sess, banOnVo);
+		Member member = getPerunBl().getMembersManagerBl().getMemberById(sess, banOnVo.getMemberId());
+		getPerunBl().getAuditer().log(sess, new MemberSuspended(member));
```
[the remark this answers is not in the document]
bodnara: I am not sure either, I assumed it just generates log messages.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/VosManagerEntryIntegrationTest.java
```diff
@@ -558,6 +559,26 @@ public void getBansForVo() throws Exception {
...
 
+	@Test
+	public void updateBan() throws Exception {
+		System.out.println(CLASS_NAME + "updateBan");
+		Vo vo = vosManagerEntry.createVo(sess, myVo);
+		Member member = createMemberFromExtSource(vo);
+
+		BanOnVo banOnVo = new BanOnVo();
+		banOnVo.setMemberId(member.getId());
+		banOnVo.setVoId(vo.getId());
```
Kuliak, line 570: Technically, you don't have to set VO's ID here because it is set later in the setBan method using member's voId value. However, it's just a tiny detail.

Kuliak (commented):
I like the solution, good job.
There are some unused imports (as a leftover from some other commits as well) in files that might have been removed as a cleanup. You can have it done automatically as well, so you don't have to solve this problem in the future manually.

Maybe just a small nitpick: Since we use BanAlreadyExists exception for VO, we might want to update its javadoc that says it is used for facility/resource only.

bodnara:
> I like the solution, good job. There are some unused imports (as a leftover from some other commits as well) in files that might have been removed as a cleanup. You can have it done automatically as well, so you don't have to solve this problem in the future manually.

Removed unused imports from affeted files.

> Maybe just a small nitpick: Since we use BanAlreadyExists exception for VO, we might want to update its javadoc that says it is used for facility/resource only.

Updated javadoc.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
```diff
@@ -65,17 +64,7 @@
...
-import java.util.Collections;
-import java.util.Date;
-import java.util.GregorianCalendar;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
xflord, line 67: Please no wildcard imports

xflord (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
Kuliak, line 67: we don't want to use wildcard imports. Please, use explicit imports without asterisk
(can be set in: **_Settings - Editor - Code Style - Java - Imports - Class count to use import with '*': 500_**)

bodnara: Thanks for the settings tip.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/VosManagerBlImpl.java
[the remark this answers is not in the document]
bodnara: I created new event for the ban update after all.

xflord (approved):

Kuliak (approved):

PR #3777, opened 2022-11-22
feat(core): filter audit messages by event type

sarkapalkovicova:
Added param to getMessagesPage method according to which messages will be filtered.

### DIFF
--- a/perun-core/pom.xml
+++ b/perun-core/pom.xml
@@ -174,6 +174,11 @@
 			<artifactId>jackson-datatype-jsr310</artifactId>
 		</dependency>
 
+		<dependency>
+			<groupId>org.reflections</groupId>
+			<artifactId>reflections</artifactId>
+		</dependency>
+
 		<!-- PostgreSQL driver -->
 		<dependency>
 			<groupId>org.postgresql</groupId>
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuditMessagesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuditMessagesManager.java
@@ -62,7 +62,7 @@ public interface AuditMessagesManager {
 	List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int id, int count) throws PrivilegeException;
 
 	/**
-	 * Returns page of audit messages. Query parameter specifies offset and page size. Total count is only estimated.
+	 * Returns page of audit messages. Query parameter specifies offset and page size and allows filtering by name of event. Total count is only estimated.
 	 *
 	 * @param perunSession perun session
 	 * @return Page of audit messages
@@ -175,4 +175,12 @@ public interface AuditMessagesManager {
 	 */
 	int getAuditerMessagesCount(PerunSession perunSession);
 
+	/**
+	 * Return list of names of all possible events
+	 *
+	 * @param sess perun session
+	 * @return list of all possible events
+	 * @throws InternalErrorException When implementation fails
+	 */
+	List<String> findAllPossibleEvents(PerunSession sess);
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuditMessagesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuditMessagesManagerBlImpl.java
@@ -6,11 +6,14 @@ import cz.metacentrum.perun.core.api.AuditMessage;
 import cz.metacentrum.perun.core.api.MessagesPageQuery;
 import cz.metacentrum.perun.core.api.Paginated;
 import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.bl.AuditMessagesManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.impl.Auditer;
 import cz.metacentrum.perun.core.implApi.AuditMessagesManagerImplApi;
+import org.reflections.Reflections;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
@@ -67,6 +70,21 @@ public class AuditMessagesManagerBlImpl implements AuditMessagesManagerBl {
 	@Override
 	public Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, MessagesPageQuery query) {
 		return getAuditMessagesManagerImpl().getMessagesPage(perunSession, query);
+	}
+
+	@Override
+	public List<String> findAllPossibleEvents(PerunSession sess) {
+		try {
+			Reflections reflections = new Reflections(AuditEvent.class);
+			List<String> events = new ArrayList<>((reflections.getSubTypesOf(AuditEvent.class).stream().map(Class::getSimpleName).toList()));
+
+			// Remove subclasses outside subpackages.
+			events.remove("StringMessageEvent");
+
+			return events;
+		} catch (Exception e) {
+			throw new InternalErrorException(e);
+		}
 	}
 
 	@Override
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
@@ -42,6 +42,8 @@ import org.springframework.dao.EmptyResultDataAccessException;
 import org.springframework.jdbc.core.JdbcPerunTemplate;
 import org.springframework.jdbc.core.ResultSetExtractor;
 import org.springframework.jdbc.core.RowMapper;
+import org.springframework.jdbc.core.namedparam.MapSqlParameterSource;
+import org.springframework.jdbc.core.namedparam.NamedParameterJdbcTemplate;
 
 import javax.sql.DataSource;
 import java.io.IOException;
@@ -68,6 +70,8 @@ public class AuditMessagesManagerImpl implements AuditMessagesManagerImplApi {
 	private static final int PAGE_COUNT_PRECISION = 1000;
 
 	private final JdbcPerunTemplate jdbc;
+
+	private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
 
 	static {
 
@@ -178,6 +182,7 @@ public class AuditMessagesManagerImpl implements AuditMessagesManagerImplApi {
 
 	public AuditMessagesManagerImpl(DataSource perunPool) {
 		this.jdbc = new JdbcPerunTemplate(perunPool);
+		this.namedParameterJdbcTemplate = new NamedParameterJdbcTemplate(perunPool);
 	}
 
 	@Override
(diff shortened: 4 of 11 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/AuditMessagesManager.java
```diff
@@ -62,13 +62,14 @@ public interface AuditMessagesManager {
...
-	 * Returns page of audit messages. Query parameter specifies offset and page size. Total count is only estimated.
+	 * Returns page of audit messages filtered by name of event. Query parameter specifies offset and page size. Total count is only estimated.
 	 *
 	 * @param perunSession perun session
+	 * @param eventName name of event
 	 * @return Page of audit messages
 	 * @throws InternalErrorException When implementation fails
 	 */
-	Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, MessagesPageQuery query) throws PrivilegeException;
+	Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, MessagesPageQuery query, String eventName) throws PrivilegeException;
```
Johaney-s, line 72: instead of putting new parameters to paging methods, we put the parameters into the PageQuery object  (see `UsersPageQuery`) and then retrieve the parameters from there

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
```diff
@@ -213,12 +213,12 @@ public List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int
...
 
 	@Override
-	public Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, MessagesPageQuery query) {
+	public Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, MessagesPageQuery query, String eventName) {
 		// take exact total count up to PAGE_COUNT_PRECISION entries, estimate it otherwise
 		return jdbc.query("select " + auditMessageMappingSelectQuery
 			+ ", case when (select count(*) from (select 1 from auditer_log limit" + PAGE_COUNT_PRECISION + ") as sample) < " + PAGE_COUNT_PRECISION + " then (select count(*) from auditer_log) else " +
 			"(select 100 * count(*) FROM auditer_log TABLESAMPLE SYSTEM (1)) end as total_count "
-			+ "from auditer_log order by id " + query.getOrder().getSqlValue() + " offset " + query.getOffset()
+			+ "from auditer_log where msg::json ->>'name' ilike '%" + eventName + "%' order by id " + query.getOrder().getSqlValue() + " offset " + query.getOffset()
```
Johaney-s, line 221: consider allowing a list of event names to be passed as parameter

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AuditMessagesManagerEntryIntegrationTest.java
```diff
@@ -183,7 +183,7 @@ public void getMessagesPage_ascendingOrder() throws Exception {
 
 		MessagesPageQuery query = new MessagesPageQuery(2, 0, SortingOrder.ASCENDING);
 
-		Paginated<AuditMessage> messages = perun.getAuditMessagesManager().getMessagesPage(sess, query);
+		Paginated<AuditMessage> messages = perun.getAuditMessagesManager().getMessagesPage(sess, query, "");
 		assertThat(messages.getData().get(0).getId()).isLessThan(messages.getData().get(1).getId());
 	}
```
Johaney-s, line 189: please add test for the new use case

Johaney-s (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
Johaney-s: also, i think it would be faster to exact match the name value than to use 'like', but that requires defining available types of audit messages and passing it to GUI, where user could select them from a dropdown and in the request we would then obtain the correct type name. I think Michal mentioned on planning we could use reflection for this, or write a list of available audit messages manually.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuditMessagesManagerBlImpl.java
```diff
@@ -69,6 +76,22 @@ public Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, Messag
 		return getAuditMessagesManagerImpl().getMessagesPage(perunSession, query);
 	}
 
+	@Override
+	public List<String> findAllPossibleEvents(PerunSession sess) {
+		try {
+			List<String> events = ClassPath.from(ClassLoader.getSystemClassLoader())
+					.getAllClasses()
+					.stream()
+					.filter(clazz -> clazz.getName().matches("^cz\\.metacentrum\\.perun\\.audit\\.events\\.[^.]+\\.[A-Za-z]+$"))
```
xflord, line 85: Feel like you could use ServiceLoader here to do it with less code and without regexp

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuditMessagesManager.java
```diff
@@ -62,7 +62,7 @@ public interface AuditMessagesManager {
 	List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int id, int count) throws PrivilegeException;
 
 	/**
-	 * Returns page of audit messages. Query parameter specifies offset and page size. Total count is only estimated.
+	 * Returns page of audit messages filtered by name of event. Query parameter specifies offset and page size. Total count is only estimated.
```
xflord, line 65:
```suggestion
	 * Returns page of audit messages. Query parameter specifies offset and page size and allows filtering  by name of event. Total count is only estimated.
```
nitpick - theoretically the query parameters also includes the filtering.

xflord (commented):
Also, I believe a test testing whether the filter itself works is still missing

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuditMessagesManagerBlImpl.java
```diff
@@ -69,6 +74,23 @@ public Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, Messag
 		return getAuditMessagesManagerImpl().getMessagesPage(perunSession, query);
 	}
 
+	@Override
+	public List<String> findAllPossibleEvents(PerunSession sess) {
```
Johaney-s, line 78: I tried the following code, and it seems to be working on our devel:

	public List<String> findAllPossibleEvents(PerunSession sess) {
			Reflections reflections = new Reflections(AuditEvent.class);
			return (reflections.getSubTypesOf(AuditEvent.class).stream().map(Class::getSimpleName).collect(Collectors.toList()));
	}

perun-core/pom.xml
```diff
@@ -198,6 +198,14 @@
 			<groupId>com.oracle.database.nls</groupId>
 			<artifactId>orai18n</artifactId>
 		</dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+        </dependency>
+		<dependency>
```
Johaney-s, line 205: duplicated dependency, also maybe move it to "OTHERS" section of the pom file as it is not Oracle's package

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AuditMessagesManagerEntryIntegrationTest.java
```diff
@@ -181,12 +182,21 @@ public void getMessagesPage_ascendingOrder() throws Exception {
...
 
+	@Test
+	public void findAllPossibleEvents() throws Exception {
+		System.out.println(CLASS_NAME + "findAllPossibleEvents");
+
+		List<String> events = perun.getAuditMessagesManager().findAllPossibleEvents(sess);
+
+		assertThat(events.size()).isEqualTo(208);
```
Johaney-s, line 197: I think it would be pain to keep changing this test every time new audit event is created. Maybe just checking that some (>10?) are returned? @xflord ideas?

However, please add test case where you use the new query parameter to get audit messages (you can log some audit message with `perun.getAuditer().logWithoutTransaction` method)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuditMessagesManagerBlImpl.java
```diff
@@ -1,18 +1,27 @@
...
 import cz.metacentrum.perun.core.api.Paginated;
 import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.bl.AuditMessagesManagerBl;
 import cz.metacentrum.perun.core.bl.PerunBl;
 import cz.metacentrum.perun.core.impl.Auditer;
 import cz.metacentrum.perun.core.implApi.AuditMessagesManagerImplApi;
+import org.reflections.Reflections;
 
+import java.io.IOException;
```
Johaney-s, line 17: unused imports

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
```diff
@@ -214,12 +219,15 @@ public List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int
...
 	@Override
 	public Paginated<AuditMessage> getMessagesPage(PerunSession perunSession, MessagesPageQuery query) {
+		MapSqlParameterSource parameters = new MapSqlParameterSource();
+		parameters.addValue("selectedEvents", query.getSelectedEvents());
+
 		// take exact total count up to PAGE_COUNT_PRECISION entries, estimate it otherwise
-		return jdbc.query("select " + auditMessageMappingSelectQuery
+		return namedParameterJdbcTemplate.query("select " + auditMessageMappingSelectQuery
 			+ ", case when (select count(*) from (select 1 from auditer_log limit" + PAGE_COUNT_PRECISION + ") as sample) < " + PAGE_COUNT_PRECISION + " then (select count(*) from auditer_log) else " +
 			"(select 100 * count(*) FROM auditer_log TABLESAMPLE SYSTEM (1)) end as total_count "
```
Johaney-s, line 228: the total count should be updated as well, it should represent how many results there are after applying all the filters, not before

Johaney-s (commented):
really like it, this feature is perfect :rocket:. some thought

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AuditMessagesManagerEntryIntegrationTest.java
xflord: Larger than a set amount sounds reasonable. Also maybe check that it includes some of the essential events (MemberCreated for example), to make sure that the format of returned events is correct

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
```diff
@@ -214,12 +219,16 @@ public List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int
...
-		return jdbc.query("select " + auditMessageMappingSelectQuery
-			+ ", case when (select count(*) from (select 1 from auditer_log limit" + PAGE_COUNT_PRECISION + ") as sample) < " + PAGE_COUNT_PRECISION + " then (select count(*) from auditer_log) else " +
-			"(select 100 * count(*) FROM auditer_log TABLESAMPLE SYSTEM (1)) end as total_count "
-			+ "from auditer_log order by id " + query.getOrder().getSqlValue() + " offset " + query.getOffset()
-			+ " limit " + query.getPageSize(), getPaginatedMessagesExtractor(query));
+		return namedParameterJdbcTemplate.query("select " + auditMessageMappingSelectQuery
+			+ ", case when (select count(*) from (select 1 from auditer_log " + (!query.getSelectedEvents().isEmpty() ? "where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents) " : "") + "limit " + PAGE_COUNT_PRECISION + ") as sample) < " + PAGE_COUNT_PRECISION
+			+ " then (select count(*) from auditer_log" + (!query.getSelectedEvents().isEmpty() ? " where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents)" : "") +")"
+			+ " else (select 100 * count(*) FROM auditer_log TABLESAMPLE SYSTEM (1)) end as total_count "
```
Johaney-s, line 229:
```suggestion
			+ " else (select 100 * count(*) FROM auditer_log TABLESAMPLE SYSTEM (1)" + (!query.getSelectedEvents().isEmpty() ? " where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents)" : "") +") end as total_count "
```
the 1% sample should also be filtered by the event name, otherwise the estimation of the total count is again for the unfiltered table.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuditMessagesManagerImpl.java
```diff
@@ -214,12 +219,16 @@ public List<AuditMessage> getMessagesByIdAndCount(PerunSession perunSession, int
...
-			+ "from auditer_log order by id " + query.getOrder().getSqlValue() + " offset " + query.getOffset()
-			+ " limit " + query.getPageSize(), getPaginatedMessagesExtractor(query));
+		return namedParameterJdbcTemplate.query("select " + auditMessageMappingSelectQuery
+			+ ", case when (select count(*) from (select 1 from auditer_log " + (!query.getSelectedEvents().isEmpty() ? "where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents) " : "") + "limit " + PAGE_COUNT_PRECISION + ") as sample) < " + PAGE_COUNT_PRECISION
+			+ " then (select count(*) from auditer_log" + (!query.getSelectedEvents().isEmpty() ? " where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents)" : "") +")"
+			+ " else (select 100 * count(*) FROM auditer_log TABLESAMPLE SYSTEM (1)" + (!query.getSelectedEvents().isEmpty() ? " where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents)" : "") +") end as total_count "
+			+ "from auditer_log " + (!query.getSelectedEvents().isEmpty() ? "where split_part(msg::json ->>'name', '.', 7) IN (:selectedEvents) " : "") +  "order by id " + query.getOrder().getSqlValue() + " offset " + query.getOffset()
```
xflord, line 230: Just a cosmetic detail, but maybe I would extract the `split_part` string (and the ternary expression whole) into a local variable to make it a bit more readable.

xflord (approved):
Otherwise looks very nice

PR #3779, opened 2022-11-23
feat(core): anonymized attribute added to USER object

Kuliak:
* anonymized attribute added to USER object
* equals, hashcode methods updated for USER
* toString and serializeToString methods updated for both USER and RICHUSER to correctly list all user attributes
* impl layer adjusted to work with newly added attribute (mapping/select)
* openApi USER object updated
* test for anonymizeUser adjusted to check the value of newly added attribute as well

### DIFF
--- a/perun-auditparser/src/test/java/cz/metacentrum/perun/auditparser/AuditParserTest.java
+++ b/perun-auditparser/src/test/java/cz/metacentrum/perun/auditparser/AuditParserTest.java
@@ -29,7 +29,7 @@ public class AuditParserTest {
 	private final String textMismatch = "!@#$%^<<&*()_+<\\><:{[}][]{>} sd";
 	private final String CLASS_NAME = "AuditMessagesManagerEntry";
 	private AuditMessage createdAuditMessage = new AuditMessage();
-	private final User user = new User(5, textMismatch,textMismatch, textMismatch, textMismatch, textMismatch, false, false);
+	private final User user = new User(5, textMismatch,textMismatch, textMismatch, textMismatch, textMismatch, false, false, false);
 	private final ExtSource extSource = new ExtSource(9, textMismatch, textMismatch);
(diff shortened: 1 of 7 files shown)


Johaney-s (commented):
looks good to me, please update User and RichUser also in `parseRpcMethods.pl`

zlamalp (commented):
Please note, that also other objects extend the User object (like Candidate, CandidateSync). Same objects are also used in API, so we need to test, how it affect de/serialization of them in RPC and old GUI.
For example we should test, if we can create new User/Member from a Candidate through old GUI. Also this might affect synchronization between perun instances (CESNET,MU,Lifescience) - it will fail, if we won't deploy same version on all of them at once. So we might want to make this property optional/ignored in `cz.metacentrum.perun.rpc.deserializer.JsonDeserializer` which is used by API and ExtSourcePerun.

Kuliak:
> Please note, that also other objects extend the User object (like Candidate, CandidateSync). Same objects are also used in API, so we need to test, how it affect de/serialization of them in RPC and old GUI. For example we should test, if we can create new User/Member from a Candidate through old GUI. Also this might affect synchronization between perun instances (CESNET,MU,Lifescience) - it will fail, if we won't deploy same version on all of them at once. So we might want to make this property optional/ignored in `cz.metacentrum.perun.rpc.deserializer.JsonDeserializer` which is used by API and ExtSourcePerun.

Okay, thank you for your feedback! Do you think that _**serviceUser**_ and _**sponsoredUser**_ along with _**anonymized**_ should be added to _**JsonIgnoreProperties**_ for _**UserMixIn**_, or _**anonymized**_ only? It makes sense to me that only _**anonymized**_ will be added there.

Johaney-s:
Solved in https://github.com/CESNET/perun/pull/3785 by creating virtual attribute.

PR #3782, opened 2022-11-25
feat(registrar): allow bulk invitations

Johaney-s:
* similarily to generating sponsored members, we want to allow inviting users by provided csv

### DIFF
--- a/perun-openapi/openapi.yml
+++ b/perun-openapi/openapi.yml
@@ -15052,6 +15052,39 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /json/registrarManager/sendInvitationsFromCsv:
+    post:
+      tags:
+        - RegistrarManager
+      operationId: sendInvitationsFromCsv
+      summary: Send invitations with link to VO / Group application form from provided csv data
+          If VO or Group have non-empty attribute urn:perun:[vo/group]:attribute-def:def:applicationURL
+          content is used as link to application form. Otherwise link is automatically generated based on
+          required AUTHZ in template and registrar url set in /etc/perun/perun-registrar.properties.
+
+          General errors result in exception being thrown, single failures are skipped and added to result
+      responses:
+        '200':
+          $ref: '#/components/responses/MapStringStringResponse'
+        default:
+          $ref: '#/components/responses/ExceptionResponse'
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              title: InputInvitationsFromCsv
+              description: "input for sending invitations from csv"
+              type: object
+              required:
+                - voId
+                - invitationData
+              properties:
+                invitationData: { type: array, items: { type: string }, description: "values separated with semicolons" }
+                voId: { type: number, description: "id of VO to send invitation into" }
+                groupId: { type: number, description: "id of Group to send invitation into" }
+                language: { type: string }
+
   /urlinjsonout/registrarManager/sendInvitation/g:
     post:
       tags:
--- a/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
+++ b/perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
@@ -137,6 +137,41 @@ public enum RegistrarManagerMethod implements ManagerMethod {
 		}
 
 	},
+
+	/*#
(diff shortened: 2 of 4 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
```diff
@@ -137,6 +137,41 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * Send invitations with link to VO / Group application form from provided csv data
+	 *
+	 * If VO or Group have non-empty attribute urn:perun:[vo/group]:attribute-def:def:applicationURL
+	 * content is used as link to application form. Otherwise link is automatically generated based on
+	 * required AUTHZ in template and registrar url set in /etc/perun/perun-registrar.properties.
+	 *
+	 * General errors result in exception being thrown, single failures are skipped and added to result
+	 *
+	 * @param invitationData csv file values separated by semicolon ';'. Only [email; name] or [email] is valid format.
+	 * 		example: ["mail@mail.cz", "mail2@mail.cz;user2"]
```
zlamalp, line 151: You can use `@exampleParam invitationData ["mail@mail.cz", "mail2@mail.cz;user2"]` to include it in example method call in generated API docs.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
```diff
@@ -137,6 +137,41 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @param vo VO to link form to
```
zlamalp, line 152: `vo` param type should be `int` and not `VO` as we read it as integer in the code. Also I would prefer using `voId`, just like in standard `sendInvitation()` methods. Same goes for `group` param.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
```diff
@@ -137,6 +137,41 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * @param group Group to link form to
+	 * @param language Language used in notification (if not specified, VO settings is used, if not set, "en" is used).
```
zlamalp, line 154: Please include param type `String` for the language.

zlamalp (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
```diff
@@ -137,6 +137,41 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 * Send invitations with link to VO / Group application form from provided csv data
+	 *
+	 * If VO or Group have non-empty attribute urn:perun:[vo/group]:attribute-def:def:applicationURL
+	 * content is used as link to application form. Otherwise link is automatically generated based on
+	 * required AUTHZ in template and registrar url set in /etc/perun/perun-registrar.properties.
+	 *
+	 * General errors result in exception being thrown, single failures are skipped and added to result
+	 *
+	 * @param invitationData csv file values separated by semicolon ';'. Only [email; name] or [email] is valid format.
+	 * @exampleParam ["mail@mail.cz", "mail2@mail.cz;user2"]
```
zlamalp, line 151: I think you need to include `invitationData` for the parser to know for which parameter is this example used.

perun-openapi/openapi.yml
```diff
@@ -15052,6 +15052,39 @@ paths:
...
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              title: InputInvitationsFromCsv
+              description: "input for sending invitations from csv"
+              type: object
+              required:
+                - vo
```
zlamalp, line 15080: This should be probably `voId` too.

zlamalp (changes requested):

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/RegistrarManagerMethod.java
Johaney-s: I see, added

cuadradek (approved):

PR #3794, opened 2022-12-07
feat(cli): script for fixing member's quotas

Johaney-s:
* when resource's defaultDataQuota is raised, member's dataQuota and dataQuotaOverride should be raised too, if their value is lower than the new defaultDataQuota
* this script compares hard quota part of member's quotas, compares it to resource's hard quota and if member's quota is lower, removes it
* virtual member's attribute fallbacks to resource's default quota if dataQuota or dataQuotaOverride is not set
* quotas are stored in a map where more quota names can appear, either all quotas can be checked or inputted quota only

### DIFF
--- a/perun-cli/fixMemberQuotas
+++ b/perun-cli/fixMemberQuotas
@@ -1,0 +1,129 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use Getopt::Long qw(:config no_ignore_case);
+use Perun::Agent;
+use Perun::Common qw(printMessage);
+
+sub help {
+	return qq{
+	Removes quota (dataQuotas, dataQuotasOverride) of those members assigned to resource whose hard quota is lower than resource's default hard quota (defaultDataQuotas).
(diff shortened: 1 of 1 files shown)


zlamalp (approved):
I think the script is ok, we just wait for the feedback from requestor.

xflord (approved):

perun-cli/fixMemberQuotas
```diff
@@ -0,0 +1,130 @@
...
+}
+
+my %updatedMembers = ();
+
+if (defined $memberId) {
+	resolveAttribute($memberId, $M_R_DATA_QUOTAS_OVERRIDE);
+	resolveAttribute($memberId, $M_R_DATA_QUOTAS);
+} else {
+	my @members = $resourcesAgent->getAssignedMembers( resource => $resourceId );
+	foreach my $member (@members) {
```
zlamalp, line 65: I suggest to sort members, so they are always processed in expected order. 
```perl
foreach my $member (sort { $a->getId <=> $b->getId } @members) {
```

PR #3795, opened 2022-12-08
feat(core): filter out outdated user IdP extsource's attributes

Johaney-s:
* when retrieveing attributes from user extsources, the extsources are filtered based on the last access
* if last access is older than allowed time period (defined in coreConfig - idpLoginValidity), its attributes are skipped
* exceptions can be defined in coreConfig - idpLoginValidityExceptions

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliationsTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliationsTest.java
@@ -8,7 +8,10 @@ import cz.metacentrum.perun.audit.events.AttributesManagerEvents.AttributeSetFor
 import cz.metacentrum.perun.audit.events.AuditEvent;
 import cz.metacentrum.perun.core.api.Attribute;
 import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.BeansUtils;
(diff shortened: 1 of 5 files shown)


Johaney-s:
https://github.com/CESNET/ansible-role-perun-docker-server/pull/55

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_eduPersonScopedAffiliationsTest.java
```diff
@@ -265,4 +271,44 @@ public void resolveAttributeValueChangeTest() throws Exception {
 		assertEquals(auditEvents.get(0).getClass(), AttributeChangedForUser.class);
 	}
 
+	@Test
+	public void lastAccessValid() throws Exception {
+		ues1.getExtSource().setType(ExtSourcesManager.EXTSOURCE_IDP);
+		ues1.setLastAccess(LocalDateTime.now().minusYears(10).format(Utils.lastAccessFormatter));
```
xflord, line 277: Nitpick - shouldn't we set this value based on the current config value of idpLoginValidity, just in case it changes in the future? Or change idpLoginValidity for the test and then return it to the original value; it doesn't matter.

xflord (commented):
Looks good

Johaney-s:
just discussed with @vyskocilpavel and realized i got the idea wrong - the exceptions are not for the UESes, but for the virtual attributes. will fix.

xflord (approved):

mattjoke (approved):

PR #3797, opened 2022-12-09
feat(core): new method copyMembers

xflord:
* copyMembers copies members from one group of the VO to other groups in the same VO
* added the method to all layers
* added to RPC and openapi
* fixed some javadocs


Johaney-s (commented):
looks good

sarkapalkovicova (approved):

Johaney-s (approved):

PR #3805, opened 2022-12-14
fix(cli): fixed warning displayed on VSUP instance

zlamalp:


zlamalp:
Generating service on VSUP instance shows:
```
Scalar value @segments[1] better written as $segments[1] at /opt/perun-cli/lib//Perun/Agent.pm line 509.
```
but it doesn't happen on other instances with the same (latest) release, so it might be perl version dependant.

Johaney-s (approved):

PR #3810, opened 2022-12-16
feat(core): add new login namespace admin-meta

cuadradek:
- New login namespace admin-meta was introduced. Attribute and password manager modules were created.


xflord (approved):

Johaney-s (approved):
good job

PR #3818, opened 2022-12-23
feat(core): new attribute module for optional login in MU namespace

xflord:
* optionalLogin-namespace:mu is now deprecated
* it was replaced with optional-login-namespace:mu which extends the new generic optional-login module
* some tweaks to the generic module were made as well

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_optional_login_namespace_mu.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_optional_login_namespace_mu.java
@@ -7,13 +7,8 @@ import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.ExtSourcesManager;
 import cz.metacentrum.perun.core.api.User;
 import cz.metacentrum.perun.core.api.UserExtSource;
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_optional_login_namespace_mu.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Contains login in the MU namespace if it is available, if not the value is null
+ *
+ * @author David Flor
+ */
+public class urn_perun_user_attribute_def_virt_optional_login_namespace_mu extends urn_perun_user_attribute_def_virt_optional_login_namespace {
```
zlamalp, line 22: We should probably add `@SkipValueCheckDuringDependencyCheck` class annotation, since it is used in the original attribute module for `user:virt:optionalLogin_namespace_mu`. Also it should be used in parent class too, since both classes do not implement checkAttributeSyntax/Semantics() methods and this annotations optimize the whole checking process by skipping them.

zlamalp (changes requested):

Johaney-s (approved):

mattjoke (approved):

PR #3820, opened 2022-12-25
feat(core): application view preferences

Johaney-s:
* new attributes for saving application view preferences (names of columns)
* attribute modules checking permitted values
* new method retrieving names of IdP attributes included in FedInfo

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -7786,6 +7786,36 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		policies.add(Triple.of(Role.GROUPADMIN, READ, RoleObject.Group));
 		attributes.put(attr, createInitialPolicyCollections(policies));
 
+		//urn:perun:group:attribute-def:def:applicationViewPreferences
+		attr = new AttributeDefinition();
+		attr.setNamespace(AttributesManager.NS_GROUP_ATTR_DEF);
+		attr.setType(ArrayList.class.getName());
+		attr.setFriendlyName("applicationViewPreferences");
+		attr.setDisplayName("Application view preferences");
+		attr.setDescription("Columns to be shown in application page. Use applications page configuration dialogue to set the value.");
+		//set attribute rights (with dummy id of attribute - not known yet)
+		policies = new ArrayList<>();
+		policies.add(Triple.of(Role.VOADMIN, READ, RoleObject.Vo));
+		policies.add(Triple.of(Role.VOADMIN, WRITE, RoleObject.Vo));
+		policies.add(Triple.of(Role.GROUPADMIN, READ, RoleObject.Group));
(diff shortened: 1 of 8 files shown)


xflord (commented):
Won't do a full review since I believe these changes are still subject to change, but here are some of my thoughts:
- Shouldn't we have a method or an attribute that returns the available columns? (instead of a private field in the new attribute). That way it would be more consistent (in the sense that we can get the same columns in FE and BE via this method/attribute)
- It feels rigid to have the same preferences for all the admins of the same Vo. My initial thought was to have it customizable for each person viewing the table (maybe a member-vo attribute, since user-vo attrs don't exist?). Then observers would also be able to change the table to fit their preference.
- I also took a look at the FE draft and believe that a dropdown checkbox list of the columns would be user-friendlier than a new dialog, however that's a matter of cosmetic preference.

I wanted to finish this during the holidays but I will leave it for after New Year's so we can discuss it further.

Johaney-s:
Thanks for the thoughts!
> Shouldn't we have a method or an attribute that returns the available columns? (instead of a private field in the new attribute). That way it would be more consistent (in the sense that we can get the same columns in FE and BE via this method/attribute)
- These columns are defined in FE anyway to be used as default options, but if you think it would make sense, why not.

> It feels rigid to have the same preferences for all the admins of the same Vo. My initial thought was to have it customizable for each person viewing the table (maybe a member-vo attribute, since user-vo attrs don't exist?). Then observers would also be able to change the table to fit their preference.
- I thought it would be cool to have the settings per manager too, but discussed it beforehand with PO and it won't be necessary. The columns just need to show some easy to understand info about the user, and which columns show that does not depend on who is viewing it. The main problem is with the hashes in the "created by" columns which are sometimes understandable and sometimes not.

> I also took a look at the FE draft and believe that a dropdown checkbox list of the columns would be user-friendlier than a new dialog, however that's a matter of cosmetic preference.
- Sure, why not, there should just be some clear confirmation button to tell the user he's changing the settings for all admins.

> I wanted to finish this during the holidays but I will leave it for after New Year's so we can discuss it further.
- i'll be on vacay till beginning of second january's week, and cause i didn't manage to finish it on my own, it would be great if someone could take over and finish it. either to continue with the proposed solution or redo it completely, i can also try to sum up the discussions with PO if necessary.

xflord:
Ok now that you made things clear it doesn't seem that complex. I will try to finish the FE part.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -7786,6 +7786,36 @@ protected void initialize() {
 		policies.add(Triple.of(Role.GROUPADMIN, READ, RoleObject.Group));
 		attributes.put(attr, createInitialPolicyCollections(policies));
 
+		//urn:perun:group:attribute-def:def:applicationViewPreferences
```
xflord, line 7789: Shouldn't observer roles have 'Read' privileges, since they also view applications?

xflord (approved):

mattjoke (approved):

PR #3821, opened 2023-01-02
fix(core): fixed a bug where the first day of grace period was not co…

xflord:
…unted as grace period when extending membership

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
@@ -1981,7 +1981,7 @@ public class MembersManagerBlImpl implements MembersManagerBl {
 				}
 
               // Check if we are in grace period
-              if (gracePeriodDate.isBefore(LocalDate.now())) {
+              if (gracePeriodDate.isEqual(LocalDate.now()) || gracePeriodDate.isBefore(LocalDate.now())) {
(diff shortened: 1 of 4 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -2398,7 +2398,7 @@ protected Pair<Boolean, Date> manageMembershipExpiration(PerunSession sess, Memb
 						fieldAmount = Utils.prepareGracePeriodDate(m);
 						LocalDate gracePeriodDate = localDate.minus(fieldAmount.getLeft(), fieldAmount.getRight());
 						// Check if we are in grace period
-						if (gracePeriodDate.isBefore(LocalDate.now())) {
+						if (gracePeriodDate.isEqual(LocalDate.now()) || gracePeriodDate.isBefore(LocalDate.now())) {
```
Johaney-s, line 2401: is there a reason to not add it also to line 1984? and I think the identical problem is also in group membership grace periods in GroupsManager methods `isMemberInGracePeriod` and `extendForStaticDate`

xflord: You're right; I did a sloppy job checking the use cases.

mattjoke (approved):

Johaney-s (approved):

PR #3844, opened 2023-01-11
fix(core): synchronization updates group name

Johaney-s:
* groups are matched by group login, so group name can and should be updated by synchronization


Kuliak (approved):

sarkapalkovicova (approved):
It looks good. Just consider if now, when the group's name is also changed, the java doc of `updateExistingGroupsWhileSynchronization` method shouldn't be updated.

PR #3845, opened 2023-01-12
fix(registrar): trim bulk invite csv values

xflord:
*  trim values of leading/trailing white spaces before using them to invite members


mattjoke (approved):
Looks good :+1:

Johaney-s (approved):

PR #3853, opened 2023-01-20
fix(core): fixed sending emails in batches to MSMTP

martin-kuba:
MSMTP implementation of SMTP daemon cannot receive more than a single message in one session. Sending batches of messages was reimplemented to close the SMTP connection after each message.


perun-notification/src/main/java/cz/metacentrum/perun/notif/managers/PerunNotifEmailManagerImpl.java
```diff
@@ -56,272 +39,99 @@ public void init() throws Exception {
 		System.setProperty("client.encoding", "utf-8");
 
 		// Load properties file notif.properties
```
zlamalp, line 41: `notif.sendMessages` is the last property, which is used from `perun-notif.properties`. I suggest to move it to the `perun.properties`. Then remove all usages of `perun-notif.properties` file. Then we can also remove this comment.

perun-notification/src/main/java/cz/metacentrum/perun/notif/managers/PerunNotifEmailManagerImpl.java
```diff
@@ -56,272 +39,99 @@ public void init() throws Exception {
...
+							mimeMessage.getSubject(),
+							content
+						);
+					} catch (AuthenticationFailedException authenticationFailedException) {
+						throw new EmailAuthenticationException(authenticationFailedException);
+					} catch (MessagingException messagingException) {
+						try {
+							ByteArrayOutputStream out = new ByteArrayOutputStream();
+							mimeMessage.writeTo(out);
+							failedEmailLogger.error(out.toString(StandardCharsets.UTF_8));
```
zlamalp, line 126: I'm not sure, why we write `mimeMessage` to the log file like this. When successfully sent, we simply log only relevant parts. But it was in the original implementation too, so it's probably ok.

I'm also not sure, if we need specific logger for failed messages, since there is no specific logging configuration for it (unlike for the send messages). In such case it uses default logger anyway. We should either add specific logger configuration (at least commented as example) or simply log this in the normal logger associated with this class).

zlamalp (changes requested):

perun-notification/src/main/java/cz/metacentrum/perun/notif/managers/PerunNotifEmailManagerImpl.java
zlamalp: We will resolve this in a new pull-request.

zlamalp (approved):

perun-notification/src/main/java/cz/metacentrum/perun/notif/managers/PerunNotifEmailManagerImpl.java
```diff
@@ -50,278 +30,38 @@ public class PerunNotifEmailManagerImpl implements PerunNotifEmailManager {
...
-				transport.close();
-			} catch (MessagingException ex) {
-				if (!failedMessages.isEmpty()) {
-					throw new EmailSendException("Failed to close server connection after message failures", ex, failedMessages);
-				} else {
-					throw new EmailSendException("Failed to close server connection after message sending", ex);
+					//send message over SMTP
+					mailSender.send(message);
+					// log successful sending
+					sendMessagesLogger.info("Email sent date: {}, to: {}, from: {}, subject: {}, text: {}",
```
zlamalp, line 54: Since we don't set `sentDate` anymore we shouldn't try to log it, as it is always null.

zlamalp (changes requested):
Can you please improve logging? We log following unnecessary messages:
```
DEBUG c.m.p.n.m.PerunNotifTemplateManager - Found handling sender: cz.metacentrum.perun.notif.senders.PerunNotifEmailUserSender@e942c6e
DEBUG c.m.p.n.m.PerunNotifTemplateManager - Messages send by sender: cz.metacentrum.perun.notif.senders.PerunNotifEmailUserSender@e942c6e
```
It would be probably sufficient to know, that we found one of the sender implementations (user or group), but that is also clear from the rest of messages in the log.

Also by removing those unnecessary classes we lost a few log messages about resolved/sent emails in the main log, so they are only in `sendMessages` log.
I suggest to log at least from/to/subject in the main log too, so that we can see what was resolved by what IDs (regexes/templates) - basically see full flow. Full details of each mail can be in the sendMessages log.

zlamalp (approved):

PR #3866, opened 2023-01-30
fix(registrar): notify admins if application auto-approval fails

sarkapalkovicova:
- send notification to the relevant vo/group admins
- store error that caused approval failure in db

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -1591,7 +1591,7 @@ public class RegistrarManagerImpl implements RegistrarManager {
 			}
 
 			// send mail
-			getMailManager().sendMessage(app, MailType.APP_REJECTED_USER, reason, null);
+			getMailManager().sendMessage(app, MailType.APP_REJECTED_USER, reason,null);
 
 			perun.getAuditer().log(sess, new ApplicationRejected(app));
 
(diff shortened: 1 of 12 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3837,14 +3835,10 @@ private void tryToAutoApproveApplication(PerunSession sess, Application app) thr
 
 			}
 		} catch (Exception ex) {
-
-			ArrayList<Exception> list = new ArrayList<>();
-			list.add(ex);
-			getMailManager().sendMessage(app, MailType.APP_ERROR_VO_ADMIN, null, list);
-
+			setAutoApproveErrorToApplication(app, ex.getMessage());
+			getMailManager().sendMessage(app, MailType.APP_ERROR_VO_ADMIN, ex.getMessage(), Arrays.asList(ex));
```
zlamalp, line 3839: I'm not sure, that passing `ex.getMessage()` as a `reason` param to the notification is useful, since it was originally used only in "rejection" notification, but I checked and it can be now generally used in all notifications (replaces tag `{customMessage}`. For error messages we used exception objects with tag `{errors}`.

So you can probably use it to generally improve the notification style/content, but make sure to think about it from the perspective of hints in GUI, documentation and mention it in the release notes for new version.

zlamalp (changes requested):
Since you changed the API object Application, please make sure to update also its definition in old GUI / WUI (esp. registrar) and CLI. You should also mark this as a breaking change.

zlamalp (approved):

Johaney-s (approved):
looks good, please rebase

PR #3867, opened 2023-01-31
feat(core): add options to sort paginated members by email, organizat…

Johaney-s:
…ion and statuses

* extended membersOrderColumn with email, organization and statuses, which will enable sorting in GUI by these parameters
* preferences (order) of email and organization are taken from current GUI logic (mixing user and member attributes)
* logins are not sortable as they are not unifiable

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersOrderColumn.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersOrderColumn.java
@@ -18,7 +18,39 @@ public enum MembersOrderColumn {
 	                 "users.first_name " + getLangSql(query) + query.getOrder().getSqlValue()
 	),
 
-	ID("", "", query -> "members.id " + query.getOrder().getSqlValue());
+	ID("", "", query -> "members.id " + query.getOrder().getSqlValue()),
+	STATUS("","", query -> "members.status " + query.getOrder().getSqlValue()),
+	GROUP_STATUS("", "", query -> "groups_members.source_group_status " + query.getOrder().getSqlValue()),
+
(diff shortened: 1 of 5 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersOrderColumn.java
```diff
@@ -18,7 +18,39 @@ public enum MembersOrderColumn {
...
-	ID("", "", query -> "members.id " + query.getOrder().getSqlValue());
+	ID("", "", query -> "members.id " + query.getOrder().getSqlValue()),
+	STATUS("","", query -> "members.status " + query.getOrder().getSqlValue()),
+	GROUP_STATUS("", "", query -> "source_group_status " + query.getOrder().getSqlValue()),
+
+	// 1. user preferred mail, 2. member mail
+	EMAIL(
+			", uav.attr_value, mav.attr_value ",
+			" left join " +
+			   	"(select attr_value, member_id, attr_id from member_attr_values) as mav " +
```
xflord, line 29: the aliases 'mav' and 'uav' are used later on in the WHERE string, could this be an issue (or at the very least be confusing) ?

xflord (commented):
looks like it should work, will test tomorrow

perun-base/src/main/java/cz/metacentrum/perun/core/api/MembersOrderColumn.java
Johaney-s: yeah, didn't realize they are used later on. renamed.

xflord (approved):

PR #3868, opened 2023-01-31
Get user roles by source

Kuliak:


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
```diff
@@ -1,33 +1,7 @@
...
-import cz.metacentrum.perun.core.api.User;
-import cz.metacentrum.perun.core.api.Vo;
-import cz.metacentrum.perun.core.api.exceptions.AlreadyAdminException;
-import cz.metacentrum.perun.core.api.exceptions.GroupNotAdminException;
-import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
-import cz.metacentrum.perun.core.api.exceptions.PolicyNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.RoleAlreadySetException;
-import cz.metacentrum.perun.core.api.exceptions.RoleManagementRulesNotExistsException;
-import cz.metacentrum.perun.core.api.exceptions.RoleNotSetException;
-import cz.metacentrum.perun.core.api.exceptions.UserNotAdminException;
```
Johaney-s, line 30: please fix your IDE configuration to not use wildcard imports

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
```diff
@@ -146,21 +115,27 @@ public AuthzResolverImpl(DataSource perunPool) {
...
+						"where authz.user_id=?",
+					AUTHZROLE_MAPPER, user.getId());
 
 				for (Pair<String, Map<String, Set<Integer>>> pair : authzRolesPairs) {
 					authzRoles.putAuthzRoles(pair.getLeft(), pair.getRight());
 				}
 
+				if (getAuthorizedGroupBasedRoles) {
+					// add user roles based on membership in authorized groups
+					authzRoles.putAll(getRolesObtainedFromAuthorizedGroupMemberships(user));
```
Johaney-s, line 136: I think this removes existing values for keys (roles) and puts new ids to them (so objects are not merged, but overwritten with group-gained roles).

Kuliak: Very nice! Previous tests were not built to test both direct and inherited roles presence. I updated them and fixed the problem with removing already existing roles. Thank you.

xflord (approved):
There are still some wildcard imports, otherwise well done

Johaney-s (commented):
besides the imports looks good now, but i'd say you can squash the commits - we don't really use "base" or "rpc" scopes...

edit: discussed on daily that backend might need adjustments in case the view should contain authoritative group ids... added don't merge label.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -402,7 +402,19 @@ perun_policies:
...
+    include_policies:
+      - default_policy
+
+  getRoleComplementaryObjectsWithAuthorizedGroups_int_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  getRolesObtainedFromAuthorizedGroupMemberships_int_policy:
```
Johaney-s, line 417: These policies would not be enough for the roles view in "myProfile" section - every user can view his roles there. I think "`SELF: User`" policy needs to be added to the new policies.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
```diff
@@ -87,6 +91,43 @@ public class AuthzResolverImpl implements AuthzResolverImplApi {
 
 	private static final Pattern columnNamesPattern = Pattern.compile("^[_0-9a-zA-Z]+$");
 
+
+	private static final RowMapper<Map<String, Map<Integer, Group>>> AUTHZ_COMPLEMENTARY_OBJECTS_AUTH_GROUPS_MAPPER = (rs, i) -> {
```
Johaney-s, line 95: we should be able to represent the use case, when user is member of group A and group B, which are both voadmins of VO - then you need a list of groups to be related with one voadmin role. now only one of them is returned.

perun-core/src/main/java/cz/metacentrum/perun/core/implApi/AuthzResolverImplApi.java
```diff
@@ -1,21 +1,7 @@
 package cz.metacentrum.perun.core.implApi;
 
-import cz.metacentrum.perun.core.api.Facility;
```
Johaney-s, line 3: still wildcard imports

perun-core/src/test/java/cz/metacentrum/perun/core/impl/AuthzResolverImplIntegrationTest.java
```diff
@@ -1,37 +1,19 @@
...
-import cz.metacentrum.perun.core.api.Role;
-import cz.metacentrum.perun.core.api.User;
-import cz.metacentrum.perun.core.api.UserExtSource;
-import cz.metacentrum.perun.core.api.Vo;
-import cz.metacentrum.perun.core.api.exceptions.AlreadyMemberException;
-import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
-import cz.metacentrum.perun.core.api.exceptions.RoleAlreadySetException;
-import cz.metacentrum.perun.core.api.exceptions.RoleNotSetException;
-import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
-import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
```
Johaney-s, line 20: wildcard imports

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -1,35 +1,19 @@
...
-import cz.metacentrum.perun.core.api.Member;
-import cz.metacentrum.perun.core.api.PerunBean;
-import cz.metacentrum.perun.core.api.PerunPolicy;
-import cz.metacentrum.perun.core.api.PerunPrincipal;
-import cz.metacentrum.perun.core.api.PerunSession;
-import cz.metacentrum.perun.core.api.Resource;
-import cz.metacentrum.perun.core.api.RichUser;
-import cz.metacentrum.perun.core.api.Role;
-import cz.metacentrum.perun.core.api.RoleManagementRules;
-import cz.metacentrum.perun.core.api.SecurityTeam;
```
Johaney-s, line 21: wildcard imports

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/AuthzResolverMethod.java
```diff
@@ -78,7 +62,47 @@ public List<String> call(ApiCaller ac, Deserializer parms ) throws PerunExceptio
...
+		public AuthzRoles call(ApiCaller ac, Deserializer parms) throws PerunException {
+			return cz.metacentrum.perun.core.api.AuthzResolver.getRolesObtainedFromAuthorizedGroupMemberships(ac.getSession(), parms.readInt("userId"));
+		}
+	},
+	/*#
+	 * Returns map of role complementary objects (perun beans) together with authorized group where user is member.
+	 *
+	 * @param userId int Id of a user
+	 * @return Map<String, Map<Integer, Group>> complementary objects with associated authorized groups
+	 * @exampleResponse {"Vo": { 28: 9645, 37: 9645 }}
```
Johaney-s, line 100: please update the example according to the returned type annotation

Johaney-s (changes requested):
please rebase it - db update

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
```diff
@@ -87,6 +91,43 @@ public class AuthzResolverImpl implements AuthzResolverImplApi {
...
+								rs.getInt("authz_authorized_group_id"),
+								rs.getString("groups_name"),
+								rs.getString("groups_description"),
+								rs.getInt("groups_vo_id")
+							));
+					}
+				}
+			}
+
+			return perunBeansWithAuthorizedGroups;
```
xflord, line 124: This code is quite difficult to read, so I might be wrong but I think you need to return a structure like this: 
Pair<RoleName, Map<AuthGroupId, Pair<BeanName, Set\<ObjectId>>>>
Otherwise you completely lose the information about which role is associated with which authorized group / object, no?

perun-core/src/test/java/cz/metacentrum/perun/core/impl/AuthzResolverImplIntegrationTest.java
```diff
@@ -78,24 +60,100 @@ public void getRolesOnlyValid() throws Exception {
...
+		// set role for group
+		AuthzResolver.setRole(sess, testGroup, createdVo, Role.VOADMIN);
+
+		// check that user has VO
+		AuthzRoles userRoles = authzResolverImpl.getRoles(testUser, false);
+		assertFalse(userRoles.hasRole(Role.VOADMIN, createdVo));
+	}
+
+	@Test
+	public void getRoleComplementaryObjectsWithAuthorizedGroupsTest() throws Exception {
```
xflord, line 123: Afterwards an edit to this test to cover multiple authorized groups / different roles would be nice

xflord (commented):
I might be misunderstanding the purpose of the  `getRoleComplementaryObjectsWithAuthorizedGroups` method so correct me if I'm wrong

perun-base/src/main/resources/perun-roles.yml
Johaney-s: yes, they should

perun-core/src/test/java/cz/metacentrum/perun/core/impl/AuthzResolverImplIntegrationTest.java
Kuliak: Yes, good suggestion. I have adjusted the test already. Thank you!

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
Kuliak: I admit, code is not the best one to read and yes, if you used only this method, you would lose information about which role is connected to which object.
However, when you get all roles for user, they contain an id of complementary object connected to it, so it can be used to obtain specific groups from map returned by `getRoleComplementaryObjectsWithAuthorizedGroups` method.

xflord: Right, but can't a user have multiple roles in the same complementary object through different authorized groups? How do you then tell which group provides which role?

xflord: I guess there could be a way to get this information through a combination of the methods, but that seems like extra logic for no reason.

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -892,24 +893,63 @@ public static List<String> getUserRoleNames(PerunSession sess, User user) throws
...
-	 * @return AuthzRoles object which contains all roles with perunbeans
 	 */
-	public static AuthzRoles getUserRoles(PerunSession sess, int userId) throws UserNotExistsException, PrivilegeException {
+	public static AuthzRoles getUserRoles(PerunSession sess, int userId, boolean getAuthorizedGroupBasedRoles) throws UserNotExistsException, PrivilegeException {
 		Utils.checkPerunSession(sess);
 		User user = ((PerunBl) sess.getPerun()).getUsersManagerBl().getUserById(sess, userId);
 
 		//Authorization
-		if (!authorizedInternal(sess, "getUserRoles_int_policy"))
+		if (!authorizedInternal(sess, "getUserRoles_int_boolean_policy"))
```
Johaney-s, line 911:
```suggestion
		if (!authorizedInternal(sess, "getUserRoles_int_boolean_policy", user))
```

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -892,24 +893,63 @@ public static List<String> getUserRoleNames(PerunSession sess, User user) throws
...
+	 * @param sess   perun session
+	 * @param userId id of a user
+	 * @return Map<String, Map<Integer, List<Group>>> roles with map of complementary objects with associated authorized groups
+	 */
+	public static Map<String, Map<Integer, List<Group>>> getRoleComplementaryObjectsWithAuthorizedGroups(PerunSession sess, int userId) throws UserNotExistsException, PrivilegeException {
+		Utils.checkPerunSession(sess);
+		User user = ((PerunBl) sess.getPerun()).getUsersManagerBl().getUserById(sess, userId);
+
+		//Authorization
+		if (!authorizedInternal(sess, "getRoleComplementaryObjectsWithAuthorizedGroups_int_policy"))
```
Johaney-s, line 930:
```suggestion
		if (!authorizedInternal(sess, "getRoleComplementaryObjectsWithAuthorizedGroups_int_policy", user))
```

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -892,24 +893,63 @@ public static List<String> getUserRoleNames(PerunSession sess, User user) throws
...
+	 * @param userId                       id of a user
+	 * @return AuthzRoles object which contains roles with perunbeans
+	 * @throws InternalErrorException
+	 */
+	public static AuthzRoles getRolesObtainedFromAuthorizedGroupMemberships(PerunSession sess, int userId) throws UserNotExistsException, PrivilegeException {
+		Utils.checkPerunSession(sess);
+		User user = ((PerunBl) sess.getPerun()).getUsersManagerBl().getUserById(sess, userId);
+
+		//Authorization
+		if (!authorizedInternal(sess, "getRolesObtainedFromAuthorizedGroupMemberships_int_policy"))
```
Johaney-s, line 949:
```suggestion
		if (!authorizedInternal(sess, "getRolesObtainedFromAuthorizedGroupMemberships_int_policy", user))
```

xflord (approved):
good job

Johaney-s (approved):

PR #3870, opened 2023-02-01
feat(core): added HTML escaping class

mattjoke:
* HTMLParser utilizes OWASP's HTML sanitization
* Allowed tags and attributes are the same as in GUI

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/InvalidHtmlInputException.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/InvalidHtmlInputException.java
@@ -1,0 +1,13 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+import cz.metacentrum.perun.core.impl.HTMLParser;
+
+public class InvalidHtmlInputException extends WrongAttributeValueException{
+	public InvalidHtmlInputException(String message, String[] unsafeTags) {
+		super(message + " " + HTMLParser.getMessage(unsafeTags));
+	}
+
+	public InvalidHtmlInputException(String message, String rejectReason) {
+		super(message + " " + rejectReason);
+	}
+}
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/HTMLParser.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/HTMLParser.java
@@ -1,0 +1,221 @@
+package cz.metacentrum.perun.core.impl;
+
+import org.owasp.html.CssSchema;
+import org.owasp.html.HtmlChangeListener;
+import org.owasp.html.HtmlPolicyBuilder;
+import org.owasp.html.PolicyFactory;
+
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+/*
+ * Class for sanitizing HTML input using OWASP Java HTML Sanitizer.
+ * @author: Matej Hakoš <492968@mail.muni.cz>
+ */
+public class HTMLParser {
+	private static final String[] allowedTags = {"a", "article", "aside", "b", "blockquote", "br", "button", "caption", "center", "cite", "decorator", "del", "details", "div", "em", "footer", "h1", "h2", "h3", "h4", "h5", "h6", "header", "i", "img", "kbd", "label", "li", "ol", "p", "pre", "section", "select", "span", "strong", "sup", "table", "tbody", "td", "textarea", "tfoot", "th", "thead", "tr", "ul"};
+	private static final String[] allowedAttributes = {"align", "class", "color", "disabled", "height", "hidden", "href", "id", "label", "size", "span", "src", "srcset", "style", "width"};
+	private static final String[] allowedStyles = {"color", "background-color", "font-size", "font-family", "text-align", "margin", "padding", "border", "width", "height", "display", "position", "top", "bottom", "left", "right", "overflow", "float", "clear", "z-index"};
+	private static PolicyFactory policy = null;
+
+	private List<String> escapedTags = new ArrayList<>();
+	private List<String> escapedAttributes = new ArrayList<>();
+	private List<String> escapedStyles = new ArrayList<>();
+
+	private String rawHTML = "";
+	private String escapedHTML = "";
+	private String[] escapedStrings = new String[]{"", "", ""};
+	private boolean isInputValid = true;
+
+	public HTMLParser() {
+		policy = generatePolicy();
+	}
+
+	/**
+	 * Creates policy defined by the allowedTags, allowedAttributes and allowedStyles.
+	 * Used to sanitize HTML input.
+	 *
+	 * @return policy - PolicyFactory object
+	 */
+	private static PolicyFactory generatePolicy() {
+		HtmlPolicyBuilder p = new HtmlPolicyBuilder();
+		p.disallowUrlProtocols();
+		for (String tag : allowedTags) {
+			p.allowElements(tag);
+		}
+		for (String attr : allowedAttributes) {
+			p.allowAttributes(attr).globally();
+		}
+		CssSchema cssWhitelist = CssSchema.withProperties(Arrays.stream(allowedStyles).toList());
+		p.allowStyling(cssWhitelist);
+		return p.toFactory();
+	}
+
+	/**
+	 * Computes the difference between all style attributes in the escaped and unescaped input.
+	 *
+	 * @param input   - unescaped input
+	 * @param escaped - escaped input
+	 */
+	private void computeEscapedStyles(String input, String escaped) {
+		Pattern pattern = Pattern.compile("style=\"(.*?)\"");
+		Matcher matcher = pattern.matcher(input);
+		while (matcher.find()) {
+			String style = matcher.group(1);
+			String[] styles = style.split(";");
+			for (String s : styles) {
+				String[] split = s.split(":");
+				if (split.length == 2) {
+					escapedStyles.add(split[0].trim());
+				}
+			}
+		}
+		matcher = pattern.matcher(escaped);
+		while (matcher.find()) {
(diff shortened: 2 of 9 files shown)


mattjoke:
I am not sure where I should put the HTMLParser class. Please check if the placement of the checks in appropriate.

xflord (commented):
I would probably place the HTMLParser class under perun/core/impl, since it's not really module related. Also a test would be nice to have (probably in the Registrar test classes)

perun-core/src/main/java/cz/metacentrum/perun/core/impl/Utils.java
```diff
@@ -156,6 +167,12 @@ public class Utils {
...
 	private static final String A_UES_ORGANIZATION = AttributesManager.NS_UES_ATTR_DEF + ":o";
 
+	private static final String[] allowedTags = {"a", "article", "aside", "b", "blockquote", "br", "button", "caption", "center", "cite", "decorator", "del", "details", "div", "em", "footer", "h1", "h2", "h3", "h4", "h5", "h6", "header", "i", "img", "kbd", "label", "li", "ol", "p", "pre", "section", "select", "span", "strong", "sup", "table", "tbody", "td", "textarea", "tfoot", "th", "thead", "tr", "ul"};
+	private static final String[] allowedAttributes = {"align", "class", "color", "disabled", "height", "hidden", "href", "id", "label", "size", "span", "src", "srcset", "style", "width"};
+	private static final String[] allowedStyles = {"color", "background-color", "font-size", "font-family", "text-align", "margin", "padding", "border", "width", "height", "display", "position", "top", "bottom", "left", "right", "overflow", "float", "clear", "z-index"};
+	private static final PolicyFactory policy = generatePolicy();
+	private static final List<List<String>> escapedTagsAndAttributes = new ArrayList<>();
```
xflord, line 174: Since you're keeping the data in this field, I believe it would be better practice to keep the parsing logic in a separate class (as you had before).

perun-core/src/main/java/cz/metacentrum/perun/core/bl/AttributesManagerBl.java
```diff
@@ -1635,7 +1636,7 @@ void setRequiredAttributes(
 	 * @throws WrongAttributeAssignmentException if attribute is not vo attribute or if it is core attribute
 	 * @throws WrongReferenceAttributeValueException
 	 */
-	void setAttribute(PerunSession sess, Vo vo, Attribute attribute) throws WrongAttributeValueException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException;
+	void setAttribute(PerunSession sess, Vo vo, Attribute attribute) throws WrongAttributeValueException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException, InvalidHtmlInputException;
```
xflord, line 1639: declare the exception in the entry level as well

xflord (changes requested):
Sanitizing for mail notifications is missing here

perun-core/src/test/java/cz/metacentrum/perun/core/impl/HTMLParserTest.java
```diff
@@ -0,0 +1,51 @@
...
+	public void checkSafeHtmlInput() {
+		System.out.println("HTMLParserTest.checkSafeHtmlInput");
+
+		String safeInput = """
+			<p class="helo">HELLO !</p>
+			<p style="background-color:blue; color:red;">Hello 2!</p>""";
+
+		String[] unsafeTags = HTMLParser.checkEscapedHTML(safeInput, safeInput);
+		
+		System.out.println("Unsafe tags: " + unsafeTags.length);
```
xflord, line 44: Unnesaccary prints

perun-core/src/test/java/cz/metacentrum/perun/core/impl/HTMLParserTest.java
```diff
@@ -0,0 +1,51 @@
...
+		for (String tag : unsafeTags) {
+			System.out.println(tag);
+		}
+
+		assert(unsafeTags.length == 0);
```
xflord, line 49: Please use junit asserts (assertEquals/assertThat). They provide more information

perun-core/src/test/java/cz/metacentrum/perun/core/impl/HTMLParserTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+	@Test
+	public void checkSafeHtmlInput() {
+		System.out.println("HTMLParserTest.checkSafeHtmlInput");
+
+		String safeInput = """
+			<p class="helo">HELLO !</p>
+			<p style="background-color:blue; color:red;">Hello 2!</p>""";
+
+		String[] unsafeTags = HTMLParser.checkEscapedHTML(safeInput, safeInput);
```
xflord, line 42: Shouldn't you call sanitize first before checking the escaped tags?

xflord (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -1910,13 +1912,22 @@ public boolean setAttributeWithoutCheck(PerunSession sess, Facility facility, At
...
 	@Override
-	public void setAttribute(PerunSession sess, Vo vo, Attribute attribute) throws WrongAttributeValueException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException {
+	public void setAttribute(PerunSession sess, Vo vo, Attribute attribute) throws WrongAttributeValueException, WrongAttributeAssignmentException, WrongReferenceAttributeValueException, InvalidHtmlInputException {
 		convertEmptyAttrValueToNull(attribute);
 		if (attribute.getValue() == null) {
 			removeAttribute(sess, vo, attribute);
 			return;
 		}
 		checkAttributeSyntax(sess, vo, attribute);
+		if (attribute.getFriendlyName().equals("htmlMailFooter")){
```
Johaney-s, line 1922: this is not the right place to check attribute value - this is usually handled in separate attribute module (checkAttributeSyntax method of VoAttributesModuleImplApi inheriting classes)

perun-core/src/main/java/cz/metacentrum/perun/core/api/exceptions/InvalidHtmlInputException.java
```diff
@@ -0,0 +1,7 @@
+package cz.metacentrum.perun.core.api.exceptions;
+
+public class InvalidHtmlInputException extends PerunException{
```
Johaney-s, line 3: this exception could extend `WrongAttributeValueException` because the meaning of the exception is identical - the attribute value is not allowed. then we wouldn't need to add separately InvalidHtmlInputException to thrown method's exceptions.

mattjoke: Sounds like a good idea. 👍

perun-core/src/main/java/cz/metacentrum/perun/core/impl/HTMLParser.java
```diff
@@ -0,0 +1,162 @@
...
+ * Class for sanitizing HTML input using OWASP Java HTML Sanitizer.
+ * @author: Matej Hakoš <492968@mail.muni.cz>
+ */
+public class HTMLParser {
+	private static final String[] allowedTags = {"a", "article", "aside", "b", "blockquote", "br", "button", "caption", "center", "cite", "decorator", "del", "details", "div", "em", "footer", "h1", "h2", "h3", "h4", "h5", "h6", "header", "i", "img", "kbd", "label", "li", "ol", "p", "pre", "section", "select", "span", "strong", "sup", "table", "tbody", "td", "textarea", "tfoot", "th", "thead", "tr", "ul"};
+	private static final String[] allowedAttributes = {"align", "class", "color", "disabled", "height", "hidden", "href", "id", "label", "size", "span", "src", "srcset", "style", "width"};
+	private static final String[] allowedStyles = {"color", "background-color", "font-size", "font-family", "text-align", "margin", "padding", "border", "width", "height", "display", "position", "top", "bottom", "left", "right", "overflow", "float", "clear", "z-index"};
+	private static PolicyFactory policy = null;
+
+	private static List<String> escapedTags = new ArrayList<>();
```
Johaney-s, line 26: i think this might be problematic in case several threads call sanitize method - they all would change the same list and there would be inconsistency in the results. @xflord has a nice idea to make this class builder-like, where every sanitization would create separate class instance. we can discuss it further

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -325,9 +348,29 @@ public void updateMailById(PerunSession sess, ApplicationMail mail) throws FormN
...
+			if (inputSubject != null && inputText != null) {
+				// Check if html text contains invalid tags (subject and text)
+				HTMLParser parser = new HTMLParser()
+					.sanitizeHTML(inputSubject)
+					.checkEscapedHTML();
+				if (!parser.isInputValid()) {
+					throw new InvalidHtmlInputException("HTML Subject contains unsafe HTML tags or styles. Remove them and try again.", parser.getEscaped());
+				}
+				htmlMessage.setSubject(parser.getEscapedHTML());
+
```
xflord, line 365: You need to include the second sanitization call here (which is why the tests are failing). Also, I believe it would be better practice to create a new HTMLParser instance for EACH string.

xflord (changes requested):
I feel like you somehow overwrote the check for Application form items

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
mattjoke: You are right it is a good practice to do new instance per string

Johaney-s (approved):
neat!

xflord (approved):
well done

PR #3873, opened 2023-02-06
fix(core): unique constraint to vos shortName

sarkapalkovicova:
Dropped unique constraint to name and added a constraint to shortName instead.

### DIFF
--- a/perun-base/src/test/resources/test-schema.sql
+++ b/perun-base/src/test/resources/test-schema.sql
@@ -1,4 +1,4 @@
--- database version 3.2.10 (don't forget to update insert statement at the end of file)
+-- database version 3.2.11 (don't forget to update insert statement at the end of file)
 CREATE EXTENSION IF NOT EXISTS "unaccent";
 CREATE EXTENSION IF NOT EXISTS "pgcrypto";
 
@@ -14,7 +14,7 @@ create table vos (
 					  created_by_uid integer,
 					  modified_by_uid integer,
 					  constraint vo_pk primary key (id),
-					  constraint vo_u unique (name)
+					  constraint vo_u unique (short_name)
 );
 
 -- USERS - information about user as real person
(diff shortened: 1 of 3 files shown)


perun-core/src/main/resources/postgresChangelog.txt
```diff
@@ -6,6 +6,10 @@
 -- Directly under version number should be version commands. They will be executed in the order they are written here.
 -- Comments are prefixed with -- and can be written only between version blocks, that means not in the lines with commands. They have to be at the start of the line.
 
+3.2.11
+ALTER TABLE vos DROP constraint vo_u unique (name);
```
HejdaJakub, line 10: I think that if you drop the constraint you should just define a name without the rest of the definition part.
```suggestion
ALTER TABLE vos DROP constraint vo_u;
```

perun-base/src/test/resources/test-schema.sql
```diff
@@ -1881,7 +1881,7 @@ create index idx_fk_attr_critops ON attribute_critical_actions(attr_id);
 create index app_state_idx ON application (state);
 
 -- set initial Perun DB version
-insert into configurations values ('DATABASE VERSION','3.2.10');
+insert into configurations values ('DATABASE VERSION','3.2.11');
```
HejdaJakub, line 1884: Update also comment on the first line of this file, please.

perun-db/postgres.sql
```diff
@@ -1988,7 +1988,7 @@ grant all on attribute_critical_actions to perun;
 grant all on app_notifications_sent to perun;
 
 -- set initial Perun DB version
-insert into configurations values ('DATABASE VERSION','3.2.10');
+insert into configurations values ('DATABASE VERSION','3.2.11');
```
HejdaJakub, line 1991: The same. Update also comment on the first line of this file, please.

HejdaJakub (changes requested):

HejdaJakub (approved):

xflord (commented):
Mark the commit as breaking change, as we're updating DB (you can also add this to release notes for deployment)

xflord (approved):

PR #3883, opened 2023-02-16
feat(core): consents for resource

sarkapalkovicova:
Added methods to obtain consents for members assigned to the chosen resource.

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -7934,6 +7934,16 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getConsentHubByResource_Resource_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+      - FACILITYADMIN: Facility
+      - FACILITYOBSERVER: Facility
+      - VOADMIN: Vo
+      - VOOBSERVER: Vo
+    include_policies:
+      - default_policy
+
   filter-getConsentHub_policy:
     policy_roles:
       - PERUNOBSERVER:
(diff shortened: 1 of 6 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/ConsentsManagerEntryIntegrationTest.java
```diff
@@ -197,6 +197,21 @@ public void getConsentsForUser() throws Exception {
 		assertEquals(0, consentsManagerEntry.getConsentsForUser(sess, user.getId(), ConsentStatus.GRANTED).size());
 	}
 
+	@Test
+	public void getConsentsForConsentHubByResource() throws Exception {
```
xflord, line 201: I would rather have this test create a resource under a facility, assign some members to it, create consents for them and then assert that the method only returns the consents of these assigned members.

perun-core/src/main/java/cz/metacentrum/perun/core/api/ConsentsManager.java
```diff
@@ -49,6 +50,19 @@ public interface ConsentsManager {
 	 */
 	List<Consent> getConsentsForConsentHub(PerunSession sess, int id) throws PrivilegeException, ConsentHubNotExistsException;
 
+	/**
+	 * Get all consents for ConsentHub corresponding to the facility of the chosen resource.
```
xflord, line 54: also edit the docs. (we want to return only the consents of the members of that resource, not all consents for the facility)

xflord (approved):
well done

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -7934,6 +7934,16 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getConsentHubByResource_Resource_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+      - FACILITYADMIN: Facility
+      - FACILITYOBSERVER: Facility
+      - VOADMIN: Vo
+      - VOOBSERVER: Vo
```
Kuliak, line 7943: Shouldn't we allow also RESOURCEADMIN and RESOURCEOBSERVER to use this method?

Kuliak (commented):
Good job! I have only one suggestion to check.

mattjoke (approved):
:+1:

Kuliak (approved):

PR #3886, opened 2023-02-21
feat(core): remove redundant (deprecated) code

Kuliak:
* openAPI updated to exclude deprecated methods and related objects
* RPC, Entry, Business layers updated
* policies concerning deprecated methods removed from perun-roles file
* perun-core XML file updated
* ServicesAgent.pm file updated
* tests for deprecated methods removed

breaking change: getDataWithVos, getDataWithGroups, getFlatData, getHierarchicalData methods cannot be used anymore


perun-openapi/openapi.yml
```diff
@@ -2456,13 +2456,6 @@ components:
             items:
               $ref: "#/components/schemas/MemberCandidate"
 
-    ServiceAttributesResponse:
```
Johaney-s, line 2459: `ServiceAttributes` were only used in this response type, we can remove it too

xflord (approved):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ServicesManagerEntryIntegrationTest.java
```diff
@@ -1295,709 +1292,6 @@ public void getDestinationsWhenFacilityNotExists() throws Exception {
...
-		//group has 3 members but member2 and member3 are expired
-		assertEquals(groupMembers.size(), 1);
-
-		//group2 has 2 members but member3 is expired
-		assertEquals(group2Members.size(), 1);
-
-		//resource has 3 members but member3 is expired in all groups of given resource
-		assertEquals(members.size(), 2);
-	}
-
```
zlamalp, line 2000: It's not strictly related to this pull-request, but I think we don't have much tests for the new methods. For example to test usage of expired members within the service. We might want to add them later.

zlamalp (changes requested):
I think we can also remove whole `ServiceAttributes` object from the core/cli or is there any reason why not?

perun-core/src/test/java/cz/metacentrum/perun/core/entry/ServicesManagerEntryIntegrationTest.java
Kuliak: I think this might be discussed in the future and tasks to create these tests could be part of a sprint.

perun-openapi/openapi.yml
Johaney-s: still here

zlamalp (changes requested):
Please remove also `ServiceAttributes` object definition from the CLI and openapi.yml. Otherwise its ok.

zlamalp (approved):

PR #3892, opened 2023-02-28
fix(core): HTMLParser NullPointerException

mattjoke:
* added null check for computeEscapedStyles
* Fix: when creating new form item, the HTML check was not run

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -1112,6 +1112,23 @@ public class RegistrarManagerImpl implements RegistrarManager {
 			// is item to create ? => create
 			if (item.getId() <= 0 && !item.isForDelete()) {
 				int temporaryId = item.getId();
+				// Check if newly created item contains allowed tags (in case of Header and HTML Comment
+				if (item.getType() == HTML_COMMENT || item.getType() == HEADING) {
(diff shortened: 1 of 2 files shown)


HejdaJakub (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1112,6 +1112,23 @@ public int updateFormItems(PerunSession sess, ApplicationForm form, List<Applica
 			// is item to create ? => create
 			if (item.getId() <= 0 && !item.isForDelete()) {
 				int temporaryId = item.getId();
+				// Check if newly created item contains allowed tags (in case of Header and HTML Comment
+				if (item.getType() == HTML_COMMENT || item.getType() == HEADING) {
```
Johaney-s, line 1116: we could extract the repetitive logic from creating and updating to separate method to avoid code duplication, what do you think?

mattjoke: The difference between these two blocks is that this one directly replaces the i18n of the item, whereas the following function does not. The main reason for it is there is a partial update in the following function (i.e. only ItemTexts is directly updated).

Johaney-s: Aha, I see. thanks for clarification

Johaney-s (approved):

PR #3894, opened 2023-03-01
feat(core): un/block login per namespace or globally

HejdaJakub:
* Now is possible to block login for given namespace or also block login globally (for all namespaces on instance).
* It is not possible to block already blcoked login or login which is already used. It is also not possible to ublock NOT blocked login.
* We can block login for some namespace(s) and also block it globally - we don't want to unblock them automatically during the new (different) blocking for the same login.

BREAKING CHANGE: update DB

### DIFF
--- a/perun-base/src/main/resources/perun-roles.yml
+++ b/perun-base/src/main/resources/perun-roles.yml
@@ -6207,6 +6207,40 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getAllBlockedLoginsInNamespaces_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  isLoginBlocked_String_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  isLoginBlockedGlobally_String_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  isLoginBlockedForNamespace_String_String_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  blockLogins_List<String>_String_policy:
+    policy_roles: []
+    include_policies:
+      - default_policy
+
+  unblockLogins_List<String>_String_policy:
+    policy_roles: []
+    include_policies:
+      - default_policy
+
   getUsersWithoutVoAssigned_policy:
     policy_roles:
       - PERUNOBSERVER:
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
@@ -37,6 +37,8 @@ import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.LoginIsAlreadyBlockedException;
+import cz.metacentrum.perun.core.api.exceptions.LoginIsNotBlockedException;
 
 import java.util.List;
 import java.util.Map;
@@ -795,6 +797,62 @@ public interface UsersManager {
 	 */
 	boolean isLoginAvailable(PerunSession sess, String loginNamespace, String login) throws InvalidLoginException;
 
+	/**
+	 * Returns all pairs of blocked login in namespace (if namespace is null, then this login is blocked globally)
+	 *
+	 * @param sess
+	 * @return list of pairs login and namespace - List<Pair<login, namespace>>
+	 */
+	List<Pair<String, String>> getAllBlockedLoginsInNamespaces(PerunSession sess) throws PrivilegeException;
+
+	/**
+	 * Return true if login is blocked (globally - for all namespaces per instance OR for some namespace), false if not
+	 *
+	 * @param sess
+	 * @param login login to check
+	 * @return true if login is blocked
+	 */
+	boolean isLoginBlocked(PerunSession sess, String login) throws PrivilegeException;
+
+	/**
+	 * Return true if login is blocked globally (for all namespaces per instance - represented by namespace = null), false if not
+	 *
+	 * @param sess
+	 * @param login login to check
+	 * @return true if login is blocked globally
+	 */
+	boolean isLoginBlockedGlobally(PerunSession sess, String login) throws PrivilegeException;
+
+	/**
+	 * Return true if login is blocked for given namespace, false if not
+	 * When the namespace is null, then the method behaves like isLoginBlockedGlobally(), so it checks if the login is blocked globally
+	 *
+	 * @param sess
+	 * @param login login to check
+	 * @param namespace namespace for login
+	 * @return true if login is blocked for given namespace (or globally for null namespace)
+	 */
+	boolean isLoginBlockedForNamespace(PerunSession sess, String login, String namespace) throws PrivilegeException;
+
+	/**
+	 * Block logins for given namespace or block logins globally (if no namespace is selected)
+	 *
+	 * @param sess
+	 * @param logins list of logins to be blocked
+	 * @param namespace namespace where the logins should be blocked (null means block the logins globally)
+	 * @throws LoginIsAlreadyBlockedException
+	 * @throws LoginExistsException
+	 */
+	void blockLogins(PerunSession sess, List<String> logins, String namespace) throws PrivilegeException, LoginIsAlreadyBlockedException, LoginExistsException;
+
+	/**
+	 * Unblock logins for given namespace or unblock logins globally (if no namespace is selected)
+	 * @param sess
+	 * @param logins list of logins to be unblocked
+	 * @param namespace namespace where the logins should be unblocked (null means unblock the logins globally)
+	 * @throws LoginIsNotBlockedException
+	 */
+	void unblockLogins(PerunSession sess, List<String> logins, String namespace) throws PrivilegeException, LoginIsNotBlockedException;
 
 	/**
 	 * Returns all users who have set the attribute with the value. Searching only def and opt attributes.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
@@ -51,6 +51,9 @@ import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
 import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.LoginIsAlreadyBlockedException;
+import cz.metacentrum.perun.core.api.exceptions.LoginExistsException;
+import cz.metacentrum.perun.core.api.exceptions.LoginIsNotBlockedException;
 import cz.metacentrum.perun.core.impl.PerunSessionImpl;
 import cz.metacentrum.perun.core.implApi.modules.pwdmgr.PasswordManagerModule;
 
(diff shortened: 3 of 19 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
```diff
@@ -1213,6 +1216,58 @@ void deletePassword(PerunSession sess, User user, String loginNamespace)
...
+	 * Return true if login is blocked globally (for all namespaces per instance - represented by namespace = null), false if not
+	 *
+	 * @param sess
+	 * @param login login to check
+	 */
+	boolean isLoginBlockedGlobally(PerunSession sess, String login);
+
+	/**
+	 * Return true if login is blocked for given namespace, false if not
+	 *
```
Kuliak, line 1244: Even though it is clear when we look on other methods, I would specify that globally blocked logins are not taken into account in this method (just to be sure).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/UsersManagerBlImpl.java
```diff
@@ -1149,6 +1152,41 @@ public boolean isLoginAvailable(PerunSession sess, String loginNamespace, String
...
+
+	@Override
+	public boolean isLoginBlockedForNamespace(PerunSession sess, String login, String namespace) {
+		return getUsersManagerImpl().isLoginBlockedForNamespace(sess, login, namespace);
+	}
+
+	@Override
+	public void blockLogin(PerunSession sess, String login, String namespace) throws LoginIsAlreadyBlockedException, LoginExistsException {
+		if (getPerunBl().getAttributesManagerBl().isLoginAlreadyUsed(sess, login, namespace)) {
+			throw new LoginExistsException("Login: "+login+" is already in use.");
```
Kuliak, line 1179:
```suggestion
			throw new LoginExistsException("Login: " + login + " is already in use.");
```

perun-core/src/main/java/cz/metacentrum/perun/core/api/UsersManager.java
```diff
@@ -795,6 +797,57 @@ UserExtSource getUserExtSourceByExtLogin(PerunSession perunSession, ExtSource so
 	 */
 	boolean isLoginAvailable(PerunSession sess, String loginNamespace, String login) throws InvalidLoginException;
 
+	/**
+	 * Returns all pairs of blocked login in namespace (if namespace is null, then this login is blocked globally)
+	 *
+	 * @param sess
+	 */
```
Kuliak, line 804: I know it's a technicality, but we could add @return tag into javadoc as well. 

Also, closer specification of returned object in the method description might be useful for future use, for this one let's say something like: List<Pair< LOGIN, NAMESPACE >>

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -701,6 +706,131 @@ public void isAlreadyReservedLogin() throws Exception {
...
+	@Test
+	public void getAllBlockedLoginsInNamespaces() throws Exception {
+		System.out.println(CLASS_NAME + "getAllBlockedLoginsInNamespaces");
+
+		perun.getUsersManager().blockLogin(sess, "login", null);
+		perun.getUsersManager().blockLogin(sess, "login", "namespace");
+		perun.getUsersManager().blockLogin(sess, "login2", "namespace2");
+
+		List<Pair<String, String>> listOfBlockedLogins = perun.getUsersManager().getAllBlockedLoginsInNamespaces(sess);
+		assertEquals(listOfBlockedLogins.size(), 3);
```
Kuliak, line 718: I think we might want to check that obtained blocked login-namespace pairs really are those above with correct login value and namespace.

Kuliak (changes requested):
Good job, I have few suggestions.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -1248,6 +1248,25 @@ public List<Attribute> getAttributes(PerunSession sess, User user, List<String>
...
+				return jdbc.queryForInt("select count(*) from attr_names as attr join user_attr_values attr_val on attr.id=attr_val.attr_id\n" +
+					"where attr.friendly_name like 'login-namespace:%' \n" +
+					"    and attr.friendly_name not like '%persistent%' \n" +
+					"    and attr_val.attr_value like ?", login) > 0;
+			} else {
+				return jdbc.queryForInt(String.format("select count(*) from attr_names as attr join user_attr_values attr_val on attr.id=attr_val.attr_id\n" +
+					"where attr.friendly_name like 'login-namespace:%s' \n" +
+					"    and attr.friendly_name not like '%%persistent%%' \n" +
+					"    and attr_val.attr_value like ?", namespace), login) > 0;
+			}
```
bodnara, line 1264: Could we just set the namespace to empty string if it is null and keep the else clause ? Since right now both queries are almost identical.

HejdaJakub: Good point, I just need to replace it with `%` instead of an empty string if the namespace is NULL, but yes, now it is more readable, thanks.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1146,6 +1157,83 @@ public boolean isLoginReserved(PerunSession sess, String namespace, String login
...
+		}
+	}
+
+	@Override
+	public void unblockLogin(PerunSession sess, String login, String namespace) throws LoginIsNotBlockedException {
+		Utils.notNull(login, "userLogin");
+
+		try {
+			if (namespace == null) {
+				int numAffected = jdbc.update("delete from blocked_logins where login=? and namespace is null", login);
```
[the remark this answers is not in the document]
HejdaJakub: I am not sure how to clearly and safely replace this branching with ternary syntax. The `is null` is not interchangeable with something like `=null` or by an empty string, am I right? And `log.info(...)` methods have a different structure, so for one of them, it doesn't make sense to put an empty string or something like that instead of the null namespace.

So how do you mean the replacement with the ternary operator? Can you give me an example, please? :face_with_head_bandage:

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1146,6 +1157,99 @@ public boolean isLoginReserved(PerunSession sess, String namespace, String login
...
```
xflord, line 1236:
```suggestion
				int numAffected = jdbc.update("delete from blocked_logins where login=? and namespace " + namespace == null ? "is null" : "=?", login);
```
The logging makes it bit more awkward now that you mention it, but then you could just leave the log calls in the if-else, which would still be less duplicate code.
It's not big of a deal, go with what seems better readable to you

HejdaJakub:
Methods for un/blocking logins should be implemented as bulk methods.

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6207,6 +6207,40 @@ perun_policies:
     include_policies:
       - default_policy
 
+  getAllBlockedLoginsInNamespaces_policy:
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  isLoginBlocked_policy:
```
Kuliak, line 6216: I think in the name of conventions, we should add type of method parameters into policy names.
```suggestion
  isLoginBlocked_String_policy:
```

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6207,6 +6207,40 @@ perun_policies:
...
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  isLoginBlockedGlobally_policy:
```
Kuliak, line 6222:
```suggestion
  isLoginBlockedGlobally_String_policy:
```

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6207,6 +6207,40 @@ perun_policies:
...
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  isLoginBlockedForNamespace_policy:
```
Kuliak, line 6228:
```suggestion
  isLoginBlockedForNamespace_String_String_policy:
```

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6207,6 +6207,40 @@ perun_policies:
...
+    policy_roles:
+      - PERUNOBSERVER:
+    include_policies:
+      - default_policy
+
+  blockLogins_policy:
```
Kuliak, line 6234:
```suggestion
  blockLogins_List<String>_String_policy:
```

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -6207,6 +6207,40 @@ perun_policies:
...
+    policy_roles: []
+    include_policies:
+      - default_policy
+
+  unblockLogins_policy:
```
Kuliak, line 6239:
```suggestion
  unblockLogins_List<String>_String_policy:
```

perun-openapi/openapi.yml
```diff
@@ -9922,6 +9922,36 @@ paths:
...
+          $ref: '#/components/responses/ExceptionResponse'
+
+  /urlinjsonout/usersManager/unblockLogins:
+    post:
+      tags:
+        - UsersManager
+      operationId: unblockLogins
+      summary: Unblock logins for given namespace or unblock logins globally (if no namespace is selected)
+      parameters:
+        - { name: "login[]", schema: { type: array, items: { type: string } }, in: query, required: false, description: "list of logins to be blocked" }
```
xflord, line 9947:
```suggestion
        - { name: "logins[]", schema: { type: array, items: { type: string } }, in: query, required: false, description: "list of logins to be blocked" }
```

perun-openapi/openapi.yml
```diff
@@ -9922,6 +9922,36 @@ paths:
...
           $ref: '#/components/responses/ExceptionResponse'
 
+  /urlinjsonout/usersManager/blockLogins:
+    post:
+      tags:
+        - UsersManager
+      operationId: blockLogins
+      summary: Block logins for given namespace or block logins globally (if no namespace is selected)
+      parameters:
+        - { name: "login[]", schema: { type: array, items: { type: string } }, in: query, required: false, description: "list of logins to be blocked" }
```
xflord, line 9932:
```suggestion
        - { name: "logins[]", schema: { type: array, items: { type: string } }, in: query, required: false, description: "list of logins to be blocked" }
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -972,6 +975,73 @@ public boolean isLoginAvailable(PerunSession sess, String loginNamespace, String
...
+		}
+
+		return getUsersManagerBl().getAllBlockedLoginsInNamespaces(sess);
+	}
+
+	@Override
+	public boolean isLoginBlocked(PerunSession sess, String login) throws PrivilegeException {
+		Utils.checkPerunSession(sess);
+
+		if (!AuthzResolver.authorizedInternal(sess, "isLoginBlocked_policy")) {
```
Kuliak, line 993: Change to match added parameter types in policy names.
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "isLoginBlocked_String_policy")) {
```

perun-openapi/openapi.yml
xflord: to match parameter name in usersManagerMethod

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -972,6 +975,73 @@ public boolean isLoginAvailable(PerunSession sess, String loginNamespace, String
...
+		}
+
+		return getUsersManagerBl().isLoginBlocked(sess, login);
+	}
+
+	@Override
+	public boolean isLoginBlockedGlobally(PerunSession sess, String login) throws PrivilegeException {
+		Utils.checkPerunSession(sess);
+
+		if (!AuthzResolver.authorizedInternal(sess, "isLoginBlockedGlobally_policy")) {
```
Kuliak, line 1004: Adjust policy name.
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "isLoginBlockedGlobally_String_policy")) {
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -972,6 +975,73 @@ public boolean isLoginAvailable(PerunSession sess, String loginNamespace, String
...
+
+		return getUsersManagerBl().isLoginBlockedGlobally(sess, login);
+
+	}
+
+	@Override
+	public boolean isLoginBlockedForNamespace(PerunSession sess, String login, String namespace) throws PrivilegeException {
+		Utils.checkPerunSession(sess);
+
+		if (!AuthzResolver.authorizedInternal(sess, "isLoginBlockedForNamespace_policy")) {
```
Kuliak, line 1016: Policy name.
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "isLoginBlockedForNamespace_String_String_policy")) {
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -972,6 +975,73 @@ public boolean isLoginAvailable(PerunSession sess, String loginNamespace, String
...
+		}
+
+		return getUsersManagerBl().isLoginBlockedForNamespace(sess, login, namespace);
+	}
+
+	@Override
+	public void blockLogins(PerunSession sess, List<String> logins, String namespace) throws PrivilegeException, LoginIsAlreadyBlockedException, LoginExistsException {
+		Utils.checkPerunSession(sess);
+
+		if (!AuthzResolver.authorizedInternal(sess, "blockLogins_policy")) {
```
Kuliak, line 1027: Policy name.
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "blockLogins_List<String>_String_policy")) {
```

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -972,6 +975,73 @@ public boolean isLoginAvailable(PerunSession sess, String loginNamespace, String
...
+		}
+
+		getUsersManagerBl().blockLogins(sess, logins, namespace);
+	}
+
+	@Override
+	public void unblockLogins(PerunSession sess, List<String> logins, String namespace) throws PrivilegeException, LoginIsNotBlockedException {
+		Utils.checkPerunSession(sess);
+
+		if (!AuthzResolver.authorizedInternal(sess, "unblockLogins_policy")) {
```
Kuliak, line 1038: Policy name.
```suggestion
		if (!AuthzResolver.authorizedInternal(sess, "unblockLogins_List<String>_String_policy")) {
```

Kuliak (changes requested):

xflord (approved):

Kuliak (approved):

PR #3895, opened 2023-03-07
feat: sql pooling

xflord:
rebase and minor fixes of https://github.com/CESNET/perun/pull/2678


xflord:
Tested on devel, works

HejdaJakub (approved):
From my side, it looks good. Although I think that I am not able to see all the possible consequences, I tried to read it in detail and I didn't find any problems in this code.

PR #3897, opened 2023-03-09
feat(core): added copySponsoredMembers() method

xflord:
* copySponsoredMembers() creates new sponsorships with user A as sponsor for members sponsored by user B
* added the new method to openapi
* wrote test

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
@@ -68,6 +68,7 @@ import cz.metacentrum.perun.core.api.SponsoredUserData;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.time.LocalDate;
 import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -573,6 +573,39 @@ public Void call(ApiCaller ac, Deserializer params) throws PerunException {
...
+
+			boolean copyValidity = false;
+			if (params.contains("copyValidity")) {
+				copyValidity = params.readBoolean("copyValidity");
+			}
+			LocalDate validityTo = null;
+			if (params.contains("validityTo")) {
+				validityTo = params.readLocalDate("validityTo");
+			}
+
```
Johaney-s, line 600: a nitpick - i think we could check that `copyValidity` and `validityTo` are not both set at once and have separate javadocs in this file for them

perun-core/src/main/java/cz/metacentrum/perun/core/entry/MembersManagerEntry.java
```diff
@@ -1459,6 +1460,22 @@ public void sponsorMembers(PerunSession session, List<Member> sponsored, User sp
 		}
 	}
 
+	@Override
+	public void copySponsoredMembers(PerunSession session, List<MemberWithSponsors> sponsored, User copyFrom, User copyTo, boolean copyValidity, LocalDate validityTo) throws SponsorshipDoesNotExistException, UserNotInRoleException, AlreadySponsorException, PrivilegeException, MemberNotSponsoredException {
+		for (MemberWithSponsors member : sponsored) {
```
Johaney-s, line 1465: if i'm not mistaken, it seems MemberWithSponsors object is not retrieved from backend, but relies on the object provided in the request - we should then check if that information is valid, same for `validityTo`. It would be also more user-friendly for API callers to specify just the member id instead of whole MemberWithSponsor object.

Johaney-s (approved):

mattjoke (approved):

PR #3898, opened 2023-03-10
feat(core): use blocked logins logic

Kuliak:
* when creating new login (login attribute), value is checked if not blocked (namespace, globally, default)
* already existing methods for blocked logins adjusted to work with ignoreCase flag
* global logins are ALWAYS case-insensitive
* logins blocked by default (internal components) are ALWAYS case-sensitive.
* tests created/adjusted

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
@@ -5,6 +5,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
@@ -836,5 +837,23 @@ public class CoreConfig {
 
 	public void setIdpLoginValidityExceptions(List<String> idpLoginValidityExceptions) {
 		this.idpLoginValidityExceptions = idpLoginValidityExceptions;
+	}
+
+	/**
+	 * Get all logins blocked by default (used by internal components).
+	 *
+	 * @return set of logins used by instance
+	 */
+	public Set<String> getBlockedLogins() {
+		Set<String> blockedLogins = new HashSet<>();
+
+		blockedLogins.addAll(admins);
+		blockedLogins.addAll(enginePrincipals);
+		blockedLogins.addAll(notificationPrincipals);
+		blockedLogins.addAll(registrarPrincipals);
+		blockedLogins.addAll(dontLookupUsers);
+		blockedLogins.add(rpcPrincipal);
+
+		return blockedLogins;
 	}
 }
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
@@ -1217,6 +1217,18 @@ public interface UsersManagerBl {
 	void checkReservedLogins(PerunSession sess, String namespace, String login, boolean ignoreCase) throws AlreadyReservedLoginException;
 
 	/**
+	 * Check if login is blocked. Login can be blocked by default (used by internal components),
+	 * globally or in namespace.
+	 *
+	 * @param sess       session
+	 * @param namespace  attribute
+	 * @param userLogin  login
+	 * @param ignoreCase ignore case (work as case-insensitive)
+	 * @throws LoginIsAlreadyBlockedException when login is blocked
+	 */
+	void checkBlockedLogins(PerunSession sess, String namespace, String userLogin, boolean ignoreCase) throws LoginIsAlreadyBlockedException;
+
+	/**
 	 * Returns all pairs of blocked login in namespace (if namespace is null, then this login is blocked globally)
 	 *
 	 * @param sess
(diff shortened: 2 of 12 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
```diff
@@ -837,4 +837,22 @@ public List<String> getIdpLoginValidityExceptions() {
...
 		this.idpLoginValidityExceptions = idpLoginValidityExceptions;
 	}
+
+	/**
+	 * Get all logins blocked by default (used by internal components).
+	 *
+	 * @return list of logins used by instance
+	 */
+	public List<String> getBlockedLogins() {
+		List<String> blockedLogins = new ArrayList<>();
```
Johaney-s, line 847:
```suggestion
		Map<String> blockedLogins = new HashMap<>();
```

Johaney-s: my bad, i meant set, no map :D

perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
```diff
@@ -1216,6 +1216,17 @@ void deletePassword(PerunSession sess, User user, String loginNamespace)
...
+	/**
+	 * Check if login is blocked. Login can be blocked by default (used by internal components),
+	 * globally or in namespace.
+	 *
+	 * @param sess session
+	 * @param userLogin login
+	 * @param attribute attribute
+	 * @throws LoginIsAlreadyBlockedException when login is blocked
+	 */
+	void checkBlockedLogins(PerunSessionImpl sess, String userLogin, Attribute attribute) throws LoginIsAlreadyBlockedException;
```
Johaney-s, line 1228: one more thing - it would make sense to me pass login and namespace parameters directly instead of login and attribute containing the namespace, so it is also usable when we'd just like to check login without namespace etc.

also - i see reserved logins can ignore case of the login - maybe we'd have a use case for that also here? should discuss it...

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1167,36 +1167,49 @@ public List<Pair<String, String>> getAllBlockedLoginsInNamespaces(PerunSession s
...
 
 	@Override
-	public boolean isLoginBlockedGlobally(PerunSession sess, String login) {
+	public boolean isLoginBlockedGlobally(PerunSession sess, String login, boolean ignoreCase) {
 		Utils.notNull(login, "userLogin");
 
 		try {
-			return jdbc.queryForInt("select count(1) from blocked_logins where login=? and namespace is null", login) == 1;
+			if (ignoreCase) {
+				System.out.println("ignoring case");
```
xflord, line 1190: leftover from testing

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -127,7 +127,7 @@ public class UsersManagerImpl implements UsersManagerImplApi {
 	}
 
 	private JdbcPerunTemplate jdbc;
-	private NamedParameterJdbcTemplate  namedParameterJdbcTemplate;
```
Johaney-s, line 130: lots of changed indentation here, could you please undo those changes?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1167,11 +1173,15 @@ public List<Pair<String, String>> getAllBlockedLoginsInNamespaces(PerunSession s
 	}
 
 	@Override
-	public boolean isLoginBlocked(PerunSession sess, String login) {
+	public boolean isLoginBlocked(PerunSession sess, String login, boolean ignoreCase) {
 		Utils.notNull(login, "userLogin");
 
 		try {
-			return jdbc.queryForInt("select count(1) from blocked_logins where login=?", login) == 1;
+			if (ignoreCase) {
```
Johaney-s, line 1180: i'd say let's be consistent about handling the globally blocked logins - apply the ignoreCase only to the namespaced logins and always match globally blocked logins case insensitively. would be nice to mention it in the javadocs too

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1166,42 +1166,50 @@ public List<Pair<String, String>> getAllBlockedLoginsInNamespaces(PerunSession s
...
-	}
+    @Override
+    public boolean isLoginBlocked(PerunSession sess, String login, boolean ignoreCase) {
+        Utils.notNull(login, "userLogin");
+
+        try {
+            if (ignoreCase) {
+                return jdbc.queryForInt("select count(1) from blocked_logins where UPPER(login)=?", login.toUpperCase()) == 1;
+            } else {
+                return jdbc.queryForInt("select count(1) from blocked_logins where login=? OR (UPPER(login)=? AND namespace IS NULL)", login, login) == 1;
```
Johaney-s, line 1177:
```suggestion
                return jdbc.queryForInt("select count(1) from blocked_logins where login=? OR (UPPER(login)=? AND namespace IS NULL)", login, login.toUpperCase()) == 1;
```

Johaney-s (approved):

xflord (approved):

PR #3900, opened 2023-03-13
feat(core): blocked logins methods

sarkapalkovicova:
- Added method for obtaining all existing namespaces in which login can be blocked.
- Added method for unblocking logins by id without dependency on namespace.


mattjoke (approved):
Looks good 👍

PR #3902, opened 2023-03-14
test(core): getHashedData tests added

Kuliak:


perun-core/src/test/java/cz/metacentrum/perun/core/entry/ServicesManagerEntryIntegrationTest.java
```diff
@@ -1344,7 +1345,7 @@ public void addDestinationSameDestinationDifferentAdmin() throws Exception {
 	}
 
 	@Test
-	public void testGetHashedDataWithGroups() throws Exception {
+	public void getHashedDataWithGroups() throws Exception {
 		System.out.println(CLASS_NAME + "testGetHashedDataWithGroups");
```
Johaney-s, line 1349:
```suggestion
		System.out.println(CLASS_NAME + "getHashedDataWithGroups");
```

Johaney-s (commented):
i'm not a fan of verifying the structure before setting the service to skip expired members, as it seems to me to duplicate the already existing basic tests. what do you think, @Kuliak, @mattjoke ?

mattjoke:
> i'm not a fan of verifying the structure before setting the service to skip expired members, as it seems to me to duplicate the already existing basic tests. what do you think, @Kuliak, @mattjoke ?

Yes, personally I would skip the verification before as it is redundant.

Johaney-s (approved):

mattjoke (approved):
Looks good 👍

PR #3907, opened 2023-03-20
fix(core): add SPONSOR role to privileged_roles_to_read

HejdaJakub:
* SPONSOR now has its own role between roles that are privileged to read this role for given Vo.


HejdaJakub:
Related to https://github.com/CESNET/perun-web-apps/pull/1282

xflord (approved):

PR #3908, opened 2023-03-21
feat(core): interconnect the deletion of user with blocking user's logins

HejdaJakub:
* The current functionality for user deletion was extended by the new functionality of blocking user's logins.
* There is also new config property that switches between methods deleteUser()/anonymizeUser() and just only one of them will be allowed for the instance. The behavior of the new GUI will adjust to this property as well.
* Database was updated - column related_user_id was added to the table blocked_logins.

BREAKING CHANGE: According to the new config property [userDeletionForced] for the instance will be disabled to execute the deleteUser() method OR anonymizeUser() method. Just one of them will be executable according to the config. Also the database was updated.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
@@ -1279,10 +1279,11 @@ public interface UsersManagerBl {
 	 * @param sess
 	 * @param logins list of logins to be blocked
 	 * @param namespace namespace where the logins should be blocked (null means block the logins globally)
+	 * @param relatedUserId id of the user related to the login or null if the relatedUserId should not be stored
 	 * @throws LoginIsAlreadyBlockedException
 	 * @throws LoginExistsException
 	 */
-	void blockLogins(PerunSession sess, List<String> logins, String namespace) throws LoginIsAlreadyBlockedException, LoginExistsException;
+	void blockLogins(PerunSession sess, List<String> logins, String namespace, Integer relatedUserId) throws LoginIsAlreadyBlockedException, LoginExistsException;
 
 	/**
 	 * Get page of blocked logins.
@@ -1319,6 +1320,16 @@ public interface UsersManagerBl {
 	 * @return id of login blocked in specified namespace
 	 */
 	int getIdOfBlockedLogin(PerunSession sess, String login, String namespace);
+
+	/**
+	 * Get user id of the user who was related to the given login in the past
+	 *
+	 * @param sess session
+	 * @param login blocked login
+	 * @param namespace namespace where the login is blocked
+	 * @return user id or null if there is no related user id
+	 */
+	Integer getRelatedUserIdByBlockedLoginInNamespace(PerunSession sess, String login, String namespace) throws LoginIsNotBlockedException;
 
 	void checkUserExists(PerunSession sess, User user) throws UserNotExistsException;
 
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
@@ -1244,13 +1244,13 @@ public class UsersManagerImpl implements UsersManagerImplApi {
     }
 
 	@Override
-	public void blockLogin(PerunSession sess, String login, String namespace) throws LoginIsAlreadyBlockedException {
+	public void blockLogin(PerunSession sess, String login, String namespace, Integer relatedUserId) throws LoginIsAlreadyBlockedException {
 		Utils.notNull(login, "userLogin");
 
 		try {
 			int newId = Utils.getNewId(jdbc, "blocked_logins_id_seq");
-			jdbc.update("insert into blocked_logins(id, login, namespace) values (?,?,?)",
-				newId, login, namespace);
+			jdbc.update("insert into blocked_logins(id, login, namespace, related_user_id) values (?,?,?,?)",
+				newId, login, namespace, relatedUserId);
 			if (namespace == null) {
 				log.info("Login {} globally blocked", login);
 			} else {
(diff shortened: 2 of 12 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -394,6 +394,9 @@ public void updateUserWithNullValueInLastName() throws Exception {
 	public void deleteUser() throws Exception {
 		System.out.println(CLASS_NAME + "deleteUser");
 
+		// Enable deletion of users
+		BeansUtils.getCoreConfig().setUserDeletionForced(true);
```
xflord, line 398: please use try-finally to return CoreConfig properties to their original values (see the 'anonymizeUser()' test for reference)

xflord (changes requested):

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
HejdaJakub: Tests rewrote to set the `userDeletionForced` property back to the original value.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -390,22 +390,127 @@ public void updateUserWithNullValueInLastName() throws Exception {
...
+		boolean originalUserDeletionForced = BeansUtils.getCoreConfig().getUserDeletionForced();
+		BeansUtils.getCoreConfig().setUserDeletionForced(true);
+
+		try {
+			usersManager.deleteUser(sess, user, true);  // force delete
+			usersManager.getUserById(sess, user.getId());
+			// should NOT be able to get deleted user by his id - should throw UserNotExistsException
+			fail();
+		} catch (UserNotExistsException ex) {
+			// this is expected
```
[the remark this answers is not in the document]
xflord: After further thought, I believe having 'BeansUtils.getCoreConfig().setUserDeletionForced(true)' (and all the code that follows) inside the try block (like it is in ConsentsManager tests) is safer because it avoids missing other exceptions (which would result in the property not being set back to the original value). Please rewrite 'anonymizeUserWhenAnonymizationNotSupported()' (and the other tests you added ofc) as well.

HejdaJakub:
As the login is blocked during the process of deleting some user, the userId of the previously related user was added to the table.  Since this is a relatively edge case, I didn't add the new property `relatedUserId` to the object of blockedLogin neither to the paginated methods. `RelatedUserId` should be used primarily as a log for further debugging if needed.

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
```diff
@@ -393,32 +392,155 @@ public void updateUserWithNullValueInLastName() throws Exception {
...
+		} finally {
+			// set userDeletionForced back to the original value
+			BeansUtils.getCoreConfig().setUserDeletionForced(originalUserDeletionForced);
+		}
+
+		// check that user id is correctly stored with the block login
+		assertEquals(user.getId(), usersManager.getRelatedUserIdByBlockedLoginInNamespace(sess, "login2", "namespace2").intValue());
+		assertThatExceptionOfType(LoginIsNotBlockedException.class).isThrownBy(
+			() -> usersManager.getRelatedUserIdByBlockedLoginInNamespace(sess, "login3", "namespace2"));
+	}
```
bodnara, line 522: Not sure what is our convention but right now the test tests for two different things ? That login was correctly blocked with user id and that if some login is not blocked exception is thrown. If you see this as a problem please create a separate test for one of the cases.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
```diff
@@ -1279,10 +1279,11 @@ void deletePassword(PerunSession sess, User user, String loginNamespace)
 	 * @param sess
 	 * @param logins list of logins to be blocked
 	 * @param namespace namespace where the logins should be blocked (null means block the logins globally)
+	 * @param relatedUserId id of the user related to the login
```
xflord, line 1282: please mention that this can be null

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1368,6 +1368,23 @@ public int getIdOfBlockedLogin(PerunSession sess, String login, String namespace
...
+	@Override
+	public Integer getRelatedUserIdByBlockedLoginInNamespace(PerunSession sess, String login, String namespace) throws LoginIsNotBlockedException {
+		try {
+			int userId;
+			if (namespace == null) {
+				userId =  jdbc.queryForInt("SELECT related_user_id FROM blocked_logins WHERE login=? AND namespace IS NULL", login);
+			} else {
+				userId = jdbc.queryForInt("SELECT related_user_id FROM blocked_logins WHERE login=? AND namespace=?", login, namespace);
+			}
+			return userId == 0 ? null : userId;
```
xflord, line 1380: is this check necessary?

xflord (commented):
Also please add info about db update into commit message

perun-core/src/test/java/cz/metacentrum/perun/core/entry/UsersManagerEntryIntegrationTest.java
HejdaJakub: We discussed it with @xflord and we agreed that there it is probably not necessary - these asserts are both related to the same method and this method even is not in openapi, so I consider it just as a side assert.

perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
HejdaJakub: Added.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
xflord: This was a nitpick to begin with, but the behavior you're looking for is the 'queryForObject()' method from jdbc. 'queryForInt' was implemented to supplement a jdbc method that has been deprecated (and all it does is add the conversion from null to 0, that you revert here :D ) 

Screenshot from 2023-04-03 10-13-53

xflord (approved):

PR #3911, opened 2023-03-22
feat(core): new mfaStatus Attribute

mattjoke:
* value is dependent on the value of the mfaEnforceSettings:namespace attribute
* ENFORCED_ALL, ENFORCED_PARITALLY or <empty string>

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -8253,6 +8253,18 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 			policies.add(Triple.of(Role.FACILITYADMIN, READ, RoleObject.Facility));
 			attributes.put(attr, createInitialPolicyCollections(policies));
 
+			// mfaStatus namespace
+			attr = new AttributeDefinition();
+			attr.setNamespace(AttributesManager.NS_USER_ATTR_VIRT);
+			attr.setType(String.class.getName());
+			attr.setFriendlyName("mfaStatus:"+namespace);
+			attr.setDisplayName("MFA status in namespace: "+namespace);
+			attr.setDescription("Contains user friendly version of MFA status in " + namespace + " namespace.");
+
+			policies = new ArrayList<>();
+			policies.add(Triple.of(Role.PERUNADMIN, READ, RoleObject.User));
+			attributes.put(attr, createInitialPolicyCollections(policies));
+
 			// pwd-reset templates
 
 			attr = new AttributeDefinition();
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
@@ -1,0 +1,47 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.AttributeDefinition;
+import cz.metacentrum.perun.core.api.AttributesManager;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.impl.PerunSessionImpl;
+import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleAbstract;
+import org.slf4j.LoggerFactory;
+import org.slf4j.Logger;
+
+/**
+ * Attribute module for mfaStatus attribute.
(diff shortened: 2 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * @author Matej Hakoš <492968@mail.muni.cz>
+ */
+public class urn_perun_user_attribute_def_virt_mfaStatus extends UserVirtualAttributesModuleAbstract {
+	private static Logger log = LoggerFactory.getLogger(urn_perun_user_attribute_def_virt_mfaStatus.class);
+
+	@Override
+	public Attribute getAttributeValue(PerunSessionImpl perunSession, User user, AttributeDefinition attribute) {
+		Attribute attr = new Attribute(attribute);
+		String namespace = attribute.getFriendlyName().split(":")[1];
```
xflord, line 26:
```suggestion
		String namespace = attr.getFriendlyNameParameter()
```

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
```diff
@@ -0,0 +1,56 @@
...
+		String mfaEnforceSettings = AttributesManager.NS_USER_ATTR_DEF + ":mfaEnforceSettings:" + namespace;
```
xflord, line 27: We usually define the names of other necessary attributes as static fields

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
```diff
@@ -0,0 +1,56 @@
...
+		try {
+			Attribute mfaSettings = perunSession.getPerunBl().getAttributesManagerBl().getAttribute(perunSession, user, mfaEnforceSettings);
+			if (mfaSettings.getValue() == null) {
```
xflord, line 30:
```suggestion
			if (mfaSettings.getValue() == null || mfaSettings.getValue() == "") {
```
Pretty sure the value can also be an empty string, no?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
```diff
@@ -0,0 +1,56 @@
...
+				attr.setValue("");
+			} else if (mfaSettings.getValue() == "{\"all\":true}") {
+				attr.setValue("ENFORCED_ALL");
+			} else {
+				attr.setValue("ENFORCED_PARTIALLY");
+			}
+		} catch (AttributeNotExistsException e) {
+			// Log non-existing attribute
+			log.warn("Attribute {} does not exist.", namespace);
```
xflord, line 39: I don't think we need to log this, but if you decide to log, please include whole attribute name

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
```diff
@@ -0,0 +1,56 @@
...
+		} catch (WrongAttributeAssignmentException e) {
+			// If the attribute does not exist, mfaStatus is empty string
+		}
+		return attr;
+	}
+
+	@Override
+	public AttributeDefinition getAttributeDefinition() {
```
xflord, line 47: When I think about it, this method is redundant since the attribute will not be used without a namespace

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_mfaStatus.java
mattjoke: I'll leave the logging, but I'll pass the correct values to the log.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -8253,6 +8253,18 @@ protected void initialize() {
 			policies.add(Triple.of(Role.FACILITYADMIN, READ, RoleObject.Facility));
 			attributes.put(attr, createInitialPolicyCollections(policies));
 
+			// mfaStatus namespace
+			attr = new AttributeDefinition();
+			attr.setNamespace(AttributesManager.NS_USER_ATTR_VIRT);
+			attr.setType(String.class.getName());
+			attr.setFriendlyName("mfaStatus:"+namespace);
+			attr.setDisplayName("MFA status in namespace: "+namespace);
+			attr.setDescription("Contains the MFA status in " + namespace + " namespace (more human readable string).");
```
bodnara, line 8262: Just a suggestion.
```suggestion
			attr.setDescription("Contains user friendly version of MFA status in " + namespace + " namespace.");
```

xflord (approved):

Kuliak (approved):

PR #3915, opened 2023-03-29
fix(core): blocked logins search string

sarkapalkovicova:
Updated SQL query to search blocked logins by substring and case-insensitive.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/UsersManagerImpl.java
```diff
@@ -1305,7 +1305,7 @@ public Paginated<BlockedLogin> getBlockedLoginsPage(PerunSession sess, BlockedLo
 			if (hasSearchString || hasNamespaces) {
 				sqlQuery += " WHERE";
 				if (hasSearchString) {
-					sqlQuery += " login like (:login)";
+					sqlQuery += " login like ('%%' || :login || '%%')";
```
bodnara, line 1308: Shouldn't we also make the database content lowercase or is it not necessary ?
```suggestion
					sqlQuery += " LOWER(login) like ('%%' || :login || '%%')";
```

sarkapalkovicova: At first, I thought it too. I tested it, and it seems to be already case-insensitive as it is. But I should confess that I have no idea how.  :smile:

bodnara (approved):

PR #3921, opened 2023-04-05
fix(core): disallow to block already reserved logins

HejdaJakub:
* Added the check to blockLogins method for already reserved logins. If the namespace is given, check reserved login for this namespace. If the login should be blocked globally, check reserved logins for all namespaces.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/bl/UsersManagerBl.java
@@ -1211,10 +1211,10 @@ public interface UsersManagerBl {
 
 
 	/**
-	 * Check if login in specified namespace exists.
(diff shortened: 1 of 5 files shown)


sarkapalkovicova (commented):
Maybe it would be useful to add a test to `UsersManagerEntryIntegrationTest` where you try to block reserved login but otherwise, it looks good.

HejdaJakub:
I am not sure if there is some easy way how to test it, because the process of reservation logins is related to Registrar, which means creating an application form, submitting an application, etc. and I am not sure how to set up all these things. :disappointed_relieved: - if there is some way, then is also a question, if it should be in `UsersManagerEntryIntegrationTest` or in `RegistrarBaseIntegrationTest` :eyes:

sarkapalkovicova (approved):

bodnara (approved):

PR #3926, opened 2023-04-11
fix: allow correct usage of getAllBlockedLoginsInNamespaces in openapi

HejdaJakub:
* This method used to return list of pairs (left: loginValue, right: namespaceValue), but now it returns list of BlockedLogin.
* All interface and test were fixed for this return type and rpc and openapi specification for this method was added.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
@@ -979,7 +979,7 @@ public class UsersManagerEntry implements UsersManager {
 	}
 
 	@Override
-	public List<Pair<String, String>> getAllBlockedLoginsInNamespaces(PerunSession sess) throws PrivilegeException {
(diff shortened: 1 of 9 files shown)


sarkapalkovicova (approved):

bodnara (approved):

HejdaJakub:
According to the conversation with @xflord, the response type of the method was converted from `List<Map<String,String>>` to `List<BlockedLogin>`.

sarkapalkovicova (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -82,6 +82,7 @@
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
```
xflord, line 85: Unused import

xflord (commented):
GJ, please merge after removing the import

PR #3927, opened 2023-04-11
feat(core): LSAAI changePassword

mattjoke:
* added changePassword to LifescienceidusernamePasswordManagerModule
* changePassword is not constrained to only Registrar
* on th end super.changePassword is called


xflord (approved):

bodnara (commented):
Would it make sense to extract the shared logic for validation and password change so we don't have it duplicated ?

xflord (approved):

mattjoke:
Successfully tested on an instance, and the user is successfully able to change the password on an LSAAI instance.

bodnara:
Personally I would prefer for the `addUserToVo` method to return `true` when successful, but feel free to disagree. In any way, at least add a comment about the expected behavior of the method.

PR #3928, opened 2023-04-11
feat: 🎸 Allow embedded group from item in group applications

dBucik:
Use embedded group form items on the group application form as well. A check for the configured subgroups being subgroups in the context of VO or Group on whose application form is used has been added. Also, this change introduces the possibility of configuring more than one embedded group's application form item on a single application form.

**DISCLAIMER**: has been tested locally but without running the registrar instance to verify that things are drawn properly and work properly.
Storing and retrieving items has been tested and seems to work OK.


dBucik:
Discovered conceptual issue. The item is now re-useable on a single form, however, a link between the configured groups and the form is missing. Need to re-think how to establish such a link in the most breaking-change-free way.

dBucik:
This should be it. I have reworked the whole concept to tie the configured groups to a particular application form item. This allows having multiple instances of this form item on a single form while keeping the relation of the configured groups.

PR #3932, opened 2023-04-14
feat(core): ServicesManager bulk methods

sarkapalkovicova:
Added methods for forcing propagation, blocking and unblocking multiple services at once.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/ServicesManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/ServicesManagerEntryIntegrationTest.java
@@ -34,8 +34,10 @@ import cz.metacentrum.perun.core.api.exceptions.FacilityNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.RelationExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ServiceAlreadyBannedException;
 import cz.metacentrum.perun.core.api.exceptions.ServiceAttributesCannotExtend;
 import cz.metacentrum.perun.core.api.exceptions.ServiceExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ServiceIsNotBannedException;
 import cz.metacentrum.perun.core.api.exceptions.ServiceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ServicesPackageExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ServicesPackageNotExistsException;
@@ -54,6 +56,7 @@ import static org.assertj.core.api.AssertionsForClassTypes.assertThatExceptionOf
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThrows;
 import static org.junit.Assert.assertTrue;
 
 /**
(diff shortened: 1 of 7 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/entry/ServicesManagerEntryIntegrationTest.java
```diff
@@ -1187,6 +1190,59 @@ public void blockAndUnblockServiceOnDestinations() throws Exception {
 		assertFalse("Service should NOT be blocked on the richDestination3", perun.getServicesManager().isServiceBlockedOnDestination(sess, service2, destination3.getId()));
 	}
 
+	@Test
+	public void blockAndUnblockServicesOnFacility() throws Exception {
+		System.out.println(CLASS_NAME + "blockAndUnblockServicesOnFacility");
```
bodnara, line 1195: Wouldn't it be better to split the test for block and unblock ?

sarkapalkovicova: There are more cases when we test some action and its opposite in the same test since most of the code would be duplicated. But if it makes more sense to you, I can split it with no problem.

bodnara: In that case it is fine with me.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -299,6 +342,38 @@ public Integer call(ApiCaller ac, Deserializer parms) throws PerunException {
...
 	},
 
+	/*#
+	 * Forces services propagation on defined facility.
+	 *
+	 * @param facility int Facility <code>id</code>
+	 * @param services List<Integer> <code>id</code> of service
+	 *
+	 * @throw FacilityNotExistsException if there is no such facility
+	 * @throw PrivilegeException insufficient permissions
```
HejdaJakub, line 352:
```suggestion
	 * @throw PrivilegeException insufficient permissions
	 * @throw ForceServicePropagationDisabledException when forcing propagation is not possible
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/ServicesManagerMethod.java
```diff
@@ -299,6 +342,38 @@ public Integer call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+	 */
+	/*#
+	 * Forces services propagation on all facilities where the services are defined on.
+	 *
+	 * @param services List<Integer> <code>id</code> of service
+	 *
+	 * @throw PrivilegeException insufficient permissions
```
HejdaJakub, line 359:
```suggestion
	 * @throw PrivilegeException insufficient permissions
	 * @throw ForceServicePropagationDisabledException when forcing propagation is not possible
```

perun-openapi/openapi.yml
```diff
@@ -17395,6 +17426,21 @@ paths:
         default:
           $ref: '#/components/responses/ExceptionResponse'
 
+  /urlinjsonout/servicesManager/forceServicePropagationBulk:
+    post:
+      tags:
+        - ServicesManager
+      operationId: forceServicePropagationBulk
+      summary: Forces services propagation on defined facility.
```
HejdaJakub, line 17434: If we want to have it as a single method with an optional facilityId parameter, I would at least prefer to define it in summary, e.g. : 
```suggestion
      summary: Forces services propagation on all facilities, where the services are defined, or just on the given facility.
```
or something like that.

HejdaJakub (changes requested):

HejdaJakub (approved):

PR #3935, opened 2023-04-19
BBMRI-ID Virtual attribute module

dBucik:
Attribute module to extract original BBMRI ID as an attribute value from the UserExtSource. Serves for the purposes of migrating BBMRI AAI into the LS AAI. The module requires naming the ExtSource "BBMRI id" and being of type "ExtSourceInternal".

This attribute will be used by the BBMRI ProxyIdP temporarily until it migrates to LS ID. When we do so, the attribute module can be deprecated and consequently removed.

The attribute also needs to be exposed in LDAP.


perun-ldapc/src/main/resources/perun-ldapc.xml
```diff
@@ -1273,6 +1273,16 @@ http://www.springframework.org/schema/aop http://www.springframework.org/schema/
 						</bean>
 					</property>
 				</bean>
+				<bean class="cz.metacentrum.perun.ldapc.model.impl.PerunAttributeDesc">
```
Johaney-s, line 1276: doesn't `perun-schema.ldif` also need to be updated?

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_bbmriUserIdTest.java
```diff
@@ -0,0 +1,144 @@
...
+			Arrays.asList(ues1, ues2, correctUes)
+		);
+
+		Attribute receivedAttr = classInstance.getAttributeValue(session, user, classInstance.getAttributeDefinition());
+		assertTrue(receivedAttr.getValue() instanceof Integer);
+		assertEquals("received attribute name is wrong",
+			classInstance.getAttributeDefinition().getFriendlyName(), receivedAttr.getFriendlyName());
+
+		@SuppressWarnings("unchecked")
+		Integer actual = (Integer) receivedAttr.getValue();
```
Johaney-s, line 72:
```suggestion
		Integer actual = receivedAttr.valueAsInteger();
```

dBucik: Fixed on all places.

perun-ldapc/src/main/resources/perun-ldapc.xml
dBucik: Please see the updated version.

Johaney-s:
Solved in https://github.com/CESNET/perun/pull/3943

PR #3941, opened 2023-04-24
fix(core): correctly delete user with persistent shadow login attribute

HejdaJakub:
* There was an issue that deletion of user didn't work correctly if user had some persistent attributes. The problem was that during the process of user deletion all ext sources were removed, but further the method removeAllAttributes() triggered changedAttributeHook() for persistent shadow attributes and the ext source was created again - this caused an SQL error (FK constraint). Deletion of ext sources was moved lower in the whole process which fixes this error. New test for this use case was also created.
* During the deletion of user are no longer blocked logins in all login-namespace attribute, but only in those that are stored in attributesToKeep config property.
* It was allowed to block the same login globally several times - it has been also fixed.


mattjoke (approved):
Looks good, however as mentioned, it should be also tested on other instances.

HejdaJakub:
Tested on devel and the previous bug was fixed.

PR #3944, opened 2023-05-02
feat(core): Added tags and attributes to HTML Parser

mattjoke:
* added new tags and attributes for parsing


perun-base/src/main/java/cz/metacentrum/perun/core/api/CoreConfig.java
HejdaJakub, line 1: New property in CoreConfig should be marked as `BREAKING CHANGE` according to our DoD. Maybe our new footer `DEPLOYMENT NOTE` could better fit this type of change, but for now, I would use the `BREAKING CHANGE` defined in our DoD.

HejdaJakub (changes requested):
It looks good otherwise.

PR #3950, opened 2023-05-12
feat(registrar): 🎸 BBMRIResources reg. module

dBucik:
Introduce generalized module for registrar replacing BBMRI Collections and BBMRI Networks. Previously used modules are marked as deprecated.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
@@ -1,0 +1,314 @@
+package cz.metacentrum.perun.registrar.modules;
+
+import com.google.common.base.Strings;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.Group;
+import cz.metacentrum.perun.core.api.Member;
+import cz.metacentrum.perun.core.api.PerunSession;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.exceptions.AlreadyMemberException;
+import cz.metacentrum.perun.core.api.exceptions.AttributeNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.ExternallyManagedException;
+import cz.metacentrum.perun.core.api.exceptions.GroupNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
+import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.NotGroupMemberException;
+import cz.metacentrum.perun.core.api.exceptions.PerunException;
+import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
+import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.VoNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
+import cz.metacentrum.perun.core.api.exceptions.WrongReferenceAttributeValueException;
+import cz.metacentrum.perun.core.bl.PerunBl;
+import cz.metacentrum.perun.registrar.exceptions.CantBeApprovedException;
+import cz.metacentrum.perun.registrar.exceptions.RegistrarException;
+import cz.metacentrum.perun.registrar.model.Application;
+import cz.metacentrum.perun.registrar.model.ApplicationFormItemData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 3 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
```diff
@@ -0,0 +1,292 @@
...
+		throws VoNotExistsException, UserNotExistsException, PrivilegeException,
+		MemberNotExistsException, RegistrarException, GroupNotExistsException,
+		AttributeNotExistsException, WrongAttributeAssignmentException, ExternallyManagedException,
+		WrongAttributeValueException, WrongReferenceAttributeValueException, NotGroupMemberException
+	{
+		// get perun and beans from session
+		PerunBl perun = (PerunBl)session.getPerun();
+		Vo vo = app.getVo();
+		User user = app.getUser();
+		Member member = perun.getMembersManagerBl().getMemberByUser(session, vo, user);
```
zlamalp, line 87: If you expect this to work only for the group applications when user is already a member of the VO, then you should probably check application type (if group is present) and also handle MemberNotExistsException in a better way.
Eg. you can silently skip it and do nothing or you can log the warning messages about it or throw better (custom) exception, which will be clear to the support team when they will be handling the error report.

dBucik: Think I have done my best how to really handle it. Thanks for the tip!

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
```diff
@@ -0,0 +1,317 @@
...
+		} catch (AttributeNotExistsException | WrongAttributeAssignmentException ex) {
+			//OK, we consider it as disabled, try manually configured resource root group
+		}
+
+		try {
+			try {
+				String resourceOriginGroupName = perun.getAttributesManager()
+					.getAttribute(session, app.getGroup(), RESOURCES_ROOT_GROUP_ATTR_NAME)
+					.valueAsString();
+				return perun.getGroupsManager().getGroupByName(session, app.getVo(), resourceOriginGroupName);
```
zlamalp, line 217: I suggest to use Bl instead of Entry layer for this kind of operations (getting groups, attributes, subgroups), since for example in this case you sort of redirect the root of resources to a completely different group, where approving person might not be the group manager and hence might not be able to retrieve relevant data. Same goes for the adding member to such group (if it will be outside of the sub-group tree).

zlamalp (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
dBucik: Thanks for the suggestion, I have applied BL for every call :)

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
```diff
@@ -0,0 +1,318 @@
...
+		WrongAttributeValueException, WrongReferenceAttributeValueException, NotGroupMemberException
+	{
+		if (app.getGroup() == null) {
+			throw new RegistrarException(
+				"Invalid usage of registrar module - module '" + this.getClass().getName() +
+					"' should be used on group level only");
+		}
+
+		// get perun and beans from session
+		PerunBl perun = (PerunBl) session.getPerun();
```
zlamalp, line 90: I believe you have changed this to the correct type `PerunBl`, but in order to actually call the BL layer you need to also call proper method with `Bl` suffix, eg.: `getMembersManager()` -> `getMembersManagerBl()`.

zlamalp (changes requested):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
dBucik: Updated as well. Hopefully this is the last needed change :).

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/modules/BBMRIResources.java
```diff
@@ -0,0 +1,314 @@
...
+		} catch (AttributeNotExistsException | WrongAttributeAssignmentException ex) {
+			//OK, we consider it as disabled, try manually configured resource root group
+		}
+
+		try {
+			try {
+				String resourceOriginGroupName = perun.getAttributesManagerBl()
+					.getAttribute(session, app.getGroup(), RESOURCES_ROOT_GROUP_ATTR_NAME)
+					.valueAsString();
+				return perun.getGroupsManager().getGroupByName(session, app.getVo(), resourceOriginGroupName);
```
zlamalp, line 215: This call is still using Entry instead of Bl, it is the last usage.

zlamalp (changes requested):

xflord (approved):

PR #3953, opened 2023-05-15
feat(core): default blocked login checker

sarkapalkovicova:
Added check that none of the default blocked logins is already used.

### DIFF
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/impl/DefaultBlockedLoginCheckerTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/impl/DefaultBlockedLoginCheckerTest.java
@@ -1,0 +1,101 @@
+package cz.metacentrum.perun.core.impl;
+
+import cz.metacentrum.perun.core.AbstractPerunIntegrationTest;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.CoreConfig;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.AttributeDefinitionExistsException;
+import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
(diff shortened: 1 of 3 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/impl/DefaultBlockedLoginCheckerTest.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	@Test
+	public void defaultBlockedLoginAlreadyUsed() {
+		System.out.println(CLASS_NAME + "defaultBlockedLoginIsAlreadyUsed");
+
+		assertThrows(InternalErrorException.class, () -> defaultBlockedLoginChecker.checkDefaultBlockedLogins());
+	}
+
+	private void addDefaultBlockedLoginToCoreConfig(String defaultBlockedLogin) {
```
xflord, line 47: When setting CoreConfig properties, we use a try block to return the property to its original value (to not tamper with the results of other tests). See 'anonymizeUserWhenAnonymizationNotSupported' for example. This would also mean removing this method and doing so in each test.

perun-core/src/test/java/cz/metacentrum/perun/core/impl/DefaultBlockedLoginCheckerTest.java
```diff
@@ -0,0 +1,74 @@
...
+		defaultBlockedLoginChecker = new DefaultBlockedLoginChecker(perun);
+
+		String login = "testLogin";
+
+		setUser();
+		setLoginNamespaceAttribute(login);
+		addDefaultBlockedLoginToCoreConfig(login);
+	}
+
+	@Test
```
xflord, line 40: Maybe add another test for a case when the exception is not thrown? If not, it feels unnecessary to have private setup methods.

xflord (commented):
Good job, I really like this solution. Also, could you make sure you test this locally/on devel, not just via the tests?

HejdaJakub (approved):

PR #3958, opened 2023-05-18
fix(core): updated perunDefaultBlockedLoginChecker

sarkapalkovicova:
Edited error message to contain list of user ids.
Check usage of all default blocked logins at once.


bodnara (approved):

mattjoke (approved):
Looks good 👍

PR #3961, opened 2023-05-23
fix(core): fix randomly failing test

michalberky:
* Changed RichGroup class equals method to compare sorted attribute lists

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichGroup.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/core/api/RichGroup.java
@@ -112,12 +112,20 @@ public class RichGroup extends Group {
 			return false;
 		}
 		if (attributes == null) {
-			if (other.getAttributes() != null) {
-				return false;
-			}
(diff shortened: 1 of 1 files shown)


perun-base/src/main/java/cz/metacentrum/perun/core/api/RichGroup.java
```diff
@@ -115,8 +116,20 @@ public boolean equals(Object obj) {
...
+		} else {
+			List<Attribute> sortedThis = this.getAttributes().stream()
+				.sorted()
+				.collect(Collectors.toList());
+
+			List<Attribute> sortedOther = other.getAttributes().stream()
+				.sorted()
+				.collect(Collectors.toList());
+
+			if (!sortedThis.equals(sortedOther)) {
```
sarkapalkovicova, line 130:
```suggestion
			retrun sortedThis.equals(sortedOther);
```

sarkapalkovicova (commented):
The last condition can be replaced with a return statement which would look a little nicer in my opinion. Otherwise, good job.

PR #3963, opened 2023-05-24
feat: use introspection data saved in request headers for MFA

xflord:
* mfaTimestamp for additionalInformation is now retrieved from the request headers filled in by Apache server
* removed the logic retrieving MFA timestamp from userinfo endpoint


HejdaJakub (approved):
I think it looks good.

xflord:
Will test one more time with the changed property names (however should still work)

HejdaJakub (approved):

bodnara (approved):

PR #3964, opened 2023-05-24
feat: allow embedded group form item in group applications

HejdaJakub:
* Now is possible to use embedded group form item also in group applications. It means that we can specify in group form for which subgroups should be created also embedded application.
* There are several new methods which interacts with new table. Old functionality is deprecated.
* Old test were moved from GroupsManagerEntryIntegrationTest file to the RegistrarBaseIntegrationTest. New tests for subgroups were added, too.
* Implementation into the old gui is a part of this commit.

BREAKING CHANGE: new database table auto_registration_group was created, data from the old table groups_to_register should be migrated into this new table with appropriate relation to form_item_id


HejdaJakub:
It was tested via this implementation of the old GUI. Embedded group item form is possible to set for Vo and also for Group. Embedded applications were correctly created in both use cases. It is also possible to copy embedded app form items also with related groups if in the destination vo/group is a group with the same name.

This PR is based on https://github.com/CESNET/perun/pull/3928, so @dBucik prepared a huge part of this code, thanks :tada:

bodnara (commented):
Form just going over the changes it looks to be correct (did not check any changes to gui tho). 

Not sure if we take into consideration that groups can be moved in the structure. So it is possible some group can become top level and no longer be a subgroup. It seems to me that at that point it will get stuck in the auto registration ?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/GroupsManagerImpl.java
```diff
@@ -1310,6 +1333,15 @@ public void deleteGroupFromAutoRegistration(PerunSession sess, Group group) {
 		}
 	}
 
+	@Override
+	public void deleteGroupFromAutoRegistration(PerunSession sess, Group group, ApplicationFormItem formItem) {
+		try {
+			jdbc.update("delete from auto_registration_groups where group_id=? and application_form_item_id=?", group.getId(), formItem.getId());
```
xflord, line 1339: A check that the form item exists would be nice (or at least throw a comprehensive exception if not)

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3689,6 +3734,46 @@ public void deleteGroupsFromAutoRegistration(PerunSession sess, List<Group> grou
 		perun.getGroupsManagerBl().deleteGroupsFromAutoRegistration(sess, groups);
 	}
 
+	@Override
+	public void deleteGroupsFromAutoRegistration(PerunSession sess, List<Group> groups, ApplicationFormItem formItem) throws GroupAlreadyRemovedException, GroupNotExistsException, PrivilegeException {
```
xflord, line 3738: Add the check that form item exists here ( I know its in RegistrarManagerMethod, but in case someone calls this method not through RPC)

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3716,6 +3801,65 @@ public void addGroupsToAutoRegistration(PerunSession sess, List<Group> groups) t
...
+		}
+
+		//Authorization
+		for (Group group : groups) {
+			if (!AuthzResolver.authorizedInternal(sess, "addGroupsToAutoRegistration_List<Group>_ApplicationFormItem_policy", group)) {
+				throw new PrivilegeException(sess, "addGroupsToAutoRegistration");
+			}
+		}
+
+		// Create application form if non exists
```
xflord, line 3819: There are a few lines of code in submitEmbeddedGroupApplications() that also create a new form if it does not exist, which are obsolete now (lines 1909-1913 I believe)

xflord (changes requested):
Do not have time to check WUI today, let me know if it is necessary and I will take a look another day

perun-db/postgres.sql
```diff
@@ -1612,6 +1612,13 @@ create table authz (
 	     ((user_id is not null and authorized_group_id is null) or (user_id is null and authorized_group_id is not null))
 );
 
+create table auto_registration_groups (
```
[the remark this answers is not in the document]
xflord: Also code to move the groups from "groups_to_register" to the new table should be in the changelog. (If I understand correctly that the tables carry the same information)

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
HejdaJakub: Removed from msubmitEmbeddedGroupApplications().

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -3689,6 +3734,46 @@ public void deleteGroupsFromAutoRegistration(PerunSession sess, List<Group> grou
...
+		List<Group> subGroups = perun.getGroupsManagerBl().getAllSubGroups(sess, registrationGroup);
+		for (Group group : groups) {
+			perun.getGroupsManagerBl().checkGroupExists(sess, group);
+			if (!subGroups.contains(group)) {
+				throw new GroupIsNotASubgroupException();
+			}
+		}
+
+		//Authorization
+		for (Group group : groups) {
```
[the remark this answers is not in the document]
HejdaJakub: It was replaced by the new array to match the `List<PerunBean>` type correctly. Fixed for delete/add methods.

perun-db/postgres.sql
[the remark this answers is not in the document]
HejdaJakub: As we discussed on a daily, I'll create a migration script that will connect old values (group ids) with the form item ids and store them in that new table `auto_registration_groups`.

HejdaJakub:
The migration script is prepared and the use case for moving groups is fixed (I hope).

bodnara (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -12,39 +12,16 @@
...
-import cz.metacentrum.perun.core.api.PerunClient;
-import cz.metacentrum.perun.core.api.PerunPrincipal;
-import cz.metacentrum.perun.core.api.PerunSession;
-import cz.metacentrum.perun.core.api.Role;
-import cz.metacentrum.perun.core.api.RoleObject;
-import cz.metacentrum.perun.core.api.Status;
-import cz.metacentrum.perun.core.api.User;
-import cz.metacentrum.perun.core.api.UserExtSource;
-import cz.metacentrum.perun.core.api.Vo;
+import cz.metacentrum.perun.core.api.*;
```
xflord, line 15: wildcard import

xflord (approved):
After external discussion, the migration script can stay as it is.
One last nitpick, otherwise very well done.

PR #3966, opened 2023-05-25
feat(core): add PROXY role

Johaney-s:
* to remove perunadmin role for proxy, new role is created
* it covers necessary method calls and attribute rights


bodnara (approved):
Looks good. However I can't speak whether new role was added to all required methods, and correct rights were given to all attribute definitions.

PR #3972, opened 2023-06-01
fix(core): fixed attributes parsing for sql group structure sync

balcirakpeter:
The current parsing does not include additional group attributes like `g:d:adName`. 
I have used the previous implementation of the process except for getting the DB connection which is still taken from the DB pool.


zlamalp (approved):
I think that we don't have to re-introduce so much of the old code (e.g. for handling blobs), since we are using better implementation in `querySource()` already, but it should work. Since it passed the manual testing I'm ok with it to be a quickfix.

PR #3974, opened 2023-06-07
fix(core): refresh principal roles more effectively

HejdaJakub:
* In the method refreshAuthz were all roles removed from principal and then consecutively added according to several conditions. This can caused for some parallel methods that the roles in principal were for a moment empty and due to this behavior some methods threw notAuthorized error.
* Now the roles are prepared aside and then replaced at one line.


sarkapalkovicova (approved):
I tested it, and it is working. The code looks good to me too.

bodnara (approved):

PR #3975, opened 2023-06-08
fix(core): MFA timeout

HejdaJakub:
* If MFA is not valid anymore, then the new exception is thrown - it must be specifically managed by client (now just by the new GUI).

BREAKING CHANGE: The unit of the config property 'mfaAuthTimeout' has been changed from hours to minutes (for easier testing).

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -55,6 +55,7 @@ import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MfaInvalidRolesException;
 import cz.metacentrum.perun.core.api.exceptions.MfaPrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.MfaRolePrivilegeException;
+import cz.metacentrum.perun.core.api.exceptions.MfaTimeoutException;
 import cz.metacentrum.perun.core.api.exceptions.PolicyNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.RoleAlreadySetException;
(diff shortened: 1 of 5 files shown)


HejdaJakub:
The use case when the MFA is expired works well, but we need to figure out also the use case when the user hasn't authenticated with the second factor yet, but the timestamp (first auth time) + timeout is before now, so we want to throw an exception (and force authentication from the client) also for this case.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -4357,34 +4355,47 @@ private static Map<String, Integer> createMappingOfValues(PerunBean complementar
...
+	 * Check if the auth time + mfa timeout is not older than the current time
+	 * auth time = time of the first authentication
+	 * mfa timeout = amount of time defined in the config for how long the MFA should be valid (since SFA)
 	 *
 	 * @param sess session
+	 * @param throwError if this method should throw errors or just return boolean
 	 * @return true if principal authorized by MFA in allowed limit, false otherwise
 	 */
-	private static boolean isAuthorizedByMfa(PerunSession sess) {
+	private static boolean isAuthorizedByMfa(PerunSession sess, boolean throwError) {
```
bodnara, line 4367: The `throwError` argument seems redundant. The caller know if exception needs to be thrown or not.

HejdaJakub: I am looking at the code again and I think it is not probably redundant. In this method (`isAuthorizedByMfa()`) can be returned false in more cases:
```
if (!BeansUtils.getCoreConfig().isEnforceMfa()) {
	return false;
}
```
and
```
// check if the auth time + mfa timeout > the current time
if (mfaValidUntil.isAfter(Instant.now())) {
	// if user has MFA and it is still valid
	return sess.getPerunPrincipal().getAdditionalInformations().containsKey(ACR_MFA);
```
so without a duplicate check, I wouldn't be able to know from the caller when to throw an exception.

bodnara (approved):

PR #3981, opened 2023-06-14
feat(core): lsaai login policy

sarkapalkovicova:
- Updated login policy to include not only numbers and start with a letter.
- Made check for duplicates case-insensitive.


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_login_namespace_lifescienceid_username.java
```diff
@@ -4,24 +4,80 @@
...
 	public void changedAttributeHook(PerunSessionImpl sess, User user, Attribute attribute) {
 		trySetAttribute(sess, user, attribute, elixirUsername);
 		trySetAttribute(sess, user, attribute, bbmriUsername);
 	}
 
+	@Override
+	public void checkAttributeSyntax(PerunSessionImpl sess, User user, Attribute attribute) throws WrongAttributeValueException {
+		super.checkAttributeSyntax(sess, user, attribute);
+
+		if (attribute.getValue() != null) {
```
mattjoke, line 39: This may be a nitpick, but if you'd have reversed the condition, you wouldn't need this big of an if. 
```suggestion
		if (attribute.getValue() == null) {
		    return;
		}
```

bodnara (approved):

mattjoke (approved):

PR #3984, opened 2023-06-20
feat(core): use percentage property to force MFA log in

HejdaJakub:
* We want to force log in with both factors earlier (e.g. 75% of mfaAuthTimeout) due to the first executed MFA since authentication time.
* We want to avoid situation when the validity is e.g. 60 minutes, user executes MFA (just second factor) after 59 minutes and after one minute he/she would need to log in again with both factors.
* Fix correct handling of MfaTimeoutException also for MFA critical roles.

BREAKING CHANGE: new configuration property introspectionEndpoint.mfaAuthTimeoutPercentageForceLogIn

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -55,6 +55,7 @@ import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.MfaInvalidRolesException;
 import cz.metacentrum.perun.core.api.exceptions.MfaPrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.MfaRolePrivilegeException;
+import cz.metacentrum.perun.core.api.exceptions.MfaRoleTimeoutException;
 import cz.metacentrum.perun.core.api.exceptions.MfaTimeoutException;
 import cz.metacentrum.perun.core.api.exceptions.PolicyNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceNotExistsException;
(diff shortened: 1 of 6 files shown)


bodnara (approved):

bodnara (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -4380,21 +4406,18 @@ private static boolean isAuthorizedByMfa(PerunSession sess, boolean throwError)
...
-			if (sess.getPerunPrincipal().getAdditionalInformations().containsKey(ACR_MFA)) {
-				// MFA is no longer valid
-				throw new MfaTimeoutException("Your MFA timestamp " + returnedAuthTime + " is not valid anymore, you'll need to reauthenticate");
-			} else {
-				// user is authenticated by SFA but the mfa timeout would cause an error, so we need to reauthenticate this user
-				throw new MfaTimeoutException("Your single factor authentication timestamp " + returnedAuthTime + " is not valid anymore, you'll need to reauthenticate");
+		double mfaTimeoutPercentage = 1;
+		// if the current session is SFA, we want to force log in with both factors earlier (e.g. 75% of mfaAuthTimeout) due to the first executed MFA since authentication time
+		// -> we want to avoid situation when the validity is e.g. 60 minutes, user executes MFA (just second factor) after 59 minutes and after one minute he/she would need to log in again with both factors
+		if (!sess.getPerunPrincipal().getAdditionalInformations().containsKey(ACR_MFA)) {
```
bodnara, line 4412: We have this check on many places, but it is quite ambiguous. Would it be possible to create method that would return if session is SFA with a more explanatory name ? Would be also helpful in case that the check changes, only one place needs to be modified.

bodnara (approved):
I bless this code, so it does work first time as intended :pray:

PR #3988, opened 2023-06-25
fix(core): periodic update of roles

sarkapalkovicova:
Update roles in periodic configurable intervals in the scope of minutes.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -96,6 +96,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -139,6 +139,13 @@ public static boolean authorized(PerunSession sess, String policyDefinition, Lis
 			refreshAuthz(sess);
 		}
 
+		if (System.currentTimeMillis() - sess.getPerunPrincipal().getRolesUpdatedAt() >= (long) BeansUtils.getCoreConfig().getRoleUpdateInterval() * 60000) {
```
HejdaJakub, line 142: I know that this is really easy math, but IMHO it can be more readable with something like this: TimeUnit.MINUTES.toMillis(_myMinutes_) - it also returns long, so you don't need to change the type then;
```suggestion
		if (System.currentTimeMillis() - sess.getPerunPrincipal().getRolesUpdatedAt() >= TimeUnit.MINUTES.toMillis(BeansUtils.getCoreConfig().getRoleUpdateInterval())) {
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
HejdaJakub, line 1: I don't know if the `authorized` method is the only one potentially problematic place. I'm not sure, but it looks like there can be the same threat also in methods `authorizedToManageRole`, `authorizedToReadRole` (both related to `un/setRoles`), and `hasAccessByDefault` (related to all `isAuthorizedForAttribute` methods).

Am I right or this is not a problem? What do you think? :thinking:

HejdaJakub (approved):

mattjoke (approved):
Looks good :+1:

PR #3991, opened 2023-06-28
feat(core): give perun user a perunadmin role in DB

zlamalp:
- Do not use perun.properties config for assignment of privileged role, since it prevents existence of the same login in a different namespaces.
- "perunadmin" role is created during the DB init and assigned to the "Master Perun" user based on matching basic auth identity (the default one).
- Updated DB to make sure existing users are assigned with this role.

DEPLOYMENT NOTES: Remove 'perun' from the 'perun.admins' property in perun.properties config file.


martin-kuba (approved):

HejdaJakub (commented):
Please add also `BREAKING CHANGE` footer to the commit message due to DB change. It looks good otherwise.

HejdaJakub (approved):

PR #3993, opened 2023-06-29
feat(core): virtual attribute for eligibilities

Kuliak:
* new user virtual attributed created for aggregating eligibilities attributes from obtained user ext sources (only the highest value is used for each key)


perun-ldapc/src/main/java/cz/metacentrum/perun/ldapc/model/PerunAttribute.java
```diff
@@ -74,6 +74,7 @@ public interface PerunAttributeNames {
 		public static final String ldapAttrAdminOfGroup = "adminOfGroup";
 		public static final String ldapAttrAdminOfFacility = "adminOfFacility";
 		public static final String ldapAttrUuid = "uuid";
+		public static final String ldapAttrUserEligibilities = "userEligibilities";
```
michalberky, line 77: Here you should be able to use the variable _perunAttrUserEligibilities_ defined on line 27 instead of defining the same string again. Should be a non-issue, but would be more consistent with other Perun attribute names also used in this way in LDAP attribute names definitions.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_userEligibilities.java
```diff
@@ -0,0 +1,135 @@
...
+//		} else if (message instanceof AttributeChangedForUser) {
+//			AttributeDefinition attrVirtUserEligibilitiesDefinition = perunSession.getPerunBl().getAttributesManagerBl().getAttributeDefinition(perunSession, A_U_V_USER_ELIGIBILITIES);
+//			resolvingMessages.add(new AttributeChangedForUser(new Attribute(attrVirtUserEligibilitiesDefinition), ((AttributeChangedForUser) message).getUser()));
+//		} else if (message instanceof AttributeRemovedForUser) {
+//			AttributeDefinition attrVirtUserEligibilitiesDefinition = perunSession.getPerunBl().getAttributesManagerBl().getAttributeDefinition(perunSession, A_U_V_USER_ELIGIBILITIES);
+//			resolvingMessages.add(new AttributeRemovedForUser(new Attribute(attrVirtUserEligibilitiesDefinition), ((AttributeRemovedForUser) message).getUser()));
+//		}
+//
+//		return resolvingMessages;
+//	}
```
xflord, line 117: this can be removed

xflord (commented):
I'm unsure whether the changes to `PerunAttribute.java` and `addProxyRolePolicies` are necessary.
However, I tested this manually, and the propagation works
Screenshot from 2023-07-12 09-27-28

balcirakpeter:
Good points, but the comments do not represent critical issues and we need to merge the PR ASAP so I would proceed and resolve the comments another time.

PR #4000, opened 2023-07-07
feat(core): skip MFA for internal components

xflord:
* skip the MFA check when refreshing authz roles of service component principals (defined in perun.properties).

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -362,6 +362,7 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 		List<String> perunAdmins = new ArrayList<>(BeansUtils.getCoreConfig().getAdmins());
 		perunAdmins.addAll(BeansUtils.getCoreConfig().getRegistrarPrincipals());
 		if (perunAdmins.contains(sess.getPerunPrincipal().getActor())) {
+			log.debug("skipped MFA policy check for {}", sess.getPerunPrincipal().getActor());
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


sarkapalkovicova (approved):
It looks like it could work. I recommend that someone else also tests it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -2593,7 +2594,11 @@ public static synchronized void refreshAuthz(PerunSession sess) {
 			}
 		}
 
-		checkMfaForHavingRole(sess, sess.getPerunPrincipal().getRoles());
+		if (!serviceRole) {
```
dBucik, line 2597: @xflord @balcirakpeter Please re-think if PerunAdmin should be considered as a service role in the context. 
 I agree with the internal roles like engine, registrar, etc., but do not see why it should be not checked for the PerunAdmin role, given that this role is represented by a person capable of performing Multi-Factor Authentication.

xflord: (am on holiday and might be out of the loop but will try to help when possible)

HejdaJakub (commented):
I tested it and the skip part looks good, just maybe the debug log can be a little bit noisy. I tried it on devel and especially for dispather there were thousands of these debug lines in a few minutes. :eyes:

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
dBucik: Ah, ignore my comment, please. I misunderstood the code above. I did not realize that it actually checks the PerunAdmin for service roles, not for every role.
(and of course, I should have put it a few lines above 😁 )

PR #4005, opened 2023-07-14
fix(core): filter candidates bases on group related roles

sarkapalkovicova:
Filter out candidates to which the user does not have a correct privilege.


HejdaJakub (approved):
Tested on devel and it looks good. :+1:

PR #4006, opened 2023-07-17
fix(core): filter get rich users with attributes

sarkapalkovicova:
Filter out users to which the user does not have a correct privilege.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
@@ -1425,8 +1425,37 @@ public class UsersManagerEntry implements UsersManager {
 			}
 		}
 
-		return getPerunBl().getUsersManagerBl().filterOnlyAllowedAttributes(sess, getUsersManagerBl().findRichUsersWithAttributes(sess, searchString, attrNames));
+		List<RichUser> users = getPerunBl().getUsersManagerBl().filterOnlyAllowedAttributes(sess, getUsersManagerBl().findRichUsersWithAttributes(sess, searchString, attrNames));
+		return filterRichUsers(sess, users);
+	}
 
+	/**
+	 * Filters rich users to which the principal has access.
+	 *
+	 * @param sess session
+	 * @param users list of all rich users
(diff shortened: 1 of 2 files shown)


mattjoke (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -1425,8 +1425,36 @@ public List<RichUser> findRichUsersWithAttributes(PerunSession sess, String sear
...
+			List<Vo> vos = perunBl.getUsersManagerBl().getVosWhereUserIsMember(sess, user);
+			List<Group> groups = perunBl.getGroupsManagerBl().getUserGroups(sess, user);
+			List<Facility> facilities = perunBl.getFacilitiesManagerBl().getAssignedFacilities(sess, user);
+			List<Resource> resources = perunBl.getUsersManagerBl().getAllowedResources(sess, user);
+
+			if (AuthzResolver.authorizedInternal(sess, "filter-findRichUsersWithAttributes_policy") ||
+				vos.stream().anyMatch(vo -> AuthzResolver.authorizedInternal(sess, "filter-findRichUsersWithAttributes_policy", vo)) ||
+				groups.stream().anyMatch(group -> AuthzResolver.authorizedInternal(sess, "filter-findRichUsersWithAttributes_policy", group)) ||
+				facilities.stream().anyMatch(facility -> AuthzResolver.authorizedInternal(sess, "filter-findRichUsersWithAttributes_policy", facility)) ||
+				resources.stream().anyMatch(resource -> AuthzResolver.authorizedInternal(sess, "filter-findRichUsersWithAttributes_policy", resource))) {
```
dBucik, line 1452: Perhaps extract `filter-findRichUsersWithAttributes_policy` into a String variable before the `if` statement so you do not need to repeat it everywhere and just use the variable in method calls instead.
This will prevent future changes from missing modification of all occurrences in the future and will be clearer that we are referring to the same policy definition.

dBucik (commented):
One suggestion, otherwise GTG by me :).

dBucik (approved):

PR #4008, opened 2023-07-19
fix(cli): fixed encoding

mattjoke:
* setValueFromArray and getValueAsScalar now escape to/from UTF-8
* Dumper uses Unicode sequence and therefore it is replaced


sarkapalkovicova (approved):
It seems to work.

michalberky (approved):
Works as it should

PR #4016, opened 2023-07-24
feat(core): filter getMembersPage

mattjoke:
* getMembersPage now adhere to policy
* added filter `filter-getMembersPage_policy` policy
* added getPolicyByName to AuthzResolver
* added support for VOADMIN and VOOBSERVER

### DIFF
--- a/perun-base/src/test/resources/test-roles.yml
+++ b/perun-base/src/test/resources/test-roles.yml
@@ -205,5 +205,38 @@ perun_policies:
       - PROXY:
     include_policies: []
 
+  test_filter-getMembersPage_policy:
+    policy_roles:
+      - GROUPADMIN: Group
+      - GROUPOBSERVER: Group
+      - GROUPMEMBERSHIPMANAGER: Group
+      - PERUNOBSERVER:
+      - VOADMIN:
+      - VOOBSERVER:
+    include_policies:
+      - default_policy
+
+  test_filter-getMembersPage_policy-vo:
+    policy_roles:
+      - GROUPADMIN: Group
+      - GROUPOBSERVER: Group
+      - GROUPMEMBERSHIPMANAGER: Group
+      - PERUNOBSERVER:
+      - VOADMIN: Vo
+      - VOOBSERVER:
+    include_policies:
+      - default_policy
+
+  test_filter-getMembersPage_policy-voobserver:
+    policy_roles:
+      - GROUPADMIN: Group
+      - GROUPOBSERVER: Group
+      - GROUPMEMBERSHIPMANAGER: Group
+      - PERUNOBSERVER:
+      - VOADMIN:
+      - VOOBSERVER: Vo
+    include_policies:
+      - default_policy
+
 perun_roles_management: {}
 ...
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
@@ -24,6 +24,7 @@ import cz.metacentrum.perun.core.api.exceptions.ParentGroupNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordCreationFailedException;
 import cz.metacentrum.perun.core.api.exceptions.NotificationMemberMailNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.PasswordStrengthException;
+import cz.metacentrum.perun.core.api.exceptions.PolicyNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.PrivilegeException;
 import cz.metacentrum.perun.core.api.exceptions.ResourceNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.SponsorshipDoesNotExistException;
@@ -1521,7 +1522,23 @@ public interface MembersManager {
 	 * @throws GroupNotExistsException if there is no such query group
 	 * @throws PrivilegeException insufficient permission
 	 */
-	Paginated<RichMember> getMembersPage(PerunSession sess, Vo vo, MembersPageQuery query, List<String> attrNames) throws VoNotExistsException, PrivilegeException, GroupNotExistsException;
+	Paginated<RichMember> getMembersPage(PerunSession sess, Vo vo, MembersPageQuery query, List<String> attrNames) throws VoNotExistsException, PrivilegeException, GroupNotExistsException, PolicyNotExistsException;
+
+	/**
+	 * Get page of members from the given vo, with the given attributes, based on policy.
+	 *
+	 * @param sess session
+	 * @param vo vo
+	 * @param query query with page information
+	 * @param attrNames attribute names
+	 * @param policy policy to use
+	 * @return page of requested rich members
+	 * @throws VoNotExistsException if there is no such vo
+	 * @throws GroupNotExistsException if there is no such query group
+	 * @throws PrivilegeException insufficient permission
+	 */
+	Paginated<RichMember> getMembersPage(PerunSession sess, Vo vo, MembersPageQuery query, List<String> attrNames, String policy) throws VoNotExistsException, PrivilegeException, GroupNotExistsException, PolicyNotExistsException;
+
 
 	/**
 	 * Update the sponsorship of given member for given sponsor.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -2289,6 +2289,16 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 	}
 
 	/**
+	 * Returns true if the perun principal inside the perun session is Perun Observer.
+	 *
+	 * @param sess perun session
+	 * @return true if the perun principal is top group creator.
+	 */
+	public static boolean isPerunObserver(PerunSession sess) {
+		return sess.getPerunPrincipal().getRoles().hasRole(Role.PERUNOBSERVER);
+	}
+
+	/**
 	 * Returns true if the perun principal inside the perun session is top group creator.
 	 *
 	 * @param sess perun session
@@ -2306,6 +2316,16 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 	 */
 	public static boolean isPerunAdmin(PerunSession sess) {
 		return sess.getPerunPrincipal().getRoles().hasRole(Role.PERUNADMIN);
+	}
+
+	/**
+	 * Returns true if perun principal is Vo admin or Vo observer of specific Vo.
+	 * @param sess - perun session
+	 * @param vo -specific vo
+	 * @return bolean
+    **/
+	public static boolean isVoAdminOrObserver(PerunSession sess, Vo vo) {
+		return authzResolverImpl.isVoAdminOrObserver(sess, vo);
 	}
 
 	/**
@@ -4471,5 +4491,15 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 	 */
 	private static boolean sessionHasMfa(PerunSession sess) {
 		return sess.getPerunPrincipal().getAdditionalInformations().containsKey(ACR_MFA);
+	}
+
+	/**
+	 * Return id of the role by its name.
+	 *
+	 * @param name - name of the role
+	 * @return - id of the role
+	 */
+	public static int getRoleIdByName(String name) {
+		return authzResolverImpl.getRoleIdByName(name);
 	}
 }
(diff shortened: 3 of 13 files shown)


mattjoke:
Manual test needed. I'll add automatic tests that use the filter policy.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/MembersManagerBlImpl.java
```diff
@@ -3059,6 +3060,21 @@ public Paginated<RichMember> getMembersPage(PerunSession sess, Vo vo, MembersPag
...
 
+	private Paginated<Member> filterMembersBasedOnPolicy(PerunSession sess, Paginated<Member> paginatedMembers, MembersPageQuery query) {
+		// Filter out based on the policy, if necessary load another data to fill the page
+		List<Member> members = paginatedMembers.getData();
+
+		List<Member> filteredMembers = new ArrayList<>();
+
+		for (Member member: members) {
+
+		}
```
dBucik: Empty for loop?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -884,6 +890,46 @@ private String getSQLWhereForMembersPage(MembersPageQuery query, MapSqlParameter
...
+	}
+
+	private String getWhereConditionBasedOnThePolicy(MembersPageQuery query, MapSqlParameterSource namedParams) {
+		PerunPolicy policy = AuthzResolverBlImpl.getPolicyByName("filter-getMembersPage_policy");
+		if (policy == null) {
+			return " WHERE members.vo_id = (:voId)";
+		}
+
+		StringBuilder sql = new StringBuilder();
+		sql.append(" WHERE members.vo_id = (:voId) AND (");
```
dBucik, line 909: Perhaps in the logical organization of the code having these two lines would be better later in the method as they might not be used if roles are evaluated empty.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -884,6 +890,46 @@ private String getSQLWhereForMembersPage(MembersPageQuery query, MapSqlParameter
...
+			"			JOIN authz ON g.id = authz.group_id" +
+			"			JOIN groups_members gm ON g.id = gm.group_id" +
+			"		 WHERE authz.user_id = (:userId))" +
+			"		AS members_group ON members.id = members_group.member_id AND members.vo_id = members_group.vo_id";
+	}
+
+	private String getWhereConditionBasedOnThePolicy(MembersPageQuery query, MapSqlParameterSource namedParams) {
+		PerunPolicy policy = AuthzResolverBlImpl.getPolicyByName("filter-getMembersPage_policy");
+		if (policy == null) {
+			return " WHERE members.vo_id = (:voId)";
```
dBucik, line 905: this return statement is repetitively used. Either define it as a commonly returned variable or reorganize the code a bit so it does not appear multiple times.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -884,6 +890,46 @@ private String getSQLWhereForMembersPage(MembersPageQuery query, MapSqlParameter
...
+		List<String> roles = new ArrayList<>();
+		for (Map<String,String> role : policy.getPerunRoles()) {
+			for (Map.Entry<String,String> entry : role.entrySet()) {
+				// Do nothing
+				if (entry.getValue() == null) continue;
+				if (entry.getValue().equals("Group")) {
+					switch (entry.getKey()) {
+						case "GROUPADMIN" -> roles.add("members_group.role_id=23");
+						case "GROUPMEMBERSHIPMANAGER" -> roles.add("members_group.role_id=174");
+						case "GROUPOBSERVER" -> roles.add("members_group.role_id=170");
```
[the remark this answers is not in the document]
mattjoke: This is a temporary solution, I wanted to start the review on other parts of the code. I'll add more robust solution.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -4452,4 +4452,14 @@ private static boolean checkAuthValidityForMFA(PerunSession sess) {
...
 		return sess.getPerunPrincipal().getAdditionalInformations().containsKey(ACR_MFA);
 	}
+
+	/**
+	 * Returns PerunPolicy object by its name.
+	 *
+	 * @param name - name of the policy
+	 * @return PerunPolicy object
+	 */
+	public static PerunPolicy getPolicyByName(String name) {
```
HejdaJakub, line 4462: I know that in this class is not the best code organization, but I would move this public method higher in the code between other general methods (it is often mixed in this class, but we usually put private methods at the end of the class).

HejdaJakub (changes requested):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AuthzResolverImpl.java
```diff
@@ -1272,4 +1272,22 @@ private String prepareSelectQueryString(Map<String, Integer> mappingOfValues) {
...
 		return StringUtils.join(listOfConditions, " and ");
 	}
+
+
+	/**
+	 * Returns policy based on its name, null if not exists
+	 *
+	 * @param name - name of the policy
+	 * @return PerunPolicy
+	 */
```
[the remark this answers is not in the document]
mattjoke: I'll use `getPerunPolicy` instead of my now useless code.

sarkapalkovicova:
Such a situation occurred during testing. I had filled the policy, and as VoAdmin of Vo (without PerunAdmin role), I could not see any Vo members. Please check this scenario, and if my setup was correct and there is a bug, then fix it.

perun-base/src/test/resources/test-roles.yml
```diff
@@ -205,5 +205,14 @@ perun_policies:
       - PROXY:
     include_policies: []
 
+  test_filter-getMembersPage_policy:
+    policy_roles:
+      - GROUPADMIN: Group
+      - GROUPOBSERVER: Group
+      - GROUPMEMBERSHIPMANAGER: Group
+      - PERUNOBSERVER: Group
```
sarkapalkovicova, line 213: The PerunObserver role is not related to any object.

```suggestion
      - PERUNOBSERVER: 
```

HejdaJakub (changes requested):
I tried several combinations with Group related roles and what was working before is now broken again.. :disappointed:
When I am just GROUPADMIN in the group with one member and the filter policy is set to this:
```
perun_policies:
  filter-getMembersPage_policy:
    policy_roles:
      - GROUPADMIN: Group
    include_policies:
      - default_policy 
```
I can still see all VO members on the organization-members page.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -884,6 +905,52 @@ private String getSQLWhereForMembersPage(MembersPageQuery query, MapSqlParameter
...
+		}
+
+		boolean ignoreGroupRelation = AuthzResolverBlImpl.isPerunAdmin(sess) || AuthzResolverBlImpl.isVoAdmin(sess) || AuthzResolverBlImpl.isVoObserver(sess) || AuthzResolverBlImpl.isPerunObserver(sess);
+		if (roles.isEmpty() || query.getGroupId() != null || ignoreGroupRelation) {
+			System.out.println("ignoreGroupRelation: " + ignoreGroupRelation);
+			System.out.println("isPerunAdmin: " + AuthzResolverBlImpl.isPerunAdmin(sess));
+			System.out.println("isVoAdmin: " + AuthzResolverBlImpl.isVoAdmin(sess));
+			System.out.println("isVoObserver: " + AuthzResolverBlImpl.isVoObserver(sess));
+			System.out.println("isGroupAdmin: " + AuthzResolverBlImpl.isGroupAdmin(sess));
+			System.out.println("isPerunObserver: " + AuthzResolverBlImpl.isPerunObserver(sess));
```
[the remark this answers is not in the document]
mattjoke: Yes, it's just a debug output... I'll remove them

HejdaJakub (approved):
I tested it again and from my side, it looks good now. :+1:

sarkapalkovicova (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -884,6 +906,47 @@ private String getSQLWhereForMembersPage(MembersPageQuery query, MapSqlParameter
...
+					roles.add("members_group.role_id=" + roleId);
+				}
+			}
+		}
+
+		// Check if user is VO admin in vo
+		boolean ignoreGroupRelation = AuthzResolverBlImpl.isPerunAdmin(sess) || AuthzResolverBlImpl.isPerunObserver(sess) || isVoAdminOrObserver(sess, vo);
+		if (roles.isEmpty() || query.getGroupId() != null || ignoreGroupRelation) {
+			return " WHERE members.vo_id = (:voId)";
+		}
```
xflord, line 944: We could do this before the for loop, no?

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -912,4 +975,13 @@ private String getGroupStatusSQLConditionForMembersPage(MembersPageQuery query,
...
 
+	private boolean isVoAdminOrObserver(PerunSession sess, Vo vo) {
+		try {
+			var query = jdbc.query("SELECT 1 FROM authz WHERE user_id=? AND vo_id=? AND (role_id=? OR role_id=?)", (rs, i) -> true,sess.getPerunPrincipal().getUserId(), vo.getId(), AuthzResolverBlImpl.getRoleIdByName(Role.VOADMIN), AuthzResolverBlImpl.getRoleIdByName(Role.VOOBSERVER));
+			return !query.isEmpty();
+		} catch (InternalErrorException e) {
+			log.error("Error during checking if user is vo admin of vo {}", vo, e);
+		}
+		return false;
+	}
```
xflord, line 986: This probably belongs in AuthzResolver, could you please move it there? Methods like `isVoAdmin()` already exist, just need to be extended a bit.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
```diff
@@ -1506,7 +1507,23 @@ List<Map<String, String>> createSponsoredMembersFromCSV(PerunSession sess, Vo vo
...
+	 * @param vo vo
+	 * @param query query with page information
+	 * @param attrNames attribute names
+	 * @param policy policy to use
+	 * @return page of requested rich members
+	 * @throws VoNotExistsException if there is no such vo
+	 * @throws GroupNotExistsException if there is no such query group
+	 * @throws PrivilegeException insufficient permission
+	 */
+	Paginated<RichMember> getMembersPage(PerunSession sess, Vo vo, MembersPageQuery query, List<String> attrNames, String policy) throws VoNotExistsException, PrivilegeException, GroupNotExistsException, PolicyNotExistsException;
```
xflord, line 1525: Just a question, as I am not too aware of the task's context. Do we need the policy argument? As I understand, the default policy (`filter-getMembersPage_policy`) should be the thing that determines how members are filtered, and admins of each instance can configure this as they please.

xflord (commented):
Few code separation details that I noticed. Otherwise looks like the best possible solution. Good job with the tests

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
mattjoke: The method `isVoAdmin` does not check if the user is VoAdmin for concrete VO. It only checks the session, if the user has the role but nothing else.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
mattjoke: I used the policy argument for easier handling of the tests. It is not strictly needed for the task, but it makes the tests more accessible and easier to use.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
mattjoke: Well, yes and no. The `roles.isEmpty()` is dependent on the outcome of the loop. There would be a kind of duplicity there. I mean, both of the solutions are not ideal.

perun-core/src/main/java/cz/metacentrum/perun/core/api/MembersManager.java
xflord: I see, it's fine then

xflord (approved):
Really good job

PR #4022, opened 2023-07-26
feat(core): remove not allowed roles

sarkapalkovicova:
Depending on where the request comes from, remove roles that are not allowed.

### DIFF
--- a/perun-base/src/main/resources/perun-base.xml
+++ b/perun-base/src/main/resources/perun-base.xml
@@ -82,6 +82,7 @@
 		<property name="mfaAuthTimeout" value="${perun.introspectionEndpoint.mfaAuthTimeout}"/>
 		<property name="mfaAuthTimeoutPercentageForceLogIn" value="${perun.introspectionEndpoint.mfaAuthTimeoutPercentageForceLogIn}"/>
 		<property name="enforceMfa" value="${perun.enforceMfa}"/>
+		<property name="appAllowedRoles" value="#{'${perun.appAllowedRoles.apps}'.split('\s*,\s*')}"/>
 		<property name="introspectionEndpointMfaAcrValue" value="${perun.introspectionEndpoint.mfaAcrValue}"/>
 		<property name="idpLoginValidity" value="${perun.idpLoginValidity}"/>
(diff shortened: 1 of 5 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/Api.java
```diff
@@ -225,6 +225,8 @@ private static PerunPrincipal setupPerunPrincipal(HttpServletRequest req, Deseri
 		int extSourceLoa;
 		Map<String, String> additionalInformations = new HashMap<>();
 
+		String referer = req.getHeader("Referer");
```
zlamalp, line 228: I'm not really sure what this logic actually suppose to do, but HTTP request headers are set by client. Hence I can forge a request using cURL with specific `Referer` value, even it if originates from my script. If this suppose to somehow override the problems with MFA, I would be cautious.

sarkapalkovicova: According to the referer header, this code removes roles from a session that are not defined as allowed in the config. It does not skip MFA but only limits user's rights. So even if a user fakes the request, there should be no security risk.

bodnara (commented):
Should this PR also contain development note about the new config property or is it not necessary.

perun-base/src/main/resources/perun-base.xml
```diff
@@ -82,6 +82,7 @@
 		<property name="mfaAuthTimeout" value="${perun.introspectionEndpoint.mfaAuthTimeout}"/>
 		<property name="mfaAuthTimeoutPercentageForceLogIn" value="${perun.introspectionEndpoint.mfaAuthTimeoutPercentageForceLogIn}"/>
 		<property name="enforceMfa" value="${perun.enforceMfa}"/>
+		<property name="appAllowedRoles" value="#{'${perun.appAllowedRoles.apps}'.split('\s*,\s*')}"/>
```
xflord, line 85: Looks like a good solution, just please include deployment notes related to the new header and new config property (an explanation of its structure would be nice)

bodnara (approved):

xflord (approved):

PR #4023, opened 2023-08-01
ci: update renovate config

melanger:
* use predefined presets instead of packageRules
* enable vulnerability alerts
* enable lock file maintenance
* enable workarounds


melanger:
PRs by dependabot should be disabled (but keep GitHub vulnerability alters enabled, renovate will read them)

HejdaJakub (approved):

PR #4025, opened 2023-08-01
feat(core): allow to set attribute action as globally critical

HejdaJakub:
* We used to have only the option to set attribute action as critical only for critical objects.
* Now we have an option to set any attribute definition as globally critical, so it will not depend on the related objects.

BREAKING CHANGE: column 'global' was added to the attribute_critical_actions table 
Database changelog:
ALTER TABLE attribute_critical_actions ADD COLUMN global boolean default false not null; UPDATE configurations SET value='3.2.16' WHERE property='DATABASE VERSION';


perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4178,13 +4188,20 @@ public List<AttributeAction> getCriticalAttributeActions(PerunSession sess, int
...
+				if (isAttributeActionCritical(sess, attr, action) && !globalCriticalityChanged) {
 					throw new RelationExistsException("Attribute " + attr.getName() + " is already critical on " + action + " action.");
 				}
-				jdbc.update("insert into attribute_critical_actions (attr_id, action) values (?,?::attribute_action)", attr.getId(), action.toString());
+
+				if (isAttributeActionCritical(sess, attr, action)) {
+					jdbc.update("update attribute_critical_actions set global=? where attr_id=? and action=?::attribute_action", global, attr.getId(), action.toString());
+				} else {
+					jdbc.update("insert into attribute_critical_actions (attr_id, action, global) values (?,?::attribute_action,?)", attr.getId(), action.toString(), global);
+				}
```
xflord, line 4204: Nitpick - this could be done with one SQL statement  (upsert)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -1697,7 +1697,9 @@ public static boolean isMfaAuthorizedForAttribute(PerunSession sess, AttributeDe
 			return principalMfa || updatePrincipalMfa(sess);
 		}
 
-		return principalMfa || !isAnyObjectMfaCritical(sess, objects) || updatePrincipalMfa(sess);
+		boolean globallyCriticalAction  = ((PerunBl) sess.getPerun()).getAttributesManagerBl().isAttributeActionGloballyCritical(sess, attrDef, actionType);
+
+		return principalMfa || !isAnyObjectMfaCritical(sess, objects) || !globallyCriticalAction || updatePrincipalMfa(sess);
```
xflord, line 1702: If I understand correctly, we want to completely skip the critical object check when the action is globally critical, no? This way `!isAnyObjectMfaCritical(sess, objects)` would return true if no objects were critical and we would not check global criticality.

perun-openapi/openapi.yml
```diff
@@ -389,9 +389,14 @@ components:
...
-          items:
-            $ref: '#/components/schemas/AttributeAction'
+          type: object
+          description: "Represents types of actions (READ, WRITE) that users can perform upon attributes and if they are critical globally or not."
+          additionalProperties:
+            type: boolean
+          example: {
+            READ: false,
+            WRITE: true,
+          }
```
bodnara, line 399: Is it possible for the key to still be `AttributeAction` ? This way we lose the type on FE.

HejdaJakub: Unfortunately in our current version of open API, it is probably not possible to define an enum for the key in a map (https://stackoverflow.com/questions/46552863/how-to-write-openapi-3-swagger-specification-for-property-name-in-map-object/46553255#46553255). I'll discuss with others if we can upgrade this version without any conflicts.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
HejdaJakub: You're right. I think that the improved condition should resolve this problem.

xflord (approved):
Fine by me

bodnara (approved):

perun-openapi/openapi.yml
bodnara: Definitely not a block for this PR, just a minor inconvenience at worst.

HejdaJakub:
After rebasing the master I needed to change the DB version from `3.2.16` to `3.2.17` (also in the commit message for deployment).

PR #4028, opened 2023-08-02
fix(core): service account skip MFA

sarkapalkovicova:
Skip MFA when user is service user.


xflord (approved):

mattjoke (approved):

mattjoke (changes requested):
I found a bug, but I'm unsure how it relates to this problem. Whenever the MFA is turned on, I cannot log in (which is what we want, in a sense). But the problem is that the GUI would then throw an unmatched exception error. Looking at logs I can see that the error arises in checking for validity for MFA (which service account does not have).

mattjoke (approved):
Tested and appears to be working as it should.

PR #4030, opened 2023-08-03
fix(core): proper fix for group structure sync

zlamalp:
- Removed old code introduced by a quickfix for the group struc sync.


xflord (approved):

zlamalp:
Will be handled at GitLab. Closing it here.

PR #4031, opened 2023-08-04
feat(core): attribute modul for mfaEnforceSettings

bodnara:
* Added semantic and syntactic checks

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_mfaEnforceSettings.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_mfaEnforceSettings.java
@@ -1,0 +1,190 @@
+package cz.metacentrum.perun.core.impl.modules.attributes;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import cz.metacentrum.perun.core.api.Attribute;
+import cz.metacentrum.perun.core.api.User;
+import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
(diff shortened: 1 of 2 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_mfaEnforceSettings.java
```diff
@@ -0,0 +1,188 @@
...
+		try {
+			final ObjectMapper mapper = new ObjectMapper();
+			JsonNode mfaCategoriesNode = mapper.readTree(mfaSettingsValue).get("categories");
+
+			// Iterate through categories and check that all included categories exist
+			for (Iterator<Map.Entry<String, JsonNode>> catIt = mfaCategoriesNode.fields(); catIt.hasNext(); ) {
+				Map.Entry<String, JsonNode> catEntry = catIt.next();
+				includeCategories.remove(catEntry.getKey());
+
+				// Iterate through rps and check that all excluded rps exist
```
xflord, line 156: We also want to check that the excluded rps is from one of the included categories (would not make sense to exclude an rps from a category that is not included). 99% sure about this, but check what the mfa API used in the profile app returns in this scenario

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_def_mfaEnforceSettings.java
```diff
@@ -0,0 +1,188 @@
...
+		if (val == null || val.isEmpty()) return;
+
+		Set<String> includeCategories = null;
+		Set<String> excludeRps = null;
+		try {
+			final ObjectMapper mapper = new ObjectMapper();
+			JsonNode root = mapper.readTree(val);
+
+			// Syntax check ensures correctness for 'all'
+			if (root.has("all")) return;
```
xflord, line 130: Nitpick - the check for `true` should theoretically be here since we're checking the semantics

PR #4037, opened 2023-08-14
feat(core): free logins when deleting login namespace attribute

bodnara:
* blocked and reserved logins are freed for given namespace

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
@@ -2590,8 +2590,16 @@ public class AttributesManagerBlImpl implements AttributesManagerBl {
 		//Remove services' required attributes
 		//TODO
 
-		//Remove attribute and all it's values
+		// Remove all values
 		this.deleteAllAttributeAuthz(sess, attribute);
+		// Free logins for login-namespace attribute
+		if (attribute.getBaseFriendlyName().equalsIgnoreCase("login-namespace")) {
+			// Free blocked logins
+			getPerunBl().getUsersManagerBl().unblockLoginsForNamespace(sess, attribute.getFriendlyNameParameter());
+			// Free reserved logins
+			getPerunBl().getUsersManagerBl().deleteReservedLoginsForNamespace(sess, attribute.getFriendlyNameParameter());
+		}
+		// Delete attribute
 		getAttributesManagerImpl().deleteAttribute(sess, attribute);
 		getPerunBl().getAuditer().log(sess, new AttributeDeleted(attribute));
 
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTest.java
@@ -11,6 +11,7 @@ import cz.metacentrum.perun.core.api.AttributePolicyCollection;
 import cz.metacentrum.perun.core.api.AttributeRights;
 import cz.metacentrum.perun.core.api.AttributesManager;
 import cz.metacentrum.perun.core.api.BeansUtils;
+import cz.metacentrum.perun.core.api.BlockedLogin;
 import cz.metacentrum.perun.core.api.Candidate;
 import cz.metacentrum.perun.core.api.ExtSource;
 import cz.metacentrum.perun.core.api.ExtSourcesManager;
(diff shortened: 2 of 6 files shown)


dBucik:
I am not sure this is a thing which should be done. Logins should stay blocked permanently IMO. Just imagine that after some time the namespace is re-introduced. You might argue that users will no longer have previous logins, but imagine somebody assigns a previously used login. That could mean they potentially get access to places where previously the login had access.

Instead, I would prefer if all of the active logins are blocked and stored somewhere, so we prevent running into the mentioned issue. 

But perhaps my opinion is a bit overkill? Did you have some discussion about such implications?

perun-core/src/test/java/cz/metacentrum/perun/core/entry/AttributesManagerEntryIntegrationTest.java
```diff
@@ -5415,6 +5416,29 @@ public void deleteAttribute() throws Exception {
...
+		BlockedLogin del = new BlockedLogin("login1", deleteNamespace);
+		perun.getUsersManagerBl().blockLogins(sess, List.of(del.getLogin()), deleteNamespace, null);
+		String otherNamespace = "other";
+		BlockedLogin other = new BlockedLogin("login2", "other");
+		perun.getUsersManagerBl().blockLogins(sess, List.of(other.getLogin()), otherNamespace, null);
+		System.out.println(perun.getUsersManagerBl().getAllBlockedLoginsInNamespaces(sess));
+
+		AttributeDefinition attrDef = new AttributeDefinition();
+		attrDef.setFriendlyName("login-namespace:deleteNamespace");
+		attrDef.setNamespace("deleteNamespace");
```
xflord, line 5433:
```suggestion
		attrDef.setNamespace("urn:perun:user:attribute-def:def");
```
different kind of namespace hehe (this is also why the tests fail)

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -2590,8 +2590,16 @@ public void deleteAttribute(PerunSession sess, AttributeDefinition attribute) {
 		//Remove services' required attributes
 		//TODO
 
-		//Remove attribute and all it's values
+		// Remove all values
 		this.deleteAllAttributeAuthz(sess, attribute);
+		// Free logins for login-namespace attribute
+		if (attribute.getBaseFriendlyName().equalsIgnoreCase("login-namespace")) {
+			// Free blocked logins
+			getPerunBl().getUsersManagerBl().unblockLoginsForNamespace(sess, attribute.getNamespace());
```
xflord, line 2598:
```suggestion
			getPerunBl().getUsersManagerBl().unblockLoginsForNamespace(sess, attribute.getFriendlyNameParameter());
```

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AttributesManagerBlImpl.java
```diff
@@ -2590,8 +2590,16 @@ public void deleteAttribute(PerunSession sess, AttributeDefinition attribute) {
...
+			// Free reserved logins
+			getPerunBl().getUsersManagerBl().deleteReservedLoginsForNamespace(sess, attribute.getNamespace());
```
xflord, line 2600:
```suggestion
			getPerunBl().getUsersManagerBl().deleteReservedLoginsForNamespace(sess, attribute.getFriendlyNameParameter);
```

xflord (changes requested):

bodnara:
> I am not sure this is a thing which should be done. Logins should stay blocked permanently IMO. Just imagine that after some time the namespace is re-introduced. You might argue that users will no longer have previous logins, but imagine somebody assigns a previously used login. That could mean they potentially get access to places where previously the login had access.
> 
> Instead, I would prefer if all of the active logins are blocked and stored somewhere, so we prevent running into the mentioned issue.
> 
> But perhaps my opinion is a bit overkill? Did you have some discussion about such implications?

I don't remember a discussion about the idea of re-introducing a namespace. I think this is a good point but can't really comment about what's best. This is mostly on @balcirakpeter.

xflord (approved):

balcirakpeter:
@dBucik my thoughts... up to discussion of course.
- Deleting an attribute definition can be done only by Perun administrators, that should be aware of the consequences. Moreover, they are properly informed that by deleting the definition, all its values are deleted as well.
- The perun administrator can delete all values even from the blocked logins table, so it would not protect us from re-using the same logins; it would just move values to another place.
- If we do not want to delete all values, we probably do not want to delete the namespace at all. We can just keep it and don't use it.
- I am afraid that if we would want to re-introduce a namespace, only keeping old values blocked would not be enough. We would probably want to re-assign old values to existing users.
- Deleting a login namespace attribute definition (or any other definition or object) should be done only if we know that the definition is not used by any service. If there is a service that relies on the attribute, the definition should not be deleted.

HejdaJakub (approved):

PR #4038, opened 2023-08-14
fix(core): handle missing referer header

zlamalp:
- If missing in source request read value is NULL and not empty, hence creating matcher on the pattern fails.
- Perform matching only if there is non-empty input.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
@@ -2599,10 +2599,12 @@ public class AuthzResolverBlImpl implements AuthzResolverBl {
 		Map<String, List<String>> appAllowedRoles = BeansUtils.getCoreConfig().getAppAllowedRoles();
 		for (String reg : appAllowedRoles.keySet()) {
 			Pattern pattern = Pattern.compile(reg);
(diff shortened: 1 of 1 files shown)


bodnara (approved):

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
```diff
@@ -80,6 +80,7 @@
 import cz.metacentrum.perun.core.impl.Utils;
 import cz.metacentrum.perun.core.implApi.AuthzResolverImplApi;
 import cz.metacentrum.perun.registrar.model.Application;
+import org.apache.commons.lang3.StringUtils;
```
HejdaJakub, line 83: NITPICK: We usually want to import the smallest part of a package and then use these methods directly without the prefix `StringUtils.`.
```suggestion
import org.apache.commons.lang3.StringUtils.isBlank;
```

HejdaJakub (commented):
Otherwise looks good.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/AuthzResolverBlImpl.java
zlamalp: Should be fixed

PR #4039, opened 2023-08-15
feat(core): restrict deletion of the attribute definition

HejdaJakub:
* We want to restrict the deletion of the attribute definition if this attribute is required for any service.
* We want to restrict it also for the case when this attribute is a source or a destination attribute of any application form item in some application form.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
@@ -1730,10 +1730,11 @@ public interface AttributesManager {
 	 * <p>
 	 * PRIVILEGE: Only PerunAdmin can delete existing attribute.
 	 *
-	 * @throws InternalErrorException if an exception raise in concrete implementation, the exception is wrapped in InternalErrorException
-	 * @throws PrivilegeException     if privileges are not given
+	 * @throws InternalErrorException 	if an exception raise in concrete implementation, the exception is wrapped in InternalErrorException
+	 * @throws PrivilegeException     	if privileges are not given
+	 * @throws RelationExistsException 	if attribute definition has any relation to some application form item or to some service as a required attribute
 	 */
-	void deleteAttribute(PerunSession sess, AttributeDefinition attributeDefinition) throws PrivilegeException, AttributeNotExistsException;
+	void deleteAttribute(PerunSession sess, AttributeDefinition attributeDefinition) throws PrivilegeException, AttributeNotExistsException, RelationExistsException;
 
 	/**
 	 * Deletes the attribute.
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
@@ -65,6 +65,8 @@ import cz.metacentrum.perun.core.implApi.modules.attributes.UserFacilityAttribut
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserFacilityVirtualAttributesModuleImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.UserVirtualAttributesModuleImplApi;
 import cz.metacentrum.perun.core.implApi.modules.attributes.VoAttributesModuleImplApi;
+import cz.metacentrum.perun.registrar.model.ApplicationForm;
+import cz.metacentrum.perun.registrar.model.ApplicationFormItem;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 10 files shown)


xflord (commented):
Also please fix the tests that now fail

perun-core/src/main/java/cz/metacentrum/perun/core/api/AttributesManager.java
```diff
@@ -1733,7 +1733,7 @@ Map<String, Attribute> getEntitylessAttributesWithKeys(PerunSession sess, String
 	 * @throws InternalErrorException if an exception raise in concrete implementation, the exception is wrapped in InternalErrorException
 	 * @throws PrivilegeException     if privileges are not given
 	 */
-	void deleteAttribute(PerunSession sess, AttributeDefinition attributeDefinition) throws PrivilegeException, AttributeNotExistsException;
+	void deleteAttribute(PerunSession sess, AttributeDefinition attributeDefinition) throws PrivilegeException, AttributeNotExistsException, RelationExistsException;
```
sarkapalkovicova, line 1736: Maybe should be `RelationExistsException` also mentioned in `AttributesManagerMethod.java` in the doc for this method?

HejdaJakub:
> Also please fix the tests that now fail

I am not sure what the strong/inverse/all dependencies are in the context of these failing tests, so I am not sure how to properly fix it. :disappointed:  But these tests are saying that deleted attribute is required for any service, so maybe it is correct that these test doesn't work and should be marked as `@Ignore` or deleted now?

HejdaJakub:
I also tried to improve error messages for better readability (I used to paste long objects into this message):
Example of error for service required attribute:
`Attribute definition with id: 3584 is the required attribute for these services: [AttributesManagerTestService (id: 862), AttributesManagerTestService2 (id: 863)]`

Example of error for app from items relation (first for group form, second for vo form):
`Attribute definition with id: 3589 is source or destination attribute in these application forms: [Application form in Group with id: 9645 (under Vo with id: 3323), Application form in Vo with id: 3323]`

HejdaJakub:
I tried to improve the error message for the app form item relations use-case again. Now the message contains also `shortname` of the app form items that are related to the attribute. There is also one new check related to the hidden/disabled dependencies. If e.g. `item1` has a dependency on `item2` with source/destination attribute to delete, after removing source/destination attribute from `item2` there won't be any blocking relation, but we want at least inform the user in the error message, that this `item1` has also some weak relation to the attribute which should be managed (because it can break application form logic). The error message now doesn't handle each use case (source / destination / hidden or disabled dependency) separately, rather just informs generally that items in the message have some relation.


`Attribute definition with id: 3589 has a relation (as a source/destination/dependency) to these application form items: [Application form items: [name, displayName] in the application form in Group with id: 9645 (under Vo with id: 3323), Application form items: [shortname, displayName] in the application form in Group with id: 9646 (under Vo with id: 3323)]`

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4215,6 +4240,44 @@ public List<String> getAllNamespaces(PerunSession sess) {
 		return jdbc.query("SELECT friendly_name FROM attr_names WHERE friendly_name LIKE 'login-namespace:%%' AND attr_name NOT LIKE '%%def:virt%%'", ATTRIBUTE_FRIENDLY_NAMES_MAPPER);
 	}
 
+	@Override
+	public List<ApplicationForm> getAppFormsWhereAttributeRelated(PerunSession sess, AttributeDefinition attr) {
+		try {
+			String urn = attr.getNamespace() + ":" + attr.getFriendlyName();
```
xflord, line 4246: just realized there's a method for this (`attr.getName()`)

perun-core/src/main/java/cz/metacentrum/perun/core/impl/AttributesManagerImpl.java
```diff
@@ -4215,6 +4240,44 @@ public List<String> getAllNamespaces(PerunSession sess) {
...
+			return new ArrayList<>();
+		} catch (RuntimeException ex) {
+			throw new InternalErrorException(ex);
+		}
+	}
+
+	@Override
+	public List<String> getAppFormItemsForAppFormAndAttribute(PerunSession sess, int appFormId, AttributeDefinition attr) {
+		try {
+			String urn = attr.getNamespace() + ":" + attr.getFriendlyName();
```
xflord, line 4260: same here

xflord:
> There is also one new check related to the hidden/disabled dependencies. If e.g. `item1` has a dependency on `item2` with source/destination attribute to delete, after removing source/destination attribute from `item2` there won't be any blocking relation, but we want at least inform the user in the error message, that this `item1` has also some weak relation to the attribute which should be managed (because it can break application form logic). The error message now doesn't handle each use case (source / destination / hidden or disabled dependency) separately, rather just informs generally that items in the message have some relation.

This to me feels a bit confusing. I feel like the exception should only include the blocking relations, any warnings related to removing form items should be displayed WHEN removing form items (maybe we should add a warning when removing form items that other items have a dependency on?)

sarkapalkovicova (approved):

HejdaJakub:
> This to me feels a bit confusing. I feel like the exception should only include the blocking relations, any warnings related to removing form items should be displayed WHEN removing form items (maybe we should add a warning when removing form items that other items have a dependency on?)

It makes sense to me, so I removed this warning and moved it to the front end for better UX. 

Just FYI - at the front end part, I am checking the deletion of an app item and I also warn a user when he/she tries to change the source or destination attribute of an item that is marked as a dependency of another item. This should cover all main use cases related to this issue IMHO.

bodnara (approved):

sarkapalkovicova (approved):

PR #4040, opened 2023-08-15
fix(core): check open applications

sarkapalkovicova:
Do not allow change of the destination attribute when open applications exist.


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1150,6 +1154,11 @@ public int updateFormItems(PerunSession sess, ApplicationForm form, List<Applica
 
 			// else update form item
 
+			// Check that if the destination attribute was changed then there are no open applications
+			if (!Objects.equals(getFormItemById(sess, item.getId()).getPerunDestinationAttribute(), item.getPerunDestinationAttribute()) && !existingOpenApplications.isEmpty()) {
+				throw new OpenApplicationExistsException("It is impossible to change the destination attribute of the form item while some open applications still exist. First, please resolve these applications.");
+			}
```
bodnara, line 1160: I think this check should happen before deletion of the item.

xflord: This should not matter since the whole transaction gets rollbacked when an exception is thrown (but for readability's sake it could be before I guess)

xflord (approved):

bodnara (approved):

PR #4041, opened 2023-08-16
feat(core): added new role

mattjoke:
* new role for basic auth accounts
* role skips all MFA checks

DEPLOYMENT NOTE: added new role 'PERUNADMINBA'


perun-base/src/main/resources/perun-roles.yml
```diff
@@ -8181,13 +8186,33 @@ perun_roles_management:
       - PERUNADMIN:
```
xflord, line 8186: Theoretically, PERUNADMINBA should be able to set PERUNADMIN, no?

perun-base/src/main/resources/perun-roles.yml
```diff
@@ -8181,13 +8186,33 @@ perun_roles_management:
...
+      - PERUNADMINBA:
+    privileged_roles_to_read:
+      - PERUNADMIN:
+      - PERUNADMINBA:
+      - PERUNOBSERVER:
+    associated_read_roles:
+      - PERUNOBSERVER
+    assignable_to_attributes: false
+    mfa_critical_role: false
+    display_name: "Perun admin BA"
```
xflord, line 8214:
```suggestion
    display_name: "Perun admin BA"
    skip_mfa: true
```
this is missing in order to skip MFA when performing critical actions

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -1289,7 +1289,7 @@ public static void unsetRole(PerunSession sess, User user, PerunBean complementa
 
 		try {
 			if(!authorizedToManageRole(sess, complementaryObject, role)) {
-				throw new PrivilegeException("You are not privileged to use the method unsetRole.");
+				throw new PrivilegeException("You are not privileged to use the method unsetRole. "+authorizedToManageRole(sess, complementaryObject, role));
```
xflord, line 1292: Why is this here

xflord (changes requested):
Again, I would utilize the `isPerunAdmin()` method to reduce the explicit checks for PERUNADMINBA.

Also, add the changes to the perun-roles template in the ansible repo https://github.com/CESNET/ansible-role-perun-docker-server

perun-base/src/main/resources/perun-roles.yml
mattjoke: I think this is not what we want because PERUNADMINBA is supposed to be just an internal role, which will be added manually through the database.

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
mattjoke: Good catch, just a leftover from debugging

mattjoke:
I replaced role checks with `isAdmin()`. Nevertheless, most of the occurrences have explicit checks (isAuthorized, hasRole), which cannot be replaced with `isAdmin()`

perun-core/src/main/java/cz/metacentrum/perun/core/entry/UsersManagerEntry.java
```diff
@@ -376,7 +376,7 @@ public User createUser(PerunSession sess, User user) throws PrivilegeException {
 		Utils.notNull(user, "user");
 
 		// Authorization
-		if(!AuthzResolver.isAuthorized(sess, Role.PERUNADMIN)) {
+		if(!(AuthzResolver.isAuthorized(sess, Role.PERUNADMIN) || AuthzResolver.isAuthorized(sess, Role.PERUNADMINBA))) {
```
xflord, line 379: No need for this - this form of isAuthorized is deprecated

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -579,7 +579,7 @@ public void sendMessage(PerunSession sess, Application app, MailType mailType, S
 		if (mail == null) throw new RegistrarException("Notification template for "+mailType+" is not defined.");
 		if (!mail.getSend()) throw new RegistrarException("Sending of notification "+mailType+" is disabled.");
 
-		if (!AuthzResolver.hasRole(sess.getPerunPrincipal(), Role.PERUNADMIN)) {
+		if (!(AuthzResolver.hasRole(sess.getPerunPrincipal(), Role.PERUNADMIN) || AuthzResolver.hasRole(sess.getPerunPrincipal(), Role.PERUNADMINBA))) {
```
xflord, line 582: This call effectively does the same as isPerunAdmin()

xflord (commented):
I also noticed that the `isUserPerunAdmin()` exists, even though it is not used anywhere. Maybe we can mark it as deprecated to avoid misusage (since it only checks for PERUNADMIN). What do you think?

perun-base/src/main/resources/perun-roles.yml
xflord: I know that, but as an internal "super user" account, I don't see why it shouldn't be allowed to set PERUNADMIN. But I guess it's not a big deal.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
mattjoke: Well, I think it is different because it does throw the error if the role does not exist in the Perun. `isPerunAdmin()` does not check if the role exists.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/SecurityTeamsManagerBlImpl.java
```diff
@@ -52,7 +52,7 @@ public SecurityTeamsManagerBlImpl(SecurityTeamsManagerImplApi securityTeamsManag
 
 	@Override
 	public List<SecurityTeam> getSecurityTeams(PerunSession sess) {
-		if (AuthzResolverBlImpl.hasRole(sess.getPerunPrincipal(), Role.PERUNADMIN)) {
+		if (AuthzResolverBlImpl.hasRole(sess.getPerunPrincipal(), Role.PERUNADMIN) || AuthzResolverBlImpl.hasRole(sess.getPerunPrincipal(), Role.PERUNADMINBA)) {
```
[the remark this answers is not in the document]
mattjoke: Same here; the check throws an error, if the role is not in Perun.

bodnara (approved):
Should be ok like this.

sarkapalkovicova (approved):
I tested it and did not find any issues.

PR #4052, opened 2023-08-23
feat: use journald instead of syslog in perun-auditlogger

zlamalp:
Removed all usage and references to syslog, since it was replaced with journald logging.

BREAKING CHANGE: Auditlogger no longer writes audit messages to the syslog. All configuration related to usage of syslog is ignored and can be removed from /etc/perun/perun-auditlogger and /etc/perun/perun-auditlogger.properties. Make sure journald is present and configured on the machine before deploying.


bodnara (commented):
Looks ok, but I can't really comment if the configuration is correct or not.

xflord (approved):

PR #4053, opened 2023-08-23
feat: 🎸 Cascade to parent gr. when deciding gr. TO recipients

dBucik:
When sending out notification to managers or configured TO recipients for groups, try to lookup recipients in group hierarchy instead of using the VO level directly. This means that if the group does not have "TO" set and there is noone to send notification to in corresponding role, the lookup will be tried one level higher in the group hierarchy, cascading to the top level group and only then to VO level.

DEPLOYMENT NOTE: Changed behaviour might cause sending notifications to managers or configured TO recipients in parent group rather than to VO.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
@@ -1,10 +1,21 @@
 package cz.metacentrum.perun.registrar.impl;
 
-import java.io.*;
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
 import java.math.BigInteger;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -1091,66 +1086,54 @@ private void setFromAndReplyTo(MimeMessage message, String fromName, String from
...
+			do {
+				result = getToEmailsGroupLevel(currentGroup);
+				if (result.isEmpty()) {
+					try {
+						currentGroup = groupsManager.getGroupById(registrarSession, currentGroup.getParentGroupId());
+					} catch (GroupNotExistsException e) {
+						log.error("[MAIL MANAGER] Inconsistency detected - parent (in hierarchy) group with ID '{}' of group '{}' does not exist",
+							currentGroup.getParentGroupId(), app.getGroup());
+						break;
+					}
```
bodnara, line 1116: I think this does not have to be inconsistency, since there is no check if the group has a parent.

dBucik: I have added a check for null before the call. What I meant by the log is that although `currentGroup` has got `parentGroupId`, a group for such ID could not be found, which IMHO means there is some inconsistency.

bodnara: Yes with the check in place this does make sense. However I believe that now if the result is not empty for first group we get an infinite loop. Maybe we could change the condition to `!result.isEmpty()` since the current one can never actually end the loop. It only breaks through the exception.

dBucik: You are right, well spotted :) will add a the clause into the loop condition.

bodnara (approved):
Fine by me.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -2529,45 +2516,90 @@ private void setHtmlMessageWithAltPlainTextMessage (MimeMessage message, String
...
+	private Set<String> getToEmailsFromAttribute(Attribute toEmailAttr) {
 		// if there are any pre-defined VO emailTo attributes, use them
-		if (attrToEmail != null && attrToEmail.getValue() != null) {
-			ArrayList<String> value = attrToEmail.valueAsList();
+		Set<String> emails = new HashSet<>();
+		if (toEmailAttr != null && toEmailAttr.getValue() != null) {
+			ArrayList<String> value = toEmailAttr.valueAsList();
 			for (String adr : value) {
-				if (adr != null && !adr.isEmpty()) {
+				if (StringUtils.hasText(adr)) {
```
xflord, line 2571: The attribute module should probably do this check. It is this way for VO toEmail/fromEmail attrs, toEmail group attribute module is missing for some reason.
You can consider adding it but I'm okay with leaving it as it is (since it's registrar)

xflord (commented):
Just nitpicks, looks okay otherwise

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -1091,69 +1086,61 @@ private void setFromAndReplyTo(MimeMessage message, String fromName, String from
...
+		// no recipients found for Group and Vo, use backup
+		if (result.isEmpty()) {
 			result.add(getPropertyFromConfiguration("backupTo"));
 		}
-
 		return result;
 	}
 
+	private Group getParentGroupForMailAddresses(Group group) {
+		if (group != null && group.getParentGroupId() != null) {
```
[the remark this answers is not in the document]
dBucik: I don't think members should be checked specifically. IMHO:
1) members group cannot have subgroups
2) members group is just a representation of VO, so technically we are interested in VO settings instead. 

But if you wish, I can modify it.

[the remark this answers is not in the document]
xflord: What I meant is that `getParentGroup()` returns the `members` group when the passed group does not have a parent.
It would be more confusing probably, leave it like this.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
```diff
@@ -1091,69 +1130,60 @@ private void setFromAndReplyTo(MimeMessage message, String fromName, String from
...
-				for (String role : rolesToNotify) {
-					admins.addAll(AuthzResolverBlImpl.getRichAdmins(registrarSession, app.getGroup(), role));
-				}
-				result.addAll(getUserPreferredMails(registrarSession, admins));
-				result = result.stream().distinct().toList();
-
-				if (!result.isEmpty()) {
-					return result;
+			// Check group first
+			if (app.getGroup() != null) {
```
xflord, line 1146: The if is unnecessary, the loop condition does the job.

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/MailManagerImpl.java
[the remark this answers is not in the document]
dBucik: Ah, got it!

xflord (approved):

PR #4057, opened 2023-08-25
fix(core): richgroup is not supported

mattjoke:
* 'RichGroup is not supported' error when getting rich group data from database


HejdaJakub:
It would be maybe better not to send RichGroups from GUI (that has been caused by some recent change to entity storage) or at least edit the comments/types in GUI (which looks like the Group type is presented and it isn't true now). On the other hand, there is no restriction or definition in the comment in RPC nor open API of allowed PerunBean objects, which should be correct to define when it can cause bugs like this one.

But as a quick fix, this change is also acceptable and it works properly.

HejdaJakub (approved):

PR #4060, opened 2023-08-28
feat: 🎸 Allow multiple reg. modules to be configured

dBucik:
Instead of specifying single registrar module, allow to specify multiple. When storign in database, specified names are concatenated using `,` as separator. In from handling (any), instead of operating with single module, loop over all specified modules and try to process appropriate action (e.g. call `beforeApprove(...)`). Also updates Perl libs.

BREAKING CHANGE: 🧨 ApplicationForm bean property `moduleClassName` replaced with `moduleClassNames`. Type has changed from String to List<String>. Includes database version update and column `module_name` of `application_form` table being renamed to `module_names`.

DEPLOYMENT NOTE: requires database update. UI version have to work with updated model of ApplicationForm (`moduleClassName` replaced with field `moduleClassNames`).

### DIFF
--- a/perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationForm.java
+++ b/perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationForm.java
@@ -2,6 +2,10 @@ package cz.metacentrum.perun.registrar.model;
 
 import cz.metacentrum.perun.core.api.Group;
 import cz.metacentrum.perun.core.api.Vo;
+import org.springframework.util.StringUtils;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Application form of a VO. Use {@link cz.metacentrum.perun.registrar.RegistrarManager#getFormItems} for items.
@@ -16,7 +20,7 @@ public class ApplicationForm {
 	private boolean automaticApproval;
 	private boolean automaticApprovalExtension;
 	private boolean automaticApprovalEmbedded;
-	private String moduleClassName;
+	private final List<String> moduleClassNames = new ArrayList<>();
 
 	public ApplicationForm() {
 	}
@@ -69,12 +73,29 @@ public class ApplicationForm {
 		this.automaticApprovalEmbedded = automaticApprovalEmbedded;
 	}
 
-	public String getModuleClassName() {
-		return moduleClassName;
+	public List<String> getModuleClassNames() {
+		return new ArrayList<>(moduleClassNames);
 	}
 
-	public void setModuleClassName(String moduleClassName) {
-		this.moduleClassName = moduleClassName;
+	public void setModuleClassNames(List<String> moduleClassNames) {
+		this.moduleClassNames.clear();
+		for (String moduleClassName : moduleClassNames) {
+			if (StringUtils.hasText(moduleClassName)) {
+				this.moduleClassNames.add(moduleClassName);
+			}
+		}
+	}
+
+	public void addModuleClassName(String moduleClassName) {
+		if (StringUtils.hasText(moduleClassName)) {
+			this.moduleClassNames.add(moduleClassName);
+		}
+	}
+
+	public void removeModuleClassName(String moduleClassName) {
+		if (StringUtils.hasText(moduleClassName)) {
+			this.moduleClassNames.remove(moduleClassName);
+		}
 	}
 
 	/**
@@ -94,7 +115,7 @@ public class ApplicationForm {
 			", group='" + getGroup() + '\'' +
 			", automaticApproval='" + isAutomaticApproval() + '\'' +
 			", automaticApprovalExtension='" + isAutomaticApprovalExtension() + '\'' +
-			", moduleClassName='" + getModuleClassName() + '\'' +
+			", moduleClassNames='" + getModuleClassNames() + '\'' +
 			"]";
 	}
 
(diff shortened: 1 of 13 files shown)


perun-base/src/test/resources/test-schema.sql
```diff
@@ -1908,7 +1908,7 @@ create index idx_fk_attr_critops ON attribute_critical_actions(attr_id);
 create index app_state_idx ON application (state);
 
 -- set initial Perun DB version
-insert into configurations values ('DATABASE VERSION','3.2.17');
+insert into configurations values ('DATABASE VERSION','3.2.18');
```
xflord, line 1911: top of the file as well please

xflord (commented):
Seems fine. When it comes to the new GUI app, please either create a new task for the changes or make them yourself (should be a few lines)

xflord (approved):

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -4192,34 +4213,33 @@ private void setIfNotEmpty(Map<String, String> map, String value, String key) {
...
-				return module;
+			for (String regModule : form.getModuleClassNames()) {
+				try {
+					log.debug("[REGISTRAR] Attempting to instantiate class: {}", MODULE_PACKAGE_PATH + regModule);
+					module = (RegistrarModule) Class.forName(MODULE_PACKAGE_PATH + regModule)
+						.getDeclaredConstructor().newInstance();
+					module.setRegistrar(registrarManager);
+				} catch (Exception ex) {
+					log.error("[REGISTRAR] Exception when instantiating module.", ex);
+					modules.add(module);
```
xflord, line 4236: In this case we add a null module, no? And twice at that (couple lines below)

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -4192,34 +4213,33 @@ private void setIfNotEmpty(Map<String, String> map, String value, String key) {
 	 * @param form application form
 	 * @return RegistrarModule if present or null
 	 */
-	private RegistrarModule getRegistrarModule(ApplicationForm form) {
-
+	private List<RegistrarModule> getRegistrarModules(ApplicationForm form) {
```
xflord, line 4216: Set here as well, otherwise we would for example try to reject the same application twice, no?

perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationForm.java
```diff
@@ -16,7 +20,7 @@ public class ApplicationForm {
 	private boolean automaticApproval;
 	private boolean automaticApprovalExtension;
 	private boolean automaticApprovalEmbedded;
-	private String moduleClassName;
+	private final List<String> moduleClassNames = new ArrayList<>();
```
xflord, line 23: Shouldn't this we a set instead of a list?

xflord (commented):
Couple more things that need to be cleared up now that the new GUI changes are being made

perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -4192,34 +4214,33 @@ private void setIfNotEmpty(Map<String, String> map, String value, String key) {
...
+					log.debug("[REGISTRAR] Attempting to instantiate class: {}{}", MODULE_PACKAGE_PATH, regModule);
+					module = (RegistrarModule) Class.forName(MODULE_PACKAGE_PATH + regModule)
+						.getDeclaredConstructor().newInstance();
+					module.setRegistrar(registrarManager);
+				} catch (Exception ex) {
+					log.error("[REGISTRAR] Exception when instantiating module. Skipping module {}{}",
+						MODULE_PACKAGE_PATH, regModule, ex);
+				}
+				log.debug("[REGISTRAR] Class {}{} successfully created.", MODULE_PACKAGE_PATH, regModule);
+				modules.add(module);
```
xflord, line 4240: Now, since null is assigned to `module` only before the loop, should the exception be thrown (and caught), this line will still execute and add the last assigned module. Which I suppose should not be a problem since it's a set (logs would be doubled), but if it happened on the first iteration, a null value would be added to the list and cause NullPointerExceptions further on.
We need to add `continue` to the end of the catch block, no?
I'm quite tired while reading this so maybe I'm wrong, but try to take a look

dBucik: Well spotted, should now be corrected. Thanks!

perun-base/src/main/java/cz/metacentrum/perun/registrar/model/ApplicationForm.java
xflord: I'm not too familiar with registrar modules, so this is up to debate with the higher-ups. I was just concerned that if duplicates were in the list, some error might be thrown when the `module.approveApplication` method gets called twice, for example.

xflord (approved):
When it comes to the list/set discussion, I'm not too familiar with registrar modules and so it's up to the higher-ups. I was just concerned that when the same modules tries to for example approve the same application twice, issues might arise.

mattjoke (approved):
Looks good :+1:

PR #4070, opened 2023-09-05
feat(core): autoreject applications on expiration

mattjoke:
* scheduler autorejects all of the group application, if the member is switched to EXPIRED state
* setStatus also rejects all group applications for Member

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
@@ -33,6 +33,7 @@ import cz.metacentrum.perun.core.api.exceptions.ExtendMembershipException;
 import cz.metacentrum.perun.core.api.exceptions.InternalErrorException;
 import cz.metacentrum.perun.core.api.exceptions.MemberGroupMismatchException;
 import cz.metacentrum.perun.core.api.exceptions.MemberNotExistsException;
+import cz.metacentrum.perun.core.api.exceptions.PerunException;
 import cz.metacentrum.perun.core.api.exceptions.UserNotExistsException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeAssignmentException;
 import cz.metacentrum.perun.core.api.exceptions.WrongAttributeValueException;
@@ -50,7 +51,6 @@ import org.springframework.jdbc.core.JdbcPerunTemplate;
 
 import javax.sql.DataSource;
 import java.time.LocalDate;
-import java.time.Period;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
@@ -84,6 +84,8 @@ public class ExpirationNotifScheduler {
 
 	private PerunBl perun;
 	private JdbcPerunTemplate jdbc;
+
+	private RegistrarManagerImpl registrarManager;
 
 	private final DateTimeFormatter lastAccessFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
 
@@ -123,6 +125,7 @@ public class ExpirationNotifScheduler {
 		this.sess = perun.getPerunSession(
 				new PerunPrincipal(synchronizerPrincipal, ExtSourcesManager.EXTSOURCE_NAME_INTERNAL, ExtSourcesManager.EXTSOURCE_INTERNAL),
 				new PerunClient());
+		this.registrarManager = new RegistrarManagerImpl();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/ExpirationNotifScheduler.java
```diff
@@ -638,11 +641,20 @@ private void expireMembers(LocalDate date) throws WrongAttributeAssignmentExcept
 			if (member.getStatus().equals(Status.VALID)) {
 				try {
 					perun.getMembersManagerBl().expireMember(sess, member);
+					List<Group> groups = sess.getPerun().getGroupsManager().getMemberGroups(sess, member);
+					for (Group g : groups) {
+						List<Application> apps = registrarManager.getApplicationsForMember(sess, g, member);
+						for (Application app: apps) {
+							registrarManager.rejectApplication(sess, app.getId(), "Application is set to status REJECTED, because the member it is assigned to is in EXPIRED state.");
```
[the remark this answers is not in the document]
mattjoke: The problem with that is there is no easy way to call the `rejectApplication` method from the registrarManager because of the circular dependencies. I am not sure if the module rejection is also needed. I'll find out

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -979,4 +979,21 @@ private String getGroupStatusSQLConditionForMembersPage(MembersPageQuery query,
 		}
 		return groupStatusesQueryString;
 	}
+
+	public void rejectGroupApplications(PerunSession session, Member member) {
```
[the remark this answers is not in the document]
mattjoke: This method selects only applications that are NOT rejected or approved. Also, it removed the logins, which may be unwanted .

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
```diff
@@ -979,4 +979,21 @@ private String getGroupStatusSQLConditionForMembersPage(MembersPageQuery query,
...
+		try {
+			List<Integer> ids = jdbc.query(
+				"select id from application " +
+					"where user_id=? and vo_id=? and group_id is not null and state not in (?)",
+				new SingleColumnRowMapper<>(Integer.class), member.getUserId(), member.getVoId(), rejected);
+
+			// Reject all applications for member
+			for (Integer id : ids) {
```
dBucik, line 991: Just wondering if this should not be done directly in DB instead of looping through IDs. And actually, I believe it could be then merged into a single query.
```
UPDATE application SET state = ?, modified_at =?, modified_by_uid=?,
WHERE id IN
( SELECT id FROM application WHERE user_id = ? AND vo_id = ? AND group_id IS NOT NULL AND state != ?)
```
@xflord, @HejdaJakub any opinions?

mattjoke: Yes, this could be done. I'll try to implement this :+1:

perun-core/src/main/java/cz/metacentrum/perun/core/impl/MembersManagerImpl.java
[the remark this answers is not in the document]
dBucik: Just please take a look what is the order in execution of nested queries inside updates. Otherwise, we might end up with the same result as all ELIXIR users being renamed to `Deleted User` 😉 👼🏼 #notMyFail :D

mattjoke:
I decided to scrap the idea of introducing the logic of the registrar manager to the members manager.
Changes:

- removed the logic used before
- only ways to set status to EXPIRED and reject applications are as follows
    - setStatus method from API, which will do the rejection (NOTE: any other internal calls of setStatus will not trigger the rejection process)
    - scheduler will do the automatic rejection

perun-registrar-lib/src/test/java/cz/metacentrum/perun/registrar/ExpirationNotifSchedulerTest.java
```diff
@@ -683,6 +686,66 @@ public void isCesnetEligible_ignoredCases() throws Exception {
...
+		registrarManager.submitApplication(session, application, data);
+
+		// Approve all applications
+		List<Application> applications = registrarManager.getApplicationsForGroup(session, group, new ArrayList<>());
+		for (Application app: applications) {
+			registrarManager.approveApplication(session, app.getId());
+		}
+
+		// Expire member
+		perun.getMembersManagerBl().setStatus(session, member, Status.EXPIRED);
```
xflord, line 740: You will have to set the expiration date with you current implementation for this to work

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/MembersManagerMethod.java
```diff
@@ -1742,7 +1744,20 @@ public Member call(ApiCaller ac, Deserializer parms) throws PerunException {
...
+			member = ac.getMembersManager().setStatus(sess, member, status);
+			// Reject all applications for this member if member is expired
+			if (status != Status.EXPIRED) return member;
+			List<Group> groups = sess.getPerun().getGroupsManager().getMemberGroups(sess, member);
+			for (Group g : groups) {
+				List<Application> apps = ac.getRegistrarManager().getApplicationsForMember(sess, g, member);
+				for (Application app: apps) {
+					ac.getRegistrarManager().rejectApplication(sess, app.getId(), "Application is set to status REJECTED, because the member it is assigned to is in EXPIRED state.");
+				}
+			}
```
xflord, line 1759: This will also pass approved and already rejected applications into `rejectApplication()`, which will throw an exception. We only want to reject open applications. Also, maybe we can extract this into a separate method in RegistrarManager, to save some code (and include exception handling there)?

xflord (changes requested):
It feels a bit weird that setting status manually behaves differently, and this might lead to some future inconsistencies if we have some new code in the core that expires members.
However, if you've discussed this with others, I'm fine with it.

zlamalp:
Will be handled in source repository. Closing this one.

PR #4073, opened 2023-09-06
fix(core): approve applications order

sarkapalkovicova:
Sort application before approving.
First, approve applications to VOs and then to groups.


bodnara (approved):
Should handle approving of Vo and Group applications at the same time properly.

xflord (approved):

PR #4076, opened 2023-09-12
feat(core): extend method getRichAdmins for group to optionally include also indirect admins

HejdaJakub:
* Method getRichAdmins was extended to optionally include also indirect admins.
* This method now also have a parameter to define group related role (GROUPADMIN, GROUPOBSERVER, GROUPMEMBERSHIPMANAGER).

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
@@ -818,10 +818,10 @@ public interface GroupsManager {
 	List<User> getAdmins(PerunSession perunSession, Group group, boolean onlyDirectAdmins) throws PrivilegeException, GroupNotExistsException;
 
 	/**
-	 * Gets list of all richUser administrators of this group.
+	 * Gets list of all richUser administrators of this group with a specified role.
 	 * If some group is administrator of the given group, all VALID members are included in the list.
 	 *
-	 * Supported roles: GroupAdmin
+	 * Supported roles: GroupAdmin, GroupObserver, GroupMembershipManager
 	 *
 	 * If "onlyDirectAdmins" is "true", return only direct users of the group for supported role with specific attributes.
 	 * If "allUserAttributes" is "true", do not specify attributes through list and return them all in objects richUser. Ignoring list of specific attributes.
@@ -832,6 +832,7 @@ public interface GroupsManager {
 	 * @param specificAttributes list of specified attributes which are needed in object richUser
 	 * @param allUserAttributes if true, get all possible user attributes and ignore list of specificAttributes (if false, get only specific attributes)
 	 * @param onlyDirectAdmins if true, get only direct user administrators (if false, get both direct and indirect)
+	 * @param role specified role
 	 *
 	 * @return list of RichUser administrators for the group and supported role with attributes
 	 *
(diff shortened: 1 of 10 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
```diff
@@ -818,10 +818,10 @@ public interface GroupsManager {
 	List<User> getAdmins(PerunSession perunSession, Group group, boolean onlyDirectAdmins) throws PrivilegeException, GroupNotExistsException;
 
 	/**
-	 * Gets list of all richUser administrators of this group.
+	 * Gets list of all richUser administrators of this group with a specified role.
 	 * If some group is administrator of the given group, all VALID members are included in the list.
 	 *
-	 * Supported roles: GroupAdmin
+	 * Supported roles: GroupAdmin, GroupObserver, GroupMembershipManager
```
xflord, line 824: Not entirely true since user can pass any role.

perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
```diff
@@ -832,6 +832,7 @@ public interface GroupsManager {
 	 * @param specificAttributes list of specified attributes which are needed in object richUser
 	 * @param allUserAttributes if true, get all possible user attributes and ignore list of specificAttributes (if false, get only specific attributes)
 	 * @param onlyDirectAdmins if true, get only direct user administrators (if false, get both direct and indirect)
+	 * @param role specified role
```
xflord, line 835: How about changing this to a list/set of role names and then returning users with any of the roles (this would reduce calls in the GUI app, no?)

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1140,7 +1160,8 @@ public List<RichUser> call(ApiCaller ac, Deserializer parms) throws PerunExcepti
 						ac.getGroupById(parms.readInt("group")),
 						parms.readList("specificAttributes", String.class),
 						parms.readBoolean("allUserAttributes"),
-						parms.readBoolean("onlyDirectAdmins"));
+						parms.readBoolean("onlyDirectAdmins"),
+						parms.contains("role") ? parms.readString("role") : Role.GROUPADMIN);
```
xflord, line 1164: Since `role` is defined as required within the query in openapi, the ternary operator is unnecessary, no?

perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
HejdaJakub: I created it this way according to a planned usage in GUI - you can choose there just only one role at the same time when displaying managers. If we had a use case for getting all roles/more roles together, I would change it (but now I don't see this use case).

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
HejdaJakub: I am not sure if the old method is not used by any script/old GUI and the previous behavior was implemented like the only expected role is GROUPADMIN, so I wanted to let it be available in RPC also in the old form (just for back compatibility) - does it make sense or it is not necessary, what do you think?

xflord: It doesn't hurt to leave it in. Maybe mention the default value in the javadoc above

perun-core/src/main/java/cz/metacentrum/perun/core/api/GroupsManager.java
xflord: I think the list would allow more flexibility, but also it can be updated later if necessary. Up to you

HejdaJakub:
Reason for closing this PR here: https://github.com/CESNET/perun-web-apps/pull/1423

PR #4077, opened 2023-09-12
feat(engine): pass service name to send/gen script when using generic…

xflord:
… scripts

* the name of the propagated service is passed to gen/scripts, which then use this information for gen folder creation
* this is necessary when one facility has multiple services utilizing the same generic script

### DIFF
--- a/perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/GenWorkerImpl.java
+++ b/perun-engine/src/main/java/cz/metacentrum/perun/engine/scheduling/impl/GenWorkerImpl.java
@@ -46,8 +46,13 @@ public class GenWorkerImpl extends AbstractWorker<Task> implements GenWorker {
 
 		log.info("[{}] Executing GEN worker for Task with Service ID: {} and Facility ID: {}.",
 				getTask().getId(), getTask().getServiceId(), getTask().getFacilityId());
-
-		ProcessBuilder pb = new ProcessBuilder(service.getScript(), "-c", "-f", String.valueOf(getTask().getFacilityId()));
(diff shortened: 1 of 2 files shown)


zlamalp (commented):
This doesn't seem to break anything, but I would like to know if its part of some bigger change or not? Let's say migration of all generic services to shared sender? How we will manually test the services - if new param will be always required etc.

xflord:
> This doesn't seem to break anything, but I would like to know if its part of some bigger change or not? Let's say migration of all generic services to shared sender? How we will manually test the services - if new param will be always required etc.

This PR is still subject to change, I just made it to share my thoughts with others working on the generic export service. My bad; I should have put some information here as well.

- As mentioned, these changes are to support the generic export service. The `SERVICE_NAME` variable present in current gen scripts wouldn't make much sense in the generic one, and we need to get the information somewhere else to ensure correct spool folder generation (should 2 different services on one facility use the generic script, conflict would occur if the spool folder name were hard-coded in that script). The support for the `-s` parameter is already present.
- The same then goes for the send script, which needs to locate the gen folder. The send script for the new generic service will utilize the extra parameter. (That's the initial thought, at least)

melanger:
re ST-1193

melanger:
re https://perunaai.atlassian.net/browse/ST-1193

xflord:
After a discussion with @zlamalp and @balcirakpeter it was decided that the current structure of send/slave scripts should be sufficient with these changes, and the passing of service name should solve the spool folder conflict. We expect very few (probably no more than one) generic gen scripts, so changing the logic/structure for them heavily seems redundant.

mattjoke (approved):

PR #4082, opened 2023-09-15
feat(core): allow members to read their group expiration

xflord:
* fixes a bug, where users couldn't extend membership in a group through the profile app


dBucik (approved):

bodnara (approved):

zlamalp (commented):
I believe authorization rules/policies are not updated through this for existing attributes. If this is a generally required change, please add DEPLOYMENT NOTE to ensure it is updated on each instance during the deployment.

xflord:
added deployment note to commit message

xflord:
> I believe authorization rules/policies are not updated through this for existing attributes. If this is a generally required change, please add DEPLOYMENT NOTE to ensure it is updated on each instance during the deployment.

You are right, did not notice bottom of the `initialize()` method ensures this

PR #4089, opened 2023-09-19
feat(core): support authoritative groups in group structure synchronization

zlamalp:
- Set group:def:authoritativeGroup attribute to all subgroups in group structure synchronization based on the value from the base group.


zlamalp:
Merge only after #4088 is merged.

xflord (approved):

PR #4093, opened 2023-09-25
fix(core): user:virt:voPersonExternalAffiliation forces to lowercase

mattjoke:
* all of the values are forced to lowercase
* added new test to check for case duplicities

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_voPersonExternalAffiliation.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_voPersonExternalAffiliation.java
@@ -47,6 +47,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * All affiliations collected from:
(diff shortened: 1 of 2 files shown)


perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_voPersonExternalAffiliationTest.java
```diff
@@ -118,6 +128,33 @@ public void setVariables() {
...
+		);
+
+		String attributeName = classInstance.getSourceAttributeName();
+		when(session.getPerunBl().getAttributesManagerBl().getAttribute(session, ues3, attributeName)).thenReturn(
+			uesAtt3
+		);
+
+		Attribute receivedAttr = classInstance.getAttributeValue(session, user, classInstance.getAttributeDefinition());
+
+		System.out.println(receivedAttr.getValue());
```
xflord, line 148: debugging print

perun-core/src/test/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_voPersonExternalAffiliationTest.java
```diff
@@ -118,6 +128,33 @@ public void setVariables() {
...
 	}
 
+	@Test
+	public void getAttributeValueNoDuplicities() throws Exception {
+		System.out.println("getAttributeValueNoDuplicities()");
+		urn_perun_user_attribute_def_virt_voPersonExternalAffiliation classInstance = new urn_perun_user_attribute_def_virt_voPersonExternalAffiliation();
+		PerunSessionImpl session = mock(PerunSessionImpl.class, RETURNS_DEEP_STUBS);
+
+		when(session.getPerunBl().getUsersManagerBl().getUserExtSources(session, user)).thenReturn(
+			Arrays.asList(ues1, ues2, ues3)
```
xflord, line 138: It confused me a bit at first that we return three ext sources here and only get the values from one of them. Maybe it would more semantically correct if we only returned ues3 here (or mocked the attributes for the other ext sources)?

xflord (commented):
Nitpicks, looks fine though

bodnara (approved):

mattjoke:
It does not behave as it should. More discussion needed.

perun-core/src/main/java/cz/metacentrum/perun/core/impl/modules/attributes/urn_perun_user_attribute_def_virt_voPersonExternalAffiliation.java
```diff
@@ -125,8 +126,18 @@ public Attribute getAttributeValue(PerunSessionImpl sess, User user, AttributeDe
 		valuesWithoutDuplicities.addAll(getAffiliationsManuallyAssigned(sess, user));
 		valuesWithoutDuplicities.addAll(getAffiliationsFromGroups(sess, user));
 
+		// remove duplicities, by accepting only the first occurrence of each value (other occurences, case-insensitive, will be removed)
+		Set<String> valuesWithoutDuplicitiesCaseInsensitive = new HashSet<>();
```
xflord, line 130: Nitpick, but this can be an ArrayList right away.

Why was this change necessary anyway?

mattjoke: The Set performs better; it does not matter what we use.

The lower-case solution does not work because we shouldn't modify the values; we should group them case-insensitively.

xflord (approved):

PR #4094, opened 2023-09-25
feat(core): new loginWithMail_mu virtual attribute

kofzera:
Added a new attribute. Contains login in the MU namespace concatenated with @muni.cz if it is available, null otherwise.


xflord (approved):
Looks good, I would maybe consider naming the attribute 'scopedLogin' to match existing 'unscopedLogin' (also mail is more than the domain, no?) But that probably doesn't matter much

zlamalp:
I'm too for renaming to `scopedLogin` or something similar so the attribute name make sense in the used context.

bodnara (approved):

zlamalp:
@kofzera please update also pull-request name/description based on the new attribute name (commit message).

PR #4104, opened 2023-10-02
feat(core): synchronize group membership status

kofzera:
- The status of a group member is also synchronized for SQL extSource
- The default value (status column not found, unknown value) is valid
- When a previously valid member is synchronized to expired
the groupMembershipExpiration attr is set to current date
- When a new/current member is being set to valid the attr is
set based on the group policy accordingly

DEPLOYMENT NOTE: possibility to synchronize group status.
Expects "status" column in SQLExtSource with values "EXPIRED" or "VALID".
Candidate object extended with expectedSyncGroupStatus property.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -3376,12 +3376,15 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 									log.info("Switching member id {} into INVALID state from DISABLED, because there was problem with attributes {}.", richMember.getId(), e);
 								}
 							}
+							// also synchronize the group status
+							synchronizeGroupStatus(sess, richMember, subjectFromLoginSource.get("status"), group);
 							idsOfUsersInGroup.remove(user.getId());
 						} else {
 							//he is not yet in group, so we need to create a candidate
 							Candidate candidate = new Candidate(user, source);
 							//for lightweight synchronization we want to skip all update of attributes
 							candidate.setAttributes(new HashMap<>());
+							candidate.setExpectedSyncGroupStatus(subjectFromLoginSource.get("status"));
 							candidatesToAdd.add(candidate);
 						}
 						break;
@@ -3805,8 +3808,53 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 		//Synchronize userExtSources (add not existing)
 		addUserExtSources(sess, candidate, memberToUpdate);
 
+		if (isDirectGroupMember(sess, group, memberToUpdate)) {
+			synchronizeGroupStatus(sess, memberToUpdate, candidate.getExpectedSyncGroupStatus(), group);
+		}
+
 		//Set correct member Status
 		updateMemberStatus(sess, memberToUpdate);
+	}
+
+	/**
+	 * Update a group member status based on the status in candidate object.
+	 * The member-group expiration attribute is set to now when expiring a member
+	 * and updated based on the group policy if validating a member.
+	 *
+	 * @param sess perun session
+	 * @param memberToSynchronize member being synchronized
+	 * @param newStatusString the group status from the extSource
+	 * @param group the group being synchronized
+	 */
+	private void synchronizeGroupStatus(PerunSession sess, Member memberToSynchronize, String newStatusString, Group group) {
+		MemberGroupStatus previousStatus = getDirectMemberGroupStatus(sess, memberToSynchronize, group);
+		MemberGroupStatus newStatus;
+		if (newStatusString == null) {
+			// default policy is VALID
+			newStatus = MemberGroupStatus.VALID;
+		}
+		else if (newStatusString.equals("VALID")) {
+			newStatus = MemberGroupStatus.VALID;
+		} else if (newStatusString.equals("EXPIRED")) {
+			newStatus = MemberGroupStatus.EXPIRED;
+		} else {
+			// unknown status value default policy VALID
+			log.warn("Unknown value {} of status while synchronizing group {} for user {}.", newStatusString, group.getName(), memberToSynchronize.getUserId());
+			newStatus = MemberGroupStatus.VALID;
+		}
+		if (newStatus.equals(MemberGroupStatus.EXPIRED) && !MemberGroupStatus.EXPIRED.equals(previousStatus)) {
+			try {
+				inactivateMember(sess, memberToSynchronize, group);
+			} catch (MemberNotExistsException e) {
+				throw new InternalErrorException(e);
+			}
+		} else if (newStatus.equals(MemberGroupStatus.VALID) && !MemberGroupStatus.VALID.equals(previousStatus)) {
+			try {
+				reactivateMember(sess, memberToSynchronize, group);
+			} catch (MemberNotExistsException e) {
+				throw new InternalErrorException(e);
+			}
+		}
 	}
 
 	/**
@@ -4097,6 +4145,7 @@ public class GroupsManagerBlImpl implements GroupsManagerBl {
 					// Shouldn't happen, group should always exist
 					throw new ConsistencyErrorException(ex);
 				}
+				synchronizeGroupStatus(sess, member, candidate.getExpectedSyncGroupStatus(), group);
 			}
 			log.info("Group synchronization {}: New member id {} added.", group, member.getId());
 		} catch (AlreadyMemberException e) {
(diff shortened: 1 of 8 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3339,6 +3339,7 @@ private void categorizeMembersForLightweightSynchronization(PerunSession sess, G
 							Candidate candidate = new Candidate(user, source);
 							//for lightweight synchronization we want to skip all update of attributes
 							candidate.setAttributes(new HashMap<>());
+							candidate.setExpectedStatus(subjectFromLoginSource.get("status"));
```
xflord, line 3342: This should work only for users that were not yet in the group - don't we want to update status for existing members as well? (Even during lightweight synchronization). Now it only validates the member, but ideally we want to check the status in the ExtSource and set accordingly.

xflord (commented):
Will take a further look later today

xflord (changes requested):
Also please add more tests - for lightweight synchro, cases where user is not a member of the VO/Group beforehand, etc.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3762,10 +3765,63 @@ public void updateExistingMemberWhileSynchronization(PerunSession sess, Group gr
 		//Synchronize userExtSources (add not existing)
 		addUserExtSources(sess, candidate, memberToUpdate);
 
+		synchronizeGroupStatus(sess, memberToUpdate, candidate.getExpectedSyncGroupStatus(), group);
```
xflord, line 3768: This will still get called before the member gets added to the group in some cases - from `addMissingMemberWhileSynchronization` (and will fail).

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3762,10 +3765,63 @@ public void updateExistingMemberWhileSynchronization(PerunSession sess, Group gr
...
+			newStatus = MemberGroupStatus.VALID;
+		} else if (newStatusString.equals("EXPIRED")) {
+			newStatus = MemberGroupStatus.EXPIRED;
+		} else {
+			// unknown status value default policy VALID
+			log.warn("Unknown value {} of status while synchronizing group {} for user {}.", newStatusString, group.getName(), memberToSynchronize.getUserId());
+			newStatus = MemberGroupStatus.VALID;
+		}
+		if (newStatus.equals(MemberGroupStatus.EXPIRED) &&
+			(previousStatus == null || !previousStatus.equals(MemberGroupStatus.EXPIRED))) {
```
xflord, line 3802:
```suggestion
			!MemberGroupStatus.EXPIRED.equals(previousStatus)) {
```
just nitpicking at this point but this is what I meant

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupAndGroupStructureSynchronizationIntegrationTest.java
```diff
@@ -1088,6 +1093,239 @@ public void synchronizeGroupUpdateExistingMemberStatus() throws Exception {
...
+		groupsManagerBl.synchronizeGroup(sess, group);
+		assertEquals(MemberGroupStatus.EXPIRED, groupsManagerBl.getGroupMembers(sess, group).get(0).getGroupStatus());
+	}
+
+	@Test
+	public void synchronizeGroupUpdateMembershipStatusToDefaultValid() throws Exception {
+		System.out.println(CLASS_NAME + "synchronizeGroupUpdateMembershipStatus");
+
+		Group groupWithExpiration = new Group("GroupsManagerTestGroup2","testovaci2");
+		groupWithExpiration = groupsManagerBl.createGroup(sess, vo, groupWithExpiration);
```
xflord, line 1156: Please create a private method to setup `groupWithExpiration` since it repeats itself in most of the tests

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupAndGroupStructureSynchronizationIntegrationTest.java
```diff
@@ -1088,6 +1093,239 @@ public void synchronizeGroupUpdateExistingMemberStatus() throws Exception {
...
+		when(essa.getGroupSubjects(anyMap())).thenReturn(subjects);
+
+		groupsManagerBl.synchronizeGroup(sess, groupWithExpiration);
+		assertEquals(MemberGroupStatus.VALID, groupsManagerBl.getGroupMembers(sess, groupWithExpiration).get(0).getGroupStatus());
+		assertEquals(LocalDate.now().plusMonths(1).toString(), attributesManagerBl.getAttribute(sess, member, groupWithExpiration, A_MG_D_MEMBERSHIP_EXPIRATION).getValue());
+	}
+
+	@Test
+	public void synchronizeGroupUpdateMembershipStatusExpirDateToNever() throws Exception {
+		System.out.println(CLASS_NAME + "synchronizeGroupUpdateMembershipStatus");
```
xflord, line 1193: update the print in all of the tests

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupAndGroupStructureSynchronizationIntegrationTest.java
```diff
@@ -1088,6 +1093,239 @@ public void synchronizeGroupUpdateExistingMemberStatus() throws Exception {
...
+		when(extSourceManagerBl.getCandidate(sess, attributes, (ExtSourceLdap)essa, "metodej")).thenReturn(new CandidateSync(candidate));
+		when(essa.getGroupSubjects(anyMap())).thenReturn(subjects);
+
+		groupsManagerBl.synchronizeGroup(sess, groupWithExpiration);
+		assertEquals(MemberGroupStatus.VALID, groupsManagerBl.getGroupMembers(sess, groupWithExpiration).get(0).getGroupStatus());
+		assertEquals(LocalDate.now().plusMonths(1).toString(), attributesManagerBl.getAttribute(sess, member, groupWithExpiration, A_MG_D_MEMBERSHIP_EXPIRATION).getValue());
+	}
+
+	@Test
+	public void synchronizeGroupUpdateMembershipStatusExpirDateToNever() throws Exception {
```
xflord, line 1192: This name does not make much sense to me

perun-core/src/test/java/cz/metacentrum/perun/core/entry/GroupAndGroupStructureSynchronizationIntegrationTest.java
```diff
@@ -1088,6 +1093,239 @@ public void synchronizeGroupUpdateExistingMemberStatus() throws Exception {
...
+
+		Attribute expirationRulesAttribute = new Attribute(attributesManagerBl.getAttributeDefinition(sess,
+			AttributesManager.NS_GROUP_ATTR_DEF + ":groupMembershipExpirationRules"));
+		Map<String, String> values = new LinkedHashMap<>();
+		values.put(AbstractMembershipExpirationRulesModule.membershipPeriodKeyName, "+1m");
+		expirationRulesAttribute.setValue(values);
+		attributesManagerBl.setAttribute(sess, groupWithExpiration, expirationRulesAttribute);
+
+		Attribute groupMembExpirAttr = attributesManagerBl.getAttribute(sess, member, groupWithExpiration, A_MG_D_MEMBERSHIP_EXPIRATION);
+		groupMembExpirAttr.setValue("2022-10-11");
```
xflord, line 1224: setting expiration of valid member to the past is not correct

xflord (commented):
One more test that assures that member is not added to group when not in VO and I'll be content. Other than the tests looks good, thanks.

kofzera:
> One more test that assures that member is not added to group when not in VO and I'll be content. Other than the tests looks good, thanks.

Shouldn't the member be added to the group (and the VO) even if he is not in the VO beforehand? There is a test for that `synchronizeExpiredGroupStatusMemberNotInVO` and it adds the member to the group even without the recent changes.

xflord (approved):
Please include a deployment note about the possibility of status synchronization. Other than that as I said before - good job.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3762,10 +3765,60 @@ public void updateExistingMemberWhileSynchronization(PerunSession sess, Group gr
...
+	 * @param memberToSynchronize member being synchronized
+	 * @param newStatusString the group status from the extSource
+	 * @param group the group being synchronized
+	 */
+	private void synchronizeGroupStatus(PerunSession sess, Member memberToSynchronize, String newStatusString, Group group) {
+		MemberGroupStatus previousStatus = getDirectMemberGroupStatus(sess, memberToSynchronize, group);
+		MemberGroupStatus newStatus;
+		if (newStatusString == null) {
+			// default policy is VALID
+			log.warn("Missing status column while synchronizing group {}.", group.getName());
```
balcirakpeter, line 3791: Most synchronizations will not use the group status, therefore this message would just spam the logs. It would be probably better to remove it.

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3762,10 +3765,60 @@ public void updateExistingMemberWhileSynchronization(PerunSession sess, Group gr
 		//Synchronize userExtSources (add not existing)
 		addUserExtSources(sess, candidate, memberToUpdate);
 
+		if (isGroupMember(sess, group, memberToUpdate)) {
```
balcirakpeter, line 3768: Be careful. This method checks whether the member is in the group, but it does not care if the membership type is direct or indirect. I think we care only about direct memberships.

balcirakpeter (changes requested):
We need to double-check (probably on review) that we want to synchronize group status in the lightweight synchronizations. Moreover, we need to check if it is ok to change the Candidate object in the OpenAPI (if it does not break something).

balcirakpeter (approved):

PR #4105, opened 2023-10-03
feat(core): filter groups by role

sarkapalkovicova:
Added methods for filtering groups by user's role and its assignment.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
@@ -50,6 +50,7 @@ import cz.metacentrum.perun.core.api.RichMember;
 import cz.metacentrum.perun.core.api.RichUser;
 import cz.metacentrum.perun.core.api.RichUserExtSource;
 import cz.metacentrum.perun.core.api.Role;
+import cz.metacentrum.perun.core.api.RoleAssignmentType;
 import cz.metacentrum.perun.core.api.SecurityTeam;
 import cz.metacentrum.perun.core.api.Status;
 import cz.metacentrum.perun.core.api.User;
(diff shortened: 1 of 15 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1544,23 +1544,27 @@ public List<Group> call(ApiCaller ac, Deserializer parms) throws PerunException{
 	 * @param vo int <code>id</code> of vo
 	 * @param fromIndex int begin index of returned subList, included
 	 * @param attrNames List<String> if attrNames is null method will return RichGroups containing all attributes
+	 * @param roles list of selected roles (if empty, then return groups by all roles)
+	 * @param types list of selected types of roles (if empty, then return by roles of all types)
```
HejdaJakub, line 1548: I am not sure but I think that you should duplicate the javadoc for the new parameters, not just change the old one - from this javadoc is created the documentation and to keep in the docs also old methods with fewer parameters you should probably duplicate it. (it should be applied for all methods you changed in this file)

HejdaJakub (changes requested):
I found just RPC documentation inconsistency in the `GroupsManagerMethod.java` file.

perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleAssigmentType.java
```diff
@@ -0,0 +1,6 @@
+package cz.metacentrum.perun.core.api;
+
+public class RoleAssigmentType {
```
xflord, line 3: Why not make this an enum? And also use it everywhere instead of `string` (would eliminate extra code and make it more readable)

xflord (commented):
Will take another look later today

perun-core/src/main/java/cz/metacentrum/perun/core/blImpl/GroupsManagerBlImpl.java
```diff
@@ -3090,6 +3120,15 @@ public List<RichGroup> getAllRichSubGroupsWithAttributesByNames(PerunSession ses
 		return convertGroupsToRichGroupsWithAttributes(sess, this.getAllSubGroups(sess, parentGroup), attrNames);
 	}
 
+	@Override
+	public List<RichGroup> getAllRichSubGroupsWithAttributesByNames(PerunSession sess, Group parentGroup, List<String> attrNames, List<String> roles, List<String> types) {
+		List<Group> subGroups = this.getAllSubGroups(sess, parentGroup);
+		if ((!roles.isEmpty() || !types.isEmpty()) && sess.getPerunPrincipal().getUser() != null) {
+			subGroups.removeIf(group -> !AuthzResolverBlImpl.groupMatchesUserRolesFilter(sess, sess.getPerunPrincipal().getUser(), group, roles, types));
```
xflord, line 3127: Calling this for every group doesn't seem very efficient. Why not pass roles and types to Impl layer and create a private method like `getSQLWhereForRolesAndTypes` which could be then reused for every modified method (you can extract the method already from `getGroupsPage`)
Just a thought, not really sure how much this actually effects speed

perun-core/src/main/java/cz/metacentrum/perun/core/entry/GroupsManagerEntry.java
```diff
@@ -1389,6 +1403,33 @@ public List<RichGroup> getMemberRichGroupsWithAttributesByNames(PerunSession ses
 		return getGroupsManagerBl().filterOnlyAllowedAttributes(sess, richGroups, member, null, true);
 	}
 
+	@Override
+	public List<RichGroup> getMemberRichGroupsWithAttributesByNames(PerunSession sess, Member member, List<String> attrNames, List<String> roles, List<String> types) throws MemberNotExistsException, PrivilegeException {
+		Utils.checkPerunSession(sess);
+		this.getPerunBl().getMembersManagerBl().checkMemberExists(sess, member);
+
+		for (String role : roles) {
+			if (!AuthzResolver.roleExists(role)) {
```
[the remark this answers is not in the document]
xflord: @sarkapalkovicova

perun-base/src/main/java/cz/metacentrum/perun/core/api/RoleAssigmentType.java
xflord: @sarkapalkovicova

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1722,16 +2529,83 @@ public List<RichGroup> call(ApiCaller ac, Deserializer parms) throws PerunExcept
...
 	getMemberRichGroupsWithAttributesByNames {
 
 		@Override
 		public List<RichGroup> call(ApiCaller ac, Deserializer parms) throws PerunException {
-
 			List<RichGroup> listOfRichGroups = ac.getGroupsManager().getMemberRichGroupsWithAttributesByNames(ac.getSession(),
-				ac.getMemberById(parms.readInt("member")),
-				parms.readList("attrNames", String.class));
+					ac.getMemberById(parms.readInt("member")),
+					parms.readList("attrNames", String.class),
```
HejdaJakub, line 2606: You're defining here in javadoc for `getMemberRichGroupsWithAttributesByNames` method the `attrNames` parameter as optional, which is not a problem if we want it now (because before this method had required parameter attrNames) - but if we want it, then there should be a null replacement for this param if it is not presented in `parms`.

HejdaJakub (changes requested):

HejdaJakub (approved):

PR #4113, opened 2023-10-12
fix: correct attribute references in enabledO365MailForward

melanger:
it should reference defined attributes, not virtual attributes

re ST-1168


kofzera (approved):

xflord (approved):
missed this, sorry

PR #4116, opened 2023-10-16
feat(core): allow customization of replyTo field of emails

xflord:
* methods using `sendEmail` (account activation, email validation) now fill in the `replyTo` field from config value
* `from` field still gets filled from the same config property
* registrar now uses default replyTo address as well

DEPLOYMENT NOTE: edit new config property perun.mailchange.replyTo and existing perun.mailchange.backupFrom to customize the respective fields of sent emails (from core API). `replyTo` (and `replyToName`) can be defined in `perun-registrar-lib.properties` to achieve the same for registrar


HejdaJakub (approved):
I didn't test it but I think that this is how it should exactly work.

mattjoke (approved):
Looks good :+1:  If tested, then it can be merged.

PR #4117, opened 2023-10-17
fix(core): ignore already blocked destinations

sarkapalkovicova:
Allow to block multiple destinations at once, even if some are already blocked.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
@@ -98,9 +98,12 @@ public class ServicesManagerEntry implements ServicesManager {
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


perun-core/src/main/java/cz/metacentrum/perun/core/entry/ServicesManagerEntry.java
```diff
@@ -100,7 +100,10 @@ public void blockServiceOnDestination(PerunSession sess, Service service, int de
 	@Override
 	public void blockServicesOnDestinations(PerunSession sess, List<RichDestination> richDestinations) throws PrivilegeException, DestinationNotExistsException, ServiceAlreadyBannedException, FacilityNotExistsException {
 		for (RichDestination richDestination : richDestinations) {
-			blockServiceOnDestination(sess, richDestination.getService(), richDestination.getId());
```
zlamalp, line 103: I think this should be considered as breaking change, since it changes expected outcome of the API call. You should update interface class for this method and also `ServicesManagerMethod.java` and RPC API docs accordingly as `ServiceAlreadyBannedException` is no longer thrown.

zlamalp (changes requested):

HejdaJakub (approved):

xflord (approved):

PR #4132, opened 2023-11-02
fix(registrar): transaction for approving multiple applications

zlamalp:
- Approval or rejection of multiple applications can't run in a single transaction since it is subject to non-revertible effects (like sending emails).
- We can't afford to rollback action on X applications just because one of them failed.
- GUI should be probably updated to handle this change and maybe also server can provide better feedback.

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -1745,11 +1745,10 @@ public class RegistrarManagerImpl implements RegistrarManager {
 	}
 
 	@Override
-	@Transactional(rollbackFor = Exception.class)
 	public void rejectApplications(PerunSession sess, List<Integer> applicationIds, String reason) throws PerunException {
 		Collections.sort(applicationIds, Collections.reverseOrder());
 		for (Integer id : applicationIds) {
-			rejectApplication(sess, id, reason);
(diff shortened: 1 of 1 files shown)


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1745,11 +1745,10 @@ public Application rejectApplication(PerunSession sess, int appId, String reason
 	}
 
 	@Override
-	@Transactional(rollbackFor = Exception.class)
 	public void rejectApplications(PerunSession sess, List<Integer> applicationIds, String reason) throws PerunException {
 		Collections.sort(applicationIds, Collections.reverseOrder());
 		for (Integer id : applicationIds) {
-			rejectApplication(sess, id, reason);
+			registrarManager.rejectApplication(sess, id, reason);
```
xflord, line 1751: if this is necessary for the calls to be run in separate transactions (?), then the change is required in `approveApplications` as well. Also `approveApplication` is missing the `Transactional` annotation

zlamalp: `@Transactional` annotation is on `approveApplicationInternal()` which is called by `approveApplication()`. So there is no transaction on `approveApplication()` itself, since it is kind of wrapper, just like new `approveApplications()`.

xflord (approved):

HejdaJakub (approved):

PR #4133, opened 2023-11-02
fix(registrar): pass registrar session when submitting embedded applications

zlamalp:
- We passed session of approving person but used methods are relying on the fact, that session is properly related to the person submitting its own application.
- This now fails to submit embedded application for a group if approving person is a member of such group - even if application is related to a different user.
- Usage of internal registrar session should be more appropriate, but should be thoroughly tested!

### DIFF
--- a/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
+++ b/perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
@@ -1955,7 +1955,7 @@ public class RegistrarManagerImpl implements RegistrarManager {
 				groupApplication.setExtSourceType(app.getExtSourceType());
 				groupApplication.setCreatedBy("Automatically generated");
 
-				submitApplication(sess, groupApplication, new ArrayList<>());
+				submitApplication(registrarSession, groupApplication, new ArrayList<>());
 			} catch (Exception e) {
 				failedGroups.put(group.getId(), e.getMessage());
 			}


perun-registrar-lib/src/main/java/cz/metacentrum/perun/registrar/impl/RegistrarManagerImpl.java
```diff
@@ -1955,7 +1955,7 @@ private void submitEmbeddedGroupApplications(PerunSession sess, List<Group> grou
 				groupApplication.setExtSourceType(app.getExtSourceType());
 				groupApplication.setCreatedBy("Automatically generated");
 
-				submitApplication(sess, groupApplication, new ArrayList<>());
+				submitApplication(registrarSession, groupApplication, new ArrayList<>());
```
xflord, line 1958: registrarSession always has LOA 0, which could be different from the LOA of the actual applicant and result in false rejection based on the membership rules of group/vo, no? Might be wrong, got this solely through static code review (`canBeMember` method)

zlamalp: Yes, It might have other side effects, that's why I suggested it should be thoroughly tested. We should probably update `submitApplication()` logic to properly handle embedded applications.

HejdaJakub: I tested it and it works correctly also with stronger LOA rules set on a group layer. But we can discuss if some other use cases are worth testing.

HejdaJakub (approved):

xflord (approved):

xflord (changes requested):
Screenshot from 2023-11-06 12-33-16

xflord:
Will update `submitApplication()` in separate PR to fix this, otherwise everything else seemed to work fine

xflord:
Continued in https://github.com/CESNET/perun/pull/4136

PR #4136, opened 2023-11-06
fix(registrar): pass registrar session when submitting embedded appli…

xflord:
…cations

* We passed session of approving person but used methods are relying on the fact, that session is properly related to the person submitting its own application.
* This now fails to submit embedded application for a group if approving person is a member of such group - even if application is related to a different user.
* Usage of internal registrar session should be more appropriate
* Do not check VO LOA membership rules for embedded applications, this is already checked in base applications
* Fixed typo that prevented application form to load in registrar if the application was "EXTENSION" and included embedded groups item
* Original PR by https://github.com/zlamalp


HejdaJakub (approved):
I believe that I tested it correctly now (for the LOA on the VO level) and it seems that everything works as expected. :raised_hands:

PR #4137, opened 2023-11-06
refactor(core): do not generate metacentrum idp identity

zlamalp:
- Internal Metacentrum IdP is no longer used, there is no reason to generate UES for each user with login.


zlamalp:
Code has been merged at GitLab, closing here.

PR #4151, opened 2023-11-07
fix(core): admins-related methods

sarkapalkovicova:
- Added getAdmins method into AuthzResolver.
- Admins-related methods in managers use universal methods from AuthzResolver.

BREAKING CHANGE: Methods for obtaining admins return only users with valid status in the authorized group and the corresponding VO.

### DIFF
--- a/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
+++ b/perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
@@ -1385,6 +1385,39 @@ public class AuthzResolver {
 	}
 
 	/**
+	 * Get all valid user administrators (for group-based rights, status must be VALID for both Vo and group) for complementary object and role.
+	 *
+	 * If <b>onlyDirectAdmins</b> is <b>true</b>, return only direct users of the complementary object for role.
+	 *
+	 * @param sess perun session
+	 * @param complementaryObject for which we will get administrator
+	 * @param role expected role to filter managers by
+	 * @param onlyDirectAdmins if true, get only direct user administrators (if false, get both direct and indirect)
+	 *
+	 * @return list of user administrators for complementary object and role.
+	 */
+	public static List<User> getAdmins(PerunSession sess, PerunBean complementaryObject, String role, boolean onlyDirectAdmins) throws PrivilegeException, RoleCannotBeManagedException {
+		Utils.checkPerunSession(sess);
+		Utils.notNull(role, "role");
+		Utils.notNull(complementaryObject, "complementaryObject");
+
+		if (!roleExists(role)) {
+			throw new InternalErrorException("Role: " + role + " does not exists.");
+		}
+
+		// Authorization
+		try {
+			if(!authorizedToReadRole(sess, complementaryObject, role)) {
+				throw new PrivilegeException("You are not privileged to use the method getRichAdmins.");
+			}
+		} catch (RoleManagementRulesNotExistsException e) {
+			throw new InternalErrorException("Management rules not exist for the role " + role, e);
+		}
+
+		return AuthzResolverBlImpl.getAdmins(sess, complementaryObject, role, onlyDirectAdmins);
+	}
+
+	/**
 	 * Get all authorizedGroups for complementary object and role.
 	 *
 	 * @param sess perun session
--- a/perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
+++ b/perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
@@ -32,6 +32,7 @@ import java.util.Set;
 
 import static cz.metacentrum.perun.core.api.AuthzResolver.MFA_CRITICAL_ATTR;
 import static java.util.Map.entry;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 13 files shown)


perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
```diff
@@ -1059,13 +1063,14 @@ public Void call(ApiCaller ac, Deserializer parms) throws PerunException {
...
-				return ac.getGroupsManager().getAdmins(ac.getSession(),
-						ac.getGroupById(parms.readInt("group")));
+			try {
+				if(parms.contains("onlyDirectAdmins")) {
+					return AuthzResolver.getAdmins(ac.getSession(), ac.getGroupById(parms.readInt("group")), Role.GROUPADMIN, parms.readBoolean("onlyDirectAdmins"));
+				} else {
+					return AuthzResolver.getAdmins(ac.getSession(), ac.getGroupById(parms.readInt("group")), Role.GROUPADMIN, false);
+				}
+			} catch (RoleCannotBeManagedException ex) {
+				throw new InternalErrorException(ex);
```
xflord, line 1073: Why can't we just throw `RoleCannotBeManagedException`?

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
```diff
@@ -1384,6 +1384,39 @@ public static List<RichUser> getRichAdmins(PerunSession sess, PerunBean compleme
...
+	 * If <b>onlyDirectAdmins</b> is <b>true</b>, return only direct users of the complementary object for role.
+	 *
+	 * @param sess perun session
+	 * @param complementaryObject for which we will get administrator
+	 * @param role expected role to filter managers by
+	 * @param onlyDirectAdmins if true, get only direct user administrators (if false, get both direct and indirect)
+	 *
+	 * @return list of user administrators for complementary object and role.
+	 */
+	public static List<User> getAdmins(PerunSession sess, PerunBean complementaryObject, String role, boolean onlyDirectAdmins) throws PrivilegeException, RoleCannotBeManagedException {
```
xflord, line 1399: Did you consider unifying this with `getRichAdmins`,since it does the same thing as `getRichAdmins` with the `specificAttributes` parameter empty? Since RichUser extends User I don't think it would cause problems.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/VosManagerMethod.java
```diff
@@ -482,20 +498,32 @@ public Void call(ApiCaller ac, Deserializer parms)
...
+					if (!AuthzResolver.roleExists(roleName)) {
+						throw new RoleNotSupportedException("Role: "+ roleName +" does not exists.", roleName);
+					}
+
+					//Role can be only supported one (TopGroupCreator, VoAdmin or VoObserver)
+					if(!roleName.equals(Role.TOPGROUPCREATOR) &&
+						!roleName.equals(Role.VOADMIN) &&
+						!roleName.equals(Role.VOOBSERVER)) {
+						throw new RoleNotSupportedException("Supported roles are VoAdmin, VoObserver and TopGroupCreator.", roleName);
+					}
```
xflord, line 519: I think something similar `getRichAdmins` already does with `objectAndRoleManageableByEntity`, but if we want to exactly preserve the old behaviour, this is fine

sarkapalkovicova: Yes, the behavior of methods should stay exactly the same. Therefore, the same kind of exception should be thrown in case of an incorrect input role.

perun-core/src/main/java/cz/metacentrum/perun/core/api/AuthzResolver.java
sarkapalkovicova: I considered it, and it seemed like a better option for me to add one new method than recast the output of the existing one. Yeah, `RichUser` is just an extension of `User`, but at the moment, we don't know where and how are these old methods used, so I wanted to be sure that it won't cause problems.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/GroupsManagerMethod.java
sarkapalkovicova: Old methods weren't throwing this exception, and under normal circumstances, this should not happen now either, right?

xflord: In this case, it does not matter which method RPC throws since it didn't use to throw either of them; it just seemed redundant, but I see your point.

xflord (approved):
well done

perun-core/src/test/java/cz/metacentrum/perun/core/api/AuthzResolverIntegrationTest.java
```diff
@@ -1345,6 +1346,61 @@ public void getVoRichAdminsWithProperRights() throws Exception {
...
+		List<User> admins = AuthzResolver.getAdmins(testSession, testVo, Role.VOADMIN, false);
+
+		assertEquals(2, admins.size());
+		assertTrue(admins.containsAll(Arrays.asList(testUser, testUser2)));
+	}
+
+	@Test
+	public void getDirectVoAdminsWithProperRights() throws Exception {
+		System.out.println(CLASS_NAME + "getVoAdminsWithProperRights");
+
```
michalberky, line 1380: I think this should be fixed to reflect the actual name of this test
```suggestion
	@Test
	public void getDirectVoAdminsWithProperRights() throws Exception {
		System.out.println(CLASS_NAME + "getDirectVoAdminsWithProperRights");

```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/VosManagerMethod.java
```diff
@@ -482,20 +498,32 @@ public Void call(ApiCaller ac, Deserializer parms)
...
+				if(parms.contains("role")) {
+					String roleName = parms.readString("role");
+					if (roleName == null) {
+						throw new RpcException(RpcException.Type.WRONG_PARAMETER, "Parameter role cannot be null.");
+					} else {
+						roleName = roleName.toUpperCase();
+					}
+
+					if (!AuthzResolver.roleExists(roleName)) {
+						throw new RoleNotSupportedException("Role: "+ roleName +" does not exists.", roleName);
```
michalberky, line 511: Typo
```suggestion
						throw new RoleNotSupportedException("Role: "+ roleName +" does not exist.", roleName);
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/VosManagerMethod.java
```diff
@@ -536,19 +566,33 @@ public List<User> call(ApiCaller ac, Deserializer parms)
...
```
michalberky, line 579: Same typo here

michalberky (approved):
Didn't test. The code looks good to me, just the two cosmetic details.

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/VosManagerMethod.java
```diff
@@ -608,10 +673,17 @@ public List<RichUser> call(ApiCaller ac, Deserializer parms) throws PerunExcepti
 	 */
 	getRichAdminsWithAttributes {
 		@Override
-		public List<RichUser> call(ApiCaller ac, Deserializer parms)
-				throws PerunException {
-			return ac.getVosManager().getRichAdminsWithAttributes(ac.getSession(),
-					ac.getVoById(parms.readInt("vo")));
+		public List<RichUser> call(ApiCaller ac, Deserializer parms) throws PerunException {
+			try {
+				return cz.metacentrum.perun.core.api.AuthzResolver.getRichAdmins(ac.getSession(),
```
HejdaJakub, line 678:
```suggestion
				return AuthzResolver.getRichAdmins(ac.getSession(),
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/VosManagerMethod.java
```diff
@@ -626,10 +698,16 @@ public List<RichUser> call(ApiCaller ac, Deserializer parms)
 	getRichAdminsWithSpecificAttributes {
 		@Override
 		public List<RichUser> call(ApiCaller ac, Deserializer parms) throws PerunException {
-
-			return ac.getVosManager().getRichAdminsWithSpecificAttributes(ac.getSession(),
+			try {
+				return cz.metacentrum.perun.core.api.AuthzResolver.getRichAdmins(ac.getSession(),
```
HejdaJakub, line 702:
```suggestion
				return AuthzResolver.getRichAdmins(ac.getSession(),
```

perun-rpc/src/main/java/cz/metacentrum/perun/rpc/methods/VosManagerMethod.java
```diff
@@ -645,10 +723,16 @@ public List<RichUser> call(ApiCaller ac, Deserializer parms) throws PerunExcepti
 	getDirectRichAdminsWithSpecificAttributes {
 		@Override
 		public List<RichUser> call(ApiCaller ac, Deserializer parms) throws PerunException {
-
-			return ac.getVosManager().getDirectRichAdminsWithSpecificAttributes(ac.getSession(),
+			try {
+				return cz.metacentrum.perun.core.api.AuthzResolver.getRichAdmins(ac.getSession(),
```
HejdaJakub, line 727:
```suggestion
				return AuthzResolver.getRichAdmins(ac.getSession(),
```

HejdaJakub (approved):

